PR #12, opened 2023-10-31
Query builder

Zhifei-Li:
Query builder migrate


static/js/query-builder.js
```diff
@@ -0,0 +1,373 @@
...
+  ) {
+    alert("Please select one of the values below");
+    return;
+  }
+  let tagContent = $("#column-first").val().trim() + $("#symbol").val().trim();
+  //   if(val.indexOf(" ") != -1){
+  //     tagContent += "\"" + val + "\"";
+  //   }else{
+  //     tagContent += val;
+  //   }
```
AndrewHess, line 194: Remove commented-out code

sonamgupta21:
1. The arrow to expand the navbar is not displayed correctly.

image


2. The query builder container should have a box shadow.

3. The spacing between the buttons is not consistent as in Figma.

image


4. The background color of the input box does not match that in Figma.

image

5. Some of the text has a different font family.


6. The border radius for the search icon is incorrect.
image


7. Clicking on the query language button (Splunk QL) changes it to code, whereas it should switch to code when selecting other languages (SQL or LogQL).
8. The save query popup does not appear correctly.
image

9. Add the download and query summary icons.

sonamgupta21:
The feedback and theme icons are not visible in the "Query Builder" screen.
On the "Code" screen, replace the search button with the magnifying glass icon.

image

Zhifei-Li:
download icon was added, but query summary icons cannot be found after I merged from upstream develop branch.

Zhifei-Li:
The current page still has scroll bars, even though I have changed the page height. This is because after I pulled the upstream develop branch from the upstream, there was too much content in the left directory bar, and I still had scroll bars when I ran the develop branch alone, so the function here was not fully effective. When the left directory bar of the main branch only occupies one page, the scroll bar of the current page will automatically disappear.

sonamgupta21:
> download icon was added, but query summary icons cannot be found after I merged from upstream develop branch.

image
We removed this summary and plan to add it to a popup. 
Add an info icon next to the download icon. On clicking the icon, display a popup with the summary details.

sonamgupta21:
Looks good to me. Just some minor CSS issues.
 @AndrewHess, can you also take a look.

AndrewHess:
1. There should be an error shown when the user tries to run a query with something in the third box (group by) but nothing in the second (aggregation). And also when there's something in the second box but nothing in the first.
2. When a click the "+" in one of the boxes, it should take me to the input box so I can immediately start typing. Right now I have to click "+" then click the input box that popped up (or hit tab), and then I can type
3. It would be nice if clicking anywhere in one of the three boxes (query, aggregation, group by) behaved the same as clicking the "+" in that box. This could potentially be done in a different PR.
4. For numeric columns, the value selected in the third step should not be wrapped in quotes. So you should get something like `http_status<=400` instead of `http_status<="400"`
5. When I'm adding something, the checkmark box to add it appears once I reach the last step but I can click on the checkmark before completing the last step and I get an error alert that pops up. I think we should instead just not show the checkmark button until something is entered in the last step, so then we never have to show that error popup
6. I think it would look better if we remove the "Group by" label and then put placeholder text inside the boxes until the user adds something to that box. Something like "Search Criteria", "Aggregations", and "Group By". What do you think Sonam and Zhifei?

Zhifei-Li:
> 1. There should be an error shown when the user tries to run a query with something in the third box (group by) but nothing in the second (aggregation). And also when there's something in the second box but nothing in the first.
> 2. When a click the "+" in one of the boxes, it should take me to the input box so I can immediately start typing. Right now I have to click "+" then click the input box that popped up (or hit tab), and then I can type
> 3. It would be nice if clicking anywhere in one of the three boxes (query, aggregation, group by) behaved the same as clicking the "+" in that box. This could potentially be done in a different PR.
> 4. For numeric columns, the value selected in the third step should not be wrapped in quotes. So you should get something like `http_status<=400` instead of `http_status<="400"`
> 5. When I'm adding something, the checkmark box to add it appears once I reach the last step but I can click on the checkmark before completing the last step and I get an error alert that pops up. I think we should instead just not show the checkmark button until something is entered in the last step, so then we never have to show that error popup
> 6. I think it would look better if we remove the "Group by" label and then put placeholder text inside the boxes until the user adds something to that box. Something like "Search Criteria", "Aggregations", and "Group By". What do you think Sonam and Zhifei?

got it, all these optimizations will be improved in the next PR. This PR mainly changes the issues mentioned in the meeting.

sunitakawane:
Rename `code` tab to `free text`

sunitakawane:
These issues can be fixed in next PR 
1. For groupBy queries , result table header and first row has same background  colour, change this
2. Header row should have bottom border line 
3. After adding groupBy or any other filter, if download results button is clicked, it downloads  all results not just the filtered results.
4. Change Background gray colour for dark theme, Top table bg colour in dark mode

sunitakawane (approved):

PR #15, opened 2023-11-01
Kebab case

Davleen14:
Renamed files in static to follow kebab case


static/js/settings.js
sunitakawane, line 1: rename this to org-settings.js

sonamgupta21: We've removed the 'org' prefix from OSS files. Should we rename 'org-settings.js' to 'settings.js'?

sonamgupta21 (approved):

PR #17, opened 2023-11-01
Add Spl statistic block: top/rare

asiL-tcefreP:
# Description

Allow parsing Splunk Statistic block: top/rare

Please notice that this is only the initial commit. I will create some unit tests, add the test data to the. csv file and commit again.

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

# Checklist:

- [ ] Newly added APIs have been documented at [API ReadMe](https://github.com/sigscalr/hyperion/blob/develop/API_README.md)
- [ ] I have validated that Kibana and `long_running.py` continue to work properly
  - Most cases, CI/CD will cover this
  - For PRs that touch the critical paths or change multiple modules this is needed
- [ ] I will edit `GETTING_STARTED.md` if changing configs, startup commands, etc.
- [x] I will **squash and merge** this branch if merging to `develop`
- [ ] I will **commit merge** this branch if merging to `main`
- [ ] I will read the [Release ReadMe](https://github.com/sigscalr/hyperion/blob/develop/RELEASE_README.md) if creating a release
- [x] I will delete this branch after merging (except for develop)


asiL-tcefreP:
SPL Rename doc: 
https://docs.splunk.com/Documentation/Splunk/9.1.1/SearchReference/Top, https://docs.splunk.com/Documentation/Splunk/9.1.1/SearchReference/Rare

For top/rare block. If there is a limit option, sort by frequency. Otherwise, sort by lexicographical value. 
I tested these queries as follows:

1. Without Stats GroupBy
- city=Boston | top app_name
- city=Boston | rare state, gender by weekday //Sory by frequency 
- city=Boston | top 10 http_method, gender by country, http_status
- city=Boston | top limit=10 state by country, http_status useother=true otherstr=testOther 
- city=Boston | rare limit=15 http_method, gender by country, http_status showperc=false showcount=false //Test Options
- city=Boston | rare 3 http_method, gender by country, http_status useother=true otherstr=testOther percentfield=http_method countfield=gender //Override Existing fields
- city=Boston | top 10 http_method, gender by country, http_status useother=true otherstr=testOther percentfield=http_status countfield=gender //Override Existing fields

2. With Stats GroupBy
- city=Boston | stats count AS gg BY http_status, weekday, gender | rare 2 http_status, gg useother=true countfield=true percentfield=weekday //Use columns generated by stats block
- city=Boston | stats count AS gg BY http_status, weekday, gender, state | rare 2 gg, state, http_status useother=true countfield=true percentfield=weekday //Override stats block's columns
- city=Boston | stats count AS gg BY http_status, weekday, gender | eval newField=(http_status - 100), aa=(http_status - 66) | rare 4 http_status, newField, gg, aa useother=true countfield=cf percentfield=pf //Use columns generated by eval block and stats block
- city=Boston | stats count AS gg BY http_status, weekday, gender | eval newField=(http_status - 100) | rare 4 http_status, weekday useother=true countfield=newField percentfield=gg //Override

pkg/ast/spl/spl.peg
```diff
@@ -376,6 +376,168 @@ SearchTerm <- term:(FieldWithNumberValue / FieldWithStringValue) {
...
+        fieldAtom := restSlice[i].([]any)
+        slice[i + 1] = fieldAtom[2].(string)
+    }
+
+    return slice, nil
+}
+
+ByClause <- SPACE "by" fieldList:FieldList {
+    return fieldList.([]string), nil
+}
```
AndrewHess, line 539: Can we use the existing `GroupbyBlock` rule instead, or is that rule different enough that we have to make these rules?

pkg/segment/aggregations/segaggs.go
```diff
@@ -126,14 +127,13 @@ func performAggOnResult(nodeResult *structs.NodeResult, agg *structs.QueryAggreg
 				return fmt.Errorf("performAggOnResult: %v", err)
 			}
 		}
-	default:
-		return errors.New("performAggOnResult: multiple QueryAggregators is currently only supported for OutputTransformType")
```
AndrewHess, line 130: Why did you remove this?

pkg/segment/structs/evaluationstructs.go
```diff
@@ -120,6 +138,13 @@ const (
 	BoolOpOr
 )
 
+type StatisticExprMode uint8
+
+const (
+	SEMTop = iota
+	SEMRare
+)
```
AndrewHess, line 146: This `SEM...` naming convention is going to make it hard to know if its for a `StatisticExprMode` or for a `StringExprMode`, so you should change one or both of these so the modes for string and statistic expressions doesn't both follow the pattern `SEM...`

pkg/segment/structs/evaluationstructs.go
```diff
@@ -472,6 +497,243 @@ func (self *ConcatExpr) GetFields() []string {
...
+		// Only return unique limit field combinations
+		// Since the results are already in order, and there is no Set in Go, we can use a string arr to record the previous field combinations
+		// If the current field combination is different from the previous one, it means we have finished processing data for one field combination (we need to process limit combinations in total, limit is a number)
+		uniqueFieldsCombination := make([]string, len(self.FieldList))
+		combinationCount := 0
+		newResults := make([]*BucketResult, 0)
+		// log.Error("fjl sort Results:", len(*results))
+		for _, bucketResult := range *results {
+			// log.Error("fjl sort bucket:", bucketResult)
+			// log.Error("fjl sort rowINdex:", rowIndex, " newRes:", newResults)
```
AndrewHess, line 615: Remove this commented-out code

pkg/segment/structs/evaluationstructs.go
```diff
@@ -472,6 +497,243 @@ func (self *ConcatExpr) GetFields() []string {
...
+				}
+			}
+
+			// If there is a stats groupby block before statistic groupby block. E.g. ... | stats count BY http_status, gender | rare 1 http_status,
+			// In this case, each http_status will be divided by two genders, so we should merge them into one row here
+			//Fields combination does not exist
+			if !combinationExist {
+				combinationCount++
+				if combinationCount > limit {
+					// *Results = (*Results)[:rowIndex]
```
AndrewHess, line 639: Don't add commented-out code

pkg/segment/structs/evaluationstructs.go
```diff
@@ -472,6 +497,243 @@ func (self *ConcatExpr) GetFields() []string {
...
+	bucketKey := make([]string, 0)
+	switch bucketResult.BucketKey.(type) {
+	case []string:
+		bucketKeyStrs := bucketResult.BucketKey.([]string)
+
+		for _, field := range groupByCols {
+			for rowIndex, groupByCol := range bucketResult.GroupByKeys {
+				if field == groupByCol {
+					groupByKeys = append(groupByKeys, field)
+					// bucketKeyIndex = append(bucketKeyIndex, rowIndex)
```
AndrewHess, line 678: Don't add commented-out code

AndrewHess (commented):
Looks pretty good. I added a few comments, and also you should add some unit tests

pkg/segment/aggregations/segaggs.go
asiL-tcefreP: For Stats GroupBy, it does not enter the `performAggOnResult() function`, so it does not trigger this default option. However, for this top/rare GroupBy, it does enter performAggOnResult() because we want to process it and obtain the desired results. Since its agg.PipeCommandType is not equal to structs.OutputTransformType, it triggers this default option and returns an error.

My interpretation of this default option is similar to its error message, which states, "multiple QueryAggregators are currently only supported for OutputTransformType" However, at present, we can process another GroupByType after the stats GroupBy, not just OutputTransformType. That's why I removed this default option.

pkg/ast/spl/spl.peg
asiL-tcefreP: Yes, it is almost the same. In the Splunk doc, keyword "by" in example is in lower case. If it does not matter, I can use GroupBlock instead.

pkg/segment/aggregations/segaggs.go
AndrewHess: Then you should add a `case` to this switch statement instead of removing the `default`. We basically always want to have a `default` in all `switch` statements so we can write an error message if we get somewhere unexpected.

pkg/ast/spl/spl.peg
AndrewHess: In the Splunk documentation here: https://docs.splunk.com/Documentation/Splunk/9.1.1/SearchReference/Top
They have this:
> <by-clause\>
Syntax: BY <field-list>
Description: The name of one or more fields to group by.

But in the examples they use lowercase `by`. So we should handle both upper and lower case. For the `GroupbyBlock` I think lowercase should also be allowed but I'm not quite sure. So you can change `GroupbyBlock` to allow upper and lower case, and then reuse that rule here.

AndrewHess (commented):
Looks good. There's a bunch of merge conflicts you'll need to fix, possibly because the Rename PR got merged already and had some similar changes.

Also it looks like you implemented the `top` functionality defined in the Splunk docs here: https://docs.splunk.com/Documentation/Splunk/9.1.1/SearchReference/Top and did the opposite for `rare`, as the `rare` documentation at https://docs.splunk.com/Documentation/Splunk/9.1.1/SearchReference/Rare says it behaves identically to `top` except that it finds the least frequent values. However the syntax for the `rare` command is in a slightly different order in the Splunk docs and it doesn't have some options that `top` does (e.g., `useother` and `otherstr`). So I think what you have is good, we'll just want to be aware that the Splunk documentation is inconsistent on this

AndrewHess (approved):
Looks good

PR #26, opened 2023-11-02
Eval like match

Davleen14:
Eval like,match


pkg/ast/spl/spl.peg
```diff
@@ -650,6 +650,36 @@ BoolExprLevel1 <- "(" first:BoolExprLevel4 ")" {
...
+         LeftValue: stringr.(*structs.ValueExpr),
+         RightValue: pattern.(*structs.ValueExpr),
+         ValueOp: "like",
+     }
+     return expr, nil
+ }
+/ "match(" stringr:ValueExpr ", " pattern:ValueExpr ")" {
+     expr := &structs.BoolExpr {
+         IsTerminal: true,
+         LeftValue: stringr.(*structs.ValueExpr),
```
AndrewHess, line 677: typo: `stringr` instead of `string`

pkg/segment/structs/evaluationstructs.go
```diff
@@ -272,6 +306,13 @@ func (self *BoolExpr) Evaluate(fieldToValue map[string]utils.CValueEnclosure) (b
 	}
 }
 
+func likePatternToRegex(pattern string) string {
+	pattern = regexp.QuoteMeta(pattern)
+	pattern = strings.Replace(pattern, "%", ".*", -1)
+	pattern = strings.Replace(pattern, "_", ".", -1)
+	return "^" + pattern + "$"
+}
```
AndrewHess, line 314: Are you calling this function anywhere?

AndrewHess (commented):
Looks pretty good, but you need to add some unit tests

pkg/segment/structs/evaluationstructs.go
Davleen14: I have incorporated the functionality directly in the case so no need of this function anymore.

pkg/ast/spl/spl.peg
Davleen14: Will make the change

pkg/segment/structs/evaluationstructs.go
AndrewHess: Ok, just make sure to delete this function then

pkg/segment/structs/evaluationstructs_test.go
AndrewHess, line 1: FYI for these tests you don't need the expressions doing an `if` statement (e.g., the `likeConditionExpr`). You can just make the expression doing the actual check (e.g., the `likeBoolExpr`) and evaluate that to a boolean and assert the value you get back. You can keep it how you have it though.

AndrewHess (approved):

PR #38, opened 2023-11-03
Add OTLP trace ingestion

AndrewHess:
This allows us to ingest OTLP traces using the http/protobuf protocol.

To ingest some sample traces, you can run siglens and follow https://opentelemetry.io/docs/demo/docker-deployment/ to send traces to siglens. Thus, if you're running siglens with `ingestPort: 4318` in the server.yaml file, you'll follow these steps:
1. `git clone https://github.com/open-telemetry/opentelemetry-demo.git`
2. `cd opentelemetry-demo/`
3. Update `src/otelcollector/otelcol-config-extras.yml` to be:
```
exporters:
  otlphttp/siglens:
    endpoint: "http://host.docker.internal:4318/otlp"

service:
  pipelines:
    traces:
```
This assumes that you're running siglens on your local machine, not in docker. Otherwise you'll have to update the `host.docker.internal` part. You might also need to update that if you're running docker on something other than OSX.
4. `make start`
After the docker containers start and you wait a few seconds, you should see traces getting ingested into siglens.

## Update (2024-01-11)
You can ignore the Note section below, it was fixed a while ago

## Note
There's currently an issue with packer.go so that when a field start getting string ingested into it and then gets a number ingested into it, siglens will crash. Since we're ingesting spans by making a column for each `Attributes.key` and some services have the same `Attributes.key` but some have string values and others have int values for that column, using the above commands to ingest traces from the opentelemetry-demo repo will crash siglens. To get around this before we fix the issue underlying issue, you can either use something else to ingest sample traces or you can force spans with numeric values to be ingested as string values by updating otlp.go; to do this, change the `extractAnyValue()`  lines from
```
case *commonpb.AnyValue_IntValue:
	return anyValue.GetIntValue(), nil
case *commonpb.AnyValue_DoubleValue:
	return anyValue.GetDoubleValue(), nil
```
to
```
case *commonpb.AnyValue_IntValue:
	return fmt.Sprintf("%v", anyValue.GetIntValue()), nil
case *commonpb.AnyValue_DoubleValue:
	return fmt.Sprintf("%v", anyValue.GetDoubleValue()), nil
```

### DIFF
--- a/pkg/otlp/otlp.go
+++ b/pkg/otlp/otlp.go
@@ -87,7 +87,7 @@ func ProcessTraceIngest(ctx *fasthttp.RequestCtx) {
 				lenJsonData := uint64(len(jsonData))
 				err = writer.ProcessIndexRequest(jsonData, now, indexName, lenJsonData, shouldFlush, localIndexMap, orgId)
 				if err != nil {
-					fmt.Errorf("ProcessTraceIngest: failed to process ingest request: %v", err)
+					log.Errorf("ProcessTraceIngest: failed to process ingest request: %v", err)
 					numFailedSpans++
 					continue
 				}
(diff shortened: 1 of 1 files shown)


sunitakawane:
why is the change to extractAnyValue func(mentioned in notes) not part of this PR?

AndrewHess:
> why is the change to extractAnyValue func(mentioned in notes) not part of this PR?

Because we want to ingest numbers as numbers rather than as strings so that we can do inequality comparisons, call numeric functions on them, etc. Once we fix the issue causing siglens to crash when a single column has both strings and numbers, the workaround mentioned in the Notes section won't be necessary. Also, this issue of ingesting strings and numbers into one column is an issue with the current packer.go, it's not specific to ingesting traces.

nkunal (approved):

Issue #31, opened 2023-11-02
Get query port from environment variable


AndrewHess:
Update `GetQueryPort()` in config.go. If the `PORT` environment variable is set, `GetQueryPort()` should return that value. Otherwise, it should return the value of the `queryPort` in the config file (e.g., server.yaml).

girish332:
Hey Can I work on this issue ?

AndrewHess:
Yep!

girish332:
https://github.com/siglens/siglens/pull/39/commits

Kindly review the PR and let me know if you have any changes in mind. I can pick https://github.com/siglens/siglens/issues/14 next after this issue is done

AndrewHess:
Completed by https://github.com/siglens/siglens/pull/39

PR #39, opened 2023-11-03
#31 Updated GetQueryport func to get value from env 

girish332:
PR For the issues : https://github.com/siglens/siglens/issues/31

Updated the GetQueryPort func to get value from env variable

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -258,19 +258,7 @@ func GetIngestPort() uint64 {
 // returns the configured query port
 // if the node is not a query node, this will not be set
 func GetQueryPort() uint64 {
-	var port uint64
-	value := os.Getenv("PORT")
-	if value != "" {
-		n, err := strconv.Atoi(value)
-		if err != nil {
-			log.Error("Error while converting port to int")
-			n = 80
-		}
-		port = uint64(n)
-	} else {
-		port = runningConfig.QueryPort
-	}
-	return port
+	return runningConfig.QueryPort
 }
 
 // returns the psql port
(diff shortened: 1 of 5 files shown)


pkg/config/config.go
```diff
@@ -258,7 +258,19 @@ func GetIngestPort() uint64 {
...
-	return runningConfig.QueryPort
+	var port uint64
+	value := os.Getenv("PORT")
+	if value != "" {
+		n, err := strconv.Atoi(value)
+		if err != nil {
+			log.Error("Error while converting port to int")
+			n = 80
+		}
+		port = uint64(n)
```
AndrewHess, line 269: In this error case where `PORT` is defined but not a number, we can just fall back to `runningConfig.QueryPort` instead of using `80` as the default.

AndrewHess (commented):
Looks pretty good. Can you move this logic from `GetQueryPort()` into `derivedConfig.go` so that it is computed in `InitDerivedConfig()` and then all calls to `GetQueryPort()` will return this computed value, even if the `PORT` environment variable changes while siglens is running. Sorry, I didn't add this when I made the GitHub issue.

nkunal:
@girish332 what you did is perfect, it meets what was defined in this issue. 
We have some other issues that also need to be fixed in order for your changes to "truly" take effect.

This part of the code: 
```
	err = config.InitDerivedConfig(nodeID)
	if err != nil {
		log.Errorf("Error initializing derived configurations! %v", err)
		return err
	}
```
https://github.com/siglens/siglens/blob/develop/cmd/startup/startup.go#L76-L81
needs to be moved from where it is currently and need to be place right here : https://github.com/siglens/siglens/blob/develop/cmd/siglens/main.go#L60
along with that you will need the 
```	nodeID := localnodeid.GetRunningNodeID()
``` which is here https://github.com/siglens/siglens/blob/develop/cmd/siglens/main.go#L98C7-L98C7 need to be initialized before you call the `InitDerivedConfig...`
hope this makes sense.

nkunal (approved):

PR #40, opened 2023-11-03
Eval substr tostring tonumber

Davleen14:
Eval substr tostring tonumber


pkg/ast/spl/spl.peg
```diff
@@ -523,6 +523,81 @@ TextExpr <- (opName:("lower") "(" stringExpr:StringExpr ")") {
     }
     return node, nil
 }
+/ (opName:("substr") "(" stringExpr:StringExpr "," startIndex:NumericExpr lengthParam:(", " NumericExpr)? ")") {
```
AndrewHess, line 526: Make sure you allow a space after the comma (you did this for the second comma but not the first)

pkg/ast/spl/spl.peg
```diff
@@ -523,6 +523,81 @@ TextExpr <- (opName:("lower") "(" stringExpr:StringExpr ")") {
...
+    var formatExpr *structs.StringExpr
+    if format != nil {
+        formatSlice := format.([]interface{})
+        formatExpr, _ = formatSlice[1].(*structs.StringExpr) 
+    }
+    return &structs.TextExpr{
+        IsTerminal: false,
+        Op:         "tostring",
+        Val:      value.(*structs.ValueExpr),
+        Format:     formatExpr,
```
AndrewHess, line 598: nit: `value.(*structs.ValueExpr),` should be indented a little more

pkg/segment/structs/evaluationstructs.go
```diff
@@ -683,8 +689,50 @@ func (self *TextExpr) EvaluateText(fieldToValue map[string]utils.CValueEnclosure
...
+				if convErr != nil {
+					return "", fmt.Errorf("TextExpr.Evaluate: failed to convert value '%s' to integer for hex formatting: %v", valueStr, convErr)
+				}
+				return fmt.Sprintf("%#x", num), nil
+			case "commas":
+				num, convErr := strconv.ParseFloat(valueStr, 64)
+				if convErr != nil {
+					return "", fmt.Errorf("TextExpr.Evaluate: failed to convert value '%s' to float for comma formatting: %v", valueStr, convErr)
+				}
+				return fmt.Sprintf("%0.2f", num), nil
```
AndrewHess, line 719: Will this add the commas?

pkg/segment/structs/evaluationstructs.go
```diff
@@ -683,8 +689,50 @@ func (self *TextExpr) EvaluateText(fieldToValue map[string]utils.CValueEnclosure
...
 
+	} else if self.Op == "tostring" {
+		valueStr, err := self.Val.EvaluateToString(fieldToValue)
+		if err != nil {
+			return "", fmt.Errorf("TextExpr.Evaluate: failed to evaluate value for 'tostring' operation: %v", err)
+		}
+		boolExpr, err := ParseAndEvalBooleanExpression(valueStr)
+		if err == nil {
+			return strconv.FormatBool(boolExpr), nil
+		}
```
AndrewHess, line 701: Do we need to evaluate boolean expressions? If so, we should probably handle that with our existing `BoolExpr` so that you don't have to reimplement boolean expressions in this `ParseAndEvalBooleanExpression()` function

pkg/segment/structs/evaluationstructs.go
```diff
@@ -771,10 +908,16 @@ func (self *ConditionExpr) EvaluateCondition(fieldToValue map[string]utils.CValu
...
 	if self.IsTerminal || (self.Op != "max" && self.Op != "min") {
-		return self.Value.GetFields()
+		if self.Value != nil {
+			fields = append(fields, self.Value.GetFields()...)
+		}
+		if self.Val != nil {
+			fields = append(fields, self.Val.GetFields()...)
+		}
+		return fields
 	}
```
AndrewHess, line 920: Don't you need to call `GetFields()` on all the fields you added to the `TextExpr`?

pkg/ast/spl/spl.peg
Davleen14: Okay will do that

pkg/ast/spl/spl.peg
Davleen14: Okay will do that

pkg/segment/structs/evaluationstructs.go
Davleen14: No this dosent add commas, as per the documentation it will do this 

tostring(<value>,"commas") | Formats <value> with commas. If the number includes decimals, the function rounds to nearest two decimal places.

pkg/segment/structs/evaluationstructs.go
Davleen14: Yes made the changes

pkg/segment/structs/evaluationstructs.go
Davleen14: Yes made the changes

pkg/segment/structs/evaluationstructs.go
AndrewHess: Ok... so it's supposed to add commas but your code doesn't? Splunk has an example:
```
... | eval n=tostring(1==1) + " " + tostring(15, "hex") + " " + tostring(12345.6789, "commas")
```
which should return `True 0xF 12,345.68` (so 12345.6789 became "12,345.68", so it got rounded and added commas)

Davleen14: Okay will do that

pkg/ast/spl/spl.peg
```diff
@@ -586,6 +586,81 @@ TextExpr <- (opName:("lower") "(" stringExpr:StringExpr ")") {
...
+            }
+        }
+    } 
+    node := &structs.TextExpr {
+        Op: opNameStr,
+        Value: stringExprConverted,
+        BaseExpr: baseExprConverted,
+    }
+    return node, nil
+}
```
AndrewHess, line 650: Shouldn't this be in the `NumericExpr` rule rather than the `TextExpr` rule? I think the `TextExpr` is only for things that return strings and the `NumericExpr` is for things that return numbers.

pkg/segment/structs/evaluationstructs.go
```diff
@@ -814,8 +831,48 @@ func (self *TextExpr) EvaluateText(fieldToValue map[string]utils.CValueEnclosure
...
+					return "", fmt.Errorf("TextExpr.Evaluate: failed to convert value '%s' to integer for hex formatting: %v", valueStr, convErr)
+				}
+				return fmt.Sprintf("%#x", num), nil
+			case "commas":
+				num, convErr := strconv.ParseFloat(valueStr, 64)
+				if convErr != nil {
+					return "", fmt.Errorf("TextExpr.Evaluate: failed to convert value '%s' to float for comma formatting: %v", valueStr, convErr)
+				}
+				roundedNum := math.Round(num*100) / 100
+				formattedNum := humanize.CommafWithDigits(roundedNum, 2)
```
AndrewHess, line 858: Do you need `roundedNum := math.Round(num*100) / 100` since you're using `humanize.CommafWithDigits(roundedNum, 2)`? Does `CommafWithDigits()` not properly round the number how we want it?

AndrewHess (commented):
Looks good. I just have a few questions but after that it can be merged

pkg/ast/spl/spl.peg
Davleen14: As per the documentation tonumber takes a string value hence wrote it under text expr
https://docs.splunk.com/Documentation/SCS/current/SearchReference/ConversionFunctions#tonumber.28.26lt.3Bstr.26gt.3B.2C_.26lt.3Bbase.26gt.3B.29

AndrewHess: Ok, but I think we want to write rules based on the value they return. That way, when a rule has a `TextExpr` as a subpart of the rule, we know the `TextExpr` part will return something with a string value. The `tonumber()` function should return a `NumericExpr` struct, but the `TextExpr` rule should match things that return strings. So I think this `tonumber()` parsing should be moved to the `NumericEvalExpr` rule

pkg/segment/structs/evaluationstructs.go
Davleen14: It was an oversight. Fixed it.

Davleen14: If I use only humanize.CommafWithDigits(roundedNum, 2) it rounds off the number 12345.6789 to 12,345.67 but as per documentation the  number 12345.6789 is rounded to 12,345.68 that is why I used roundedNum := math.Round(num*100) / 100

AndrewHess (approved):

PR #47, opened 2023-11-07
Eval now

Davleen14:
Added functionality for eval now


pkg/ast/spl/spl.peg
```diff
@@ -539,6 +539,12 @@ TextExpr <- (opName:("lower") "(" stringExpr:StringExpr ")") {
 
     return node, nil
 }
+/"now" "()" {
+    return &structs.TextExpr{
+        IsTerminal: true,
+        Op:         "now",
+    }, nil
+}
```
AndrewHess, line 547: Why is this a `TextExpr`, shouldn't it be a `NumericExpr`?

Also, you could use `"now()"` instead of `"now" "()"` but you can keep it how you have it if you prefer

Davleen14: Initially as per my understanding I thought it would return the time search as a string value.

AndrewHess: I'm not finding a concrete type mentioned in the docs here: https://docs.splunk.com/Documentation/Splunk/9.1.1/SearchReference/DateandTimeFunctions#now.28.29
But since it returns unix time, I think returning it as an integer makes more sense, so you should do that and then also put it in the `NumericEvalExpr` rule

AndrewHess (approved):

AndrewHess (approved):

PR #48, opened 2023-11-07
created install_with_docker.sh

sunitakawane:


README.md
```diff
@@ -22,26 +22,39 @@ Come say Hi to us on <a href="https://www.siglens.com/slack" target="_blank">Sla
...
+### Install Using SigLens Docker
+
+- SigLens can be installed on Linux or macOS machine. 
+- On macOS, Docker Engine should be installed before you run the install script. 
+- Git clone the SigLens repository and cd into the siglens directory 
+```
+    git@github.com:siglens/siglens.git
+    cd siglens
+```
+- Run the install.sh script:
```
nkunal, line 44: change this to Run the installation script

install_with_docker.sh
```diff
@@ -0,0 +1,158 @@
...
+     ;;
+esac
+
+arch=$(uname -m)
+if [[ $arch == x86_64* ]]; then
+    arch="amd64"
+elif  [[ $arch == arm* ]]; then
+    arch="arm64"
+else
+    echo 'Not Supported Architecture'
```
nkunal, line 53: we should exit here

install_with_docker.sh
```diff
@@ -0,0 +1,158 @@
...
+                exit 1;
+            fi
+
+            echo -e "Got Sudo access.\n"
+        fi
+	fi
+}
+
+
+install_docker() {
```
nkunal, line 77: this func should be called `install_docker_for_non_mac` or something, since the brew case is not handled here, and its confusing to why brew is not handled in this function, then below there is a check that calls this function only for non brew cases

nkunal (approved):
Looks pretty good. a few minor comments, we should figure out how we can make this script work for any release number, currently there is dependency that we have to modify the version number inside this file for the server.yaml

sunitakawane:
downloading server.yaml step will change once we start using github releases, where we can upload the assets.

PR #56, opened 2023-11-08
Dashboard

Zhifei-Li:
just show the progress, not finished yet


static/css/dashboard.css
```diff
@@ -64,8 +65,9 @@
   flex-direction: column;
   align-items: center;
   height: 100%;
-  background-color: var(--db-options-bg-color);
   border-radius: 5px 0 0 0;
+  border: 1px solid #DCDBDF;
```
sonamgupta21, line 69: You should use variables to define colors for both the dark and light themes. This way, you won't need to redefine the styles for the dark theme separately.

sonamgupta21:
1. Please remove the CSS styles and images that are not being used.
2. The arrow for the panel options is not visible in dark mode.
image
3. Remove this inner box 
image
4. On hover the icon size should not increase
image
5. Give border-radius to all four corners
image
6. The color on hover is different than figma.
image
7.Increase the panel display height a little more
image
8. 
image
image
9. Provide border when select table view
image
image

sonamgupta21 (approved):

PR #64, opened 2023-11-10
Alert and Contact Point UI 

sonamgupta21:
1. Update alert and contact points per Figma design.
2. Reduce padding in the navbar for additional icon placement.
3. Implement some changes to the cluster page.


sonamgupta21:
Dark Theme- 
image
image
image
image

static/org-settings.html
```diff
@@ -53,9 +53,6 @@ <h1 class="myOrg-heading">My Org</h1>
                             <div class="section-button"><a href="./test-data.html">Test Data</a></div>
                         </div>
                     </div>
-                    <div>
-                        <button type="button" id="logout-btn" class="btn">Logout</button>
-                    </div>
```
AndrewHess, line 58: Why don't we need this anymore?

AndrewHess (commented):
The alert rules page looks really nice on Safari, but when I view it in Chrome it doesn't seem to render properly.
Safari:
image 2
Chrome:
image

A few other things:
* In the “Evaluate every X min for Y min” part, can you add an i-icon to explain what that means
* In Notifications when I click Add New and enter some information, I can click off the pop-up and it closes the pop-up and doesn’t save the information I added. You might have a better understanding of how we should handle this, but my intuition is that nothing should happen when the user clicks off the popup, so they have to click either Cancel or Save to exit it.
* When making a contact point, if I enter nothing and click Save I get a message popping up on the input box saying “Please fill out this field” which looks nice, but when I enter a contact point name and don’t set the slack Channel and then click Save, I get a popup on the top right of the screen that looks worse to me; so probably make all of the popups for incomplete information be like the “Please fill out this field” one.
* When making a contact point if I set the name and add information for the contact type and then click “Add new contact type” but either don't add anything or don't fully populate it and click Save, it saves the contact point (with just the first contact type, discarding the second) and takes me back to the contact point overview view; I think it should probably instead give me an error saying to either remove the additional contact type or to enter info into the next field of it.
* We might want to allow creating an alert with no labels. What do you think?
* Did we get rid of email alerts? I’m only seeing Slack and webhook
* nit: not sure if we need to fix this, but in Safari when I go into "Alert rules" and enter a rule name and then hit tab, it goes to the next input box and highlights it. But in Chrome when I do that it goes to the Logs dropdown but doesn't highlight it, and when I hit tab a few more times to get to the "Enter a query" input it does highlight it like Safari

static/org-settings.html
sonamgupta21: Since there is no org functionality, and the logout button does not do anything, it does not make sense to have it. Therefore, I removed it.

sonamgupta21:
- Yes, email support has been removed by Sunita. I'm not sure, but it was not scalable when she was trying to run multiple alerts concurrently.
- Regarding labels, Sunita asked me to add a default label, "alerting=true." We can ask for her opinion on creating alerts without labels.
- Issue of not rendering the UI properly on Chrome, I am unable to reproduce it. Could you please try again in an incognito window to see if it works properly?

sonamgupta21:
@AndrewHess, I made a small backend change because the contact point name wasn't updating. Due to this, the CI/CD test is failing. Could you please review it?

AndrewHess (approved):
Looks good except for one thing, the i-icons you added for the "evaluate every X for Y" part seem to be anchored to the right side of the screen instead of to the text, so if I change the size of my window those icons move and are out of place. We could also have just one i-icon for that line, but do whatever you think will be best

When I run it in incognito Chrome it looks how it does in Safari, so that must have just been an issue on my side.

I'm not sure how to push to this PR, but to fix the automatic testing you have to format the Go code. In this case I think you just need to add a space after the comma, so 
```
..., editedAlert.AlertInfo.ContactName, ...
```
instead of
```
...,editedAlert.AlertInfo.ContactName, ...
```
You can install `goimports` and then use it to auto-format code running `goimports -w .` to format all of the Go code in the current directory. You can see any code that isn't properly formatted with `goimports -d <my-dir>`, and if running that on the base directory of siglens outputs anything, the PR test will fail.

sunitakawane (approved):

PR #66, opened 2023-11-13
Add Service Health Backend

asiL-tcefreP:
Allow the ingestion of Red Metrics into the 'red-traces' index every 5 minutes.

```
type RedMetrics struct {
	Rate      float64 // Number of entry spans divided by 60 seconds
	ErrorRate float64 // Percentage of entry spans that errored
	P50       uint64  // p50, p90, p95, p99 latencies are calculated by the list of durations.
	P90       uint64
	P95       uint64
	P99       uint64
}
```


pkg/segment/tracing/handler/tracehandler.go
```diff
@@ -0,0 +1,344 @@
...
+		}
+		return
+	}
+
+	// Parse the JSON data from ctx.PostBody
+	var requestData map[string]interface{}
+	if err := json.Unmarshal(ctx.PostBody(), &requestData); err != nil {
+		log.Errorf("ProcessSearchTracesRequest: could not unmarshal json body, err=%v", err)
+		return
+	}
```
AndrewHess, line 59: I think you can make this both much more compact and more readable by defining a struct that has the fields we expect to get from this `RequestCtx` and unmarshaling the `ctx.PostBody()` into that struct, so you should do that. This will also help avoid potential typos because later when you access this `requestData` you would do something like `requestData.SearchText` instead of `requestData["searchText"]`

pkg/segment/tracing/handler/tracehandler.go
```diff
@@ -0,0 +1,344 @@
...
+		spanCnt += count
+		if statusCode == string(structs.Status_STATUS_CODE_ERROR) {
+			errorCnt += count
+		}
+	}
+
+	trace := &structs.Trace{
+		TraceId: traceId,
+		// StartTime: , // to be finished
+		// EndTime: , // to be finished
```
AndrewHess, line 186: Yes, you need to set these somewhere

pkg/segment/tracing/utils/lineartimefinding.go
```diff
@@ -0,0 +1,118 @@
...
+
+	sort.Slice(arr, func(i, j int) bool {
+		return arr[i] < arr[j]
+	})
+	if len(arr)%2 == 1 {
+		return arr[len(arr)/2]
+	} else {
+		return (arr[len(arr)/2-1] + arr[len(arr)/2]) / 2
+	}
+}
```
AndrewHess, line 93: Is this needed? Is it just for testing? If you only use it for testing you can move it to the testing file

pkg/segment/tracing/handler/tracehandler.go
```diff
@@ -0,0 +1,344 @@
...
+		pipeSearchResponseOuter := pipesearch.PipeSearchResponseOuter{}
+		if err := json.Unmarshal(rawTraceCtx.Response.Body(), &pipeSearchResponseOuter); err != nil {
+			log.Errorf("ProcessSearchTracesRequest: could not unmarshal json body for trace=%v, err=%v", traceId, err)
+			continue
+		}
+		// To be modified
+		// Only process traces which start and end in this period [startEpoch, endEpoch]
+		// if (startEpoch*1e6 > uint64(startTime)) || (endEpoch*1e6 < uint64(endTime)) {
+		// 	continue
+		// }
```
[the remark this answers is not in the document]
asiL-tcefreP: Yes, I will remove it. I wrote this comment in the PR of search traces. I will finish that in that PR.

pkg/segment/tracing/utils/lineartimefinding.go
asiL-tcefreP: I think this method will be called when this algorithm needs to pick a pivot. 
This algorithm just return the median when there are < 5 items. The author proved that why use this way to select a pivot is better than just randomly pick one. (https://rcoh.me/posts/linear-time-median-finding/)

pkg/segment/tracing/handler/tracehandler.go
asiL-tcefreP: Since this method is belong to the PR of search traces. I will modify that in that PR.

pkg/segment/tracing/structs/tracestruct.go
```diff
@@ -16,47 +16,37 @@ type Status_StatusCode string
 
 const (
 	// The default status.
-	Status_STATUS_CODE_UNSET Status_StatusCode = "0"
-	Status_STATUS_CODE_OK    Status_StatusCode = "1"
-	Status_STATUS_CODE_ERROR Status_StatusCode = "2"
+	Status_STATUS_CODE_UNSET Status_StatusCode = "STATUS_CODE_UNSET"
+	Status_STATUS_CODE_OK    Status_StatusCode = "STATUS_CODE_OK"
+	Status_STATUS_CODE_ERROR Status_StatusCode = "STATUS_CODE_ERROR"
```
AndrewHess, line 21: Why do we want this change, and do we also need to have the 0-2 values or can we just use the `STATUS_CODE_...` values? I think the 0-2 values were added in your Search Traces PR if I'm not mistaken

pkg/segment/tracing/handler/tracehandler.go
```diff
@@ -208,3 +211,154 @@ func processSearchRequest(searchRequestBody *structs.SearchRequestBody, myid uin
...
+}
+
+func ProcessRedTracesIngest() {
+	// Initial request
+	ctx := &fasthttp.RequestCtx{}
+	requestDataMap := make(map[string]interface{})
+	requestDataMap["queryLanguage"] = "Splunk QL"
+	requestDataMap["searchText"] = ""
+	requestDataMap["startEpoch"] = "now-5m"
+	requestDataMap["endEpoch"] = "now"
```
AndrewHess, line 231: You also need to set the index, right?

AndrewHess (commented):
Looks good, just a couple minor issues

pkg/segment/tracing/structs/tracestruct.go
AndrewHess: Ok, sounds good

AndrewHess (approved):
Looks good

PR #71, opened 2023-11-15
Backend dependency graph

Davleen14:
Added initial logic for backend dependency graph


cmd/startup/startup.go
AndrewHess, line 1: Most of this probably doesn't belong in startup.go. The pkg/segment/tracing folder Jiale made in #70 is probably a good place, or maybe somewhere else

cmd/startup/startup.go
```diff
@@ -140,6 +144,8 @@ func StartSiglensServer(nodeType config.DeploymentType, nodeID string) error {
 	if queryNode {
 		startQueryServer(queryServer)
 	}
+	go makeTracesDependancyGraph()
+	go mock()
```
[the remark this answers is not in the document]
Davleen14: Okay

cmd/startup/startup.go
```diff
@@ -165,6 +171,118 @@ func ShutdownSiglensServer() {
...
+	}
+
+	dependencyMatrix := CalculateDependencyMatrix(spans)
+	dependencyMatrixJSON, err := json.Marshal(dependencyMatrix)
+	if err != nil {
+		fmt.Println("Error marshaling dependency matrix:", err)
+		return
+	}
+	fmt.Println("Dependency Matrix:", string(dependencyMatrixJSON))
+}
```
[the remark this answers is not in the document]
Davleen14: Yes will do that

cmd/startup/startup.go
```diff
@@ -165,6 +171,118 @@ func ShutdownSiglensServer() {
...
+		return
+	}
+
+	dependencyMatrix := CalculateDependencyMatrix(spans)
+	dependencyMatrixJSON, err := json.Marshal(dependencyMatrix)
+	if err != nil {
+		fmt.Println("Error marshaling dependency matrix:", err)
+		return
+	}
+	fmt.Println("Dependency Matrix:", string(dependencyMatrixJSON))
```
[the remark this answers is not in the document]
Davleen14: Yes doing that

cmd/startup/startup.go
```diff
@@ -40,6 +40,7 @@ import (
 	"github.com/siglens/siglens/pkg/ssa"
 	"github.com/siglens/siglens/pkg/usageStats"
 	usq "github.com/siglens/siglens/pkg/usersavedqueries"
+
```
AndrewHess, line 43: Don't add this line

pkg/segment/tracing/handler/tracehandler.go
```diff
@@ -365,3 +365,83 @@ func redMetricsToJson(redMetrics structs.RedMetrics, service string) ([]byte, er
 	result["p99"] = redMetrics.P99
 	return json.Marshal(result)
 }
+
+func DependencyGraphThread() {
+	time.Sleep(1 * time.Minute) // Wait for initial traces ingest first
+	for {
+		MakeTracesDependancyGraph()
+		time.Sleep(1 * time.Hour)
+	}
```
AndrewHess, line 374: Can you change this so that it computes the graph the first time after one minute (like you have it), but then every time after that it computes the graph at the top of the hour?

pkg/segment/tracing/handler/tracehandler.go
```diff
@@ -365,3 +365,83 @@ func redMetricsToJson(redMetrics structs.RedMetrics, service string) ([]byte, er
...
+}
+
+func MakeTracesDependancyGraph() {
+	nowTs := putils.GetCurrentTimeInMs()
+	startEpoch := nowTs - (60 * 60 * 1000)
```
AndrewHess, line 379: You're using a one-hour magic number here and above in `time.Sleep(1 * time.Hour)`. You should instead make a constant at the top of this file with a value of one hour and then read that constant for determining the `startEpoch` and how long to sleep. That way it's easier to change this later if we want to.

AndrewHess (commented):
Looks pretty good, just a couple minor suggestions

AndrewHess (approved):

PR #73, opened 2023-11-16
Fix ingestion issue

AndrewHess:
This fixes an issue where siglens would crash when a single block had a column where both string and numeric values were being ingested. In particular, this caused siglens to crash when ingesting example traces when following the instructions in #38 but now you can follow those instructions to ingest example traces and siglens handles it properly.

Now when records are being added to a block, different records can have a string or numeric value for the same column. When the block is flushed, we try to convert all those values to numbers but if any of those conversions fail we abort and instead convert all the values to strings.

### DIFF
--- a/pkg/segment/writer/segstore.go
+++ b/pkg/segment/writer/segstore.go
@@ -270,6 +270,7 @@ func convertColumnToNumbers(wipBlock *WipBlock, colName string) bool {
 				copy(newColWip.cbuf[newColWip.cbufidx:], utils.VALTYPE_ENC_INT64[:])
 				copy(newColWip.cbuf[newColWip.cbufidx+1:], toputils.Int64ToBytesLittleEndian(intVal))
 				newColWip.cbufidx += 1 + 8
+				addIntToRangeIndex(colName, intVal, wipBlock.columnRangeIndexes[colName].Ranges)
 				continue
 			}
 
@@ -280,6 +281,7 @@ func convertColumnToNumbers(wipBlock *WipBlock, colName string) bool {
 				copy(newColWip.cbuf[newColWip.cbufidx:], utils.VALTYPE_ENC_FLOAT64[:])
 				copy(newColWip.cbuf[newColWip.cbufidx+1:], toputils.Float64ToBytesLittleEndian(floatVal))
 				newColWip.cbufidx += 1 + 8
+				addFloatToRangeIndex(colName, floatVal, wipBlock.columnRangeIndexes[colName].Ranges)
 				continue
 			}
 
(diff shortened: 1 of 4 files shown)


pkg/segment/writer/segstore.go
```diff
@@ -220,7 +221,174 @@ func (segstore *SegStore) resetSegStore(streamid string, virtualTableName string
...
+			floatVal := toputils.BytesToFloat64LittleEndian(oldColWip.cbuf[i : i+8])
+			i += 8
+
+			stringVal := strconv.FormatFloat(floatVal, 'f', -1, 64)
+			newColWip.WriteSingleString(stringVal)
+			bloom.uniqueWordCount += addToBlockBloom(bloom.Bf, []byte(stringVal))
+
+		case utils.VALTYPE_ENC_BACKFILL[0]:
+			// This is a null value.
+			newColWip.WriteSingleString("")
```
nkunal, line 358: for backfill, we don't write any actual value ? we just store the valType, so we just need to copy the VALTYPE_...BACKFILL (just that 1 byte)

pkg/segment/writer/segstore.go
```diff
@@ -220,7 +221,174 @@ func (segstore *SegStore) resetSegStore(streamid string, virtualTableName string
...
+			bloom.uniqueWordCount += addToBlockBloom(bloom.Bf, []byte(""))
```
nkunal, line 359: we don't need to add the "" to the bloom for backfill

pkg/segment/writer/segstore.go
```diff
@@ -220,7 +221,174 @@ func (segstore *SegStore) resetSegStore(streamid string, virtualTableName string
...
+			var stringVal string
+			if boolVal == 0 {
+				stringVal = "false"
+			} else {
+				stringVal = "true"
+			}
+
+			newColWip.WriteSingleString(stringVal)
+			bloom.uniqueWordCount += addToBlockBloom(bloom.Bf, []byte(stringVal))
+
```
nkunal, line 375: have we copied the VALTYPE_ENC_BOOL in the newColWip ?

nkunal (commented):
Looks fantastic, just had a couple of questions/scenarios that we should take care of.

pkg/segment/writer/segstore.go
AndrewHess: No I was writing the string `"true"` or `"false"` here. Should we allow strings and booleans in the same field? If so I can copy this over as a boolean instead, and it should already be in the bloom.

nkunal: for booleans we are planning to have a different type of microindex (other than bloom, range) sometime in the future. For now we can keep it the way as the usual booleans are handled

AndrewHess: Ok, so are you saying since we currently put booleans in the bloom, we can store booleans and strings together for now so I should change this to encode them as booleans like we do in `encodeSingleBool()`?

nkunal (approved):

PR #87, opened 2023-11-17
Fix search traces issues

asiL-tcefreP:
Fix search traces issues


AndrewHess (commented):
Why is this necessary? What issue does this fix?

asiL-tcefreP:
> Why is this necessary? What issue does this fix?

I modified the description of this PR.

AndrewHess:
Why does it ever return 0 results? Shouldn't it always return at least one result because the search should match at least the root span even if it's just a 1-span trace?

asiL-tcefreP:
> Why does it ever return 0 results? Shouldn't it always return at least one result because the search should match at least the root span even if it's just a 1-span trace?

I still need more time to fix this issue. Let me optimize search traces and to see if this bug will reproduce again

PR #88, opened 2023-11-17
Remove P95 for Red Metrics

asiL-tcefreP:
Remove P95 for Red Metrics


AndrewHess:
@nkunal did you want us to remove the P95 entirely from the backend, or just have the frontend ignore it for the service health overview tab but show it for the service health details graphs? Are there any other quantiles we should calculate?

nkunal:
> @nkunal did you want us to remove the P95 entirely from the backend, or just have the frontend ignore it for the service health overview tab but show it for the service health details graphs? Are there any other quantiles we should calculate?

Just front end should ignore it.
The current quantiles are fine

PR #91, opened 2023-11-20
added install_with_binary.sh , fix darwin binary

sunitakawane:


install_with_binary.sh
```diff
@@ -0,0 +1,63 @@
...
+# Step 2: Detect the latest version from siglens release
+display_step 2 "Getting lastest version of siglens release"
+
+latest_version=`\
+    curl  --silent "https://api.github.com/repos/siglens/siglens/releases/latest" |
+    grep '"tag_name":' |
+    sed -E 's/.*"([^"]+)".*/\1/'`
+
+# Step 3: Fetching latest binary based on OS and Arch
+display_step 3 "fetching latest binary for $os-$arch and version $latest_version"
```
AndrewHess, line 53: nit: `fetching` is not capitalized, but the first word of your other Step messages are capitalized

install_with_binary.sh
```diff
@@ -0,0 +1,63 @@
...
+     ;;
+esac
+
+arch=$(uname -m)
+if [[ $arch == x86_64* ]]; then
+    arch="amd64"
+elif  [[ $arch == arm* || $arch == "aarch64" ]]; then
+    arch="arm64"
+else
+    echo 'Not Supported Architecture'
```
nkunal, line 41: this log line should be `arch {value_of_arch} not Supported`

install_with_binary.sh
```diff
@@ -0,0 +1,63 @@
...
+
+# Step 3: Fetching latest binary based on OS and Arch
+display_step 3 "fetching latest binary for $os-$arch and version $latest_version"
+
+wget "https://github.com/siglens/siglens/releases/download/$latest_version/siglens-$latest_version-$os-$arch.tar.gz"
+tar -xvf "siglens-$latest_version-$os-$arch.tar.gz"
+PORT=80 
+display_step 4 "Running the Server on http://localhost:$PORT"
+if [ $PORT == 80 ]; then 
+   echo "If you are not able to access localhost:80 to changing the PORT=8090"
```
nkunal, line 60: nit: change comment to `... try setting the PORT variable to ...`

nkunal (approved):

install_with_binary.sh
AndrewHess, line 60: nit: it might be clearer to say something like `... try running "export PORT=8090" and rerunning this script` so they can copy-paste the command

install_with_binary.sh
```diff
@@ -0,0 +1,63 @@
...
+esac
+
+arch=$(uname -m)
+if [[ $arch == x86_64* ]]; then
+    arch="amd64"
+elif  [[ $arch == arm* || $arch == "aarch64" ]]; then
+    arch="arm64"
+else
+    echo 'Not Supported Architecture'
+fi
```
AndrewHess, line 42: Probably add `exit 1` here

install_with_binary.sh
```diff
@@ -0,0 +1,63 @@
...
+   Linux\ Mint*)
+     os="linux" ;;
+   Red\ Hat*)
+     os="linux" ;;
+   CentOS*)
+     os="linux" ;;
+   *)
+     os="Not Found: $os_name"
+     echo 'Not Supported OS'
+    exit 1
```
AndrewHess, line 31: nit: `exit 1` is not indented the same as the other lines

install_with_binary.sh
```diff
@@ -0,0 +1,63 @@
...
+    sed -E 's/.*"([^"]+)".*/\1/'`
+
+# Step 3: Fetching latest binary based on OS and Arch
+display_step 3 "fetching latest binary for $os-$arch and version $latest_version"
+
+wget "https://github.com/siglens/siglens/releases/download/$latest_version/siglens-$latest_version-$os-$arch.tar.gz"
+tar -xvf "siglens-$latest_version-$os-$arch.tar.gz"
+PORT=80 
+display_step 4 "Running the Server on http://localhost:$PORT"
+if [ $PORT == 80 ]; then
```
AndrewHess, line 59: So `PORT` is always 80? Should you instead try reading from the `PORT` environment variable but if there's nothing there use 80 as the default?

install_with_binary.sh
```diff
@@ -0,0 +1,63 @@
...
+   Debian*)
+     os="linux" ;;
+   Linux\ Mint*)
+     os="linux" ;;
+   Red\ Hat*)
+     os="linux" ;;
+   CentOS*)
+     os="linux" ;;
+   *)
+     os="Not Found: $os_name"
```
AndrewHess, line 29: I don't see `os_name` defined anywhere. Is it set somehow by this switch statement?

install_with_binary.sh
```diff
@@ -0,0 +1,63 @@
...
+
+wget "https://github.com/siglens/siglens/releases/download/$latest_version/siglens-$latest_version-$os-$arch.tar.gz"
+tar -xvf "siglens-$latest_version-$os-$arch.tar.gz"
+PORT=80 
+display_step 4 "Running the Server on http://localhost:$PORT"
+if [ $PORT == 80 ]; then 
+   echo "If you are not able to access localhost:80 to changing the PORT=8090"
+fi
+
+PORT=$PORT "siglens-$latest_version-$os-$arch/siglens" --config "siglens-$latest_version-$os-$arch/server.yaml"
```
AndrewHess, line 63: Do you need to prepend this with `env` so that it's `env PORT=$PORT ...` or will this work as is?

install_with_binary.sh
```diff
@@ -0,0 +1,63 @@
...
+
+latest_version=`\
+    curl  --silent "https://api.github.com/repos/siglens/siglens/releases/latest" |
+    grep '"tag_name":' |
+    sed -E 's/.*"([^"]+)".*/\1/'`
+
+# Step 3: Fetching latest binary based on OS and Arch
+display_step 3 "fetching latest binary for $os-$arch and version $latest_version"
+
+wget "https://github.com/siglens/siglens/releases/download/$latest_version/siglens-$latest_version-$os-$arch.tar.gz"
```
AndrewHess, line 55: nit: you might want to check the return code from `wget` in case there's an internet issue

install_with_binary.sh
sunitakawane: this works as is

install_with_binary.sh
sunitakawane: I need to look into this. Is it ok if I address this in another PR? We have similar code in install_with_git script, so we will have to change in both places.

AndrewHess (approved):
It looks like you made the change to the line "If you are not able to access localhost:80, ..." in install_with_git.sh but not in install_with_binary.sh

Otherwise, this looks good to me

nkunal (approved):

PR #117, opened 2023-11-21
feat: querying for boolean values

### LINKED ISSUE #103: Querying of boolean values
# Description
Currently, any query with true will always consider it a string and will try to search accordingly.

We should also support for true and false queries to match boolean values as well. We can have true match both string and bool .

We should make necessary edits to segsearch so that we can query boolean values.

ParichayDidwania:
# Description
Summarize the change.

Fixes #103 (link all the GitHub issues this addresses)

# Testing

I added a new column in sigscalr-client (reader.go) which has random boolean values. I then ingested the generated data to the running server and tested the query like so, <column-name>=true (without quotes)

# Checklist:

### DIFF
--- a/pkg/ast/spl/spl.peg
+++ b/pkg/ast/spl/spl.peg
@@ -393,7 +393,7 @@ ClauseLevel1 <- L_PAREN clause:ClauseLevel4 R_PAREN {
 
 // Parsing the text of a Number as a String would succeed, so we need to try
 // parsing as a FieldWithNumberValue first.
-SearchTerm <- term:(FieldWithNumberValue / FieldWithStringValue) {
+SearchTerm <- term:(FieldWithNumberValue / FieldWithBooleanValue / FieldWithStringValue) {
     return term, nil
 }
 
@@ -1858,6 +1858,19 @@ UnnamedFieldWithNumberValue <- value:Number {
     return node, nil
 }
 
+FieldWithBooleanValue <- key:FieldName op:EqualityOperator value:Boolean {
+    node := &ast.Node{
+        NodeType: ast.NodeTerminal,
+        Comparison:ast.Comparison{
+            Op: op.(string),
+            Field: key.(string),
+            Values: value,
+        },
+    }
+
+    return node, nil
+}
+
 FieldWithStringValue <- keyValuePair:(NamedFieldWithStringValue / UnnamedFieldWithStringValue) {
     return keyValuePair, nil
 }
@@ -1919,6 +1932,11 @@ String <- str:(QuotedString / UnquotedString) {
 QuotedString <- '"' [^"]* '"' {
     // The returned string has quotes as the first and last character.
     return string(c.text), nil
+}
+
+Boolean <- ("true" / "false") {
+    boolValue, _ := strconv.ParseBool(string(c.text))
+    return boolValue, nil
 }
 
 UnquotedString <- ("*" / (!(MAJOR_BREAK / EOF) .))+ {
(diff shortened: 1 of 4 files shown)


pkg/ast/spl/spl.peg
```diff
@@ -1565,7 +1565,11 @@ QuotedString <- '"' [^"]* '"' {
     return string(c.text), nil
 }
 
-UnquotedString <- ("*" / (!(MAJOR_BREAK / EOF) .))+ {
+UnquotedString <- ("true" / "false") {
+    boolValue, _ := strconv.ParseBool(string(c.text))
+    return boolValue, nil
+}
+/ ("*" / (!(MAJOR_BREAK / EOF) .))+ {
```
AndrewHess, line 1572: The `UnquotedString` rules should return a string, so you'll need to move this logic to a different rule (or possibly make a new rule) that returns a boolean. Then you may need to update where that rule can appear as part of a different rule so that when we get a search like `isOpen=true` we parse `true` as a boolean but if we get something like `isOpen=tomorrow` we parse `tomorrow` as a string. In PEG grammars, the first option of a rule that matches succeeds, so this means if you had a rule like:
```
FieldAndValue <- field:FieldName "=" value:(Boolean / String) {
    ...
}
```
It's important that the order is `Boolean / String` because if it was `String / Boolean` then when you got `isOpen=true` it would try parsing `true` as a string and that would succeed, so it would never try to parse it as a boolean

ParichayDidwania: Got it, I will make the changes as soon as possible.

nbhavana: @ParichayDidwania Any update on this PR?

ParichayDidwania: @nbhavana I will try and finish it by eod

ParichayDidwania: @nbhavana please check now

AndrewHess (commented):
This is looking good. I ingested some sample boolean data and the queries seemed to work correctly.

Can you also add some unit tests in pkg/ast/spl/tests/splParser_test.go to make sure that the query is getting parsed how we expect it to?

ParichayDidwania:
@AndrewHess I added the unit test and tested it as well. Please check the commit.

pkg/ast/spl/tests/splParser_test.go
```diff
@@ -262,6 +262,40 @@ func Test_searchFieldGreaterThanOrEqualToQuotedString(t *testing.T) {
...
+
+	assert.NotNil(t, filterNode)
+	assert.Equal(t, "=", filterNode.Comparison.Op)
+	assert.Equal(t, "status", filterNode.Comparison.Field)
+	assert.Equal(t, true, filterNode.Comparison.Values)
+
+	expressionFilter := extractExpressionFilter(t, filterNode)
+	assert.Equal(t, "status", expressionFilter.LeftInput.Expression.LeftInput.ColumnName)
+	assert.Equal(t, utils.Equals, expressionFilter.FilterOperator)
+	assert.Equal(t, "true", expressionFilter.RightInput.Expression.LeftInput.ColumnValue.StringVal)
```
AndrewHess, line 279: The value should be a boolean, so you should probably assert the `ColumnValue.BoolVal` instead of the `ColumnValue.StringVal`. You can also check that the `ColumnValue.Dtype` is `SS_DT_BOOL`

Same comment for line 296

ParichayDidwania: @AndrewHess i made the changes

AndrewHess (approved):
Looks good

PR #160, opened 2023-11-27
Service health one service

Davleen14:
Red Metrics Graphs for services


sonamgupta21:
1. The view currently opens only when the user clicks on the service column of the row, but it should open when the user clicks anywhere on the row.
2. Time-picker is not working correctly.
3. The Graph view should open in the same window/tab.

static/js/service-health-graphs.js
```diff
@@ -0,0 +1,229 @@
...
+   
+}
+var RateCountChart;
+var ErrCountChart;
+var LatenciesChart;
+$(document).ready(() => {
+    displayNavbar();
+    setupEventHandlers();
+    $(".theme-btn").on("click", themePickerHandler);
+    $('.theme-btn').on('click',getAllService() );
```
sonamgupta21, line 14: Remove this line.

sonamgupta21 (approved):

PR #184, opened 2023-11-30
Use Segment to record install statistics

asiL-tcefreP:
# Description
Add install statistics logic in .sh file. Modify user id to computer-specific identifier in ssa.go

# Testing

# Checklist:


install_with_binary.sh
```diff
@@ -58,9 +58,36 @@ if [[ $? -ne 0 ]]; then
     echo "wget failed to get latest binary from $url"
     exit 1; 
 fi
+
+# Step 4: Install statistics
+display_step 4 "Install statistics"
+computer_specific_identifier=$(ifconfig 2>/dev/null | grep -o -E '([0-9a-fA-F]{2}:){5}([0-9a-fA-F]{2})' | head -n 1)
```
AndrewHess, line 64: Can you add a comment saying what this is getting?

AndrewHess (commented):
I think we don't want the Basic auth token in version control, but I'm not sure how else we would make this work. Any thoughts on this @sunitakawane or @nkunal?

Also it looks like your logic is the same for each of the install scripts, so you should probably have this in a separate script that you call from each of the install scripts, so that if we want to change it later we'd just have to make one change instead of three identical changes. However, if you expect that we'll need different logic for the different installation methods you can keep them separate like you have them

asiL-tcefreP:
Yes, their logic is the same. The reason I didn't separate this code into a standalone file initially was that I thought users would only download a single .sh file instead of two. If this doesn't have a significant impact, I will extract the logic of the code into a separate .sh file.

AndrewHess:
> Yes, their logic is the same. The reason I didn't separate this code into a standalone file initially was that I thought users would only download a single .sh file instead of two. If this doesn't have a significant impact, I will extract the logic of the code into a separate .sh file.

Oh, never mind. Yes you should keep it how you have it so they only have to download one file

AndrewHess (approved):

PR #191, opened 2023-12-06
Dp UI error

Davleen14:
If the data is being ingested then the dependency graph will be {} and UI will display the msg "Data is being ingested"


sunitakawane:
fixed this as part of another [PR](https://github.com/siglens/siglens/pull/196)

Issue #132, opened 2023-11-23
SPL: Allow eval statements in aggregate functions


sunitakawane:
# Description

For example, allow `... | stat count(eval(<field>=<value>))`

PR #193, opened 2023-12-06
Spl aggregate eval support

Jiale-Fang:
# Description

Please notice that: this pr is not ready to be merged, but can be reviewed

Link this issue: https://github.com/siglens/siglens/issues/132
Allow eval statements in aggregate functions (just finished functionalities for agg without group by cols). It should work for all of the aggregate functions: `Count / DistinctCount / Avg / Min / Max / Range / Sum / Values`)

For eval statements inside aggregate functions:
1. `min, max, sum, avg, and range` can only evaluate boolComparisonExpr(`field-name><comparison-operator><value>`), as they need to compute the field value inside the expression. There should be only one field.
2. `count` can evaluate boolExpr because it does not care about the field value inside the expression. It only cares about whether the current row should be counted or not.
3. `distinctCount and values` can evaluate valueExpr because they just need to find out the distinct values of the result of the expression. As for the type of the expression result, it is not important.

# Testing
Will add test cases after finishing all the functionalities.

# Checklist:


pkg/segment/structs/segstructs.go
```diff
@@ -438,6 +451,26 @@ func (qa *QueryAggregators) HasQueryAggergatorBlockInChain() bool {
...
+
+// To determine whether it contains Aggregate Func: Values()
+func (qa *QueryAggregators) HasValuesFunc() bool {
+	for _, agg := range qa.MeasureOperations {
+		if agg.MeasureFunc == utils.Values {
+			return true
+		}
+	}
+	return false
+}
```
AndrewHess, line 472: Why do we need to know if it has specifically the `Values()` aggregation?

AndrewHess (commented):
Looks good so far

pkg/segment/structs/segstructs.go
asiL-tcefreP: In the `func applyAggOpOnSegments()`, Values func is similar to Eval statements. They all need to recompute the raw data to get the results, since segment only store statistical data such as min, max... 

So i use this condition to judge if current QueryAggregators should be recomputed or not

pkg/ast/spl/spl.peg
```diff
@@ -1386,6 +1402,10 @@ Aggregator <- aggFunc:AggFunction asField:AsField? {
     return agg, nil
 }
 
+// For evaluation statements inside aggregate functions:
+// 1. min, max, sum, avg, and range can only evaluate boolComparisonExpr, as they need to compute the field value inside the expression. There should be only one field.
+// 2. count can evaluate boolExpr because it does not care about the field value inside the expression. It only cares about whether the current row should be counted or not.
+// 3. distinctCount and values can evaluate valueExpr because they just need to find out the distinct values of the result of the expression. As for the type of the expression result, it is not important.
```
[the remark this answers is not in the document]
asiL-tcefreP: I google it but I can only find examples in here [Use stats with eval expressions and functions](https://docs.splunk.com/Documentation/SplunkCloud/9.1.2308/Search/Usestatswithevalexpressionsandfunctions)

In the first example, 
`status=* | stats dc(eval(if(status=404, clientip, NULL()))) AS dc_ip_errors`

In here, you can see the eval statements will return a string or number value. Which might has some difference with min, max, etc. Since they just support `boolComparisonExpr`

pkg/segment/structs/segstructs.go
AndrewHess: Ok that makes sense. Can we not reuse whatever logic we have for computing the `Cardinality` but get the actual values rather than just the number of values?

asiL-tcefreP:
I have already test Aggregate functions which has group by cols for these query:
```
* | stats min(latitude), min(eval(latitude < 0)), range(latitude), max(eval(latitude < 0)), range(eval(latitude < 0)), range(eval(latitude >= 0)) BY weekday // min, max, range

* | stats max(eval(http_status < 400)), sum(http_status), sum(eval(http_status < 500)), sum(eval(http_status = 500)) BY weekday // sum

* | stats count(eval(http_status >= 100 AND latitude > 0 AND http_status < 500)), count(http_status), count(eval(http_status < 100 OR latitude <= 0 OR http_status >= 500)) BY weekday // count

* | stats sum(eval(latitude < 0)), count(eval(latitude < 0)), avg(eval(latitude < 0)) BY state // avg

* | stats values(app_name), values(eval(lower(app_name))), dc(app_name), dc(eval(lower(app_name))) BY city // cardinality and values

* | stats values(eval(if(len(state) > 5, job_title, city))) BY state
```

I will test more cases for aggs without group by cols and write test cases for this pr.

asiL-tcefreP:
The commits above have solve these issues:
- Fix: agg Values() func cannot be implemented in columns that yield numeric results (e.g., latitude).
- Fix some issues when multiple aggregates are used together
- Fix: agg max() func cannot evaluate negative values correctly.

Optimizations:
- Previously, in a query, if one aggregate column had an eval statement, all columns would add raw data to segstats, which is a waste of performance. Currently, only columns with eval statements add raw data to segstats.
- Implement a similar optimization for aggregate values func().

I have already test Aggregate functions without group by cols for these query:
```
city=Boston | stats max(latitude), range(eval(latitude >= 0)), min(eval(latitude >= 0)) // min, max, range

* | stats sum(longitude), max(eval(latitude < 0)), sum(eval(longitude < 0)), sum(eval(longitude >= 0)) // sum

* | stats sum(eval(latitude < 0)), count(eval(latitude < 0)), avg(eval(latitude < 0)) // avg

* | stats dc(app_name), values(app_name), dc(eval(state)), values(state), values(eval(lower(app_name))), dc(eval(lower(app_name))) // dc values

city=Boston | stats max(latitude), range(eval(latitude >= 0)), min(eval(latitude >= 0)), sum(longitude), sum(eval(longitude < 0)), sum(eval(longitude >= 0)), sum(eval(latitude < 0)), count(eval(latitude < 0)), avg(eval(latitude < 0)), dc(app_name), values(app_name), dc(eval(lower(state)))
```

I will push again after i finished all the unit tests.

AndrewHess (approved):
Looks good!

PR #196, opened 2023-12-07
tracing-change dependancy graph logic

sunitakawane:
# Description
When request comes in from UI, Get dependency graph by querying traces directly.
Added check to fix runtime error in ProcessRedTracesIngest

# Testing

# Checklist:

### DIFF
--- a/pkg/segment/tracing/handler/tracehandler.go
+++ b/pkg/segment/tracing/handler/tracehandler.go
@@ -543,7 +543,7 @@ func ProcessDependencyRequest(ctx *fasthttp.RequestCtx, myid uint64) {
 			pipesearch.SetBadMsg(ctx)
 			return
 		}
-
+		writeDependencyMatrix(depMatrix)
 		for key, value := range depMatrix {
 			for k, v := range value {
 				if processedData[key] == nil {
--- a/pkg/segment/tracing/handler/tracehandler.go
+++ b/pkg/segment/tracing/handler/tracehandler.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"regexp"
+	"strings"
 	"time"
 
 	jsoniter "github.com/json-iterator/go"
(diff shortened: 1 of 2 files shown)


pkg/segment/tracing/handler/tracehandler.go
```diff
@@ -515,44 +524,27 @@ func MakeTracesDependancyGraph() {
...
-	pipeSearchResponseOuter, err := processSearchRequest(searchRequestBody, myid)
-	if err != nil {
-		log.Errorf("ProcessSearchRequest: %v", err)
-		return
-	}
 	processedData := make(map[string]interface{})
-	if pipeSearchResponseOuter.Hits.Hits == nil || len(pipeSearchResponseOuter.Hits.Hits) == 0 {
+	depMatrix := MakeTracesDependancyGraph()
+	if len(depMatrix) == 0 {
 		log.Errorf("pipeSearchResponseOuter: received empty response")
```
AndrewHess, line 531: nit: replace `pipeSearchResponseOuter` with `ProcessDependencyRequest`

pkg/segment/tracing/handler/tracehandler.go
```diff
@@ -515,44 +524,27 @@ func MakeTracesDependancyGraph() {
...
-	searchRequestBody.SearchText = "*"
 
-	pipeSearchResponseOuter, err := processSearchRequest(searchRequestBody, myid)
-	if err != nil {
-		log.Errorf("ProcessSearchRequest: %v", err)
-		return
-	}
 	processedData := make(map[string]interface{})
-	if pipeSearchResponseOuter.Hits.Hits == nil || len(pipeSearchResponseOuter.Hits.Hits) == 0 {
+	depMatrix := MakeTracesDependancyGraph()
```
AndrewHess, line 529: This would make it so we never use our precomputed dependency graph. The reason for precomputing it is that computing it may take a while when there's lots of spans. Maybe we should change this so that if we have no precomputed graph then we compute it in this request (and write it to the index so we could use it for future requests), but if we do have precomputed graphs we just return the most recent precomputed one.

AndrewHess (approved):

PR #210, opened 2023-12-13
Add Gantt chart endpoint

asiL-tcefreP:
# Description
Add Gantt chart endpoint

# Testing

# Checklist:


pkg/segment/tracing/handler/tracehandler.go
```diff
@@ -587,3 +587,143 @@ func ProcessDependencyRequest(ctx *fasthttp.RequestCtx, myid uint64) {
...
+
+			operationName, exists := spanMap["name"]
+			if !exists {
+				continue
+			}
+
+			parentSpanId, exists := spanMap["parent_span_id"]
+			if !exists {
+				continue
+			}
```
AndrewHess, line 709: You should probably log an error when these fields don't exist, unless there's a reason not to

pkg/segment/tracing/handler/tracehandler.go
```diff
@@ -587,3 +587,143 @@ func ProcessDependencyRequest(ctx *fasthttp.RequestCtx, myid uint64) {
...
+
+			idToParentId[span.SpanID] = parentSpanId.(string)
```
AndrewHess, line 711: Does this cause issues for the root span, since it doesn't have a value for `parentSpanId`?

pkg/segment/tracing/handler/tracehandler.go
AndrewHess, line 709: Won't this exclude the root span? I don't think we want to exclude it

pkg/segment/tracing/handler/tracehandler.go
```diff
@@ -587,3 +587,143 @@ func ProcessDependencyRequest(ctx *fasthttp.RequestCtx, myid uint64) {
...
+
+			// Remove all attributes that are not belong to tags, leaving only the attributes that are tags
```
AndrewHess, line 713: nit: can change this to just `Remove all non-tag fields`

pkg/segment/tracing/utils/buildspantree.go
```diff
@@ -0,0 +1,31 @@
+package utils
+
+import (
+	"github.com/siglens/siglens/pkg/segment/tracing/structs"
+	log "github.com/sirupsen/logrus"
+)
+
+func BuildSpanTree(spanMap map[string]*structs.GanttChartSpan, idToParentId map[string]string) []*structs.GanttChartSpan {
+	results := make([]*structs.GanttChartSpan, 0)
```
AndrewHess, line 9: Shouldn't `results` just have one span (the root span) with all the other spans nested below that? So then the return type can be `*structs.GanttChartSpan` instead of `[]*structs.GanttChartSpan`

AndrewHess (commented):
You should also update `API_README.md` to document this new endpoint

pkg/segment/tracing/utils/buildspantree.go
asiL-tcefreP: Ok, I will modify that. Since the search text for gantt chart endpoint for now is fixed: `trace_id=...`

pkg/segment/tracing/handler/tracehandler.go
asiL-tcefreP: I use `idToParentId` in the `func BuildSpanTree`. If  `parentSpanId ==""`, which means we find a root span.

pkg/segment/tracing/handler/tracehandler.go
asiL-tcefreP: I don't think I will exclude the root span because the attribute `parent_span_id` exists in the resultMap; it just has a value of `""`. In the subsequent checks, I will determine that if `parent_span_id == ""`, it indicates that it is a root span.

API_README.md
```diff
@@ -621,7 +621,155 @@ For Gantt chart data specific to a trace ID, modify the request body accordingly
...
+                "operation_name": "/",
+                "tags": {
+                    "_index": "traces",
+                    "app.ads.ad_request_type": null,
+                    "app.ads.ad_response_type": null,
+                    "app.ads.category": null,
+                    "app.ads.contextKeys": null,
+                    "app.ads.contextKeys.count": null,
+                    "app.ads.count": null,
+                    "app.cart.items.count": null,
```
AndrewHess, line 653: A few things for the `tags`:
1. You can remove the `_index` since that's just from siglens and not part of the original span. I think you can also remove the `timestamp` below (line 760) for the same reason, but double check me on that
2. You should remove all the tags whose value is null.

Also update the code to make sure these changes happen when you actually call the endpoint

AndrewHess (commented):
Looks good, I just noticed one more thing. Whenever you get an error in `ProcessGanttChartRequest` that can't be recovered from, you should let the caller know just before you return. So do something like `SetBadMsg(ctx)` or `ctx.SetStatusCode(fasthttp.StatusBadRequest)`. You did this in some places but not in others.

AndrewHess (approved):

PR #217, opened 2023-12-15
Fix status code

Davleen14:
Fixed status code panic error in traces


pkg/otlp/otlp.go
```diff
@@ -138,7 +138,11 @@ func spanToJson(span *tracepb.Span, service string) ([]byte, error) {
 	result["dropped_attributes_count"] = uint64(span.DroppedAttributesCount)
 	result["dropped_events_count"] = uint64(span.DroppedEventsCount)
 	result["dropped_links_count"] = uint64(span.DroppedLinksCount)
-	result["status"] = span.Status.Code.String()
+	if span.Status != nil {
+		result["status"] = span.Status.Code.String()
```
AndrewHess, line 142: Just to be extra safe, can you make the condition `span.Status != nil && span.Status.Code != nil`?

Davleen14: Yes I will do that

AndrewHess: Oh that's my bad. In that case it's good how it is

AndrewHess (approved):

PR #224, opened 2023-12-17
FIX: Allowing creation of multiple dashboards with the same name

### LINKED ISSUE #223: [BUG] Handling Creation of Multiple Dashboards with the Same Name
# Description
We are able to create multiple dashboards having the same name, the current logic does not check for the uniqueness of the dashboard name while creating the dashboard. 
image

# Steps to Reproduce
Create a new dashboard with a certain name.
Create another or multiple such dashboards with the same name.
Hit the /listall API to verify the creation of these dashboards, note that they have the same name but different dashboard IDs.

# Expected Behavior
Should not allow the creation of multiple dashboards with the same name.

venkateshamatam:
# Description
Fixed an issue where multiple dashboards could be created with the same name. With this PR, it is not possible to create a dashboard with the same name.
Fixes #223 

# Testing
I added functionality to check  if there exists a dashboard with the same name as the one being created now. I tested the below functionalities thoroughly - 
1. Creating dashboard manually from dashboard page with same and different names
2. Creating dashboard from logs page with same and different names
3. 409 status code(conflict) to prevent the dashboard from being created
4. Display appropriate message on the UI when user tries to create a dashboard with same name 
image

# Checklist:

[x] I have self-reviewed this PR.
[x] I have removed all print-debugging and commented-out code that should not be merged.
[x] I have added sufficient comments in my code, particularly in hard-to-understand areas.
[x] I have formatted the code, if applicable. For Go, I have run goimports -w ..

This PR contains commits from old PRs since its unmerged and my branch is up-to-date to avoid merge conflicts. Please refer to this commit  for the above changes. (https://github.com/siglens/siglens/commit/2ee2d5bc3cacf97fae4917d4fdbd9d7b04b22ff7)


nbhavana (commented):
lgtm. @AndrewHess can you take a look?

pkg/dashboards/dashboards.go
```diff
@@ -215,6 +215,21 @@ func createUniqId(dname string) string {
 	return newId
 }
 
+// method to check if the dashboard name already exists
+func dashboardNameExists(dname string, orgid uint64) bool {
+	allDashboardIds, err := getAllDashboardIds(orgid)
+	if err != nil {
+		log.Printf("Error getting all dashboard IDs: %v", err)
```
AndrewHess, line 222: nit: use `log.Errorf` here and prepend the function name to the error message, just so it's easier to track down where the log came from if it gets emitted.

pkg/dashboards/dashboards.go
```diff
@@ -223,6 +238,11 @@ func createDashboard(dname string, orgid uint64) (map[string]string, error) {
 
 	newId := createUniqId(dname)
 
+	if dashboardNameExists(dname, orgid) {
+		log.Printf("Dashboard with name %s already exists", dname)
```
AndrewHess, line 242: nit: same as above comment; use something like `log.Errorf("createDashboard: Dashboard with name %s already exists", dname)` instead

AndrewHess (commented):
I'm seeing an odd bug when I run this code. If I make a new dashboard with a new name (say, `Dash1`), it gets created as expected. Then if I try to make another dashboard with that same name and click Save, the popup closes but I see "Dashboard name already exists" on that dashboard page (see the image). Then if I again try to make `Dash1` any number of times, I correctly see the error when I try to click Save, and the popup doesn't close when I click Save since it can't make the dashboard.
image

venkateshamatam:
Debugging this right away

venkateshamatam:
@AndrewHess, I'm able to replicate the above issue when I press enter instead of save, it looks like the pop up is being hidden on pressing enter but the error-tip stays. Could you confirm if it happens on pressing save as well?

AndrewHess:
You're right, I get this bug when I press enter but it works as expected when I click Save

venkateshamatam:
@AndrewHess, I just pushed changes to fix this issue. Could you please confirm if it works as expected?

AndrewHess:
Yep it works correctly now. Once you address my comments above about using `log.Errorf` instead of `log.Printf` I think this'll be good to merge

venkateshamatam:
Thanks @AndrewHess, I have committed the updated log statements.

AndrewHess (approved):

PR #216, opened 2023-12-15
Show error if dashboard name is empty or space

PratikshaPatil77:
# Description
The user will not be able to input space when creating a new dashboard or updating an existing one since the dashboard name cannot be empty.When creating or editing a dashboard name, an error warning will appear if the user attempts to enter the dashboard name as a space.

Fixes #212 

# Testing
Describe how you tested this code. How can the reviewers reproduce your tests?

# Checklist:
Before marking your pull request as ready for review, complete the following.

- [ ] I have self-reviewed this PR.
- [ ] I have removed all print-debugging and commented-out code that should not be merged.
- [ ] I have added sufficient comments in my code, particularly in hard-to-understand areas.
- [ ] I have formatted the code, if applicable. For Go, I have run `goimports -w .`.

PR #225, opened 2023-12-18
FIX: Handling updating dashboard name to an existing dashboard name

venkateshamatam:
# Description
Fixed an issue where you could update the name of a dashboard to an already existing dashboard name, causing duplicate dashboard names.
Fixes https://github.com/siglens/siglens/issues/223

# Testing
I added functionality to check if there exists a dashboard with the same name as the one being created now. I tested the below functionalities thoroughly -

Creating/updating dashboard manually from dashboard page with same and different names
Creating dashboard from logs page with same and different names
409 status code(conflict) to prevent the dashboard from being updated
Display appropriate message on the UI when user tries to update a dashboard with an already existing dashboard name

image

# Checklist:

[x] I have self-reviewed this PR.
[x] I have removed all print-debugging and commented-out code that should not be merged.
[x] I have added sufficient comments in my code, particularly in hard-to-understand areas.
[x] I have formatted the code, if applicable. For Go, I have run goimports -w ..


AndrewHess:
When I run this code and make a dashboard `Dash1`, then make another `Dash2` and then edit `Dash2` and try renaming it to `Dash1`, it correctly disallows that, but it also takes me back to my previous screen as if the saving had worked
image
I think it makes more sense if when the saving fails, I stay on the same edit page so I can choose a different name for the dashboard

venkateshamatam:
Yes, I was just looking at this. It looks like the PR (https://github.com/siglens/siglens/pull/216/files) that was raised following this had some changes that affected the flow of this. I'm working on it now

venkateshamatam:
@AndrewHess, I've made the changes and it should now work as expected.

AndrewHess (approved):

PR #248, opened 2023-12-19
Sql round

### LINKED ISSUE #67: SQL: Add Round
We want to support using `ROUND` in SQL queries as described here: https://www.postgresqltutorial.com/postgresql-math-functions/postgresql-round/

We do SQL parsing in `pkg/ast/sql/astsql.go` but you may also need to look into `github.com/xwb1989/sqlparser` as we use that to do some of the parsing. You may also want to look at how we implemented `round()` for Splunk QL by parsing it in `pkg/ast/spl/spl.peg` and evaluating it in `pkg/segment/structs/evaluationstructs.go`, which we call in `pkg/segment/aggregations/segaggs.go`

Macbeth98:
feat: sql round for aggregation

# Description
- I have made changes in `astsql.go` and `segaggs.go` file to implement the SQL Round functionality. 
- I have followed the implementation of Splunk round function.
- I was able to implement the functionality for the aggregate functions like sum, avg, etc.
- But not with simple queries like `select city, round(latitude)`.
- Please Note: I will be making more changes, once I got some feedback. After that this PR can be merged.

### Edit
- Fixed the suggested error log changes.

Fixes #67  

# Testing
- I have not witten any tests yet. I have tested my functionality with the front-end siglens playground.
- Will write tests once I have made sure my approach is correct.

### Edit
- Added test case in `sqlparser_test.go`, testing the round function parsing with group by queries. 

# Checklist:


pkg/ast/sql/astsql.go
```diff
@@ -130,6 +130,111 @@ func getAggregationSQL(agg string, qid uint64) utils.AggregateFunctions {
...
+	// Check for 'round' function and handle arguments
+	if strings.ToLower(funcName) == "round" {
+		if len(argExprs) < 1 || len(argExprs) > 2 {
+			log.Errorf("qid=%v, getMathFunctionSQL: incorrect number of arguments for ROUND function", qid)
+			return nil, fmt.Errorf("incorrect number of arguments for ROUND function")
+		}
+
+		leftExpr, err := convertToNumericExpr(argExprs[0])
+		if err != nil {
+			return nil, err
```
AndrewHess, line 144: nit: also log the error with `log.Errorf("getMathFunctionSQL: <your-message>")`

pkg/ast/sql/astsql.go
```diff
@@ -130,6 +130,111 @@ func getAggregationSQL(agg string, qid uint64) utils.AggregateFunctions {
...
+		}
+
+		var rightExpr *structs.NumericExpr
+		if len(argExprs) == 2 {
+			rightExpr, err = convertToNumericExpr(argExprs[1])
+			if err != nil {
+				return nil, err
```
AndrewHess, line 151: nit: also log the error with `log.Errorf("getMathFunctionSQL: <your-message>")`

AndrewHess (commented):
This is looking good so far. For the query without a groupby clause, I don't think we have a pathway for implementing that yet, so you may want to have this PR just focused on when there is a groupby clause and then handle the other case in a different PR. For the case without a groupby clause, you'll probably want to look in segaggs.go, where we have a bunch of functions for doing something with and without a groupby

Macbeth98:
Ahh! got it! I was trying to implement the simple version of query, as I thought that would be easier. But when I was implementing, I got a message that this was only implemented for aggregation queries, but still I tried to implement it somehow and failed. So I just started doing the implementation with aggregation functions. 

And thanks for the info, I will give a detailed look at `seaggs.go`. 

And @AndrewHess as for this PR goes, then I can submit it for final review, once I implemented your suggestions and I will try to write some tests?

Macbeth98:
@AndrewHess @nkunal Please review the PR and let me know for any further changes. Thank You!

pkg/ast/sql/astsql.go
```diff
@@ -375,6 +369,9 @@ func parseSelect(astNode *structs.ASTNode, aggNode *structs.QueryAggregators, cu
 							measureOp = &structs.MeasureAggregator{MeasureCol: measureCol, MeasureFunc: getAggregationSQL(measureFunc, qid)}
 						}
 					} else {
+
+						leftExpr.Value = "0(" + leftExpr.Value + ")"
```
AndrewHess, line 373: What's happening here? Is the 0 supposed to be there?

AndrewHess (commented):
This is looking pretty good. A few things:
1. You've included some extra files from `pkg/dashboards/data`. Please remove those from the PR
2. When I run this code and give different precisions to the ROUND function, it seems to ignore the precision I specified

AndrewHess:
> 2. When I run this code and give different precisions to the ROUND function, it seems to ignore the precision I specified

Ignore that point. I was running a query that was coming out to exactly one decimal place; when I run a query where the answer has more decimal places, the rounding is working correctly

pkg/ast/sql/astsql.go
Macbeth98: @AndrewHess Yes, it is zero because the aggregation functions like count, sum, etc., all have an id's from 1 to 6, but since the round is not an aggregation, the default will be zero, so when the actual rounding was done, the column is named as 0(column_Name), so for that reason I have changed the field Name for rounding function to match that. If you run the query with just `round(latitude),` then you will see there is a column that comes with `0(latitude)`.

Macbeth98:
@AndrewHess Can you please review this PR and let me know of any further changes? Thank You!

AndrewHess (commented):
This looks good. Can you just remove the siglens.db file from this PR? It shouldn't be version controlled.

Once you remove that file, I think this is ready to get merged. Sorry for taking so long to get back to you

Macbeth98:
@AndrewHess No worries. Thank you for taking the time to review the PR. I have removed the `siglens.db` file and pushed the commit. Let me know if any further changes are needed. 

Thank you very much!

AndrewHess (approved):

PR #253, opened 2023-12-20
Fix some issues for the docker install script

Jiale-Fang:
# Description
Fix some issues for the docker install script

# Testing
On mac os, Ubuntu 22.04.01 Lts, CentOS 7

# Checklist:

### DIFF
--- a/install_with_docker.sh
+++ b/install_with_docker.sh
@@ -150,9 +150,9 @@ start_docker
 
 echo -e "\n===> Pulling the latest docker image for SigLens"
 
-wget "https://github.com/siglens/siglens/releases/download/${SIGLENS_VERSION}/server.yaml"
+curl -O -L "https://github.com/siglens/siglens/releases/download/${SIGLENS_VERSION}/server.yaml"
 $sudo_cmd docker pull siglens/siglens:${SIGLENS_VERSION}
-$sudo_cmd mkdir data
+mkdir data
 echo ""
 echo -e "\n===> SigLens installation complete"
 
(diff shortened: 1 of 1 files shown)


install_with_docker.sh
```diff
@@ -150,9 +150,9 @@ start_docker
 
 echo -e "\n===> Pulling the latest docker image for SigLens"
 
-wget "https://github.com/siglens/siglens/releases/download/${SIGLENS_VERSION}/server.yaml"
+curl -O -L "https://github.com/siglens/siglens/releases/download/${SIGLENS_VERSION}/server.yaml"
 $sudo_cmd docker pull siglens/siglens:${SIGLENS_VERSION}
-$sudo_cmd mkdir data
+mkdir data
```
nkunal, line 155: If data already exists , this will log a failure. Pass the -p option or whatever that option is which does not complain if dir already exists

install_with_docker.sh
```diff
@@ -180,27 +180,49 @@ https://api.segment.io/v1/track \
...
+        return 0
+    fi
+}
+
+ui_port=$(check_ports)
+
+if [ ${ui_port} == "-1" ]; then
+    echo ""
+    tput bold
+    printf "\e[31mError: No available port found on port ${initial_port} or in the range from ${start_port} to ${end_port}.\e[0m\n"
```
nkunal, line 211: Change to "no available port found , checked range [startport -- endport]" ... something like that

install_with_docker.sh
```diff
@@ -180,27 +180,49 @@ https://api.segment.io/v1/track \
...
 echo -e "ingestPort(8081) and queryPort(80) can be changed using in server.yaml."
 echo -e "queryPort(80) can also be changed by setting the environment variable $PORT."
-echo ""
-echo -e "To be able to query data across restarts, set ssInstanceName in server.yaml."
-echo ""
-echo -e "The target for the data directory mounting should be the same as the data directory (dataPath configuration) in server.yaml"
+
 tput bold
-echo -e "\n===> ${bold}Frontend can be accessed on http://localhost:80"
+printf "\n===> \e[32mFrontend can be accessed on http://localhost:${ui_port}\e[0m"
```
nkunal, line 221: The colors should not be hardcoded , define some vars at the top and use it. 
Also for all variables we should use all caps

nkunal (commented):
Few comments

nkunal (dismissed):

AndrewHess:
This works on my mac (with apple silicon). On an x86 machine running Amazon Linux, I still couldn't connect to the docker daemon without running the script as sudo. On an ARM machine running Ubuntu, there was an issue installing docker, and after manually installing docker I still had to run the script with sudo

Jiale-Fang:
@nkunal Can you please take a look about that? 
@AndrewHess Are you starting the docker with your root user before you run that script?

AndrewHess:
No, docker wasn't running when I started the script

nkunal (approved):

PR #270, opened 2023-12-22
Allow Spl query using multiple spaces

Jiale-Fang:
# Description

Allow using multiple spaces for:
1. Those commands or operators which need at least 1 space on one side or both sides. 

```
E.g.
CMD_SEARCH <- "search" SPACE
OR <- SPACE "OR" SPACE
keyword: limit, like, in... (It should have at least one space at left side)
```
- Central idea: Use regex to replace multiple spaces with one (except for the spaces inside string literals)
<br>
2. For those commands or operators which can allow 0 space or multiple spaces around them.

```
E.g.
1. ',' 2. '(' 3. '=' 4. '-' ... 
```
 - Central idea: Central idea: Create a rule EMPTY_OR_SPACE <- " "?, and put this rule around these operators
```
OpPlus <- EMPTY_OR_SPACE "+" EMPTY_OR_SPACE {
    return "+", nil
}
COMMA <- EMPTY_OR_SPACE "," EMPTY_OR_SPACE
L_PAREN <- "(" EMPTY_OR_SPACE
R_PAREN <- EMPTY_OR_SPACE ")"
```

# Testing
Past all the test cases and run some customized query like:
```
Chicago     AND    GET   OR (   Female   AND    404)
city   =  Boston |     stats   count   AS   Count   BY    http_status  ,    http_method    |     eval newField     =    (  http_status-1000 )
city  =  Detroit   |   stats   count  AS   Count   BY   latitude  ,  longitude,state   |   where  latitude>89.6  AND longitude  <  0
```

# Checklist:

### DIFF
--- a/cicd/ingest.csv
+++ b/cicd/ingest.csv
@@ -42,6 +42,7 @@ search 404 OR 500,now-1d,now,*,total,gt,100,Splunk QL
 search latency<18580 AND app_name=Stresscrawl,now-1d,now,*,total,gt,0,Splunk QL
 search MintCream AND Thursday OR Monday,now-1d,now,*,total,eq,212,Splunk QL
 search (MintCream AND Thursday) OR Monday,now-1d,now,*,total,gt,212,Splunk QL
+search    (   MintCream    AND   Thursday  )   OR   Monday,now-1d,now,*,total,gt,212,Splunk QL
 search MintCream Thursday OR Monday,now-1d,now,*,total,eq,212,Splunk QL
 search NOT (city!=Boston OR NOT weekday=Friday),now-1d,now,*,total,gt,100,Splunk QL
 search batch=batch-*,now-1d,now,*,total,gt,100,Splunk QL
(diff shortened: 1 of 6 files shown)


Jiale-Fang:
There is another pre-existing issue that will replace multiple spaces in the results with one space. However, it is not related to the parser, so the above commits are ready for review.

AndrewHess (commented):
This looks really good! I think you need to update this rule as well, right?
```
 FieldSelectBlock <- PIPE CMD_FIELDS op:("- " / "+ ")? fields:FieldNameList {
```
It looks like you've updated all the other rules correctly.

I think with the preprocessing in `processSearchText`, it might get a bit complicated to know when we're in a string if we have to deal with escaped double quotes or double quotes within comments, etc. If you change `SPACE <- (" "? COMMENT " "?) / " "` back to `SPACE <- (" "? COMMENT " "?) / " "+` and then also change `EMPTY_OR_SPACE  <- " "?` to `EMPTY_OR_SPACE  <- SPACE?`, can we get rid of the preprocessing?

AndrewHess (approved):
Looks great!

PR #283, opened 2023-12-23
Fix for SQL Query Index

### LINKED ISSUE #197: [BUG] SQL not reading from the correct index
# Description
When there's multiple indexes, you can run an SQL query like:
```
SELECT * FROM `ind-2`
```
but it won't always select the data from `ind-2`. Sometimes it selects from `ind-1` or something else.

# Steps to Reproduce
1. Start siglens
2. Ingest data into multiple indexes. For example, using `sigscalr-client` you can run:
```
go run main.go ingest esbulk -t 10_000 -d http://localhost:8081/elastic --processCount 1 -n 5 -b 500 -g dynamic-user
```
3. Run the SQL search (you may need to run it multiple times)
```
SELECT * FROM `ind-2`
```

# Expected Behavior
Only data from `ind-2` should be returned. Also, if `ind-2` does not exist, no data should be returned.

# Environment
- Commit Hash: `176775e`

venkateshamatam:
# Description
- This change ensures that the correct index is selected when running an SQL query. Previously, the code did not correctly handle the case where the index name was specified in the SQL query.

- The updated code checks if there exists an index name in the query. If true, it sets indexNameIn to aggs.TableName, which is where we are parsing the tableName and storing it in aggs.

- This change gives precedence to the index name specified in the SQL query over the index name selected from the dropdown in the UI. If there is no index name in the SQL query, the index name selected from the UI will be used. However, if an index name is specified in the SQL query, the index name selected from the UI will be ignored.

Fixes #197 

# Testing
- Run the SQL search. For example, if you've ingested data into an index named ind-2, you can run:
 `` SELECT * FROM `ind-2` ``
 Verify that only data from ind-2 is returned. If ind-2 does not exist, no data should be returned.
- Repeat for other index names to ensure that the correct index is always selected.
- Test the precedence of the index name by selecting a different index from the UI dropdown and running an SQL query with a specific index name. The data returned should be from the index specified in the SQL query, not the index selected from the dropdown.

Below image shows data being selected from `proddata-test` image

# Checklist:

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -27,9 +27,9 @@ COPY static static
 COPY server.yaml .
 
 RUN chown -R $UNAME:$GID static
-RUN chown -R $UNAME:$GID /siglens
 USER $UNAME
 
 WORKDIR /$UNAME
 COPY --from=build /usr/app/cmd/siglens/build/siglens .
+RUN chown -R $UNAME:$GID /siglens
 CMD ["./siglens", "--config", "server.yaml"]
--- a/pkg/alerts/alertsqlite/alerts_sqlite.go
+++ b/pkg/alerts/alertsqlite/alerts_sqlite.go
@@ -376,9 +376,14 @@ func (p Sqlite) GetAllAlerts() ([]alertutils.AlertInfo, error) {
 		log.Errorf("getAllAlerts: unable to begin transaction, err: %+v", err)
 		return nil, err
 	}
-	// #todo: Add silence_minutes back in actual fix
-	// sqlStatement := "SELECT alert_id, alert_name, state, create_timestamp, contact_id, labels, silence_minutes FROM all_alerts;"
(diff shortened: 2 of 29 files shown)


AndrewHess (commented):
This looks good. I think you'll also need these changes in SearchHandler.go in `ProcessPipeSearchRequest()`. This function is called for queries that don't come from the UI. You can test your changes by sending a query through curl like this:
```
curl -X POST -d '{
    "searchText": "SELECT * FROM `ind-7`",
    "index": "ind-2",
    "startEpoch": "now-24h",
    "endEpoch": "now",
    "queryLanguage": "SQL"
}' http://localhost/api/search | python3 -m json.tool
```

venkateshamatam:
Thanks for the review @AndrewHess, I have updated the `ProcessPipeSearchRequest()` in SearchHandler.go and verified the response using the curl command. It works as expected, overrides the given indexName to use the parsed indexName.

AndrewHess (approved):

PR #284, opened 2023-12-24
Add function for EXTRACT Year

ashwinivibhandik18:
Please review code changes .

### DIFF
--- a/pkg/ast/sql/astsql.go
+++ b/pkg/ast/sql/astsql.go
@@ -18,6 +18,7 @@ package sql
 
 import (
 	"encoding/json"
+	"errors"
 	"fmt"
 	"regexp"
 	"strconv"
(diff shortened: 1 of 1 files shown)


nbhavana:
Thank you for submitting a PR, we will review it asap

pkg/ast/sql/astsql.go
```diff
@@ -36,6 +37,33 @@ const (
...
+}
+
+func (s *SelectExprPlaceholder) Evaluate() (interface{}, error) {
+	fmt.Println("Evaluating SelectExpr:", s.SelectExpr)
+	return nil, nil
+}
+
+type ExtractYearExpression struct {
+	Source Expression
+}
```
AndrewHess, line 61: Eventually we'll want to support the extract function in its entirety, so you should make this an `ExtractExpression` but for now you can just implement the functionality for extracting the year.

AndrewHess (commented):
This doesn't compile. Are you able to fix the issues?

PR #290, opened 2023-12-27
Persistent queries

Davleen14:
# Description

This is initial commit and I have added following:
In organization's settings page, there is a dropdown menu for "Persistent Queries" with two options: "Enabled" and "Disabled". By default, it is set to "Enabled".
When a user changes the value to disabled, a warning is displayed if they choose "Disabled,".
When the value of dropdown is changed the frontend makes an api call '/api/update' to backend to update the pqs config
The new PQS config value is saved to  runmod.cfg file, located in the data/common folder. This is done by the SavePQSConfigToRunMod function.
Reading Configuration on Startup:
When the server starts, it initially reads  configuration from server.yaml.
After reading server.yaml, it also reads from runmod.cfg file to check for any PQS configuration. This is handled by the ReadRunModConfig function.
If runmod.cfg contains a PQS configuration, it updates the server's running configuration to reflect this new value


pkg/config/config.go
```diff
@@ -516,6 +516,9 @@ func InitConfigurationData() error {
 		return err
 	}
 	runningConfig = config
+	if err := ReadRunModConfig(); err != false {
+		return fmt.Errorf("cannot read config")
```
nkunal, line 520: please use `log.Errorf` statement, and you should not return, rather continue the rest of the process

pkg/config/config.go
```diff
@@ -586,6 +589,34 @@ func InitializeTestingConfig() {
 	SetDebugMode(true)
 	SetDataPath("data/")
 }
+func ReadRunModConfig() bool {
```
nkunal, line 592: don't return bool here, but rather return the error itself. We shouldn't "convert" the error that are returned by the underlying functions and convert it to a "bool

pkg/config/config.go
```diff
@@ -586,6 +589,34 @@ func InitializeTestingConfig() {
...
+	runModFilePath := "data/common/runmod.cfg"
+	if _, err := os.Stat(runModFilePath); err == nil {
```
nkunal, line 594: long if blocks are not good. you could do something like
```
exists := os.Stat(...)
if !exist {
   return nil
}
 ... os.ReadFIle(...)
```

pkg/config/config.go
```diff
@@ -586,6 +589,34 @@ func InitializeTestingConfig() {
...
+		jsonData, err := os.ReadFile(runModFilePath)
+		if err != nil {
+			return false
+		}
+		var runModConfig struct {
+			PQSConfig string `json:"PQSConfig"`
+		}
```
nkunal, line 601: define this struct at the top of this file. Also don't create a new name, we already have a `pqsEnabled` reuse that name

pkg/querytracker/querytracker.go
```diff
@@ -646,6 +646,45 @@ func GetPQSById(ctx *fasthttp.RequestCtx) {
 	ctx.SetStatusCode(fasthttp.StatusOK)
 }
 
+type PqsConfig struct {
+	PQSConfig string `json:"pqsConfig"`
+}
```
nkunal, line 651: keep naming consistent, we are calling it pqsEnabled, in other places, we should call this the same

pkg/cfghandler/pqsupdate.go
```diff
@@ -0,0 +1,68 @@
...
+	var config PqsConfig
+	err := json.Unmarshal(ctx.PostBody(), &config)
+	if err != nil {
+		log.Errorf("Error parsing request body: %v", err)
+		ctx.Error("Bad Request", fasthttp.StatusBadRequest)
+		return
+	}
+	if err := SavePQSConfigToRunMod(config.PQSEnabled); err != nil {
+		log.Errorf("Error saving pqsEnabled: %v", err)
+
```
nbhavana, line 28: Add functions name in the error message, it makes debugging easy.

pkg/cfghandler/pqsupdate.go
```diff
@@ -0,0 +1,68 @@
...
+		ctx.Error("Internal Server Error", fasthttp.StatusInternalServerError)
+		return
+	}
+
+	ctx.SetStatusCode(fasthttp.StatusOK)
+	ctx.SetContentType("application/json")
+	_, err = ctx.WriteString(`{"status":"success"}`)
+	if err != nil {
+		log.Errorf("Error writing response: %v", err)
+		return
```
nbhavana, line 38: same here

pkg/cfghandler/pqsupdate.go
```diff
@@ -0,0 +1,68 @@
...
+
+	}
+}
+func SavePQSConfigToRunMod(pqsEnabled string) error {
+	file, err := os.OpenFile(config.RunModFilePath, os.O_CREATE|os.O_WRONLY, 0666)
+	if err != nil {
+		log.Errorf("Failed to open or create the file %s: %v", config.RunModFilePath, err)
+		return err
```
nbhavana, line 46: same here

pkg/cfghandler/pqsupdate.go
```diff
@@ -0,0 +1,68 @@
...
+	}
+	defer file.Close()
+	configData := map[string]string{"PQSEnabled": pqsEnabled}
+	encoder := json.NewEncoder(file)
+
+	err = encoder.Encode(configData)
+	if err != nil {
+		log.Errorf("Failed to encode JSON data to file %s: %v", config.RunModFilePath, err)
+		return err
```
nbhavana, line 55: same here

pkg/cfghandler/pqsupdate.go
```diff
@@ -0,0 +1,68 @@
...
+	}
+
+	if strings.ToLower(pqsEnabled) == "disabled" {
+		querytracker.ClearAllPQSData()
+		err = config.ClearPqsFiles()
+		if err != nil {
+			log.Errorf("Failed to clear PQS data and files: %v", err)
+			return err
```
nbhavana, line 63: Add function name in error message

pkg/config/config.go
```diff
@@ -516,6 +517,10 @@ func InitConfigurationData() error {
 		return err
 	}
 	runningConfig = config
+	err = ReadRunModConfig()
+	if err != nil {
+		log.Errorf("Failed to read runmod config: %v", err)
+	}
```
nbhavana, line 523: same here

pkg/config/config.go
```diff
@@ -587,6 +592,42 @@ func InitializeTestingConfig() {
 	SetDataPath("data/")
 }
 
+func ReadRunModConfig() error {
+	_, err := os.Stat(RunModFilePath)
+	if os.IsNotExist(err) {
+		return nil
+	} else if err != nil {
+		log.Errorf("Error accessing runmod.cfg: %v", err)
+		return err
```
nbhavana, line 601: same here

pkg/config/config.go
```diff
@@ -587,6 +592,42 @@ func InitializeTestingConfig() {
...
+	}
+	var pqsCheck bool
+	switch strings.ToLower(runModConfig.PQSEnabled) {
+	case "enabled":
+		pqsCheck = true
+	case "disabled":
+		pqsCheck = false
+	default:
+		log.Errorf("Invalid PQSEnabled value in runmod.cfg: %s", runModConfig.PQSEnabled)
+	}
```
nbhavana, line 625: add function name in error message

nbhavana (commented):
Some nit pick for adding function names in error messages. 
Also, can you add unit tests for this functionality?

static/org-settings.html
```diff
@@ -62,10 +62,40 @@
...
                                 <td id="retention-days-value">4 days</td>
                             </tr>
+                            <tr>
+                                <th>Persistent Queries</th>
+                                <td>
+                                    <div class="dropdown">
+                                        <button class="btn dropdown-toggle" type="button" id="contact-types" style=" border-radius: 0; padding: 0.1rem 0.1rem "
+                                            data-toggle="dropdown" aria-haspopup="true" aria-expanded="false"
+                                            data-bs-toggle="dropdown">
+                                            <span style="color: var(--health-stats-value-text-color);font-size: 18px;">Enabled</span>
```
AndrewHess, line 72: I think we recently changed the default font size. 18 is too large but it looks good if you just remove `font-size: 18px;`. I'm not sure if there's a different way you should change it to make sure it's consistent with the rest of the fonts.

static/org-settings.html
```diff
@@ -62,10 +62,40 @@
                                 <th>Retention</th>
                                 <td id="retention-days-value">4 days</td>
                             </tr>
+                            <tr>
+                                <th>Persistent Queries</th>
+                                <td>
+                                    <div class="dropdown">
+                                        <button class="btn dropdown-toggle" type="button" id="contact-types" style=" border-radius: 0; padding: 0.1rem 0.1rem "
```
AndrewHess, line 69: I think the id should be something other than `contact-types`, right?

AndrewHess (commented):
The dropdown in My Org -> Settings -> Persistent Queries doesn't seem to be working for me. I tried clicking on Enabled and also on the orange dropdown arrow, but nothing happens.

Also, how is this supposed to work when siglens is started with `pqsEnabled: false`. Right now I see PQS is enabled in My Org -> Settings. Is that expected, or should it use the `server.yaml` value when there's no `runmod.cfg`?

AndrewHess:
The new PR is https://github.com/siglens/siglens/pull/473

PR #291, opened 2023-12-27
Spl timechart (Basic Functionalities)

Jiale-Fang:
# Description
Allow using splunk timechart query.
https://docs.splunk.com/Documentation/Splunk/9.1.2/SearchReference/Timechart

Just finishing the following functionalities:
>- Finish Time Range Buckets Process Logic
>- Allow the Use of Multiple Aggregations Inside the Block
>- Set Span Length Option for Span Options
>- Allow the Use of the Split By Clause in Each Time Range Bucket

```
E.g.
* | timechart span=1h sum(latitude)
* | timechart span=1hrs sum(latitude) range(longitude) BY weekday 
```

# Testing
Have not add test cases for that. But it can pass all the existing test cases.

# Checklist:


pkg/segment/utils/timeconsts.go
```diff
@@ -0,0 +1,35 @@
...
+
+import "fmt"
+
+type TimeUnit uint8
+
+const (
+	TMUs TimeUnit = iota
+	TMMs
+	TMCs
+	TMDs
```
nkunal, line 11: Can ypu spell put what those those  Ms, Cs,Ds mean just kike you fid for second minute etc

nkunal (approved):
Looks good one minor comment

PR #307, opened 2023-12-28
Use regex search instead of word search for all columns search

venkateshamatam:
# Description
Use Regex search for all columns search instead of  exact match search.

# Testing
1. Perform a regex search like `Donk*are` and verify that it searches all columns from the logs to match this regex.  
2. Perform an exact match search like `Donkeyare` and verify that exact match logs are returned.

image

# Checklist:


pkg/utils/segutils.go
```diff
@@ -139,3 +141,17 @@ func IsSubWordPresent(haystack []byte, needle []byte) bool {
 
 	return false
 }
+
+func IsPatternPresent(haystack []byte, pattern []byte) bool {
+	// Convert pattern to string and replace '*' with '.*' to create a regex pattern
+	regexPattern := strings.Replace(string(pattern), "*", ".*", -1)
+
+	// Compile the regex
+	r, err := regexp.Compile(regexPattern)
```
nkunal, line 150: Doing regex compile on every check is very expensive. This should be done once . Take a look at the other applycsg function and how it passes the qvaldte which has the compiled regex in it

venkateshamatam: Got it! I'll use the compiled regex from qvaldte

pkg/segment/structs/segsearchstructs.go
```diff
@@ -375,6 +375,26 @@ func extractSearchQueryFromMatchFilter(match *MatchFilter) *SearchQuery {
...
+		for _, word := range match.MatchWords {
+			cval := dtu.ReplaceWildcardStarWithRegex(string(word))
+			rexpC, err := regexp.Compile(cval)
+			if err != nil {
+				log.Errorf("extractSearchQueryFromMatchFilter: regexp compile failed for MatchWord %s, err=%v", word, err)
+			} else {
+				currQuery.MatchFilter.Regexp = rexpC
+			}
+		}
+	} else if match.MatchPhrase != nil {
```
nkunal, line 388: this cannot be an else, when matchPhrase is present the parser fills in both matchwords and matchphrase.

pkg/segment/structs/segsearchstructs.go
```diff
@@ -375,6 +375,26 @@ func extractSearchQueryFromMatchFilter(match *MatchFilter) *SearchQuery {
 	} else {
 		currQuery.SearchType = MatchWords
 	}
+	if len(match.MatchWords) > 0 {
```
nkunal, line 378: you can ignore the matchwords case, because matchwords case would contain many words to be matched, and then you will have to create a regex array for each of the word. So for now just ignore this case and only fill in the `Regexp` for matchphrase

pkg/segment/structs/segsearchstructs.go
```diff
@@ -375,6 +375,26 @@ func extractSearchQueryFromMatchFilter(match *MatchFilter) *SearchQuery {
...
+			cval := dtu.ReplaceWildcardStarWithRegex(string(word))
+			rexpC, err := regexp.Compile(cval)
+			if err != nil {
+				log.Errorf("extractSearchQueryFromMatchFilter: regexp compile failed for MatchWord %s, err=%v", word, err)
+			} else {
+				currQuery.MatchFilter.Regexp = rexpC
+			}
+		}
+	} else if match.MatchPhrase != nil {
+		cval := dtu.ReplaceWildcardStarWithRegex(string(match.MatchPhrase))
```
nkunal, line 389: apart from checking for `match.MatchPhrase` not being nill, you should also check if `match.MatchPhrase` contains a `*`, only then fill in the regexp , that way we don't unnecessarily to do regex search for exact match

pkg/segment/writer/rawchecker.go
```diff
@@ -52,10 +52,18 @@ func ApplySearchToMatchFilterRawCsg(match *MatchFilter, col []byte) (bool, error
 	if match.MatchOperator == And {
 		var foundQword bool = true
 		if match.MatchType == MATCH_PHRASE {
-			foundQword = utils.IsSubWordPresent(col[idx:idx+clen], match.MatchPhrase)
+			regexp := match.Regexp
+			if regexp == nil {
```
nkunal, line 56: nit: no need for an extra regexp variable, you can just use the `match.Regexp` as is

pkg/segment/writer/rawchecker.go
```diff
@@ -52,10 +52,18 @@ func ApplySearchToMatchFilterRawCsg(match *MatchFilter, col []byte) (bool, error
...
+				return false, errors.New("Failed to compile regular expression for match phrase")
```
nkunal, line 57: this will not be an error based on the vey top comment above, we will do regexp search only if it contains *

pkg/segment/writer/rawchecker.go
```diff
@@ -52,10 +52,18 @@ func ApplySearchToMatchFilterRawCsg(match *MatchFilter, col []byte) (bool, error
...
+			}
+			foundQword = regexp.Match(col[idx : idx+clen])
 		} else {
-			for _, qword := range match.MatchWords {
-				foundQword = utils.IsSubWordPresent(col[idx:idx+clen], []byte(qword))
+			for range match.MatchWords {
+				regexp := match.Regexp
+				if regexp == nil {
```
nkunal, line 63: don;t use extra var and this is not an error cond, see msg above

pkg/segment/writer/rawchecker.go
```diff
@@ -66,8 +74,12 @@ func ApplySearchToMatchFilterRawCsg(match *MatchFilter, col []byte) (bool, error
 
 	if match.MatchOperator == Or {
 		var foundQword bool
-		for _, qword := range match.MatchWords {
-			foundQword = utils.IsSubWordPresent(col[idx:idx+clen], []byte(qword))
+		for range match.MatchWords {
+			regexp := match.Regexp
+			if regexp == nil {
```
nkunal, line 79: same comment as above

pkg/segment/writer/rawchecker.go
```diff
@@ -51,13 +51,17 @@ func ApplySearchToMatchFilterRawCsg(match *MatchFilter, col []byte) (bool, error
 	// todo MatchWords struct can store bytes
 	if match.MatchOperator == And {
 		var foundQword bool = true
-		if match.MatchType == MATCH_PHRASE {
-			foundQword = utils.IsSubWordPresent(col[idx:idx+clen], match.MatchPhrase)
+		if match.Regexp != nil {
```
nkunal, line 54: this check should happen in the `if match.MatchType == MATCH_PHRASE` as mentioned in the earlier comment:
when matchtype is match_phrase only then invoke the "new logic", and the new logic is  : if regexp is not nil then do regex match, else do the issubword match

nkunal (approved):

PR #256, opened 2023-12-20
Sql rename col for aggregation columns and round function

Macbeth98:
# Description
- Renaming of columns with alias `As` keyword implemented for aggregation columns and round function.
- Implemented the loop for renaming of cols at `colReq.RenameColumns != null` in `segaggs.go`.

Feature **#Adds Aliasing**

# Testing
- Added a test case in `sqlparser_test.go` for testing of renaming of round function. For aggregation, the tests are already there.

# Checklist:
Before marking your pull request as ready for review, complete the following.

- [x] I have self-reviewed this PR.
- [x] I have removed all print-debugging and commented-out code that should not be merged.
- [x] I have added sufficient comments in my code, particularly in hard-to-understand areas.
- [x] I have formatted the code, if applicable. For Go, I have run `goimports -w .`.

PR #318, opened 2023-12-29
SQL math functions(round, abs, ceil, sqrt)

Macbeth98:
# Description

- This PR contains the implementation of Math functions such as `round, ceil, abs, sqrt` for the group by and non-group by SQL commands.
- The application of these functions to the column values is being done at `readAllRawRecords` function in `recordreader.go`.
- This way, these Math Evaluator functions can be applied when the values are read without the need to loop again.

# Features 
- **Implements Math Functions for both Group by and Non-Group by SQL Commands.**

# Testing
- Test SQL Commands
   - `select city, round(latitude, 2)`
   - `select city, round(latitude, 2) as lat_round, abs(longitude) as long_abs, sqrt(latency) as latency_sqrt`
   - `select city, ceil(sum(latitude)) as lat_sum from t group by city`
   - `select city, abs(sum(latitude)) as lat_sum from t group by city`
- Written test cases in `sqlparser_test.go` for the verifying of the parsing of SQL Math functions.

# Checklist:

# Notice
- This is a PR on top of the PR #256. 
- I have chosen to do this way, as I worked on the addition of the SQL Round function, which is a part of the Math function, and the aliasing for aggregation and Round function columns.
- So, I wanted to make sure that aliasing will still be working for group and non-group commands. For that reason, I built this PR on top of the previous PR.   

# Reference Image
image


Macbeth98:
@AndrewHess @nkunal 

Please review this PR and let me know if there need to be any further changes. 
Also, let me know if you want me to convert this into a Draft PR, till the previous one is merged.

pkg/segment/aggregations/segaggs.go
```diff
@@ -168,7 +168,36 @@ RenamingLoop:
 	}
 
 	if colReq.RenameColumns != nil {
-		return errors.New("performColumnsRequest: processing ColumnsRequest.RenameColumns is not implemented")
+		// return errors.New("performColumnsRequest: processing ColumnsRequest.RenameColumns is not implemented")
+
```
AndrewHess, line 172: Don't just comment it out, delete it

AndrewHess (commented):
This looks great. Needs one quick fix and then I think it's ready to merge

pkg/segment/aggregations/segaggs.go
Macbeth98: Done, @AndrewHess

AndrewHess (approved):

PR #325, opened 2024-01-02
Fix: Allow the search block without using keyword "search"

Jiale-Fang:
# Description
In the past, only the first search block can execute without using keywrod `search`
E.g.:
`"Santa Ana" | search latency > 500000`

After this change, allow this type of query:
E.g.:
`"Santa Ana" | latency > 500000`

# Testing

# Checklist:

### DIFF
--- a/cicd/ingest.csv
+++ b/cicd/ingest.csv
@@ -47,6 +47,7 @@ search MintCream Thursday OR Monday,now-1d,now,*,total,eq,212,Splunk QL
 search NOT (city!=Boston OR NOT weekday=Friday),now-1d,now,*,total,gt,100,Splunk QL
 search batch=batch-*,now-1d,now,*,total,gt,100,Splunk QL
 latency<10000 | search latitude>0 | search longitude>0,now-1d,now,*,total,gt,10,Splunk QL
+latency<10000 | latitude>0 | search longitude>0,now-1d,now,*,total,gt,10,Splunk QL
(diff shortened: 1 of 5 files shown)


Jiale-Fang:
I believe the reason why we need to use keyword `search` for the following `search block` is because the rule for the `search block` is defined as follows:
`SearchBlock <- CMD_SEARCH clause:ClauseLevel4 {} `

There is a sub-rule in ClauseLevel4:
`UnnamedFieldWithStringValue <- value:String {} `

This means that if we do not use `CMD_SEARCH`, it will convert all other blocks as `SearchBlock` since all the blocks consist of strings.

Currently, if we want to support using `SearchBlock` without keyword `search`, I can only think of the following method:
`SearchBlock <- !(ALLCMD) CMD_SEARCH? clause:ClauseLevel4 {} `

Here, ALLCMD includes all commands we have:
```
ALLCMD <- (CMD_REGEX / CMD_STATS / ......) 

CMD_REGEX <- "regex" SPACE 
CMD_STATS <- "stats" SPACE 
......
```
Modifying `SearchBlock` to ensure it does not contain these existing commands should work. However, it seems somewhat arbitrary as a solution to this problem. What is your opinion on this?

AndrewHess:
> Modifying `SearchBlock` to ensure it does not contain these existing commands should work. However, it seems somewhat arbitrary as a solution to this problem. What is your opinion on this?

I'm not thinking of a better way to do this either. This should work though

AndrewHess (commented):
Looks good

AndrewHess (approved):

AndrewHess:
@Jiale-Fang can you add some tests for this?

AndrewHess (approved):

PR #330, opened 2024-01-03
Fix: Spl rex block

Jiale-Fang:
# Description
Fix bug for rex block without group by cols.

# Testing

# Checklist:

### DIFF
--- a/pkg/segment/aggregations/segaggs.go
+++ b/pkg/segment/aggregations/segaggs.go
@@ -739,8 +739,6 @@ func performRexColRequestWithoutGroupby(nodeResult *structs.NodeResult, letColRe
 		return fmt.Errorf("performRexColRequestWithoutGroupby: There are some errors in the pattern: %v", err)
 	}
 
-	nodeResult.MeasureFunctions = letColReq.RexColRequest.RexColNames
-
 	fieldName := letColReq.RexColRequest.FieldName
 	for _, record := range recs {
 		fieldValue := fmt.Sprintf("%v", record[fieldName])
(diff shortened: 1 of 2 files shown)


AndrewHess (commented):
When I run:
```
city=Boston | rex field=weekday "(?<first>[A-Z].{2})(?<second>.*)"
```
it works correctly, but when I run:
```
city=Boston | rex field=weekday "(?<first>[A-T].{2})(?<second>.*)"
```
I get a bunch of these errors and I don't see the new `first` and `second` fields
```
ERRO[2024-01-03 14:54:27] PostQueryBucketCleaning: performAggOnResult: performLetColumnsRequest: performRexColRequest: performRexColRequestWithoutGroupby: MatchAndExtractGroups: no str in field match the pattern
```

Jiale-Fang:
I think the reason is because `Wednesday` not start with [A-T], so for record which weekday is `Wednesday`. It can not find a match for that.

AndrewHess:
> I think the reason is because `Wednesday` not start with [A-T], so for record which weekday is `Wednesday`. It can not find a match for that.

Right, but shouldn't it still give some data? Either just the rows that were successfully extracted or all the rows with nil values for the extracted fields for rows like Wednesday? Or maybe I'm misremembering how rex is supposed to work

Jiale-Fang:
image

Jiale-Fang:
> > I think the reason is because `Wednesday` not start with [A-T], so for record which weekday is `Wednesday`. It can not find a match for that.
> 
> Right, but shouldn't it still give some data? Either just the rows that were successfully extracted or all the rows with nil values for the extracted fields for rows like Wednesday? Or maybe I'm misremembering how rex is supposed to work

Can you take a look of the pic above. It is the screenshot of Splunk

AndrewHess (approved):
Looks good

PR #337, opened 2024-01-03
Spl timechart limit improvement

Jiale-Fang:
# Description
Add support for non-numeric func(Values, Cardinality) for timechart limit option.

Timechart will only display N highest/lowest scoring distinct values of the split-by field. For those cols can not be displayed, they will be merged into the `other` category col.

For numeric func, we can use addition to do the merge logic. For non-numeric func, each one has its own merge logic.

So now it support these kinds of queries,
```
* | timechart span=10m dc(latitude) BY http_method limit=bottom 3
* | timechart span=2m values(first_name) BY http_method limit=bottom3
* | timechart span=2m dc(address), values(first_name) BY http_method limit=2
```

# Testing

# Checklist:


pkg/segment/aggregations/timechartagg.go
```diff
@@ -255,22 +264,37 @@ func MergeVal(eVal *utils.CValueEnclosure, eValToMerge utils.CValueEnclosure, hl
...
-		// 		err := hll.Merge(hllToMerge)
-		// 		if err != nil {
-		// 			log.Errorf("MergeVal: failed to merge hyperloglog stats: %v", err)
-		// 		}
-		// 		eVal.CVal = hll.Estimate()
-		// 		eVal.Dtype = utils.SS_DT_UNSIGNED_NUM
-		// 		return
-		// 	}
-		// case utils.Values:
+	// TODO: should merge values for not numeric agg
```
AndrewHess, line 267: This TODO is done, right?

pkg/segment/aggregations/timechartagg.go
```diff
@@ -255,22 +264,37 @@ func MergeVal(eVal *utils.CValueEnclosure, eValToMerge utils.CValueEnclosure, hl
...
+			}
+			eVal.CVal = hll.Estimate()
+			eVal.Dtype = utils.SS_DT_UNSIGNED_NUM
+			return
+		}
+	case utils.Values:
+		// Can not do addition for values func
+		if useAdditionForMerge {
+			return
+		}
```
AndrewHess, line 284: Does this need a TODO and/or a log message, or should it just be `return`?

pkg/segment/aggregations/timechartagg.go
Jiale-Fang: Yes

pkg/segment/aggregations/timechartagg.go
Jiale-Fang: It should be return. `useAdditionForMerge` means sum up their values. It is used by `limit` to score single agg. However, for single agg `values`, we can not sum up them, we use `frequency` to score them.

AndrewHess (approved):

PR #353, opened 2024-01-06
Clear colWips if it's too large

AndrewHess:
# Description
When new columns are added, they're included in `colWips`. However, old columns aren't removed, so the size could grow. If it gets larger than it should, we'll reset it.

# Testing
I've not tested this

# Checklist:

### DIFF
--- a/pkg/segment/writer/segstore.go
+++ b/pkg/segment/writer/segstore.go
@@ -105,13 +105,19 @@ func (segstore *SegStore) initWipBlock() {
 func (segstore *SegStore) resetWipBlock(forceRotate bool) error {
 
 	segstore.wipBlock.maxIdx = 0
-	for _, cwip := range segstore.wipBlock.colWips {
-		cwip.cbufidx = 0
-		cwip.cstartidx = 0
 
-		cwip.deCount = 0
-		for dword := range cwip.deMap {
-			delete(cwip.deMap, dword)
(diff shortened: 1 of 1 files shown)


pkg/segment/writer/segstore.go
```diff
@@ -105,13 +105,19 @@ func (segstore *SegStore) initWipBlock() {
...
 	segstore.wipBlock.maxIdx = 0
-	for _, cwip := range segstore.wipBlock.colWips {
-		cwip.cbufidx = 0
-		cwip.cstartidx = 0
 
-		cwip.deCount = 0
-		for dword := range cwip.deMap {
-			delete(cwip.deMap, dword)
+	if len(segstore.wipBlock.colWips) > 2000 {
+		log.Errorf("resetWipBlock: colWips size is too large: %v", len(segstore.wipBlock.colWips))
```
nkunal, line 110: we should add the `segkey` to this log msg, to be able to corelate which index/table this will be happening on.
Also we should say `....size is greater than 2000...`

nkunal (approved):
looks good, just one minor nit

pkg/segment/writer/segstore.go
AndrewHess: Made the changes here: https://github.com/siglens/siglens/pull/358

PR #391, opened 2024-01-16
fix: SPL timechart default binOptions added

### LINKED ISSUE #387: [BUG] Timechart Command in Splunk is encountering an interface conversion error.
# Description
- Run any Timechart command in SplunkQL and it encounters an interface conversion error mentioning that interface is not `*structs.BinOptions`
- Example: `app_name="Snailwill" | timechart avg(latency)`.
- Then, it is throwing an error: `Message: 1:27 (26): rule TimechartBlock: interface conversion: interface {} is nil, not *structs.BinOptions`

# Steps to Reproduce
1. Run any Timechart command in SplunkQL
2. Example: `app_name="Snailwill" | timechart avg(latency)`.
3. Then, it will throw an error: `Message: 1:27 (26): rule TimechartBlock: interface conversion: interface {} is nil, not *structs.BinOptions`
4. See the image below.

image

# Expected Behavior
The timechart command is expected return the statistics with timestamp and avg of latency.

Macbeth98:
# Description
- Fixes the SPL Timechart Bug.
- I have set the default Span Options to 1m. 

Fixes #387 

# Testing
- I have tested by running the `timechart` commands on the Siglens UI.

# Checklist:


AndrewHess (dismissed):
Later we should make this work how Splunk chooses the default span times: https://docs.splunk.com/Documentation/Splunk/9.1.1/SearchReference/Timechart#Default_time_spans

Macbeth98:
@AndrewHess Shall I add this functionality now?

- I am thinking that I will set the default values like how I set them right now in the parsing, but I will add a flag to identify whether the options set is a default one. 
- If it is default, then I will extract the time range from the request, and depending on that I will reset the values as per the Splunk doc at `searchQueryParser.go`.

What do you think about this?

AndrewHess:
Yes that should work

pkg/ast/spl/spl.peg
```diff
@@ -455,7 +467,8 @@ TimechartBlock <- PIPE CMD_TIMECHART binOptions:(BinOptions)? expr:(SingleAggExp
     }
 
     timeBucket := aggregations.InitTimeBucket(bOptions.SpanOptions.SpanLength.Num, bOptions.SpanOptions.SpanLength.TimeScalr, byField, limitExprTmp, len(measureAggs))
-    aggNode.TimeHistogram = timeBucket
+	//timeBucket.Timechart.BinOptions = bOptions
```
AndrewHess, line 470: nit: delete this commented-out code if it's not needed

pkg/ast/common.go
```diff
@@ -244,3 +245,30 @@ func ParseTimeRange(startEpoch, endEpoch uint64, aggs *QueryAggregators, qid uin
...
+	} else if duration <= 60*60*1000 {
+		return &structs.SpanOptions{SpanLength: &structs.SpanLength{Num: 1, TimeScalr: utils.TMMinute}}, nil
+	} else if duration <= 4*60*60*1000 {
+		return &structs.SpanOptions{SpanLength: &structs.SpanLength{Num: 5, TimeScalr: utils.TMMinute}}, nil
+	} else if duration <= 24*60*60*1000 {
+		return &structs.SpanOptions{SpanLength: &structs.SpanLength{Num: 30, TimeScalr: utils.TMMinute}}, nil
+	} else if duration <= 7*24*60*60*1000 {
+		return &structs.SpanOptions{SpanLength: &structs.SpanLength{Num: 1, TimeScalr: utils.TMHour}}, nil
+	} else if duration <= 180*24*60*60*1000 {
+		return &structs.SpanOptions{SpanLength: &structs.SpanLength{Num: 1, TimeScalr: utils.TMDay}}, nil
```
AndrewHess, line 269: Where are you getting these values from? I was checking https://docs.splunk.com/Documentation/Splunk/9.1.1/SearchReference/Timechart#Default_time_spans and all the other `(duration, spanLength)` pairs match with what you have, but these two don't. Was there other documentation where you found these values?

Macbeth98: @AndrewHess No, I added them because, as with the Splunk documentation

- If it is **more than 30 days**, it will default to only 1 month. And since we have **90 days** in our time range, which is three months. However, I also felt just three data points would be very low. So I set it for 180 days, which is up to 6 months, the data will be given per day. Then after that, it will be given per month as how Splunk does.
- And regarding days, we have **2 days**, as one of the options, if it is more than a day, by default Splunk would choose per day interval; which will give us two data points. Thinking in a similar way as above, I set the interval as one hour for the range of more than a day and less than or equal **to 7 days**.

AndrewHess: Ok, that makes sense. What you have is fine then

AndrewHess (approved):

PR #392, opened 2024-01-16
fix: SPL Transaction Startswith bug fixed

Macbeth98:
# Description
- SPL Transaction command was not showing results when only the StartsWith command was given.
- So the issue was that the code was not appending the results until it found another StartsWith condition.
- I changed the code to append the code when EndsWith is not given and append any leftover grouped records.

# Testing
- Tested the command with commands on UI.
- Tested with the automated tests written in `segaggs_test.go`

# Checklist:

### DIFF
--- a/pkg/segment/aggregations/segaggs.go
+++ b/pkg/segment/aggregations/segaggs.go
@@ -22,7 +22,6 @@ import (
 	"regexp"
 	"sort"
 	"strconv"
-	"strings"
 
 	"github.com/siglens/siglens/pkg/common/dtypeutils"
 	"github.com/siglens/siglens/pkg/segment/structs"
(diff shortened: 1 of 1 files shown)


pkg/segment/aggregations/segaggs.go
AndrewHess, line 1: Does this also work correctly when there's only a `transactionEndsWith`? I'm not sure how this is supposed to work in that case when the last record it sees doesn't match the `transactionEndsWith`; should it use the leftover records as a transaction or just discard them?

If it's supposed to use the leftovers as a transaction, it seems like the logic after looping through the records would just be "if there's any leftover records, then call appendGroupedRecords on them" (without needing to check whether `transactionStartsWith` and/or `transactionEndsWith` are nil). Or maybe I'm missing something

Macbeth98: When I checked with Splunk by running commands and watched some YouTube videos on Splunk I understood, that when `endswith` is given Splunk will only send back the group of results that ends with the given conditions. If no record matches with the given condition, then it will be discarded.

- But before this change, when only `startswith`, the code will not append any leftover logs that match the starts with condition.
- Now, It is checking in the case when the transaction `EndsWith` is not given. But Either the fields list or `StartsWith` is given there might be leftover grouped logs that need to be appended to the final results.
- And by default when only `EndsWith` is given it should only send the grouped logs that match the given condition. So the current setup should work fine without any issues. Even if there are any leftover grouped logs, they will be discarded.

AndrewHess (approved):

PR #416, opened 2024-01-23
Extract geolocation info

venkateshamatam:
# Description
During initialization or when an install script is run, extract the IP address and other geo details such as lat,lon,country and timezone.

# Testing
Tested this code on my machine by logging the details and also ran a test workflow to confirm the install script is working. 
image
link to the workflow - https://github.com/venkateshamatam/siglens/actions/runs/7629354720/job/20782582399

# Checklist:


pkg/ssa/ssa.go
```diff
@@ -65,7 +115,13 @@ func InitSsa() {
 		log.Errorf("Error initializing ssa: %v", err)
 		return
 	}
+	ipDetails, err := FetchIPAddressDetails()
+	if err != nil {
+		log.Fatalf("Failed to fetch IP address details: %v", err)
+	}
 
+	IPAddressInfo = ipDetails
+	log.Infof("IP address details: %+v\n", IPAddressInfo)
```
nbhavana, line 124: I think we should remove this log line

install_with_docker.sh
```diff
@@ -66,6 +66,21 @@ else
     echo 'Not Supported Architecture'
 fi
 
+fetch_ip_info() {
+    response=$(curl -s https://ipinfo.io)
+
+    ip=$(echo "$response" | awk -F'"' '/ip/{print $4}')
```
AndrewHess, line 72: When I run these in my terminal, and then `echo $ip` I get two lines; the first is my IP and the second is `https://ipinfo.io/missingauth`. I think we just want the first line. For all the other extracted parameters I got the expected results

install_with_docker.sh
AndrewHess, line 1: Do you need to do something with the IP info once you get it? It looks like it's not being used?

AndrewHess (approved):

PR #433, opened 2024-01-26
Extract unique user id for docker installs

venkateshamatam:
# Description
The purpose of this is to have the same user id while installing through docker and sending server updates. 
# Testing

Tested this by building a local docker image and running the docker install script and verified that the user ID remains same for events sent through script and the go code. The existing functionality remains unchanged. 

Also added another property in JSON to determine if the ID that is sent id the ID from the docker script or go code. 

# Checklist:


pkg/ssa/ssa.go
```diff
@@ -132,18 +133,22 @@ func waitForInitialEvent() {
 	props := analytics.NewProperties()
 
 	// Initialize computer-specific identifier
-	computerID, err := utils.GetSpecificIdentifier()
-	if err != nil {
-		log.Errorf("waitForInitialEvent: %v", err)
+	if userId = os.Getenv("CSI"); userId != "" {
+		source = "CSI"
+	} else {
+		computerID, err := utils.GetSpecificIdentifier()
```
nkunal, line 139: You can do userId , err = ...
And in the error scenario set userId to "unknown"

nkunal (commented):
Minor comment

nkunal (commented):
Once @AndrewHess approves then can be merged

AndrewHess (approved):

PR #457, opened 2024-01-31
Install with podman

### LINKED ISSUE #437: [NEW] Support podman engine for installation method
# Description
Supporting Podman engine for installation

# Benefits
As Fedora based distributions comes pre-installed with podman, it would be easier for the developers on these platforms to install SigLens. Also Podman and Podman desktop project is now gaining a lot of traction these days as a good open source alternative to Docker Desktop

# (Optional) Possible Implementation
If you have any ideas on how this should be implemented, let us know here.

Currently on Fedora 39 and getting below error if I use the docker install method - 
```

➜  curl -L https://siglens.com/install.sh | sh                                                                                                 (current-context is not set)

----------Setting up docker----------
sudo: apt-get: command not found
sudo: apt-get: command not found
sudo: apt-key: command not found
curl: (22) The requested URL returned error: 404
main: line 143: lsb_release: command not found
sudo: add-apt-repository: command not found
sudo: apt-get: command not found
Installing docker
sudo: apt-get: command not found
{
  "success": false,
  "message": "malformed JSON",
  "code": "invalid_request"
}install_docker: Docker installation failed during apt-get install on linux
```

Macbeth98:
# Description
- Install Siglens with Podman.

Fixes #437  (link all the GitHub issues this addresses)

# Testing
- I have tested the install script by running against the AWS EC2 instance.

# Checklist:


install.sh
```diff
@@ -0,0 +1,570 @@
...
+echo "===> Selected container tool: $CONTAINER_TOOL"
+
+
+# Extract the version number from pkg/config/version.go by getting everything
+# inside the quotes. Use -n to supress printing each line, and p to print the
+# modified line.
+SIGLENS_VERSION=`\
+    curl  --silent "https://api.github.com/repos/siglens/siglens/releases/latest" |
+    grep '"tag_name":' |
+    sed -E 's/.*"([^"]+)".*/\1/'`
```
AndrewHess, line 29: nit: this comment seems incorrect (it looks like this is the same comment from install_with_docker.sh, which is also incorrect)

install.sh
```diff
@@ -0,0 +1,570 @@
...
+   Darwin*)
+     os="darwin"
+     package_manager="brew" ;;
+   Linux*amzn2*)
+     os="amazon linux"
+     package_manager="yum" ;;
+   Red\ Hat*|CentOS*|Fedora*)
+     os="linux"
+     package_manager="yum" ;;
+   Ubuntu*|Pop!_OS|Debian*|Linux*|Mint*)
```
AndrewHess, line 59: Should `Linux*|Mint*` be `Linux\ Mint*`, or should it be what you have? (I'm just looking at the diff with install_with_docker.sh)

AndrewHess (commented):
This looks really good. I think you can remove the install_with_podman.sh file since all of that logic is now in the install.sh script. But don't delete the install_with_docker.sh file yet because that's still referenced by the website: https://github.com/siglens/siglens.github.io/blob/develop/install.sh

install.sh
Macbeth98: I changed this to how it is in the docker script. It is now changed to `Ubuntu*|Pop!_OS|Debian*|Linux*|Linux\ Mint*` But I do feel `Linux* | Linux\ Mint*` is redundant since `Linux*` covers the other Mint distributions also and we are assigning os as linux and package manager to `apt-get` 

So should I remove `Linux\ Mint*`?

AndrewHess: Oh I see, yeah as long as having `Linux*` does not include things we don't want included, you can just have `Linux*` and remove `Linux\ Mint*`

Macbeth98: @AndrewHess ignore the above comment.

install.sh
```diff
@@ -46,6 +47,36 @@ is_command_present() {
...
+    dist_version=$(lsb_release -rs)
+elif  [ -f /etc/os-release ]; then
+    . /etc/os-release
+    dist_id=$NAME
+    dist_version=$VERSION_ID
+elif is_command_present distro; then
+    dist_id=$(distro | awk '{print $1}')
+    dist_version=$(distro | awk '{print $2}')
+else 
+    echo "Couldn't determine the OS distribution"
```
AndrewHess, line 65: I'm getting this message on mac, but then installation (with docker) works. Is this just for linux flavors?

AndrewHess (commented):
This looks great. When I try to run the install script, first I have to run `chmod +x install.sh` and then it works, but `git diff` shows the mode change. Are you able to change the mode and push that change?

install.sh
Macbeth98: Yes, that is just for Linux flavors. I am using the  dist_id which is usually `ubuntu, rhel, fedora, centos`, etc, to download the appropriate docker repo.

Macbeth98:
> This looks great. When I try to run the install script, first I have to run `chmod +x install.sh` and then it works, but `git diff` shows the mode change. Are you able to change the mode and push that change?

I have changed the mode and pushed it. Let me know of any further changes. Thank you!

AndrewHess (approved):

PR #462, opened 2024-02-02
API to count the number of traces and infinite scroll on traces screen 

venkateshamatam:
# Description
Implement Infinite scroll on traces screen and a new endpoint to get the total number of traces for a particular search to display on the traces screen. 
It currently loads 100 traces at first and once the user scrolls to the 60% mark or above it loads the next set of traces until there are no more traces left.

# Testing
- Tested the API endpoint with default and custom search text.
- Tested the scatter-plot graph to represent newly fetched traces as well
- Tested the filtering for the traces that are present.
- Tested the download functionality to download the newly fetched traces as well

# Checklist:


pkg/segment/tracing/handler/tracehandler.go
```diff
@@ -150,6 +151,66 @@ func ProcessSearchTracesRequest(ctx *fasthttp.RequestCtx, myid uint64) {
 	ctx.SetStatusCode(fasthttp.StatusOK)
 }
 
+func ProcessTotalTracesRequest(ctx *fasthttp.RequestCtx, myid uint64) {
```
AndrewHess, line 154: It looks like this function repeats a lot of the work done by the `ProcessTotalTracesRequest` function. Would it be better to calculate this result in the `ProcessTotalTracesRequest` function and send the total count as another json parameter, so that then we don't have this duplicated code and it would probably save the backend some duplicate work? Or would that not work for some reason?

AndrewHess (commented):
The scroll panel should continue to the bottom of the screen. On my screen it gets cut off
image

pkg/segment/tracing/handler/tracehandler.go
```diff
@@ -150,6 +130,74 @@ func ProcessSearchTracesRequest(ctx *fasthttp.RequestCtx, myid uint64) {
...
+	searchText, _, _, _, _, _ := pipesearch.ParseSearchBody(readJSON, nowTs)
+
+	isOnlyTraceID, _ := ExtractTraceID(searchText)
+
+	if isOnlyTraceID {
+		// If the search text is a trace ID, the total count is 1
+		ctx.SetStatusCode(fasthttp.StatusOK)
+		ctx.SetBodyString("1")
+		return
+	}
```
AndrewHess, line 150: You'll still have to do a search to determine if the answer is 1 or 0

AndrewHess (commented):
Now I'm getting two scroll bars, which feels unintuitive
image

AndrewHess (approved):
Looks good

PR #473, opened 2024-02-07
Persistent Query

sonamgupta21:
# Description
Summarize the change.

# Testing

# Checklist:


nbhavana (commented):
I see these errors on server startup.
image

pkg/config/config.go
```diff
@@ -516,6 +521,11 @@ func InitConfigurationData() error {
 		return err
 	}
 	runningConfig = config
+	var readConfig RunModConfig
+	readConfig, err = ReadRunModConfig(RunModFilePath)
+	if err != nil {
```
AndrewHess, line 526: Change line 526 to
```
if err != nil && !os.IsNotExist(err) {
```
Since we can get there when the file doesn't exist, but that's not really an error (we'll just read from the config file in that case rather than the runmod config file in that case)

AndrewHess (commented):
A test is failing because the Go code isn't formatted correctly. If you run
```
goimports -w .
```
from the base directory of siglens, it should properly format the code for you (I think you're just missing some newlines at the end of files)

AndrewHess:
> I see these errors on server startup. 

@nbhavana I think you have an old runmod config file. In Davleen's PR, the pqsEnabled runmod flag was a string, but in this PR it's a boolean. Can you delete your runmod config file and try again? It should be at `data/common/runmod.cfg`

sonamgupta21:
@AndrewHess PR is ready for review.

AndrewHess (approved):

PR #478, opened 2024-02-07
fix: log and quit if address already in use

### LINKED ISSUE #340: [BUG] display port bind errors on stdout
# Description
When the ports 8081 or 5122 or already bound and then you try to start another instance of siglens via the go run cmd, the stdout displays that it has bound to the ports but actually it has failed. The failure messages show up in siglens.log, along with that we should also print it out on stdout and then exit the process.

these are the errors that show in siglens.log file
```
ESC[31mERROESC[0m[2024-01-04 13:21:06] Failed to start server! Error: cannot bind to "0.0.0.0:8081": address already in use 
ESC[36mINFOESC[0m[2024-01-04 13:21:06] populateMetricsMetadata: reading smFile=data/ingestnodes/kunals-imac.lan.EocRhoFGQrPELBymcJzc97/metricmeta.json 
ESC[36mINFOESC[0m[2024-01-04 13:21:06] Query server has successfully been initialized in 156.458µs 
ESC[31mERROESC[0m[2024-01-04 13:21:06] Failed to start server! Error: cannot bind to "0.0.0.0:5122": address already in use 
```

# Steps to Reproduce
1. ...

# Expected Behavior
Clearly describe what you expect to happen.

# Environment
- Commit Hash: [e.g. 9ed5f0a]
- OS: [e.g. MacOS 14.1.1 on apple silicon]
- Browser and Version (if applicable): [e.g. chrome 119.0.6045.159 arm64, safari 19616.2.9.11.7]

smaftoul:
# Description

Grab errors, check if listen failed. If so, log and exit.
Also, refactor a bit to use a single routine that checks which mode to start so code is more readable.
Finally, don't use reuseport because it doesn't fail listening on a port already listening (at least on macos).

Fixes #340 

# Testing

Tested manually launching 2 instances of siglens, or a netcat listening on one of the ports.
Not sure if there are some e2e test to update.

# Checklist:


AndrewHess (commented):
This looks good to me. There's some merge conflicts with go.mod though; if you merge in the changes from the `develop` branch and then run `go mod tidy` it should fix the go.mod to be correct

smaftoul:
@AndrewHess thanks, I was indeed rebasing while you were commeting.
Golangci-lint also catched that the code wasn't checking errors correctly, fixed in previous commit.

I now wonder if I should switch to `github.com/valyala/tcplisten` instead of `net.Listen` because removing `github.com/valyala/fasthttp/reuseport`, stopped using `SO_REUSEADDR`  which may be fine (seems needed otherwise we can't have the "adress already in use" error, at least on macos, but also, this setting is here for performance), but it also removes `TCP_DEFER_ACCEPT` and `TCP_FASTOPEN` which we may want to keep. WDYT ?

AndrewHess:
I'm not sure it's a big issue either way. But sure if you want to use `tcplisten` to disable `SO_REUSEADDR` but keep the others, that would be great.

smaftoul:
@AndrewHess pushed a version with tcplistener, what do you prefer ?
Also, I started to write tests for `pkg/server/ingest/server.go` , the `Run()` function, it starts, but I cannot get the server to stop so far.

I am wondering if my PR is fine, I think it handles the collision on 1 port, but not both, I'll do a manual test with `nc` but ideally would like to have those as go tests.

pkg/server/query/server.go
```diff
@@ -293,7 +292,7 @@ func renderTemplate(ctx *fasthttp.RequestCtx, tpl *template.Template) {
 func (hs *queryserverCfg) RunSafeServer() error {
 	hs.Router.GET("/health", hs.Recovery(getSafeHealthHandler()))
 	var err error
-	hs.ln, err = reuseport.Listen("tcp4", hs.Addr)
+	hs.ln, err = net.Listen("tcp4", hs.Addr)
```
AndrewHess, line 295: This server should also use `tcplisten` instead of `net`, right? Same comment for line 234

AndrewHess:
The changes with `tcplistener` look great; I think that's better than the `net` you had before. If you're having issues adding unit tests, I think this PR is fine without them since we aren't likely to change this code much

AndrewHess:
Hmm with these changes I'm not getting an error when I have siglens running and then try to start another siglens server. Is it working correctly for you? If not, you can revert to the previous version with `net` and that can get merged

smaftoul:
Indeed, with the tcplisten, the "address already in use" was gone (at least on osx). I rolled back the changes to the `net` version, and tested locally for both ports and I saw the error logged and the process quitting.

AndrewHess (approved):

PR #481, opened 2024-02-09
Selecting default logs option for dataSource

### LINKED ISSUE #467: [BUG] Blank panel when no Data Source is selected
# Description
When you make a new panel and edit it to run a certain query, but don't select the Data Source, there's no indication that you need to select a Data Source, the panel just silently fails to show anything. We should use the Logs option by default.

# Steps to Reproduce
1. Make a new dashboard and panel
2. Edit the panel and add a query and chart type, but don't select a data source
3. Click the Run Query button or the Apply button, and observe that the panel silently fails.

# Expected Behavior
The Data Source dropdown should have Logs selected by default for new panels. So when you make a new panel and edit it, the dropdown will show Logs instead of Data Source

# Environment
- Commit Hash: 512a448

PratikshaPatil77:
# Description
If the user doesn't select the data source choice, the logs option will be chosen as the default option.
The Data Source dropdown have Logs selected by default for new panels. So when you make a new panel and edit it, the dropdown will show Logs instead of Data Source

Fixes #467 

# Checklist:

### DIFF
--- a/static/js/edit-panel-screen.js
+++ b/static/js/edit-panel-screen.js
@@ -371,9 +371,13 @@ function editPanelInit(redirectedFromViewScreen) {
 		$('.dropDown-data-rate-options span').html('Data Rate');
 		prevSelectedDataTypeIndex = -2;
 	}
-			
+	if (selectedDataSourceTypeIndex === -1 || selectedDataSourceTypeIndex === undefined) {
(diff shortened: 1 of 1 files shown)


nbhavana (commented):
The changes look good for default data source. 
Can you also make this change as a part of this PR? : When default data source is  selected (`logs`), the default chart type should be selected as `log lines`.

nbhavana (approved):

PR #490, opened 2024-02-12
push image with latest tag

sunitakawane:
# Description
Prod pipeline failed when no tag was added so adding `latest` tag 

# Testing

# Checklist:

### DIFF
--- a/.github/workflows/publish-prod-images.yml
+++ b/.github/workflows/publish-prod-images.yml
@@ -53,13 +53,14 @@ jobs:
             platforms: linux/amd64,linux/arm64
             tags: ${{ steps.meta.outputs.tags }}
       -
-        name: Build and push without tag (to add latest tag to build)
+        name: Build and push with latest tag
         uses: docker/build-push-action@v3
         with:
             context: .
             file: Dockerfile
             push: true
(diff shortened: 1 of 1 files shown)


AndrewHess (dismissed):

Macbeth98:
@sunitakawane 
I was checking the issue as to why the build was failing while I was updating the Podman release assets.

In this fix, we are building the images twice.

If it is just to add the latest tag, I believe we can try to add the latest tag here directly instead of building the image again:
```
- name: Extract metadata (tags, labels) for Docker
        id: meta
        uses: docker/metadata-action@v5
        with:
            images: siglens/siglens
            tags: |
              type=raw,value=${{ env.TAG }}
              type=raw,value=latest
 ```
              
 and then since we have build and push step
 
 ```
 name: Build and push
        uses: docker/build-push-action@v3
        with:
            context: .
            file: Dockerfile
            push: true
            platforms: linux/amd64,linux/arm64
            tags: ${{ steps.meta.outputs.tags }}
  ```
  
  The tags here reference the Extract tags above, so it should include all the tags including the latest tag and the version tag.
So we can completely remove the step: `Build and push with the latest tag`

But if there is another reason we need this as a separate build, we need to keep it the same.

PR #506, opened 2024-02-16
Allow aggregation after Extracting fields

Macbeth98:
# Description
- Allows Aggregations after Extracting fields including the fields.

# Testing
- Tested with UI
- Will be writing more test cases soon

# Checklist:

### DIFF
--- a/pkg/ast/pipesearch/searchHandler.go
+++ b/pkg/ast/pipesearch/searchHandler.go
@@ -269,7 +269,7 @@ func ProcessAlertsPipeSearchRequest(queryParams alertutils.QueryParams) int {
 			// of them when there's a sortby option.
 			sizeLimit = math.MaxUint64
 		} else if aggs.HasRexBlockInChainWithStats() {
-			// If there's a Stats block in the chain followed by a Rex block, we need to
+			// If there's a Rex block in the chain followed by a Stats block, we need to
 			// see all the matched records before we apply or calculate the stats.
 			sizeLimit = math.MaxUint64
 		}
--- a/pkg/ast/pipesearch/searchHandler.go
+++ b/pkg/ast/pipesearch/searchHandler.go
@@ -268,6 +268,10 @@ func ProcessAlertsPipeSearchRequest(queryParams alertutils.QueryParams) int {
 			// Dedup needs to see all the matched records before it can return any
 			// of them when there's a sortby option.
 			sizeLimit = math.MaxUint64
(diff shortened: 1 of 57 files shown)


pkg/ast/pipesearch/searchHandler.go
```diff
@@ -268,6 +268,10 @@ func ProcessAlertsPipeSearchRequest(queryParams alertutils.QueryParams) int {
 			// Dedup needs to see all the matched records before it can return any
 			// of them when there's a sortby option.
 			sizeLimit = math.MaxUint64
+		} else if aggs.HasRexBlockInChainWithStats() {
+			// If there's a Stats block in the chain followed by a Rex block, we need to
+			// see all the matched records before we apply or calculate the stats.
+			sizeLimit = math.MaxUint64
```
AndrewHess, line 274: nit: I think this condition is implemented correctly, but the comment is slightly off. You're checking if there's a rex block followed by stats, not a stats block followed by rex

pkg/segment/structs/segstructs.go
```diff
@@ -259,18 +260,27 @@ type QueryCount struct {
...
+	BucketCount                   int               `json:"bucketCount,omitempty"`
+	PerformAggsOnRecs             bool              // if true, perform aggregations on records that are returned from rrcreader.go
+	RecsAggsType                  []PipeCommandType // To determine Whether it is GroupByType or MeasureAggsType
+	GroupByRequest                *GroupByRequest
+	MeasureOperations             []*MeasureAggregator
+	RecsAggsRecords               map[string]map[string]interface{}
+	RecsAggsRunningBuckets        interface{} // Evaluates to []*blockresults.RunningBucketResults
+	RecsAggsStringBucketIdx       map[string]int
+	RecsAggsProcessedSegments     uint64
+	RecsAggsProcessedSegmentsLock sync.Mutex
```
AndrewHess, line 283: For these new fields, it would be good if you could group them into a new struct type (or a few new types, if that makes more sense) just so the `NodeResult` doesn't get too bloated and do make it a little easier to understand the grouping of the fields (e.g., I assume the `RecsAggs...` fields are only used for certain queries, and for those queries all of those attributes will be needed, so you could make a new struct type for those)

Macbeth98: I have removed some of the types in the struct that I am not using.
Let me know if you still want me to reduce then I will change it accordingly.

sunitakawane (approved):

PR #596, opened 2024-02-29
fix: Timechart span Argument

Macbeth98:
# Description
- The Timechart Argument `span` will work regardless of the order.

# Testing
- Tested against the SigLens UI.
- Updated existing and have written additional tests to test the queries.
image
image

# Checklist:


AndrewHess (commented):
This looks pretty good. One note about formatting: since the spl.peg file isn't Go, each indent level should be 4 spaces instead of one tab. You should be able to configure your text editor to do this (and to display spaces and tabs, if you want). Once that's fixed, I think this is ready to merge

AndrewHess (approved):

PR #602, opened 2024-02-29
Fix: Eval command without stats

Macbeth98:
# Description
- The Eval command will now work without stats.

# Testing
- Have tested the code against the UI.
- Will be writing more tests.
image
image
image

# Checklist:

### DIFF
--- a/cmd/startup/startup.go
+++ b/cmd/startup/startup.go
@@ -32,6 +32,7 @@ import (
 	"github.com/siglens/siglens/pkg/blob"
 	local "github.com/siglens/siglens/pkg/blob/local"
 	"github.com/siglens/siglens/pkg/config"
+	commonconfig "github.com/siglens/siglens/pkg/config/common"
 	"github.com/siglens/siglens/pkg/dashboards"
 	"github.com/siglens/siglens/pkg/hooks"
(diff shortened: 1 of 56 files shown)


AndrewHess (commented):
I'm not quite sure, but I think with how you structured this, there will be issues with slightly more complex queries. For example, what if the query is like `* | rex ... | eval ...` where `rex` makes a new field and then the `eval` makes a new field based on that. Or what if the query is like `* | eval ... | stats ...` with `stats` using the field created by `eval`? If there's a path to make these queries work with how you've structured it, then you can keep what you have; otherwise it might make more sense to move most of your logic into segaggs.go where we do a lot of the post-search manipulations, and make sure that for these types of queries, rrcreader.go should have `GetJsonFromAllRrc()` call `agg.PostQueryBucketCleaning()` (which is already setup, but you might have to verify this path gets activated for these types of queries).

If necessary, we can add support for these more complicated queries in a different PR, but we should be able to build off this PR to support those queries

Macbeth98:
- I have written the tests and the PR should be ready to merge now. 
- Also, the queries below or any other aggs in the queue should now be processed fully after the Rex col, aggs.

image

image

AndrewHess (approved):

PR #607, opened 2024-03-06
add additional hook to get org_id

sunitakawane:
# Description
Summarize the change.
add additional hook to get org_id

# Testing

# Checklist:


pkg/server/query/entryHandlers.go
```diff
@@ -234,7 +244,7 @@ func pipeSearchWebsocketHandler(myid uint64) func(ctx *fasthttp.RequestCtx) {
 					return
 				}
 			}()
-			pipesearch.ProcessPipeSearchWebsocket(conn, myid, ctx)
+			pipesearch.ProcessPipeSearchWebsocket(conn, orgId, ctx)
```
AndrewHess, line 247: I think now `myid` is never used in the function, so you can also remove it from the definition on line 217

AndrewHess (commented):
When I made `CallWithOrgId` I didn't realize it was specific to ingest and not query. So I think for clarity you should either rename `CallWithOrgId` to something like `CallWithOrgIdIngest`, or move `CallWithOrgId` and `CallWithOrgIdQuery` out of utils.go and into the ingest and query server.go files as private functions. Apart from that this looks good

sunitakawane:
> `CallWithOrgId` to something like `CallWithOrgIdIngest`, or move `CallWithOrgId` and `CallWithOrgIdQuery` out of utils.go and into the ingest and query server.go files as private functions

I have not tested other ingest and query entry handlers. I think there is a mix of some route in ingest using CallWithOrgIdQuery and vice versa. I will have to  test these routes and then rename it if possible.

AndrewHess (approved):

PR #610, opened 2024-03-06
Feat: Add sort cmd without limit

### LINKED ISSUE #129: SPL: sort: sort without limit
# Description
https://docs.splunk.com/Documentation/Splunk/9.1.0/SearchReference/Sort

Sort on one or more fields, with ties in earlier fields being broken by later fields. Each field can be sorted ascending or descending, and the d or desc flag can be used to reverse the sorting of each specified field.

For now, return all results instead of just the first 10000.

Jiale-Fang:
# Description

Fixes #129 

Finished splunk sort cmd without the limit option based on: (https://docs.splunk.com/Documentation/Splunk/9.1.0/SearchReference/Sort)

# Testing
Run the test cases, and execute the commands as below:
```
* | sort -num(http_status)

* | fields http_status, city, state | sort num(http_status), -city, state

city=Boston | stats count AS Count BY http_status, city | sort +num(http_status)

city=Boston | stats count AS Count BY http_status, weekday | eval newField=(weekday . "ABC")  | sort +newField -Count

* | stats count as count by http_status, weekday | eval myip=192 . "." . tostring(ceil(http_status / 100)) . "." . tostring(ceil(http_status / 10)) . "." . 10 | sort -ip(myip)
```

# Checklist:


Jiale-Fang:
I added a flag called `SortByTimestampAtDefault` in response data to indicate whether the front end needs to sort the response based on timestamp by default. This may require the front-end developer to make some corresponding changes.
image

Jiale-Fang:
Yes, actually, I tried the cases you mentioned above before in splunk, and they all worked correctly. So, I reused the `sort-by-clause` for the sort command. What do you think?

1: Use comma for `dedup` sortby option; without using (+ or -) for sort by field
`sourcetype = access_combined_wcookie | dedup 100 status sortby other, clientip`

a64faed6cc7a5032f1d59cbfe7e0f350

2: Use spaces for the sortby option in the sort command.
`sourcetype = access_combined_wcookie | sort status other clientip`

9becb305b49d2eee3bc77e5fcda16ae7

AndrewHess:
Ok, then it must just be an inconsistency with their docs, so what you have looks good

PR #639, opened 2024-03-19
Enable SSL/TLS for Query and Ingest Service. 

Macbeth98:
# Description
- SSL/TLS for the query and ingest service can be enabled.
- Added two more variables under TLS config to take the Certificate and PrivateKey.

# Testing
- Tested by setting up a domain, generating a Certificate and private key, and Enabling the TLS.
- Ingested data from my local to hosted URL ingestion `https` endpoint.
- Ran queries against the UI and sent sample test data from UI against the `https` query endpoint.

# Checklist:

### DIFF
--- a/pkg/config/common/common.go
+++ b/pkg/config/common/common.go
@@ -33,8 +33,7 @@ type LogConfig struct {
 }
 
 type TLSConfig struct {
-	Enabled         bool   `yaml:"enabled"`
-	ACMEFolder      string `yaml:"acmeFolder"`      // folder to store acme certificates
+	Enabled         bool   `yaml:"enabled"`         // folder to store acme certificates
 	CertificatePath string `yaml:"certificatePath"` // path to certificate file
 	PrivateKeyPath  string `yaml:"privateKeyPath"`  // path to private key file
 }
--- a/pkg/config/common/common.go
+++ b/pkg/config/common/common.go
@@ -33,8 +33,10 @@ type LogConfig struct {
 }
 
 type TLSConfig struct {
-	Enabled    bool   `yaml:"enabled"`
-	ACMEFolder string `yaml:"acmeFolder"` // folder to store acme certificates
+	Enabled         bool   `yaml:"enabled"`
+	ACMEFolder      string `yaml:"acmeFolder"`      // folder to store acme certificates
+	CertificatePath string `yaml:"certificatePath"` // path to certificate file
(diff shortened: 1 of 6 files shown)


pkg/config/config.go
```diff
@@ -676,6 +686,14 @@ func ExtractConfigData(yamlData []byte) (common.Configuration, error) {
 		config.TLS.ACMEFolder = strings.Trim(config.TLS.ACMEFolder, "./")
 	}
 
+	if len(config.TLS.CertificatePath) >= 0 && strings.HasPrefix(config.TLS.CertificatePath, "./") {
+		config.TLS.CertificatePath = strings.Trim(config.TLS.CertificatePath, "./")
```
nkunal, line 690: are we assuming here that they will give a local path that starts with `./` and not a absolution path that would start with `/` , then that assumption is not correct.

pkg/server/ingest/server.go
```diff
@@ -151,11 +151,20 @@ func (hs *ingestionServerCfg) Run() (err error) {
 
 	// run fasthttp server
 	var g run.Group
-	g.Add(func() error {
-		return s.Serve(hs.ln)
-	}, func(e error) {
-		_ = hs.ln.Close()
-	})
+
+	if config.IsTlsEnabled() && config.GetTLSCertificatePath() != "" && config.GetTLSPrivateKeyPath() != "" {
```
nkunal, line 155: we do not need the second check for the certPath here.
During bootup, if tls is enabled and the certpath is not specified, then we should exit the process and log an error with appropriate message. 
This will avoid doing extra checks on every request that we receive

AndrewHess (commented):
I think with these changes `ACMEFolder` is not used anymore. If so you can remove it. Or did I miss something?

Apart from that this looks ready to merge

pkg/config/common/common.go
```diff
@@ -33,8 +33,7 @@ type LogConfig struct {
 }
 
 type TLSConfig struct {
-	Enabled         bool   `yaml:"enabled"`
-	ACMEFolder      string `yaml:"acmeFolder"`      // folder to store acme certificates
+	Enabled         bool   `yaml:"enabled"`         // folder to store acme certificates
```
AndrewHess, line 36: nit: this comment doesn't apply

AndrewHess (approved):

PR #655, opened 2024-03-21
Fix Search traces with min/max duration with timeunits

### LINKED ISSUE #574: [BUG] Min and Max Duration search doesn't work for traces
# Description
The search functionality doesn't consider the min and max duration inputs like 1.2s, 100ms, 500μs.

# Steps to Reproduce
1. Go to the search traces page at `/search-traces.html`.
2. Search for traces with a custom search or a `*` and add the min and max duration with values like these 1.2s, 100ms, 500μs and notice that it does not work as expected.

# Expected Behavior
- The search API should consider these values and search for traces only between those durations.
- It should atleast support the units `s`, `ms`, `μs`

ksankeerth:
* before this fix, siglens doesn't have support for searching with time duration of different units
* If search example 'search latency>0.5ms' would fail with errors
* the fix  avoids converting time duration with units to string

# Description
* Before convert values like "10s", "10ms", "10ds", "10cs" to string directly, PEG grammer will try to convert as time duration. If that fails only, they will be converted to String
* Added new evaluation struct `TimeDuration` and some additional grammer rules for avoid string conversions for time duration
* Moved and  modified `GetIntervalInMillis` function to different file for reusing

Fixes #574 

# Testing

* added tests 

# Checklist:

### DIFF
--- a/pkg/ast/pipesearch/searchQueryParser.go
+++ b/pkg/ast/pipesearch/searchQueryParser.go
@@ -26,10 +26,10 @@ import (
 	"github.com/siglens/siglens/pkg/ast/spl"
 	"github.com/siglens/siglens/pkg/ast/sql"
 	"github.com/siglens/siglens/pkg/config"
-	"github.com/siglens/siglens/pkg/segment/aggregations"
 	"github.com/siglens/siglens/pkg/segment/query/metadata"
 	"github.com/siglens/siglens/pkg/segment/structs"
 	. "github.com/siglens/siglens/pkg/segment/structs"
+	"github.com/siglens/siglens/pkg/segment/utils"
 
 	segment "github.com/siglens/siglens/pkg/segment"
 	. "github.com/siglens/siglens/pkg/segment/utils"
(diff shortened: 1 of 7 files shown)


AndrewHess (commented):
The code quality looks great, but I think you're overcomplicating this. This issue is just for the `/search-traces.html` page, and I think you could handle this mostly or completely in the frontend. We don't want to add time units into our SPL parser because different users might mean different things by the same field; for example, one user might have a field named `latency` that's in milliseconds and another might have a `latency` field in microseconds; then for a query like `latency < 3ms` we'd have to choose what units to convert that to so we can get just a number (without units) to compare to the actual (unitless) values in that field. So that would lead to an inconsistent experience because we can't know whether we should convert to seconds, milliseconds, etc. 

So for this issue I think a good solution would be to get the raw user input and convert that to the units we want (I believe it's unix nano, but not quite sure; OTLP traces should send their start and end time in unix nano as specified here: https://github.com/open-telemetry/opentelemetry-proto/blob/74e38fb4882dd500d77b757d44e97de296c65b05/opentelemetry/proto/trace/v1/trace.proto#L145-L159). So if the units the backend expects is nanoseconds, then just convert `123ms` to `123000000`

AndrewHess:
@ksankeerth are you able to make the changes mentioned? Or if you don't plan on continuing with this, just let us know and someone else can work on it

ksankeerth:
> @ksankeerth are you able to make the changes mentioned? Or if you don't plan on continuing with this, just let us know and someone else can work on it

@AndrewHess, Sorry for the late reply. I thought already another person started working on it (I saw a message in 574 issue). Yes, ofc. I'm little busy.

PR #675, opened 2024-03-26
Transaction Command with Sort and process by Segement.

Macbeth98:
# Description
- At `applyFopAllRequests` I have reversed the `QuerySegmentSlice` so that the oldest records will be processed first if the transaction command is present.
- Transaction command now sorts and processes the records for each segment and stores the processed results.

# Testing
- Tested against the UI by running transaction command queries.

# Checklist:


pkg/segment/aggregations/segaggs.go
```diff
@@ -2114,46 +2114,80 @@ func getAggregationResultFieldValues(fieldToValue map[string]segutils.CValueEncl
...
+			})
+
+			cols, err = processTransactionsOnRecords(nodeResult.TransactionEventRecords, nodeResult.TransactionsProcessed, nil, aggs.TransactionArguments, nodeResult.RecsAggsProcessedSegments == numTotalSegments)
+			if err != nil {
+				log.Errorf("performTransactionCommandRequest: %v", err)
+				return
+			}
+
+			nodeResult.TransactionEventRecords = nil // Clear the transaction records. Release the memory.
+			nodeResult.TransactionEventRecords = make(map[string]map[string]interface{})
```
AndrewHess, line 2156: Don't you have to keep all the records with a timestamp that's after the last record that got into a transaction, because those records might be in part of a transaction that gets finished by records in the next segment you process? Or are those records already being stored somewhere else, like inside a partial transaction?

Macbeth98: Yes, they are being stored inside a partial transaction. They are being stored in `aggs.TransactionArguments.OpenTransactionEvents`. 
So I created two fields for `TransactionArguments`; 1. To store the events/records. 2. To store the state of the Transaction.

The `nodeResult.TransactionEventRecords` is only being used to fetch all the records per segment.

AndrewHess (approved):

PR #685, opened 2024-03-26
Allow gzipped splunk ingestion of multiple records

AndrewHess:
# Description
This allows the splunk ingestion endpoint to handle data that is gzipped. It also makes it handle multiple records that are sent as a series of JSON objects (not inside a JSON list)

https://github.com/siglens/siglens/pull/673 had allowed multiple records to be ingested but it required them to be in a JSON list, which doesn't follow the Splunk documentation; see https://docs.splunk.com/Documentation/Splunk/latest/Data/FormateventsforHTTPEventCollector#Event_data

# Testing
Manually tested with
```
curl -XPOST -d '{
    "index": "example",
    "field1": "apples",
    "field2": "oranges"
    }{"field1": "thing1", "field2": "thing2"}' http://localhost:8081/services/collector/event
```
I also tested sending gzipped records

# Checklist:

### DIFF
--- a/pkg/utils/httpserverutils.go
+++ b/pkg/utils/httpserverutils.go
@@ -823,3 +823,19 @@ func GetSpecificIdentifier() (string, error) {
 
 	return hostname, nil
 }
+
+func GetDecodedBody(ctx *fasthttp.RequestCtx) ([]byte, error) {
+	contentEncoding := string(ctx.Request.Header.Peek("Content-Encoding"))
+	switch contentEncoding {
+	case "":
+		return ctx.Request.Body(), nil
+	case "gzip":
+		return gunzip(ctx.Request.Body())
+	default:
+		return nil, fmt.Errorf("GetDecodedBody: unsupported content encoding: %s", contentEncoding)
+	}
+}
+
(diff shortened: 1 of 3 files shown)


pkg/utils/httpserverutils.go
```diff
@@ -823,3 +825,42 @@ func GetSpecificIdentifier() (string, error) {
 
 	return hostname, nil
 }
+
+func GetDecodedBody(ctx *fasthttp.RequestCtx) ([]byte, error) {
+	contentEncoding := string(ctx.Request.Header.Peek("Content-Encoding"))
```
nkunal, line 830: is this header peek/check content sensitive?
In the request that I made I saw it was sent as `content-encoding: gzip`

Macbeth98 (approved):

pkg/utils/httpserverutils.go
AndrewHess: It should be case-insensitive for HTTP/1.1 (see https://datatracker.ietf.org/doc/html/rfc7230#section-3.2). Also it worked for me with this capitalization when I tested it

Issue #539, opened 2024-02-22
[STORY] Add error handling to parseAlphaNumTime


AndrewHess:
# Description
`parseAlphaNumTime()` is defined in pkg/ast/pipesearch/searchHandler.go. It converts strings like `now-1h` (meaning one hour before now) to a unix timestamp in milliseconds. However, the code is a little messy and doesn't properly handle invalid cases (e.g., strings like `Hello, World!`). You should refactor/comment this code to make it more legible and add error handling.

# Acceptance Criteria
* The new version of the function should be easy to understand and return an error if the input values are invalid.
* Running `make pr` from the siglens base directory gives no errors

abhinavg1997:
I can pick it up

abhinavg1997:
Hey Andrew, I had a question with respect to checking the length of the input value. While having a check for "less than 6 characters" makes sense, should there be an upper bound check? Based off of the possible values, the highest it can currently go would be 7 characters "now-90d" but that could change in the future.
Moreover I see test cases that are checking for 3 digit values which would fail if we keep a 7 character limit. 
Could use your opinion on this

AndrewHess:
You don't need an upper bound on the length; as long as the string is properly formatted, we should handle it. You could have a check to make sure we don't overflow the return value, but since it's a `uint64` and we're returning unix milli time, this shouldn't really ever be an issue unless some of the inputs to the function are very wrong.

Also if you want you can remove/refactor that check for less than 6 characters. That's there because we need at least something like `now-1s` which has 6 characters, but that's not immediately obvious from just quickly looking at the code, so if there's a cleaner way to handle that check that doesn't have a magic number that could help clean up the code

gouthamhusky:
Hey @AndrewHess , does this need to be assigned to me to open a PR or can I go ahead?

nkunal:
You can just go ahead

PR #693, opened 2024-03-27
Fixing issue#539

abhinavg1997:
# Description

1. Introduced comments to the code, explaining the process of time and unit extraction.
2. Introduced a Regex based check for the input string which checks if the string follows the format:
   "now-(number)[h/m/d]"
when the regex is failed, the default value and an error is logged.
3. The function now returns an error object as well which is now being propagated upto "[wsSearchHandler.go]" file, where it is being handled similarly to the cases above it.

Fixes #<539> https://github.com/siglens/siglens/issues/539

# Testing
Introduced 2 new test cases in the Test case [searchHandler_test.go] file which tests the input string against the regex function to make sure that it is working as expected.

# Checklist:

- [done ] I have self-reviewed this PR.
- [done ] I have removed all print-debugging and commented-out code that should not be merged.
- [done ] I have added sufficient comments in my code, particularly in hard-to-understand areas.
- [done ] I have formatted the code, if applicable. For Go, I have run `goimports -w .`.


pkg/ast/pipesearch/searchHandler.go
```diff
@@ -112,7 +115,7 @@ func ParseSearchBody(jsonSource map[string]interface{}, nowTs uint64) (string, u
 			startEpoch = uint64(val)
 		case string:
 			defValue := nowTs - (15 * 60 * 1000)
-			startEpoch = parseAlphaNumTime(nowTs, string(val), defValue)
+			startEpoch, err = parseAlphaNumTime(nowTs, string(val), defValue)
```
AndrewHess, line 118: You should check if `err != nil` and if so, log an error message and return early.

Same comment for lines 139, 248, 363

pkg/ast/pipesearch/searchHandler.go
```diff
@@ -527,41 +530,64 @@ func convertQueryCountToTotalResponse(qc *structs.QueryCount) interface{} {
...
+	//setting return value as default
 	retVal := defValue
 
+	//regex pattern for "now-[number]" and unit at the end which only accepts 'm', 'h' and 'd'
+	pattern := `^now-\d+[mhd]$`
+	r, patternError := regexp.Compile(pattern)
+	if patternError != nil {
+		log.Errorf("startEpoch/endEpoch Input received is invalid. Value probably modified")
+		return retVal, fmt.Errorf("startEpoch Input received is invalid.")
+	}
```
AndrewHess, line 558: If you get an error here, it's because the `pattern` you defined is invalid regex, not because `inp` didn't match it. Since this `pattern` is hard coded and not provided by the user, you can use `regexp.MustCompile(pattern)` which will panic if there's an error instead of just return the error (this is only ok because we know exactly what the `pattern` is and it's not provided by the user, so we trust it).

pkg/ast/pipesearch/searchHandler.go
```diff
@@ -527,41 +530,64 @@ func convertQueryCountToTotalResponse(qc *structs.QueryCount) interface{} {
...
+
 	strln := len(sanTime)
-	if strln < 6 {
-		return retVal
+	patternMatch := r.MatchString(sanTime)
+	if !patternMatch {
+		//impossible case reached, return error
+		log.Errorf("startEpoch/endEpoch Input received is invalid. Value probably modified")
+		return retVal, fmt.Errorf("startEpoch Input received is invalid.")
 	}
```
AndrewHess, line 566: This is not an impossible case. This is what happens if the input from the user is invalid. So change/remove the comment. Also we don't know if this is for the `startEpoch` or `endEpoch`, and we shouldn't care, so for your error messages you can just say `Input received is invalid.` You can also remove `Value probably modified` (why do you have this?)

pkg/ast/pipesearch/searchHandler.go
```diff
@@ -527,41 +530,64 @@ func convertQueryCountToTotalResponse(qc *structs.QueryCount) interface{} {
...
 
+	//extracting Unit from the string
 	unit := sanTime[strln-1]
 	num, err := strconv.ParseInt(sanTime[4:strln-1], 0, 64)
 	if err != nil {
-		return defValue
+		log.Errorf("startEpoch/endEpoch Printing error below")
+		log.Errorf(err.Error())
+		return defValue, fmt.Errorf("startEpoch Input received is invalid.")
 	}
```
AndrewHess, line 575: Again, don't mention the `start/end epoch` since this function doesn't know about that. You can also remove the error log on line 572. For the one on 573, prepend it with the name of this function so it's easier to track down where this error came from if we see it in our log file when the program is running

Similar comment for lines 585-588

pkg/ast/pipesearch/searchHandler_test.go
AndrewHess, line 1: You're ignoring the `error` field returned by some functions. Instead, you should get it and test it with `assert.Nil(t, err)` (or `assert.NotNil(t, err)` if you expect an error)

pkg/segment/tracing/handler/tracehandler.go
```diff
@@ -30,7 +30,7 @@ func ProcessSearchTracesRequest(ctx *fasthttp.RequestCtx, myid uint64) {
 	}
 
 	nowTs := putils.GetCurrentTimeInMs()
-	searchText, startEpoch, endEpoch, _, _, _ := pipesearch.ParseSearchBody(readJSON, nowTs)
+	searchText, startEpoch, endEpoch, _, _, _, _ := pipesearch.ParseSearchBody(readJSON, nowTs)
```
AndrewHess, line 33: Don't ignore the `error` field. Instead, check if it's not nil, and if so you should log an error message and return early

pkg/ast/pipesearch/searchHandler.go
abhinavg-neu: Got it, will make the change

pkg/ast/pipesearch/searchHandler.go
abhinavg-neu: Yes, I was thinking about the best way to go about the string validation and regex made sense. I shall implement the MustCompile function

pkg/ast/pipesearch/searchHandler.go
abhinavg-neu: Got it

pkg/ast/pipesearch/searchHandler_test.go
abhinavg-neu: Ah, makes sense. I'll do that

pkg/segment/tracing/handler/tracehandler.go
abhinavg-neu: Got it

abhinavg-neu:
Whoops, I believe you must've received a mail for every comment that I just made. Terribly sorry about that!

pkg/ast/pipesearch/searchHandler.go
abhinavg1997: I have replaced it with MustCompile function. I am getting an error with the test cases for this function, I did look into other MustCompile implementations in the codebase but I'm not sure exactly how to go about it. Could use some insights here @AndrewHess

pkg/ast/pipesearch/searchHandler.go
AndrewHess: Around lines 120 and 144 you should return early with the error immediately after emitting the error log. It looks like you're correctly early-returning at the other places I mentioned

pkg/ast/pipesearch/searchHandler.go
AndrewHess: Your use of `regexp.MustCompile` looks correct to me. For the test that's failing, you can check the stack trace and it's an issue with line 86 in the test file. That line is:
```
assert.NotNil(t, nil, err, "expected=%v, actual=%v", expected, err)
```
I think you're just passing `nil` when you shouldn't. You don't even need the extra parameters like `expected=...` because if that test fails you should see that printed anyways. So you should be able to change that line to simply
```
assert.NotNil(t, err)
```

PS. You can run the whole unit test suite locally with `make test`

abhinavg1997: Got it. I've replaced it to:
`assert.NotNil(t, err)`

On another note, while running `make test` , I am looking at some other errors (I just fast forward updated my repo). It looks like so:
`# go.opentelemetry.io/otel/internal/global
../../../go/pkg/mod/go.opentelemetry.io/otel@v1.24.0/internal/global/handler.go:44:18: undefined: atomic.Pointer
../../../go/pkg/mod/go.opentelemetry.io/otel@v1.24.0/internal/global/internal_logging.go:30:25: undefined: atomic.Pointer
note: module requires Go 1.20
# `

Is there some other issue going on?

Macbeth98:
@abhinavg-neu @abhinavg1997  There seem to be some tests failing; once I checked and tried to fix that, 

This is the error: 
`/usr/local/go/bin/go build -o siglens cmd/siglens/main.go
 github.com/siglens/siglens/pkg/segment/tracing/handler
pkg/segment/tracing/handler/tracehandler.go:670:38: assignment mismatch: 6 variables but pipesearch.ParseSearchBody returns 7 values
make: *** [Makefile:33: build] Error 1`

This error is coming from `tracehandler.go` where it calls `pipesearch.ParseSearchBody` but since now this function also returns an `error`, there is a assignment mismatch. Check the image below. And update it in the same way wherever this function is being called. 

And do not forget to do `make pr` before making a push.

image

abhinavg1997:
Hey Mani,

Yes, I had come across that error. I'll look into it and execute the "make
pr" command prior to the push.

PR #700, opened 2024-03-30
Enable tracing in SigLens

Macbeth98:
# Description
- Set new variables in `server.yaml` file that enables the Tracing.
- Created a tracing package that contains the functions required for tracing.
- Enabled tracing for the query server and the tracing middleware for most of the query endpoints.

# Testing
- Tested the tracing against the UI.

# Checklist:


pkg/tracing/tracing.go
```diff
@@ -0,0 +1,157 @@
...
+// InitTracing initializes the OpenTelemetry tracing.
+func InitTracing(serviceName string) func() {
+
+	if !config.IsTracingEnabled() {
+		log.Info("Tracing is disabled")
+		return func() {}
+	}
+
+	if serviceName == "" {
+		log.Errorf("Service name is required to initialize tracing")
```
nkunal, line 73: nit: better error msg `Service name was not provided in config, assuming a default service name as siglens`

pkg/tracing/tracing.go
```diff
@@ -0,0 +1,157 @@
...
+		serviceName = "unknown"
```
nkunal, line 74: use `siglens`

pkg/tracing/tracing.go
```diff
@@ -0,0 +1,157 @@
...
+		trace.WithResource(res),
+	)
+
+	otel.SetTracerProvider(tp)
+
+	return func() {
+		if err := tp.Shutdown(ctx); err != nil {
+			log.Fatalf("Failed to shut down tracer provider: %v", err)
+		}
+	}
```
nkunal, line 120: all of the above `Fatalf` maybe too harsh. If for whatever reason tracing export is not able to initialize, we should log a error, disable tracing but still continue bringing up the app. It is ok to not run with tracing rather than not able to start siglens due to some module issue with tracing

pkg/config/config.go
```diff
@@ -931,24 +968,32 @@ func ProcessForceReadConfig(ctx *fasthttp.RequestCtx) {
...
+		SetConfig(newConfig)
+		configFileLastModified = modifiedTimeSec
+	}
+}
+
+func ProcessForceRefreshConfig() {
+	refreshConfig()
+}
+
+func checkRefreshConfig() {
```
nkunal, line 993: rename this to `runRefreshConfigLoop`

server.yaml
```diff
@@ -44,4 +44,9 @@ tls:
...
 
 # SigLens server hostname
-queryHostname: ""
\ No newline at end of file
+queryHostname: ""
+
+
+tracing:
+  serviceName: "siglens"   # Service name for tracing
+  endpoint: ""  # Endpoint URL for tracing
```
nkunal, line 52: we don't need set the default values here. Leave server.yaml unchanged.

nkunal (dismissed):

pkg/config/config_test.go
```diff
@@ -71,6 +71,10 @@ func Test_ExtractConfigData(t *testing.T) {
  analyticsEnabled: false
  agileAggsEnabled: false
  safeMode: true
+ tracing:
+   enabled: true
```
nkunal, line 75: this needs to be removed

pkg/config/config_test.go
```diff
@@ -140,6 +145,10 @@ func Test_ExtractConfigData(t *testing.T) {
  PQSEnabled: F
  analyticsEnabled: bad string
  AgileAggsEnabled: bad string
+ tracing:
+   enabled: true
```
nkunal, line 149: same

pkg/tracing/tracing.go
```diff
@@ -0,0 +1,181 @@
...
+		log.Errorf("Tracing endpoint is required to initialize tracing. Disabling tracing. Please set the endpoint in the config file.")
+		config.SetTracingEnabled(false)
+		return func() {}
+	}
+
+	ctx := context.Background()
+
+	client := otlptracehttp.NewClient(
+		otlptracehttp.WithEndpointURL(config.GetTracingEndpoint()), // Your collector endpoint
+		otlptracehttp.WithInsecure(),                               // Use WithInsecure for non-production environments
```
AndrewHess, line 111: Should we have a config variable for whether to use `WithInsecure()`?

pkg/tracing/tracing.go
```diff
@@ -0,0 +1,181 @@
...
+		resource.WithAttributes(
+			semconv.ServiceNameKey.String(serviceName),
+		),
+	)
+	if err != nil {
+		log.Errorf("Failed to create resource: %v", err)
+	}
+
+	tp := trace.NewTracerProvider(
+		// trace.WithSampler(trace.AlwaysSample()), // Change this to trace.NeverSample() for production
```
AndrewHess, line 133: I think it would be good to use https://pkg.go.dev/go.opentelemetry.io/otel/sdk/trace#TraceIDRatioBased and have a config variable to specify what fraction of traces should be sampled. Then people can set that config variable as appropriate for their dev, prod, etc. workload

pkg/tracing/tracing.go
Macbeth98: yes, I will add that as a config variable.

pkg/tracing/tracing.go
Macbeth98: Got it.

AndrewHess (approved):

AndrewHess (approved):

PR #701, opened 2024-03-30
Aggregate dependency graphs for requested time period

venkateshamatam:
# Description
This change aggregates all the dependency graphs from a given time period and generates a dep graph on demand without saving it. Auto generate graphs only at the hour boundary.

# Testing
- Auto generated depedency graphs every hour for 3 hours and checked that the response is aggregated according to startEPoch and endEPoch
- Verified that the dep-graph generated for a given time period is equal to the aggregated dep graph for the same time period with hourly granularity.

API: POST `/api/traces/generate-dep-graph`
Request body: `{"startEpoch":"now-1h","endEpoch":"now"}`

API: POST `/api/traces/dependencies`
Request body: `{"startEpoch": "now-1h", "endEpoch": "now"}`

# Checklist:


nkunal:
@AndrewHess please review this

pkg/segment/tracing/handler/tracehandler.go
```diff
@@ -469,25 +469,23 @@ func redMetricsToJson(redMetrics structs.RedMetrics, service string) ([]byte, er
...
 		nextHour := now.Truncate(time.Hour).Add(time.Hour)
 		sleepDuration := time.Until(nextHour)
 
 		time.Sleep(sleepDuration)
-		depMatrix = MakeTracesDependancyGraph()
-		writeDependencyMatrix(depMatrix)
+
+		// Calculate startEpoch and endEpoch for the last hour
+		endEpoch := time.Now().UnixNano() / int64(time.Millisecond)
+		startEpoch := time.Now().Add(-time.Hour).UnixNano() / int64(time.Millisecond)
```
AndrewHess, line 481: Just use `UnixMilli()` instead of `UnixNano() / time.Millisecond`

pkg/segment/tracing/handler/tracehandler.go
```diff
@@ -537,7 +535,10 @@ func MakeTracesDependancyGraph() map[string]map[string]int {
 	return dependencyMatrix
 }
 
-func writeDependencyMatrix(dependencyMatrix map[string]map[string]int) {
+func writeDependencyMatrix(dependencyMatrix map[string]map[string]int, isAutoGenerated bool) {
+	if !isAutoGenerated {
+		return
+	}
```
AndrewHess, line 541: Why did you add the `isAutoGenerated` parameter if the function never does useful work when that is `false`?

pkg/segment/tracing/handler/tracehandler.go
```diff
@@ -621,7 +627,47 @@ func ProcessDependencyRequest(ctx *fasthttp.RequestCtx, myid uint64) {
 		return
 	}
 	ctx.SetStatusCode(fasthttp.StatusOK)
+}
+func ProcessDepgraphRequest(ctx *fasthttp.RequestCtx, myid uint64) {
```
AndrewHess, line 631: nit: can you make the names `ProcessDepgraphRequest` and `ProcessDependencyRequest` more distinguishable, or add a comment above them to specify what they do? Right now the name doesn't clarify that `ProcessDependencyRequest` aggregates already computed graphs while `ProcessDepgraphRequest` builds it from scratch

pkg/segment/tracing/handler/tracehandler.go
venkateshamatam: done i pushed the changes

pkg/segment/tracing/handler/tracehandler.go
venkateshamatam: I removed this, i actually passed false for the `/generate-dep-graph` but removed it later. We don't need this variable.

pkg/segment/tracing/handler/tracehandler.go
venkateshamatam: done, i added comments and updated the method names as well

AndrewHess (approved):

PR #711, opened 2024-04-01
feat: added time picker in dependency graph

navyayy:
# Description
This update introduces a time picker feature for dependency graphs, enabling users to precisely trace dependencies within specified time periods. It facilitates a more granular analysis by generating dependency graphs at hourly intervals.

# Testing
Generated dependency graphs at hour boundary according to the applied time selected in time picker and checked that the response is aggregated according to `startEpoch` and `endEpoch`.

# Checklist:


nbhavana (commented):
1. When you select the time from the time picker, the request payload does not reflect the time selected, it always shows `now-3h`.
image

2. By default, the time picker show 1hr selected, but when you click on the time picker, the 1 hr time is not highlighted.
image

navyayy:
> 1. When you select the time from the time picker, the request payload does not reflect the time selected, it always shows `now-3h`.
> 
> image
> 2. By default, the time picker show 1hr selected, but when you click on the time picker, the 1 hr time is not highlighted.
> 
> image

Resolved

nbhavana (commented):
The changes in the PR are not working as expected. The time picker selection has no effect on the graph. The graph displayed is always the same even when the api returns different values. Also, looks like you are not clearing old graph and 2 graphs are being displayed.
image
image
image

nkunal (approved):

PR #715, opened 2024-04-04
Fix: SPL Head command followed by Stats When querying less than total Segments

Macbeth98:
# Description
- The SPL Head command followed by stats was not working when we query or are given a limit less than the total segments.
- For example, the below queries will not work:
```SPL
* | head 250000 | stats count(*)
* | head 20000 | stats sum(latency)
* | head 30000 | stats sum(latency) by http_method
```
- The above queries won't work assuming that there are more segments than required because of the limit.

#### Fixed a bug in the Group By Recs Aggregation function: `PerformGroupByRequestAggsOnRecs` in `searchaggs.go`. 
- Before, this function only returned the aggregations that were only part of the final segment.
- Now, this is fixed. It now stores the previous aggregations in `NodeResult` in struct and appends data to it for each segment.

# Testing
- Tested the code against the UI.
- Written Tests for Stats and Stats With Group By with Size Limit in `searchaggs_test.go`

# Checklist:

### DIFF
--- a/pkg/segment/search/searchaggs_test.go
+++ b/pkg/segment/search/searchaggs_test.go
@@ -39,7 +39,7 @@ func getMockRecsAndFinalCols(numRecs uint64) (map[string]map[string]interface{},
 // Tests the case where the size limit is zero and the number of segments is 2.
 // Normal case where sizelimit should not be considered.
 func Test_PerformMeasureAggsOnRecsSizeLimit_Zero_MultiSegments(t *testing.T) {
-	numSegements := 2
+	numSegments := 2
 	sizeLimit := 0
 	recsSize := 100
 	nodeResult := &structs.NodeResult{PerformAggsOnRecs: true, RecsAggsType: structs.MeasureAggsType, MeasureOperations: []*structs.MeasureAggregator{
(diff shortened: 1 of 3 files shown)


pkg/segment/search/searchaggs_test.go
```diff
@@ -0,0 +1,363 @@
...
+		resultMap := PerformAggsOnRecs(nodeResult, aggs, recs, finalCols, uint64(numSegements), true, 1)
+
+		assert.Equal(t, uint64(i+1), nodeResult.RecsAggsProcessedSegments, "expected=%v, actual=%v", i+1, nodeResult.RecsAggsProcessedSegments)
+
+		if i == numSegements-1 {
+			assert.Equal(t, 1, len(resultMap), "expected=%v, actual=%v", 1, len(resultMap))
+			assert.True(t, resultMap["CHECK_NEXT_AGG"], "expected=%v, actual=%v", true, resultMap["CHECK_NEXT_AGG"])
+			assert.Equal(t, recsSize, len(recs), "expected=%v, actual=%v", recsSize, len(recs))
+			for _, record := range recs {
+				assert.Equal(t, uint64(2), record["count(*)"], "expected=%v, actual=%v", recsSize*numSegements, record["count(*)"])
```
AndrewHess, line 216: The two expected values (`uint64(2)` and `recsSize*numSegements`) are different

pkg/segment/search/searchaggs_test.go
```diff
@@ -0,0 +1,363 @@
...
+	finalCols["measure2"] = true
+	finalCols["measure3"] = true
+
+	return recs, finalCols
+}
+
+// Tests the case where the size limit is zero and the number of segments is 2.
+// Normal case where sizelimit should not be considered.
+func Test_PerformMeasureAggsOnRecsSizeLimit_Zero_MultiSegments(t *testing.T) {
+	numSegements := 2
```
AndrewHess, line 42: typo: should be `segment` not `segement`. Same comment elsewhere in this PR

AndrewHess (commented):
Looks good, just a few nits to fix

pkg/segment/search/searchaggs_test.go
Macbeth98: Yes, I will update.

AndrewHess (approved):
Looks good

PR #717, opened 2024-04-05
test: add ut for performArithmeticOperation

### LINKED ISSUE #561: [STORY] Add unit test for PerformArithmeticOperation
# Description
- `performArithmeticOperation` function is defined in `segaggs.go`. Add thorough unit testing for this function in `segaggs_test.go` in the same folder

**Note**: You need to name your function(s) so they start with `Test_`; that way Go's unit testing procedure will find that function when it looks for test functions

# Acceptance Criteria
- The added unit tests are clear and thoroughly test the function
- Your new unit tests run and pass when you run `make test` or `make pr` from the siglens base directory
- If adding comments to your code makes it easier to understand, you should add them. If the code is already self-explanatory, you can skip them
- Running `make pr` gives no errors

xmh1011:
# Description
Summarize the change.

Fixes #561 

# Testing
Add new ut test for function `performArithmeticOperation`.

# Checklist:

### DIFF
--- a/pkg/segment/aggregations/segaggs_test.go
+++ b/pkg/segment/aggregations/segaggs_test.go
@@ -18,6 +18,7 @@ package aggregations
 
 import (
 	"encoding/json"
+	"errors"
 	"fmt"
 	"math/rand"
 	"strings"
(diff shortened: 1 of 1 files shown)


xmh1011:
PTAL @nbhavana @AndrewHess @sunitakawane

pkg/segment/aggregations/segaggs_test.go
```diff
@@ -1546,3 +1546,223 @@ func Test_getColumnsToKeepAndRemove(t *testing.T) {
...
+	leftNum := 5
+	rightNum := 3
+	expectedNumResult := float64(leftNum + rightNum)
+	numResult, err := performArithmeticOperation(leftNum, rightNum, utils.Add)
+	if err != nil {
+		t.Errorf("Failed to add numbers: %v", err)
+	}
+	numResult = numResult.(float64)
+	if numResult != expectedNumResult {
+		t.Errorf("Addition of numbers incorrect, got: %v, want: %v", numResult, expectedNumResult)
```
Macbeth98, line 1561: - Use `assert.Equal(t, expected, actual, "error message")` instead of `t.Errorf()`

- Do the same for all others.

pkg/segment/aggregations/segaggs_test.go
```diff
@@ -1546,3 +1546,223 @@ func Test_getColumnsToKeepAndRemove(t *testing.T) {
...
+	}
+
+	// Test addition with strings
+	leftStr := "Hello, "
+	rightStr := "World!"
+	expectedStrResult := "Hello, World!"
+	strResult, err := performArithmeticOperation(leftStr, rightStr, utils.Add)
+	if err != nil {
+		t.Errorf("Failed to concatenate strings: %v", err)
+	}
```
Macbeth98, line 1571: For this use, `assert.Nil()`

Macbeth98 (changes requested):
Also, try to add a test case for any one of the operators where you send the wrong type of data other than string or number types. 

So, for example:

1. (number, string, "Add") // because the function only checks whether the right value is a string or not if the left value is a string. 
2. (string, number, "Sub") 
3. (string, string, "AnyOp") // except Add
4. (anyType, anyType, "AnyOp") // except string or number.

You don't need to do this for all the `OP`; do it for one, and it should be fine. As we are converting the values to `float` before processing the Operation

xmh1011:
> Also, try to add a test case for any one of the operators where you send the wrong type of data other than string or number types.
> 
> So, for example:
> 
> 1. (number, string, "Add") // because the function only checks whether the right value is a string or not if the left value is a string.
> 2. (string, number, "Sub")
> 3. (string, string, "AnyOp") // except Add
> 4. (anyType, anyType, "AnyOp") // except string or number.
> 
> You don't need to do this for all the `OP`; do it for one, and it should be fine. As we are converting the values to `float` before processing the Operation

OK, I'll fix it.

xmh1011:
> Also, try to add a test case for any one of the operators where you send the wrong type of data other than string or number types.
> 
> So, for example:
> 
> 1. (number, string, "Add") // because the function only checks whether the right value is a string or not if the left value is a string.
> 2. (string, number, "Sub")
> 3. (string, string, "AnyOp") // except Add
> 4. (anyType, anyType, "AnyOp") // except string or number.
> 
> You don't need to do this for all the `OP`; do it for one, and it should be fine. As we are converting the values to `float` before processing the Operation

PTAL

Macbeth98 (changes requested):
This is good. Can you add the exact cases below for Add and Subtract?

1. (string(number), string(number), "Sub") // ("2", "5", "Sub"); Do not do it add for Add, as this case is already there.
2. (number, string(number), "Add") // (2, "5", "Add"); Do the same for Sub.

Add these cases and push it. Then I think it is good to merge.

Macbeth98 (approved):
LGTM! 

@AndrewHess Can you once check this PR for any changes?

AndrewHess:
Yep, looks really good

PR #724, opened 2024-04-08
#528: add HasDedupBlock unit test

### LINKED ISSUE #528: [STORY] Add unit tests for HasDedupBlock
# Description
`HasDedupBlock` is defined in segstructs.go. Add thorough unit testing for this function in segstructs_tests.go in the same folder

Note: You need to name your function(s) so they start with `Test_`; that way Go's unit testing procedure will find that function when it looks for test functions

# Acceptance Criteria
* The added unit tests are clear and thoroughly test the function
* Your new unit tests run and pass when you run `make test` or `make pr` from the siglens base directory
* If adding comments to your code makes it easier to understand, you should add them. If the code is already self-explanatory, you can skip them
* Running `make pr` gives no errors

chilah:
# Description
add unit test

Fixes #528 

# Testing
add unit test 

# Checklist:

### DIFF
--- a/pkg/segment/structs/segstructs_test.go
+++ b/pkg/segment/structs/segstructs_test.go
@@ -126,3 +126,69 @@ func Test_HasQueryAggergatorBlock(t *testing.T) {
 
 	assert.Equal(t, true, qa.HasQueryAggergatorBlock(), "Expected true when HasQueryAggergatorBlock is true, got false")
 }
+
+func Test_HasDedupBlock_NilQueryAggregators(t *testing.T) {
+	var qa *QueryAggregators
+
+	assert.Equal(t, false, qa.HasDedupBlock(), "Expected false when QueryAggregators is nil, got true")
+}
+
+func Test_HasDedupBlock_NilOutputTransforms(t *testing.T) {
+	var ot *OutputTransforms
+
+	qa := &QueryAggregators{
+		OutputTransforms: ot,
+	}
+
+	assert.Equal(t, false, qa.HasDedupBlock(), "Expencted false when OutputTransforms is nil, got true")
+}
+
+func Test_HasDedupBlock_NilLetcolumns(t *testing.T) {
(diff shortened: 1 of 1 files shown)


pkg/segment/structs/segstructs_test.go
```diff
@@ -126,3 +126,69 @@ func Test_HasQueryAggergatorBlock(t *testing.T) {
...
+}
+
+func Test_HasDedupBlock_NilOutputTransforms(t *testing.T) {
+	var ot *OutputTransforms
+
+	qa := &QueryAggregators{
+		OutputTransforms: ot,
+	}
+
+	assert.Equal(t, false, qa.HasDedupBlock(), "Expencted false when OutputTransforms is nil, got true")
```
AndrewHess, line 143: typo: You have `Expencted` instead of `Expected`.

Also, you can remove these error messages in `assert.Equal()`; if an assert fails and there isn't a specified message, `assert` will print a simple message that's something like `expected: false, actual value: true`. So feel free to either keep or remove your custom messages, since they'll both provide the same information

AndrewHess (commented):
This looks good. There's just a minor issue to fix before it can get merged

pkg/segment/structs/segstructs_test.go
chilah: Thanks @AndrewHess for your comment, What if i will delete it, So should i delete an error message from HasQueryAggergatorBlock? cause i have put the error message too.

AndrewHess: Yep you can delete your error message instead of fixing the typo. When a custom error message isn't given, the `assert` package will generate one. Also the assert package (https://pkg.go.dev/github.com/stretchr/testify/assert) has some helpful methods like `True()` and `False()`, so here you could simply have `assert.False(t, qa.HasDedupBlock())` (and similar things for your other asserts)

PR #738, opened 2024-04-11
Add Install Age for fresh install and upgrade

venkateshamatam:
# Description
Add install age for fresh install and upgrade

# Checklist:


AndrewHess (commented):
Somewhere you should indicate that the epoch you're reading is in unix millis. Or since we don't really care about the milliseconds, you could compute it as just the normal unix epoch in seconds, and indicate that somewhere. Apart from that this looks ready to merge

AndrewHess (approved):

PR #752, opened 2024-04-16
Index dropdown

sonamgupta21:
# Description
1. The `*` option has been removed from the Index dropdown list. Users can now only select actual index names.
2. When a single index is selected, the dropdown label will display the full name of the selected index. If the index name exceeds the label width, it will be truncated and replaced with the first four characters followed by `...` (e.g., `vect...`).
3. If multiple indexes are selected, the label will display the abbreviation of the first index name followed by "..." and the count of additional indexes selected. For example, if `vector-2024-11-19` is the first selected index and there are two more selected indexes, the label will display `vect...+2`.
4. By default the first sorted index will be active and displayed.

Dashboard Bug Fix : 
1. Vertical scrolling is fixed on the dashboard screen.
2. Each panel displays the corresponding index and do search accordingly.

image
image


nbhavana (commented):
1. When  I land on the first page, I still see the `*`. After that when I select the index, the `*` goes away.
image
2. Same thing happens on the panel edit screen.
image

sonamgupta21:
On the dashboard and logs screens (when opening a saved query) if the saved indexes are present, they are displayed on the label and made active. 
If only some of them are present, only those will be shown on the label and be made active. 
If no indexes are present, the label will display "Index".

image
image

sunitakawane (approved):

PR #757, opened 2024-04-18
Fix : Theme Flickering Issue

sonamgupta21:
# Description

By placing the script in the head tag and directly adding the data-theme attribute to the HTML tag, the theme gets applied before the page content loads. This prevents flickering by ensuring the theme is set immediately when the page loads.

To test this locally, I went to Chrome browser, opened the developer tools, and selected "Slow 3G" under the Network tab.

https://github.com/siglens/siglens/assets/71771131/1e227f8e-f636-4214-b352-95d27594dd36


nbhavana (approved):
lgtm

PR #765, opened 2024-04-23
hide internal traces index

sunitakawane:
# Description
Summarize the change.
cluster stats was showing red-traces and other traces related indexes which are not relevant to cluster stats. Code added to Hide those.

# Testing

# Checklist:

### DIFF
--- a/pkg/health/clusterStatsHandler.go
+++ b/pkg/health/clusterStatsHandler.go
@@ -38,6 +38,8 @@ import (
 	"github.com/valyala/fasthttp"
 )
 
+var excludedInternalIndices = [...]string{"red-traces", "service-dependency"}
+
 func ProcessClusterStatsHandler(ctx *fasthttp.RequestCtx, myid uint64) {
 
 	var httpResp utils.ClusterStatsResponseInfo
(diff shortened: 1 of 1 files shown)


pkg/health/clusterStatsHandler.go
```diff
@@ -217,7 +219,14 @@ func parseIngestionStatsRequest(jsonSource map[string]interface{}) (uint64, usag
...
 }
-
+func isIndexExcluded(indexName string) bool {
+	for _, value := range excludedInternalIndices {
+		if strings.ReplaceAll(indexName, "*", "") == value {
+			return true
+		}
+	}
+	return false
+}
```
AndrewHess, line 229: I think you should require that this function takes only expanded index names, not wildcard index names. Then the `if` can just be `if indexName == value`. If you allow wildcard index names, it's unclear what to do with something like `*-traces` because `red-traces` is excluded but `blue-traces` is not excluded.

You should double check, but I think you're already getting the expanded index names since your iterating the `vtable.GetVirtualTableNames(myid)` that you got on line 237. So then in `isIndexExcluded()` you should probably also do a check to make sure `indexName` doesn't have a `*`.

AndrewHess (approved):

PR #769, opened 2024-04-24
fix: by default 2 option will be selected in search index dropdown

navyayy:
# Description
The default value in the dropdown will be the second option of the list (if available).

# Testing
It can be checked by opening the `Logs` tab in incognito mode, you will see the second option is selected by default (if available)

# Checklist:

### DIFF
--- a/static/js/index.js
+++ b/static/js/index.js
@@ -57,7 +57,7 @@ function renderIndexDropdown(listIndices) {
     if (Cookies.get('IndexList')) {
         selectedSearchIndex = Cookies.get('IndexList');
     }else {
-        if(sortedListIndices.length > 1){
+        if(sortedListIndices.length > 0){
             let selected = sortedListIndices[0].index
     
             if(sortedListIndices[0].index === 'default' && sortedListIndices.length >= 2){
--- a/static/js/index.js
+++ b/static/js/index.js
@@ -57,13 +57,15 @@ function renderIndexDropdown(listIndices) {
     if (Cookies.get('IndexList')) {
         selectedSearchIndex = Cookies.get('IndexList');
     }else {
-        let selected = sortedListIndices[0].index
+        if(sortedListIndices.length > 1){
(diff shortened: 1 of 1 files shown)


static/js/index.js
```diff
@@ -57,7 +57,13 @@ function renderIndexDropdown(listIndices) {
     if (Cookies.get('IndexList')) {
         selectedSearchIndex = Cookies.get('IndexList');
     }else {
-        selectedSearchIndex = sortedListIndices[0].index;
-        $("#index-btn span").html(sortedListIndices[0].index);
+        let selected = sortedListIndices[0].index
+
+        if(sortedListIndices.length >= 2){
+            selected = sortedListIndices[1].index
+        }
```
AndrewHess, line 64: This will always choose the second index if there is a second one. Kunal said it should be the second index if the first one is named `default` (or maybe I'm just unaware of more conversation about this that you had?)

Also it would be good to have another `if` statement to check there's at least one index before you try to get the first one, unless Javascript handles it ok the way you currently have it

navyayy: Made the required changes, you may review now.

AndrewHess:
This looks good to me. I'll wait for a frontend person to review it though

static/js/index.js
```diff
@@ -57,7 +57,15 @@ function renderIndexDropdown(listIndices) {
     if (Cookies.get('IndexList')) {
         selectedSearchIndex = Cookies.get('IndexList');
     }else {
-        selectedSearchIndex = sortedListIndices[0].index;
-        $("#index-btn span").html(sortedListIndices[0].index);
+        if(sortedListIndices.length > 1){
```
sonamgupta21, line 60: It should be `(sortedListIndices.length > 0)` to check at least one index is present.

navyayy: Fixed, please check

sonamgupta21 (approved):

PR #773, opened 2024-04-25
Added support for metric ingestion format through vector

Macbeth98:
# Description
- Updated `ExtractOTSDBPayload` to support fields/formats that come through vector ingestion.

# Testing
- Tested by ingesting metrics through SigClient.
- Tested by ingesting Nginx metrics through vector.

# Checklist:

### DIFF
--- a/pkg/segment/writer/metrics/metricssegment.go
+++ b/pkg/segment/writer/metrics/metricssegment.go
@@ -53,8 +53,13 @@ import (
 )
 
 var otsdb_mname = []byte("metric")
+var metric_name_key = []byte("name")
 var otsdb_timestamp = []byte("timestamp")
 var otsdb_value = []byte("value")
+var metric_value_gauge_key = []byte("gauge")
+var metric_value_counter_key = []byte("counter")
+var metric_value_histogram_key = []byte("histogram")
+var metric_value_summary_key = []byte("summary")
 var otsdb_tags = []byte("tags")
 
 var influx_value = "value"
@@ -477,7 +482,7 @@ func ExtractOTSDBPayload(rawJson []byte, tags *TagsHolder) ([]byte, float64, uin
 
 	handler := func(key []byte, value []byte, valueType jp.ValueType, off int) error {
 		switch {
-		case bytes.Equal(key, otsdb_mname):
+		case bytes.Equal(key, otsdb_mname), bytes.Equal(key, metric_name_key):
 			switch valueType {
 			case jp.String:
 				temp, err := jp.ParseString(value)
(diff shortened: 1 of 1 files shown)


pkg/segment/writer/metrics/metricssegment.go
```diff
@@ -522,6 +527,38 @@ func ExtractOTSDBPayload(rawJson []byte, tags *TagsHolder) ([]byte, float64, uin
...
+					// Determine if the number is in seconds or milliseconds
+					if toputils.IsTimeInMilli(uint64(t)) {
+						ts = uint32(t / 1000)
+					} else {
+						ts = uint32(t)
+					}
+				}
+
+				// Parse the string to time using time.Parse and multiple layouts.
+				layouts := []string{
```
AndrewHess, line 542: If you don't get an error on line 532, you can break out of this early before continuing with parsing it according to various RFCs, right?

pkg/segment/writer/metrics/metricssegment.go
```diff
@@ -53,8 +53,13 @@ import (
 )
 
 var otsdb_mname = []byte("metric")
+var metric_name_key = []byte("name")
 var otsdb_timestamp = []byte("timestamp")
 var otsdb_value = []byte("value")
+var metric_value_gauge_key = []byte("gauge")
+var metric_value_counter_key = []byte("counter")
+var metric_value_histogram_key = []byte("histogram")
+var metric_value_summary_key = []byte("summary")
```
nkunal, line 62: maybe we should name this vars `..._keyname`

pkg/segment/writer/metrics/metricssegment.go
```diff
@@ -522,6 +527,38 @@ func ExtractOTSDBPayload(rawJson []byte, tags *TagsHolder) ([]byte, float64, uin
...
+
+				// Parse the string to time using time.Parse and multiple layouts.
+				layouts := []string{
+					time.RFC3339,
+					time.RFC3339Nano,
+					time.RFC1123,
+					time.RFC1123Z,
+					time.RFC822,
+					time.RFC822Z,
+					time.RFC850,
```
nkunal, line 549: you are defining this layouts arr again and again during parsing, it will put pressure on GC. you should make this a const at the top of this file and just reuse it

pkg/segment/writer/metrics/metricssegment.go
Macbeth98: Ahh, yes, I added a return inside the `If` block.

AndrewHess (approved):

PR #788, opened 2024-05-01
added api support for delete index

sunitakawane:
# Description
Delete `test` index using curl command
curl -X "DELETE" http://localhost:5122/elastic/test

# Testing

# Checklist:

### DIFF
--- a/pkg/es/writer/esBulkHandler.go
+++ b/pkg/es/writer/esBulkHandler.go
@@ -36,6 +36,7 @@ import (
 
 	// segstructs "github.com/siglens/siglens/pkg/segment/structs"
 
+	"github.com/siglens/siglens/pkg/segment/query/metadata"
 	vtable "github.com/siglens/siglens/pkg/virtualtable"
 	log "github.com/sirupsen/logrus"
 	"github.com/valyala/fasthttp"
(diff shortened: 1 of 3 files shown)


pkg/es/writer/esBulkHandler.go
```diff
@@ -318,3 +319,56 @@ func PostBulkErrorResponse(ctx *fasthttp.RequestCtx) {
 	responsebody["status"] = 400
 	utils.WriteJsonResponse(ctx, responsebody)
 }
+
+func ProcessDeleteIndex(ctx *fasthttp.RequestCtx, myid uint64) {
```
AndrewHess, line 323: It seems like this is accepting wildcard index names. Is that right? If so, can you write a comment saying that?

pkg/es/writer/esBulkHandler.go
```diff
@@ -318,3 +319,56 @@ func PostBulkErrorResponse(ctx *fasthttp.RequestCtx) {
...
+		indexPresent := vtable.IsVirtualTablePresent(&indexName, myid)
+		if !indexPresent {
+			indicesNotFound++
+			continue
+		}
+
+		ok, _ := vtable.IsAlias(indexName, myid)
+		if ok {
+			alias, _ := vtable.GetAliasesAsArray(indexName, myid)
+			error := vtable.RemoveAliases(indexName, alias, myid)
```
AndrewHess, line 358: nit: rename `alias` to `aliases`, since it's a list

AndrewHess, line 358: I think this is not right. If `indexName` is an alias, `ok==true` but then you call `GetAliasesAsArray(indexName, myid)`; shouldn't you call it on the non-aliased index name instead? Something like:
```
ok, baseIndexName := vtable.IsAlias(indexName, myid)
if ok {
	alias, _ := vtable.GetAliasesAsArray(baseIndexName, myid)
	error := vtable.RemoveAliases(baseIndexName, alias, myid)
...
```

Or do we want to delete just this alias but not the actual underlying index?

pkg/es/writer/esBulkHandler.go
```diff
@@ -318,3 +319,56 @@ func PostBulkErrorResponse(ctx *fasthttp.RequestCtx) {
...
+	inIndexName := utils.ExtractParamAsString(ctx.UserValue("indexName"))
+
+	convertedIndexNames, indicesNotFound := deleteIndex(inIndexName, myid)
+
+	if indicesNotFound == len(convertedIndexNames) {
+		ctx.SetStatusCode(fasthttp.StatusNotFound)
+		response := make(map[string]interface{})
+		final := make(map[string]interface{})
+		var items = make([]interface{}, 0)
+		items = append(items, *utils.NewDeleteIndexErrorResponseInfo(inIndexName))
```
AndrewHess, line 333: So `items` will just have one element? Do you need to iterate over the `convertedIndexNames`?

pkg/es/writer/esBulkHandler.go
```diff
@@ -318,3 +319,56 @@ func PostBulkErrorResponse(ctx *fasthttp.RequestCtx) {
...
+		final["root_cause"] = items
+		final["error"] = *utils.NewDeleteIndexErrorResponseInfo(inIndexName)
+		final["status"] = 404
+		response["error"] = final
+		utils.WriteJsonResponse(ctx, response)
+	} else {
```
AndrewHess, line 339: Is there an HTTP response code for a partial success? If so we should send that when some indexes get deleted but others do not

pkg/es/writer/esBulkHandler.go
```diff
@@ -318,3 +319,56 @@ func PostBulkErrorResponse(ctx *fasthttp.RequestCtx) {
...
+		ctx.SetStatusCode(fasthttp.StatusNotFound)
+		response := make(map[string]interface{})
+		final := make(map[string]interface{})
+		var items = make([]interface{}, 0)
+		items = append(items, *utils.NewDeleteIndexErrorResponseInfo(inIndexName))
+		final["root_cause"] = items
+		final["error"] = *utils.NewDeleteIndexErrorResponseInfo(inIndexName)
+		final["status"] = 404
+		response["error"] = final
+		utils.WriteJsonResponse(ctx, response)
```
AndrewHess, line 338: Why is there this extra structure of `final.error`, `final.status`, `final.root_cause`? Why don't we have only `response.error` like we do in most error handling for other APIs? (It could be fine to keep this structure, I'm just not sure why we have it)

sunitakawane: removed extra structure

PR #795, opened 2024-05-01
Retention based changes

venkateshamatam:
# Description
- Delete indices without any segments as part of retention.
- Change retention to first run after 1hr

# Checklist:

### DIFF
--- a/pkg/retention/retention.go
+++ b/pkg/retention/retention.go
@@ -162,8 +162,8 @@ func DoRetentionBasedDeletion(ingestNodeDir string, retentionHours int, orgid ui
 	deleteEmptyIndices(ingestNodeDir, 0)
 }
 
-func deleteEmptyIndices(ingestNodeDir string, orgid uint64) {
-	allIndices, err := vtable.GetVirtualTableNames(orgid)
+func deleteEmptyIndices(ingestNodeDir string, myid uint64) {
+	allIndices, err := vtable.GetVirtualTableNames(myid)
 	if err != nil {
 		log.Errorf("DeleteEmptyIndices: Error in getting virtual table names, err: %v", err)
 		return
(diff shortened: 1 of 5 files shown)


pkg/retention/retention.go
```diff
@@ -159,6 +159,42 @@ func DoRetentionBasedDeletion(ingestNodeDir string, retentionHours int, orgid ui
 	// Delete all segment data
 	DeleteSegmentData(currentSegmeta, segmentsToDelete, true)
 	DeleteMetricsSegmentData(currentMetricsMeta, metricSegmentsToDelete, true)
+	DeleteEmptyIndices(ingestNodeDir)
+}
+
+func DeleteEmptyIndices(ingestNodeDir string) {
+	allIndices, err := vtable.GetVirtualTableNames(0)
```
AndrewHess, line 166: I think this should be `GetVirtualTableNames(orgId)`, right?

AndrewHess (commented):
I think there would be an issue if `DeleteEmptyIndices()` gets called and it starts deleting an index but just then, more data gets ingested into that index. Would we lose some data in that case? You may need to hold a lock during part of this, unless it's not possible for something bad to happen in that case

pkg/retention/retention.go
```diff
@@ -159,21 +159,18 @@ func DoRetentionBasedDeletion(ingestNodeDir string, retentionHours int, orgid ui
 	// Delete all segment data
 	DeleteSegmentData(currentSegmeta, segmentsToDelete, true)
 	DeleteMetricsSegmentData(currentMetricsMeta, metricSegmentsToDelete, true)
-	DeleteEmptyIndices(ingestNodeDir)
+	deleteEmptyIndices(ingestNodeDir, 0)
```
AndrewHess, line 162: Pass with `orgid` (from line 78) instead of `0`

nkunal (approved):

PR #800, opened 2024-05-02
feat: Read Metric Names per segment

Macbeth98:
# Description
- Read Metric Names per Segment.

# Testing
- Tested by writing an unit test case that Flushes the Metrics Map to a file and then reads from the file.

# Checklist:

### DIFF
--- a/pkg/segment/reader/metrics/series/seriesreader.go
+++ b/pkg/segment/reader/metrics/series/seriesreader.go
@@ -333,3 +333,45 @@ func (tssr *TimeSeriesSegmentReader) loadTSGFile(fileName string, rbuf []byte) (
 	}
 	return rbuf, nil
 }
+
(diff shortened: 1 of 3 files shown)


pkg/segment/reader/metrics/series/seriesreader.go
```diff
@@ -333,3 +333,45 @@ func (tssr *TimeSeriesSegmentReader) loadTSGFile(fileName string, rbuf []byte) (
...
+	defer fd.Close()
+
+	finfo, err := fd.Stat()
+	if err != nil {
+		log.Errorf("GetAllMetricNames: error when trying to stat file=%+v. Error=%+v", filePath, err)
+		return nil, err
+	}
+
+	fileSize := finfo.Size()
+	buf := make([]byte, fileSize)
```
nkunal, line 356: in next PR (todo) , you should try to use buffer pools for such kinds of memory accesses, it will reduce GC pressures.

nkunal (approved):

PR #801, opened 2024-05-02
Update install.sh

gnubyte:
fixed pip install because python pip made some dumb change where you have to manually add --break-system-packages despite 10-15 years of this working fine

# Description
Summarize the change.

# Testing

# Checklist:

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -506,7 +506,7 @@ install_podman_compose() {
     install_python_and_pip
 
     # Install podman-compose with pip
-    $sudo_cmd pip3 install podman-compose || {
+    $sudo_cmd pip3 install podman-compose --break-system-packages || {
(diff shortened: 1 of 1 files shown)


Macbeth98:
Hi @gnubyte , Thanks for making the PR and informing us about this change. 

Yeah, it seems from Python version 3.11, this change is applied. 
So I was wondering, can you modify the above one by following this [Stack Overflow](https://stackoverflow.com/a/75696359) may be 1st or 2nd bullet point in the accepted answer?

image
 
So to do it as per the second way, may be we can do something like this?

```shell
...

python3 -m venv .venv
source .venv/bin/activate
python3 -m pip install podman
deactivate

... # continue with rest of the script
```

Let me know your thoughts. Thanks!

PR #807, opened 2024-05-03
feat: Metrics Get All Tags API 

Macbeth98:
# Description
- Get All Tag Keys and values Set for the given Metric Name over the time range.

# Testing
- Tested the code by calling the API Endpoint through Postman.

# Checklist:

### DIFF
--- a/pkg/integrations/prometheus/promql/metricsSearchHandler.go
+++ b/pkg/integrations/prometheus/promql/metricsSearchHandler.go
@@ -370,15 +370,14 @@ func ProcessGetAllMetricTagsRequest(ctx *fasthttp.RequestCtx, myid uint64) {
 		return
 	}
 
-	start, ok := readJSON["start"].(float64)
-	if !ok {
-		utils.SendError(ctx, "Failed to parse 'start' from JSON body", fmt.Sprintf("ProcessGetAllMetricTagsRequest: Failed to parse 'start' from JSON body with value: %v", readJSON["start"]), errors.New("failed to parse 'start' from JSON body"))
+	startTime, err := parseTimeStringToUint32(readJSON["start"])
+	if err != nil {
+		utils.SendError(ctx, "Failed to parse 'start' from request body", fmt.Sprintf("ProcessGetAllMetricsRequest: Failed to parse 'start' from JSON body with value: %v", readJSON["start"]), errors.New("failed to parse 'start' from JSON body"))
 		return
 	}
-
-	end, ok := readJSON["end"].(float64)
-	if !ok {
-		utils.SendError(ctx, "Failed to parse 'end' from JSON body", fmt.Sprintf("ProcessGetAllMetricTagsRequest: Failed to parse 'end' from JSON body with value: %v", readJSON["end"]), errors.New("failed to parse 'end' from JSON body"))
+	endTime, err := parseTimeStringToUint32(readJSON["end"])
+	if err != nil {
+		utils.SendError(ctx, "Failed to parse 'end' from request body", fmt.Sprintf("ProcessGetAllMetricsRequest: Failed to parse 'end' from JSON body with value: %v", readJSON["end"]), errors.New("failed to parse 'end' from JSON body"))
 		return
 	}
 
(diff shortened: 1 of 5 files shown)


pkg/segment/results/mresults/metricresults.go
```diff
@@ -375,6 +375,36 @@ func (r *MetricsResult) GetResultsPromQl(mQuery *structs.MetricsQuery, pqlQueryt
...
+	tagKeyValueSet := make([]string, 0)
+
+	for _, series := range res.AllSeries {
+		tagValues := strings.Split(series.grpID.String(), tsidtracker.TAG_VALUE_DELIMITER_STR)
+		if len(uniqueTagKeys) != len(tagValues)-1 {
+			err := fmt.Errorf("GetResults: the length of tag key and tag value pair must match. UniqueTagKeys length: %v,  TagValues Length: %v", len(uniqueTagKeys), len(tagValues)-1)
+			return nil, nil, err
+		}
+		for index, val := range tagValues[:len(tagValues)-1] {
+			tagKeyValueSet = append(tagKeyValueSet, fmt.Sprintf("%s:%s", uniqueTagKeys[index], val))
```
Macbeth98, line 401: - Should I return here? Or Log the Error and continue to process the next series?

- Also, wouldn't there be a case or scenario where the Unique Tag Keys for the Metric may not match with the Tag Values?
   - I meant to say that for a Metric (`Metric_1`), some values may include two tag keys and values. For example: `method: post, respCode: 200`. But some may only include `method: post` with no `respCode`. In this scenario , the `uniqueTagKeysMap` will have both `method` and `respCode`, but for some series, there will not be value for `respCode`.
   - Or is there a general assumption that the above scenario won't happen. Meaning for a given Metric, the tag set will remain constant?

pkg/integrations/prometheus/promql/metricsSearchHandler.go
```diff
@@ -390,10 +392,36 @@ func ProcessGetAllMetricTagsRequest(ctx *fasthttp.RequestCtx, myid uint64) {
...
+
+	searchText := fmt.Sprintf("(%v)", metricName)
+
+	metricQueryRequest, _, _, err := convertPqlToMetricsQuery(searchText, timeRange.StartEpochSec, timeRange.EndEpochSec, myid)
+	if err != nil {
+		utils.SendError(ctx, "Failed to parse the Metric Name as a Query", fmt.Sprintf("Metric Name: %+v; qid: %v", metricName, qid), err)
+		return
+	}
+
+	metricQueryRequest[0].MetricsQuery.ExitAfterRawSearch = true
```
nkunal, line 408: nit: maybe rename this field to `ExitAfterTagsSearch`

pkg/integrations/prometheus/promql/metricsSearchHandler.go
```diff
@@ -390,10 +392,36 @@ func ProcessGetAllMetricTagsRequest(ctx *fasthttp.RequestCtx, myid uint64) {
...
+
+	segment.LogMetricsQuery("PromQL metrics query parser", &metricQueryRequest[0], qid)
```
nkunal, line 410: nit: log as "Tags Request PromQL...`

pkg/integrations/prometheus/promql/metricsSearchHandler.go
```diff
@@ -390,10 +392,36 @@ func ProcessGetAllMetricTagsRequest(ctx *fasthttp.RequestCtx, myid uint64) {
...
+	res := segment.ExecuteMetricsQuery(&metricQueryRequest[0].MetricsQuery, &metricQueryRequest[0].TimeRange, qid)
+
+	uinqueTagKeys, tagKeyValueSet, err := res.GetMetricTagsResultSet(&metricQueryRequest[0].MetricsQuery)
```
nkunal, line 413: typo in the name of the uniqueTag.. var

pkg/segment/results/mresults/metricresults.go
```diff
@@ -375,6 +375,36 @@ func (r *MetricsResult) GetResultsPromQl(mQuery *structs.MetricsQuery, pqlQueryt
...
+			uniqueTagKeys = append(uniqueTagKeys, tag.TagKey)
+		}
+	}
+
+	tagKeyValueSet := make([]string, 0)
+
+	for _, series := range res.AllSeries {
+		tagValues := strings.Split(series.grpID.String(), tsidtracker.TAG_VALUE_DELIMITER_STR)
+		if len(uniqueTagKeys) != len(tagValues)-1 {
+			err := fmt.Errorf("GetResults: the length of tag key and tag value pair must match. UniqueTagKeys length: %v,  TagValues Length: %v", len(uniqueTagKeys), len(tagValues)-1)
```
nkunal, line 397: func name in log is not right

nkunal (dismissed):

nkunal (approved):

PR #810, opened 2024-05-06
get unique tag key value pairs for metrics

sunitakawane:
# Description
Summarize the change.

# Testing

# Checklist:

### DIFF
--- a/pkg/segment/results/mresults/metricresults.go
+++ b/pkg/segment/results/mresults/metricresults.go
@@ -389,7 +389,7 @@ func (res *MetricsResult) GetMetricTagsResultSet(mQuery *structs.MetricsQuery) (
 		}
 	}
 
-	tagKeyValueSet := make([]string, 0)
+	tagKeyValueSet := make(map[string]bool)
 
 	for _, series := range res.AllSeries {
 		tagValues := strings.Split(series.grpID.String(), tsidtracker.TAG_VALUE_DELIMITER_STR)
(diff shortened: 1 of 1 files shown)


pkg/segment/results/mresults/metricresults.go
```diff
@@ -389,7 +389,7 @@ func (res *MetricsResult) GetMetricTagsResultSet(mQuery *structs.MetricsQuery) (
 		}
 	}
 
-	tagKeyValueSet := make([]string, 0)
+	tagKeyValueSet := make(map[string]bool)
```
AndrewHess, line 392: You never need these `bool` values, right? If not, I think it would be clearer to make this map to empty structs. Then the type could be `map[string]struct{}` and you set values like `myMap["apple"] = struct{}{}`

pkg/segment/results/mresults/metricresults.go
```diff
@@ -398,11 +398,15 @@ func (res *MetricsResult) GetMetricTagsResultSet(mQuery *structs.MetricsQuery) (
...
 		}
 		for index, val := range tagValues[:len(tagValues)-1] {
-			tagKeyValueSet = append(tagKeyValueSet, fmt.Sprintf("%s:%s", uniqueTagKeys[index], val))
+			tagKeyValueSet[fmt.Sprintf("%s:%s", uniqueTagKeys[index], val)] = true
 		}
 	}
+	uniqueTagKeysSet := make([]string, 0)
+	for key := range tagKeyValueSet {
+		uniqueTagKeysSet = append(uniqueTagKeysSet, key)
+	}
```
AndrewHess, line 407: Might want to move this into the utils package if you think we'll convert sets to slices in other places

pkg/segment/results/mresults/metricresults.go
nkunal: ah  I like this!

sunitakawane:
added to PR #811

PR #845, opened 2024-05-10
Feat: Add promql funcs increase/delta/idelta

Jiale-Fang:
# Description
Add promql funcs increase/delta/idelta, based on https://stackoverflow.com/questions/54494394/do-i-understand-prometheuss-rate-vs-increase-functions-correctly and https://prometheus.io/docs/prometheus/latest/querying/functions/#delta

# Testing

# Checklist:


pkg/segment/results/mresults/seriesresult.go
```diff
@@ -361,41 +361,28 @@ func ApplyRangeFunction(ts map[uint32]float64, function structs.Function) (map[u
...
 		var dx, dt float64
-		resetIndex := -1
 		for i := 1; i < len(sortedTimeSeries); i++ {
-			timeWindowStartTime := sortedTimeSeries[i].downsampledTime - timeWindow
-			preIndex := sort.Search(len(sortedTimeSeries), func(j int) bool {
-				return sortedTimeSeries[j].downsampledTime >= timeWindowStartTime
-			})
-
-			if i <= preIndex { // Can not find the second point within the time window
+			timeDff := sortedTimeSeries[i].downsampledTime - sortedTimeSeries[i-1].downsampledTime
```
AndrewHess, line 370: Are you supposed to use the downsampled time here for `irate` that you're doing? If so, you'll need to handle the case `timeDff == 0` so you don't divide by zero on line 388 when you do `ts[sortedTimeSeries[i].downsampledTime] = dx / dt`

pkg/segment/results/mresults/seriesresult.go
```diff
@@ -404,37 +391,42 @@ func ApplyRangeFunction(ts map[uint32]float64, function structs.Function) (map[u
...
+			preIndex := sort.Search(len(sortedTimeSeries), func(j int) bool {
+				return sortedTimeSeries[j].downsampledTime >= timeWindowStartTime
+			})
 
-		var dx, dt float64
+			if i <= preIndex { // Can not find the second point within the time window
+				delete(ts, sortedTimeSeries[i].downsampledTime)
+				continue
+			}
+			ts[sortedTimeSeries[i].downsampledTime] = sortedTimeSeries[i].dpVal - sortedTimeSeries[preIndex].dpVal
```
AndrewHess, line 413: Do you have to handle resets with `Delta`?

pkg/segment/results/mresults/seriesresult.go
```diff
@@ -642,3 +634,44 @@ func convertStrToFloat64(toNearestStr string) (float64, error) {
...
+		if i <= preIndex { // Can not find the second point within the time window
+			delete(ts, sortedTimeSeries[i].downsampledTime)
+			continue
+		}
+
+		if sortedTimeSeries[i].dpVal < sortedTimeSeries[i-1].dpVal {
+			// This metric was reset.
+			dx = sortedTimeSeries[i].dpVal
+			dt = float64(sortedTimeSeries[i].downsampledTime - sortedTimeSeries[i-1].downsampledTime)
+			ts[sortedTimeSeries[i].downsampledTime] = dx / dt
```
AndrewHess, line 659: I think `dt` could be 0, which would cause problems. Or is that not possible? What is supposed to happen in that case?

Same comment for lines 670-671

pkg/segment/results/mresults/seriesresult.go
Jiale-Fang: I think if there are elements with different indexes in sortedTimeSeries, timeDiff would not be the same because ts is a map and sortedTimeSeries was converted from that map.

https://github.com/siglens/siglens/blob/d42e43040c7e596fffb426e05455f49f6b78e337/pkg/segment/results/mresults/seriesresult.go#L289-L303

pkg/segment/results/mresults/seriesresult.go
Jiale-Fang: - Rate should only be used with counters and native histograms where the components behave like counters
- Delta should only be used with gauges and native histograms where the components behave like gauges (so-called gauge histograms).

Based on [doc](https://prometheus.io/docs/concepts/metric_types/)
> A counter is a cumulative metric that represents a single [monotonically increasing counter], while gauge can arbitrarily go up and down.

So i think we do not need to handle reset logic for a gauge metric

pkg/segment/results/mresults/seriesresult.go
Jiale-Fang: `dt = float64(sortedTimeSeries[i].downsampledTime - sortedTimeSeries[i-1].downsampledTime)`
`dt = float64(sortedTimeSeries[i].downsampledTime - sortedTimeSeries[preIndex].downsampledTime)`

I have a if statement at the beginning to make sure preIndex would not equal to i. So in these two situations, those two elements have different indexes. Which is the same as i mentioned in the first comment.

nkunal (approved):

PR #852, opened 2024-05-13
Loki label values api

Macbeth98:
# Description
- Added parsing and handling of Arguments for Loki Label Values API
- Added a fix for Measure Column Exists check before accessing it. 

# Testing
- Tested by executing the API in Postman

# Checklist:

### DIFF
--- a/pkg/ast/common.go
+++ b/pkg/ast/common.go
@@ -202,18 +202,44 @@ func CreateTermFilterCriteria(k string, v interface{}, opr FilterOperator, qid u
 	return &criteria
 }
 
-// Executes simple query to return a single column values in a given table
-func GetColValues(cname string, table string, qid uint64, orgid uint64) ([]interface{}, error) {
+func getDefaultAstAndAggNode(qid uint64, timeRange *dtu.TimeRange) (*structs.ASTNode, *structs.QueryAggregators, error) {
 	aggNode := structs.InitDefaultQueryAggregations()
-	astNode, err := query.GetMatchAllASTNode(qid)
+	astNode, err := query.GetMatchAllASTNode(qid, timeRange)
 	if err != nil {
 		log.Errorf("qid=%v, GetColValues: match all ast node failed! %+v", qid, err)
-		return nil, err
+		return nil, nil, err
 	}
 	aggNode.OutputTransforms = &structs.OutputTransforms{OutputColumns: &structs.ColumnsRequest{}}
+
+	return astNode, aggNode, nil
+}
+
+// Executes simple query to return a single column values in a given table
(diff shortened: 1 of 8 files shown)


pkg/ast/common.go
```diff
@@ -202,18 +202,44 @@ func CreateTermFilterCriteria(k string, v interface{}, opr FilterOperator, qid u
...
+		if aggNode.OutputTransforms.OutputColumns == nil {
+			aggNode.OutputTransforms.OutputColumns = &structs.ColumnsRequest{}
+		}
+	}
+
 	aggNode.OutputTransforms.OutputColumns.IncludeColumns = append(make([]string, 0), cname)
 
-	ti := structs.InitTableInfo(table, orgid, false)
+	ti := structs.InitTableInfo(indexNameIn, orgid, false)
+	log.Infof("qid=%v, GetColValues: table info %+v", qid, ti.GetQueryTables())
```
nkunal, line 242: do we need this log line ?

pkg/integrations/loki/loki.go
```diff
@@ -56,6 +56,7 @@ const (
 	DefaultLimit       = 100
 	MsToNanoConversion = 1_000_000
 	DAY_IN_MS          = 86400000
+	HOUR_IN_MS         = 3600000
```
nkunal, line 59: use `_` as the thousands seperator for this and the above day_in_ms, makes code reading easier

pkg/segment/aggregations/segaggs.go
```diff
@@ -1580,7 +1580,11 @@ func performStatisticColRequestOnMeasureResults(nodeResult *structs.NodeResult,
 	resTotal := uint64(0)
 	if letColReq.StatisticColRequest.StatisticOptions.ShowPerc {
 		for _, bucketHolder := range nodeResult.MeasureResults {
-			resTotal += bucketHolder.MeasureVal[countName].(uint64)
+			value, exists := bucketHolder.MeasureVal[countName]
+			if !exists {
+				continue
```
nkunal, line 1585: if value does not exist and this is a new bucket key, then we should create an entry in the bucket for this new key

nkunal: actually ignore this comment, we are just counting the resTotal if there is value for this bucket key

pkg/ast/common.go
Macbeth98: No I will delete it.

pkg/segment/aggregations/segaggs.go
Macbeth98: But we are actually accessing the same key from the same map below. So it might be better to create an entry here?

PR #857, opened 2024-05-14
fix: filename of log and traces export will be 3-254 chars

### LINKED ISSUE #851: [BUG] Filename limited to max. 16 characters when exporting logs
# Description
Filename of log export can not be longer than 16 characters.

# Steps to Reproduce
Go to the GUI -> Logs and then hit the export button. Now specify a filename longer than 16 characters. You will get an error that the filename should be between 3 and 16 characters. On modern file systems such a limitiation of the filename length shouldn't be necessary.
grafik

# Expected Behavior
On modern file systems a limitation of a filename to 16 characters shouldn't be necessary any more. I suggest to increase the limit to 254 characters.

# Environment
- Docker Image: based on siglens 0.2.4 (siglens/siglens:latest, f2cc2b0b8d6b)
- OS: Ubuntu 24.04, x86_64
- Browser and Version (if applicable): Microsoft Edge, Version 124.0.2478.80 (Official build) (64-bit)

navyayy:
# Description
Earlier, filename of logs and traces export were limited between 3 to 16 characters but now the limit has been extended to 254 characters.

Fixes #851 

# Testing
I have tested by exporting logs and traces with filename of 100 characters and also for 280 characters. Logs and traces exported in the former case while it produced error for the latter one.

# Checklist:

### DIFF
--- a/static/component/download/download.js
+++ b/static/component/download/download.js
@@ -231,7 +231,7 @@ $("#cancel-loading").on("click", cancelDownload);
       allFields.removeClass("ui-state-error");
       tips.removeClass("ui-state-highlight");
       tips.text("");
-      valid = valid && checkLength(qname, "download name", 3, 16);
+      valid = valid && checkLength(qname, "download name", 3, 254);
       valid =
         valid &&
         checkRegexp(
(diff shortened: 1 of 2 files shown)


static/component/download/download.js
```diff
@@ -231,7 +231,7 @@ $("#cancel-loading").on("click", cancelDownload);
       allFields.removeClass("ui-state-error");
       tips.removeClass("ui-state-highlight");
       tips.text("");
-      valid = valid && checkLength(qname, "download name", 3, 16);
+      valid = valid && checkLength(qname, "download name", 3, 254);
```
nkunal, line 234: min length can be 1

nkunal (approved):

PR #858, opened 2024-05-14
Update promql response structure

venkateshamatam:
# Description
Update promql response structure.
Update promql time parsing to handle float values

# Testing
Tested using postman 

# Checklist:


nkunal (commented):
lgtm

pkg/segment/results/mresults/metricresults.go
```diff
@@ -371,37 +366,26 @@ func (r *MetricsResult) GetResultsPromQl(mQuery *structs.MetricsQuery, pqlQueryt
 				err := errors.New("GetResultsPromQl: the length of tag key and tag value pair must match")
```
Macbeth98, line 366: Why are we making this check? The Tag Keys and the Tag Values may not be the same.
Or is there something I missed?

Macbeth98: Also, The Tag Keys is always unique, if it goes through the same Metrics Query Flow as SigLens. The `mQuery.ReorderTagFilters()` will make sure that the Tag Keys in the TagFilters are unique.
So you don't need `uniqueTagKeys` Map.

Macbeth98 (approved):

PR #903, opened 2024-05-17
feat: added error handling to ParseSearchBody function

### LINKED ISSUE #534: [STORY] Add error handling to ParseSearchBody
# Description
`ParseSearchBody()` is defined in pkg/ast/pipesearch/searchHandler.go. This function takes some a map (which is keys/values extracted from a JSON request) and then converts each field to the expected type and returns the results. You should add error handling to this function so that it returns an error if it's unable to convert a value to the expected type for that field, or if it gets a field name that it does not expect.

# Acceptance Criteria
* The new version of ParseSearchBody should log an error and return the error if it's unable to convert a value to the expected type, or if it gets a field name that it doesn't expect. Otherwise, it should continue working like it already does
* If adding comments to your code makes it easier to understand, you should add them. If the code is already self-explanatory, you can skip them
* Running `make pr` from the siglens base directory gives no errors

gouthamhusky:
# Description
Added error handling to the `ParseSearchBody()` fn in `searchHandler.go`. While we are returning from the function on account of parsing errors of the various fields in the JSON payload, there is still room for improvement in terms of handling the error in the callers of `ParseSearchBody`

Fixes #534 

# Testing
Added more scenarios in the `Test_parseSearchBody()` method. All of them are pretty much self-explanatory

# Checklist:

- [✅] I have self-reviewed this PR.
- [✅] I have removed all print-debugging and commented-out code that should not be merged.
- [✅] I have added sufficient comments in my code, particularly in hard-to-understand areas.
- [✅] I have formatted the code, if applicable. For Go, I have run `goimports -w .`.


pkg/ast/pipesearch/searchHandler.go
```diff
@@ -55,19 +55,39 @@ Example incomingBody
...
+
+	knownFields := map[string]bool{
+		"searchText": true,
+		"indexName":  true,
+		"startEpoch": true,
+		"endEpoch":   true,
+		"size":       true,
+		"scroll":     true,
+		"from":       true,
+	}
```
nkunal, line 72: This is not correct. Some of these fields are optional . This is probably why the end 2 end tests might be failing

PR #915, opened 2024-05-20
Influx API handlers

venkateshamatam:
# Description
Influx API handlers

# Testing
- Tested using postman
# Checklist:


pkg/influx/query/query.go
```diff
@@ -0,0 +1,48 @@
...
+					]
+				}
+			]
+		}`
+		ctx.SetContentType("application/json")
+		_, err := ctx.Write([]byte(databaseResponse))
+		if err != nil {
+			log.Errorf("GetQueryHandler: failed to write response, err=%v", err)
+		}
+	} else if q == "CREATE DATABASE benchmark WITH REPLICATION 1" {
```
nkunal, line 33: we should just check if `q.contains("CREATE DATABASE")` 
we don't care about what the dbName is and we should just return 200 for this

pkg/influx/query/query.go
```diff
@@ -0,0 +1,48 @@
...
+		// Return status code 200
+		ctx.SetStatusCode(fasthttp.StatusOK)
+	} else {
+		ctx.Error("Unsupported query", fasthttp.StatusBadRequest)
+	}
+}
+func PostQueryHandler(ctx *fasthttp.RequestCtx, myid uint64) {
+	q := string(ctx.QueryArgs().Peek("q"))
+	if q == "drop database benchmark" {
```
nkunal, line 42: same comment as above. also make sure it case insensitive check

pkg/influx/query/query.go
venkateshamatam: I've pushed the changes for this to check if the query contains `create database`. Do we want to return an error if it just contains `create database` without a database name?

pkg/influx/query/query.go
```diff
@@ -0,0 +1,50 @@
...
+
+	log "github.com/sirupsen/logrus"
+	"github.com/valyala/fasthttp"
+)
+
+func GetQueryHandler(ctx *fasthttp.RequestCtx, myid uint64) {
+
+	q := string(ctx.QueryArgs().Peek("q"))
+	if strings.ToLower(q) == "show databases" {
+		databaseResponse := `{
```
nkunal, line 14: make this as a const at the top of file, so that we dont have to recreate this memory

nkunal: name it `INFLUX_SHOW_DB_RESP`

nkunal (approved):

PR #917, opened 2024-05-20
fix: Influx ingestion format

Macbeth98:
# Description
- Fixes the Influx Ingestion format.

# Testing
- Tested by Adding the test cases and by ingesting the influx line format from tsbs.

# Checklist:


pkg/segment/writer/segwriter.go
```diff
@@ -288,17 +288,26 @@ func AddTimeSeriesEntryToInMemBuf(rawJson []byte, signalType SIGNAL_TYPE, orgid
...
 		}
-		err = metrics.EncodeDatapoint(mName, tagsHolder, dp, ts, uint64(len(rawJson)), orgid)
-		if err != nil {
-			metrics.ReturnTagsHolder(tagsHolder)
-			return err
+
+		size := uint64(len(rawJson))
+		ingestedMetricsCount := uint16(0) // will remove this in the final depending on how we handle errors or return of this.
+
+		for _, mip := range metricsIngestPayload {
```
nkunal, line 300: we should just move this for loop into metricssegment.go #709, so that we don;t need to create the `metricsIngestPayload` and avoid the memory allocations.

pkg/segment/writer/metrics/metricssegment.go
```diff
@@ -679,7 +680,7 @@ func ExtractInfluxPayload(rawCSV []byte, tags *TagsHolder) ([]byte, float64, uin
 			}
 			for index, value := range tag_set {
 				if index == 0 {
-					mName = []byte(value)
+					continue
```
nkunal, line 683: add a comment here, the first element in the tag_set is actually a dbName, which we ignore

pkg/segment/writer/metrics/metricssegment.go
```diff
@@ -690,27 +691,61 @@ func ExtractInfluxPayload(rawCSV []byte, tags *TagsHolder) ([]byte, float64, uin
...
-				value = kvPair[1]
-				if key == influx_value {
-					fltVal, err := strconv.ParseFloat(value, 64)
-					if err != nil {
-						return nil, 0, 0, fmt.Errorf("failed to convert value to float! %+v", err)
-					}
-					dpVal = fltVal
+			for _, metricValueSet := range field_set {
+				kvPair := strings.Split(metricValueSet, "=")
+				metricName := kvPair[0]
```
nkunal, line 696: add a check here `if len(kvPair) < 2` then continue, else this is going to crash if someone sends bad data

nkunal: do the error counting as well

Macbeth98: We can get the errors count from doing `len(errors)`? I am actually storing the list/slice of errors.

nkunal (approved):

PR #928, opened 2024-05-21
Promql query support for sigclient

venkateshamatam:
# Description
This PR adds code to run queries against the promQL API and lets you verify the result against the expected result if a particular query should fail or succeed.

To test using queries from a file using sigclient run the below command:

`go run main.go query promql -d http://localhost:5122/promql/api/v1/query_range -f ../../cicd/promql_test.csv`

To test a single query run the below command:

`go run main.go query promql -d http://localhost:5122/promql/api/v1/query_range -q "testmetric0 start:1716239444 end:1716239544"`

the `-q` flag accepts the query along with the start and end epoch in seconds in this format - `"<query> start:<startEpoch> end:<endEpoch>"`

# Testing
- Tested the above commands with different queries
# Checklist:

### DIFF
--- a/tools/sigclient/cmd/sigclient.go
+++ b/tools/sigclient/cmd/sigclient.go
@@ -178,18 +178,20 @@ var promQLQueryCmd = &cobra.Command{
 		promQLQueryWithStartEnd, _ := cmd.Flags().GetString("query")
 		var promQLQuery, startEpoch, endEpoch string
 		if promQLQueryWithStartEnd != "" {
-			// Split the string into parts
-			parts := strings.Split(promQLQueryWithStartEnd, " ")
-
-			// Check that the length of parts is exactly 3
-			if len(parts) != 3 {
+			parts := strings.SplitN(promQLQueryWithStartEnd, "start:", 2)
+			if len(parts) != 2 {
 				log.Fatalf("Invalid query format: %v. Expected format: <query> start:<startEpoch> end:<endEpoch>", promQLQueryWithStartEnd)
 			}
 
-			// Extract the individual arguments
-			promQLQuery = parts[0]
-			startEpoch = strings.TrimPrefix(parts[1], "start:")
-			endEpoch = strings.TrimPrefix(parts[2], "end:")
+			promQLQuery = strings.TrimSpace(parts[0])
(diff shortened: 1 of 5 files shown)


tools/sigclient/cmd/sigclient.go
```diff
@@ -216,10 +258,10 @@ func init() {
 	queryCmd.PersistentFlags().BoolP("validateMetricsOutput", "y", false, "check if metric querries return any results")
 	queryCmd.PersistentFlags().StringP("filePath", "f", "", "filepath to csv file to use to run queries from")
 	queryCmd.PersistentFlags().BoolP("randomQueries", "", false, "generate random queries")
-
+	queryCmd.PersistentFlags().StringP("query", "q", "", "query to run")
```
nkunal, line 261: update `promql query to run`

tools/sigclient/pkg/query/promqlquery.go
```diff
@@ -0,0 +1,109 @@
...
+			return
+		}
+
+		query, step := rec[0], rec[3]
+
+		nowTs := uint64(time.Now().Unix())
+		defValue := nowTs - 60*60 // default value is current time - 1 hour
+
+		start := utils.ConvertStringToEpochSec(nowTs, rec[1], defValue)
+		end := utils.ConvertStringToEpochSec(nowTs, rec[2], defValue)
```
nkunal, line 56: if end is not given then would we still use the same default value for end and start ?

tools/sigclient/pkg/query/promqlquery.go
```diff
@@ -0,0 +1,109 @@
...
+	var result map[string]interface{}
+	json.Unmarshal(body, &result)
+
+	if result["status"] == "success" {
+		log.Printf("Query: %v was successful.", query)
+		return true
+	} else if expectedResult {
+		log.Errorf("Non-200 HTTP status code: %v, query: %v response body: %v", resp.StatusCode, u.RawQuery, string(body))
+	} else {
+		log.Printf("Query: %v failed as expected.", query)
```
nkunal, line 106: why are we saying query failed as expected ?

nkunal (approved):

PR #929, opened 2024-05-21
Feat: Promql <aggregation>_over_time funcs Completion

Jiale-Fang:
# Description
Add promql funcs as below:
- quantile_over_time(scalar, range-vector): the φ-quantile (0 ≤ φ ≤ 1) of the values in the specified interval.
- stddev_over_time(range-vector): the population standard deviation of the values in the specified interval.
- stdvar_over_time(range-vector): the population standard variance of the values in the specified interval.
- last_over_time(range-vector): the most recent point value in the specified interval.
- present_over_time(range-vector): the value 1 for any series in the specified interval.

# Testing

# Checklist:

### DIFF
--- a/pkg/segment/results/mresults/seriesresult_test.go
+++ b/pkg/segment/results/mresults/seriesresult_test.go
@@ -523,6 +523,144 @@ func Test_applyRangeFunctionCount(t *testing.T) {
 	assert.True(t, dtypeutils.AlmostEquals(val, 2))
 }
 
+func Test_applyRangeFunctionStdvarOverTime(t *testing.T) {
+	result := make(map[string]map[uint32]float64)
+	ts := map[uint32]float64{
+		1000: 10,
+		1001: 20,
+		1002: 30,
+		1013: 40,
+		1018: 50,
+		1019: 60,
+		1020: 70,
+	}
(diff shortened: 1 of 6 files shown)


pkg/segment/results/mresults/seriesresult_test.go
AndrewHess, line 1: Aren't you missing tests for `last_over_time` and `present_over_time`?

AndrewHess (commented):
Looks pretty good. A few small things to fix.

pkg/segment/results/mresults/seriesresult_test.go
Jiale-Fang: Yeah, I will add test for `present_over_time`. For `last_over_time` If we take the very last sample from every element of a range vector, the resulting vector will be identical to a regular instant vector query. So i just return the initial `key, value` pair without changing anything.  I am not sure if i need add tests for that, what do you think?

AndrewHess: Yeah go ahead and add a simple test for it. It may help catch issues if we later refactor some of the code

AndrewHess (approved):

PR #948, opened 2024-05-23
Store all incoming queries to query.log

venkateshamatam:
# Description
Store all incoming queries to query.log

# Testing
- Tested the logging of query data on local machine

# Checklist:

### DIFF
--- a/pkg/ast/pipesearch/searchHandler.go
+++ b/pkg/ast/pipesearch/searchHandler.go
@@ -324,6 +324,13 @@ func ProcessPipeSearchRequest(ctx *fasthttp.RequestCtx, myid uint64) {
 		false,
 		"access.log",
 	)
+	utils.AddQueryLogEntry(
+		time.Now(),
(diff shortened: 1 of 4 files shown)


AndrewHess:
What is the purpose of this PR? Isn't everything you would log to query.log already logged to access.log?

nkunal:
We log to access log once the query has been completed, but if we crash
before it completes then we have no way of knowing what query it was. If
there is another better alternative to track this and then thats fine too.


*Best Regards*
*-- Kunal Nawale ,**Founder/CEO, SigScalr*

AndrewHess (commented):
I think you should reuse `AddAccessLogEntry()` instead of defining `AddQueryLogEntry()` which is nearly identical. You can have `AddAccessLogEntry()` write to query.log if we want to make sure every line in access.log follows the normal Access Log format (e.g., no restart markers). Or you could just write everything to access.log

PR #949, opened 2024-05-23
Fix: Increase tolerance for approximate equality in metrics test

Jiale-Fang:
# Description
Floating-point operations can sometimes lead to precision loss, so the tolerance value has been increased.
image

# Testing

# Checklist:


Jiale-Fang:
`predict_linear(v range-vector, t scalar)` can predict the value after `t` seconds. In the past, the `t` in test cases(`1000000`) is too large, which lead to Floating-point inaccuracies have also been magnified `1000000` times. So i modified the test case.

AndrewHess (approved):

PR #953, opened 2024-05-24
Feat: Add promql comparison operators

Jiale-Fang:
# Description
Add Comparison binary operators
The following binary comparison operators exist in Prometheus: https://prometheus.io/docs/prometheus/latest/querying/operators/#comparison-binary-operators
```
== (equal)
!= (not-equal)
> (greater-than)
< (less-than)
>= (greater-or-equal)
<= (less-or-equal)
```

Test with these queries:
```
1. >
(testmetric0) > 132
132 > (testmetric0)

2. >=
(testmetric6) >= 666
(testmetric6) >= 667
667 >= (testmetric6)

3. <
131 < (testmetric0)
(testmetric0) < 10
(testmetric0) < 1000

4. <=
667 <= (testmetric6)
(testmetric6) <= 668

5. ==
(testmetric6) === 666

6. !=
(testmetric6) != 66

(testmetric0) + (testmetric0)
(testmetric0) - (testmetric0)
(testmetric0) * (testmetric0)
(testmetric0) / (testmetric0)
(testmetric0) > (testmetric0)
(testmetric0) >= (testmetric0)
(testmetric0) < (testmetric0)
(testmetric0) <= (testmetric0)
(testmetric0) == (testmetric0)
(testmetric0) != (testmetric0)

```

# Testing

# Checklist:

### DIFF
--- a/cicd/metrics_test.csv
+++ b/cicd/metrics_test.csv
@@ -50,7 +50,6 @@ day_of_year(testmetric0),now-90d,now,"367,367,367",lt
 day_of_year(testmetric0),now-90d,now,"0,0,0",gt
 days_in_month(testmetric0),now-90d,now,"32,32,32",lt
 days_in_month(testmetric0),now-90d,now,"27,27,27",gt
-"predict_linear(testmetric0[48h], 1000000)",now-7d,now,"76.28998388512214,322.0191498768958",approx
 "predict_linear(testmetric0[48h], 100)",now-7d,now,"127.06355282392336,159.54968553618554",approx
 "(testmetric0) > 132",now-90d,now,"159.53343696489338",approx
 "(testmetric0) >= 131",now-90d,now,"131.45590577322224,159.53343696489338",approx
--- a/cicd/metrics_test.csv
+++ b/cicd/metrics_test.csv
@@ -41,5 +41,4 @@ changes(testmetric0[48h]),now-90d,now,"0,1,2",eq
 "(testmetric0) >= 131",now-90d,now,"131.45590577322224,159.53343696489338",approx
 "(testmetric0) < 131",now-90d,now,"127.06863068860355",approx
 "(testmetric0) <= 1000",now-90d,now,"131.45590577322224,127.06863068860355,159.53343696489338",approx
-"(testmetric0) == 1000",now-90d,now,"",approx
 "(testmetric0) != 1000",now-90d,now,"131.45590577322224,127.06863068860355,159.53343696489338",approx
--- a/cicd/metrics_test.csv
+++ b/cicd/metrics_test.csv
@@ -37,3 +37,9 @@ avg_over_time(testmetric0[24h]),now-90d,now,"131.45590577322233,129.262268230913
 max_over_time(testmetric0[48h]),now-90d,now,"131.45590577322233,131.45590577322233,159.53343696489338",approx
 changes(testmetric0[48h]),now-90d,now,"0,1,2",eq
 "predict_linear(testmetric0[48h], 1000000)",now-7d,now,"76.28998388512214,322.0191498768958",approx
+"(testmetric0) > 132",now-90d,now,"159.53343696489338",approx
+"(testmetric0) >= 131",now-90d,now,"131.45590577322224,159.53343696489338",approx
+"(testmetric0) < 131",now-90d,now,"127.06863068860355",approx
+"(testmetric0) <= 1000",now-90d,now,"131.45590577322224,127.06863068860355,159.53343696489338",approx
+"(testmetric0) == 1000",now-90d,now,"",approx
+"(testmetric0) != 1000",now-90d,now,"131.45590577322224,127.06863068860355,159.53343696489338",approx
--- a/pkg/integrations/prometheus/promql/metricsSearchHandler.go
+++ b/pkg/integrations/prometheus/promql/metricsSearchHandler.go
@@ -929,6 +929,7 @@ func convertPqlToMetricsQuery(searchText string, startTime, endTime uint32, myid
 
 		mquery.Aggregator = structs.Aggreation{}
 		parser.Inspect(es.Expr, func(node parser.Node, path []parser.Node) error {
+			// If there is no child node, just return nil
 			if node == nil {
 				return nil
 			}
@@ -998,6 +999,7 @@ func convertPqlToMetricsQuery(searchText string, startTime, endTime uint32, myid
 		// Since we currently handle evaluation logic only in sub-elements like MatrixSelector or VectorSelector, if we add a default case in the switch statement,
 		// traversal would stop prematurely due to an error being returned before reaching sub-nodes such as MatrixSelector
 		parser.Inspect(expr, func(node parser.Node, path []parser.Node) error {
+			// If there is no child node, just return nil
 			if node == nil {
 				return nil
 			}
@@ -1184,7 +1186,7 @@ func convertPqlToMetricsQuery(searchText string, startTime, endTime uint32, myid
 			arithmeticOperation.RHS = rhsRequest[0].MetricsQuery.HashedMName
 
 		}
-		arithmeticOperation.Operation = getArithmeticOperation(expr.Op)
+		arithmeticOperation.Operation = getLogicalAndArithmeticOperation(expr.Op)
 		if rhsValType == parser.ValueTypeVector {
 			lhsValType = parser.ValueTypeVector
 		}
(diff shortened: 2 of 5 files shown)


pkg/integrations/prometheus/promql/metricsSearchHandler.go
```diff
@@ -562,11 +562,11 @@ func ProcessUiMetricsSearchRequest(ctx *fasthttp.RequestCtx, myid uint64) {
...
-	for _, metricQuery := range metricQueryRequest {
-		hashList = append(hashList, metricQuery.MetricsQuery.HashedMName)
-		metricQueriesList = append(metricQueriesList, &metricQuery.MetricsQuery)
-		segment.LogMetricsQuery("PromQL metrics query parser", &metricQuery, qid)
-		timeRange = &metricQuery.TimeRange
+	for i := range metricQueryRequest {
+		hashList = append(hashList, metricQueryRequest[i].MetricsQuery.HashedMName)
+		metricQueriesList = append(metricQueriesList, &metricQueryRequest[i].MetricsQuery)
+		segment.LogMetricsQuery("PromQL metrics query parser", &metricQueryRequest[i], qid)
+		timeRange = &metricQueryRequest[i].TimeRange
```
AndrewHess, line 569: What's the purpose of this change?

Same comment for lines 738-742

pkg/segment/results/mresults/tests/metricsresults_test.go
```diff
@@ -404,3 +406,154 @@ func TestCalculateInterval(t *testing.T) {
...
+		},
+		[]structs.QueryArithmetic{
+			{
+				LHS:        1,
+				ConstantOp: true,
+				Operation:  utils.LetGreaterThanOrEqualTo,
+				Constant:   100,
+			},
+		})
+}
```
AndrewHess, line 436: I'm confused reading these tests. I _think_ `10800` is comes from `test_GetResults_Ops()` downsampling to `3 * 3600` and because in that function you're making points from `i + 4 * 3600` so they'll downsample to `3 * 3600`, but it's not very clear and I'm not sure where `1045` is coming from.

Can you restructure your tests so it's easier to understand what they do and so it's clear that the values they should evaluate to are obviously the correct values?

pkg/integrations/prometheus/promql/metricsSearchHandler.go
```diff
@@ -1236,16 +1238,28 @@ func convertPqlToMetricsQuery(searchText string, startTime, endTime uint32, myid
...
+	case parser.GTE:
+		return segutils.LetGreaterThanOrEqualTo
+	case parser.LSS:
+		return segutils.LetLessThan
+	case parser.LTE:
+		return segutils.LetLessThanOrEqualTo
+	case parser.EQLC:
+		return segutils.LetEquals
+	case parser.NEQ:
+		return segutils.LetNotEquals
```
AndrewHess, line 1262: Why are you calling these `Let...`?

cicd/metrics_test.csv
```diff
@@ -37,3 +37,8 @@ avg_over_time(testmetric0[24h]),now-90d,now,"131.45590577322233,129.262268230913
 max_over_time(testmetric0[48h]),now-90d,now,"131.45590577322233,131.45590577322233,159.53343696489338",approx
 changes(testmetric0[48h]),now-90d,now,"0,1,2",eq
 "predict_linear(testmetric0[48h], 100)",now-7d,now,"127.06355282392336,159.54968553618554",approx
+"(testmetric0) > 132",now-90d,now,"159.53343696489338",approx
+"(testmetric0) >= 131",now-90d,now,"131.45590577322224,159.53343696489338",approx
+"(testmetric0) < 131",now-90d,now,"127.06863068860355",approx
+"(testmetric0) <= 1000",now-90d,now,"131.45590577322224,127.06863068860355,159.53343696489338",approx
+"(testmetric0) != 1000",now-90d,now,"131.45590577322224,127.06863068860355,159.53343696489338",approx
```
AndrewHess, line 44: Can you also test `==`?

Also the promQL documentation you linked mentions these comparison operators can be followed by a `bool` option so the output will be `0` or `1` rather than acting as a filter. Is supporting that option planned for a separate PR?

pkg/segment/results/mresults/tests/metricsresults_test.go
Jiale-Fang: Yes, will do that

pkg/integrations/prometheus/promql/metricsSearchHandler.go
Jiale-Fang: I think there is a bug in here. `var metricQueriesList []*structs.MetricsQuery`. 

In the past, the code is like:
```
for _, mQ := range metricQueryRequest {
......
metricQueriesList = append(metricQueriesList, &metricQuery.MetricsQuery)
}
```

If there are 2 metricQueries `a` and `b` in the metricQueryRequest. After first iteration, `metricQueriesList = {mQ.MetricsQuery}`. The references of `a` and `mQ` point to the same address. However, when it comes to the second iteration, `mQ` will point to b. Which will lead to previous `mQ.MetricsQuery` point to `b` as well. Please correct me if i am wrong.

Simple test:
```
	for _, metricQuery := range metricQueryRequest {
		metricQueriesList = append(metricQueriesList, &metricQuery.MetricsQuery)
		log.Error("fjl test1111 mQuery:", metricQuery.MetricsQuery.MetricName)
	}

	for _, mQuery := range metricQueriesList {
		log.Error("fjl test22222 mQuery:", mQuery.MetricName)
	}
```
image

pkg/integrations/prometheus/promql/metricsSearchHandler.go
Jiale-Fang: In the past, `Operation` is `ArithmeticOperator` type. Since `ArithmeticOperator` only contains `+ - * /`, so I changed  `Operation` type from `ArithmeticOperator` to `LogicAndArithmeticOperator`
```
type QueryArithmetic struct {
Operation   utils.ArithmeticOperator
}
```

cicd/metrics_test.csv
Jiale-Fang: Yes, i think i will create a separate PR.

For the `==`, i think i need to modify the cicd test as well. `==` is to match a exact value. But due to precision issue of the floating point, its results will change everytime. So i may need to think a good way to test that.

pkg/integrations/prometheus/promql/metricsSearchHandler.go
AndrewHess: I'm not understanding that. Maybe you can show me sometime Tuesday

pkg/integrations/prometheus/promql/metricsSearchHandler.go
AndrewHess: I see. I didn't realize those were already defined before this PR

pkg/segment/results/mresults/tests/metricsresults_test.go
```diff
@@ -404,3 +406,167 @@ func TestCalculateInterval(t *testing.T) {
...
+
+func Test_GetResults_LSS(t *testing.T) {
```
AndrewHess, line 446: nit: what does `LSS` mean? Maybe rename this to clarify it's testing "less than"

pkg/segment/results/mresults/tests/metricsresults_test.go
```diff
@@ -404,3 +406,167 @@ func TestCalculateInterval(t *testing.T) {
...
+			{
+				LHS:        1,
+				ConstantOp: true,
+				Operation:  utils.LetLessThanOrEqualTo,
+				Constant:   1045,
+			},
+		})
+}
+
+func Test_GetResults_EQLC(t *testing.T) {
```
AndrewHess, line 485: nit: what is `EQLC`? Maybe rename this to clarify it's testing "equals"

pkg/segment/results/mresults/tests/metricsresults_test.go
```diff
@@ -404,3 +406,167 @@ func TestCalculateInterval(t *testing.T) {
...
+		map[uint32]float64{},
+		[]structs.QueryArithmetic{
+			{
+				LHS:        1,
+				ConstantOp: true,
+				Operation:  utils.LetGreaterThan,
+				Constant:   1045,
+			},
+		})
+}
```
AndrewHess, line 425: Add some more test cases and/or points to test boundary conditions. For example, test with values of 45, 1045, and 1046

Similar comment for tests for the other operators

pkg/segment/results/mresults/tests/metricsresults_test.go
```diff
@@ -404,3 +406,167 @@ func TestCalculateInterval(t *testing.T) {
...
+	var tsGroupId *bytebufferpool.ByteBuffer = bytebufferpool.Get()
+	defer bytebufferpool.Put(tsGroupId)
+	_, err := tsGroupId.Write([]byte("color:yellow"))
+	assert.NoError(t, err)
+	series := mresults.InitSeriesHolder(mQuery, tsGroupId)
+
+	// The entries' timestamp interval is larger than the data source interval (3 hours), so they can become separate key-value pairs.
+	for timestamp, val := range initialEntries {
+		series.AddEntry(timestamp, val)
+	}
```
AndrewHess, line 547: You don't know this comment is correct in general because `initialEntries` is a parameter and you haven't done any validation on it yet.

You might want to move the downsampler config out of this function (and pass it as a parameter) so that it's clear from the tests (e.g., `Test_GetResults_GTE()`) how downsampling is working. Or find a different way to organize this to make it more clear

AndrewHess (approved):

PR #960, opened 2024-05-28
feat: Implemented new parser function; includes parsing of nested functions.

Macbeth98:
# Description
- Created a new parser function that will parse the Metric Queries along with Nested functions.
- Fixed a typo in a struct name `Aggregation`
- Written tests to test the parser function.

# Testing
- Tested by writing the test cases with various different queries.

# Checklist:

### DIFF
--- a/pkg/integrations/prometheus/promql/parser.go
+++ b/pkg/integrations/prometheus/promql/parser.go
@@ -1,0 +1,549 @@
+package promql
+
+import (
+	"fmt"
+
+	"github.com/cespare/xxhash"
+	"github.com/prometheus/prometheus/model/labels"
+	"github.com/prometheus/prometheus/promql/parser"
+	dtu "github.com/siglens/siglens/pkg/common/dtypeutils"
+	"github.com/siglens/siglens/pkg/segment/results/mresults"
+	"github.com/siglens/siglens/pkg/segment/structs"
+	segutils "github.com/siglens/siglens/pkg/segment/utils"
+	log "github.com/sirupsen/logrus"
+)
+
+func extractSelectors(expr parser.Expr) [][]*labels.Matcher {
+	var selectors [][]*labels.Matcher
+	parser.Inspect(expr, func(node parser.Node, _ []parser.Node) error {
+		var vs interface{}
+		vs, ok := node.(*parser.VectorSelector)
+		if ok {
+			selectors = append(selectors, vs.(*parser.VectorSelector).LabelMatchers)
+		}
+		vs, ok = node.(parser.Expressions)
+		if ok {
+			for _, entry := range vs.(parser.Expressions) {
+				expr, ok := entry.(*parser.MatrixSelector)
+				if !ok {
+					continue
+				}
+				vectorSelector, ok := expr.VectorSelector.(*parser.VectorSelector)
+				if !ok {
+					continue
+				}
+				selectors = append(selectors, vectorSelector.LabelMatchers)
+			}
+		}
+		return nil
+	})
+	return selectors
+}
+
+func extractTimeWindow(args parser.Expressions) (float64, float64, error) {
+	if len(args) > 0 {
+		if ms, ok := args[0].(*parser.MatrixSelector); ok {
+			return ms.Range.Seconds(), 0, nil
+		} else if subQueryExpr, ok := args[0].(*parser.SubqueryExpr); ok {
+			return subQueryExpr.Range.Seconds(), subQueryExpr.Step.Seconds(), nil
+		} else {
+			return 0, 0, fmt.Errorf("extractTimeWindow: can not extract time window from args: %v", args)
+		}
+	} else {
+		return 0, 0, fmt.Errorf("extractTimeWindow: can not extract time window")
+	}
+}
+
+func parsePromQLQuery(query string, startTime, endTime uint32, myid uint64) ([]*structs.MetricsQueryRequest, parser.ValueType, []*structs.QueryArithmetic, error) {
+	expr, err := parser.ParseExpr(query)
+	if err != nil {
+		log.Errorf("parsePromQLQuery: Error parsing promql query: %v", err)
+		return []*structs.MetricsQueryRequest{}, "", []*structs.QueryArithmetic{}, err
+	}
+
+	pqlQuerytype := expr.Type()
+	var mQuery structs.MetricsQuery
+	mQuery.Aggregator = structs.Aggregation{}
+	selectors := extractSelectors(expr)
+	//go through labels
+	for _, lblEntry := range selectors {
+		for _, entry := range lblEntry {
+			if entry.Name != "__name__" {
+				tagFilter := &structs.TagsFilter{
+					TagKey:          entry.Name,
+					RawTagValue:     entry.Value,
+					HashTagValue:    xxhash.Sum64String(entry.Value),
+					TagOperator:     segutils.TagOperator(entry.Type),
+					LogicalOperator: segutils.And,
+				}
+				mQuery.TagsFilters = append(mQuery.TagsFilters, tagFilter)
+			} else {
+				mQuery.MetricName = entry.Value
+			}
+		}
+	}
+
+	timeRange := &dtu.MetricsTimeRange{
+		StartEpochSec: startTime,
+		EndEpochSec:   endTime,
+	}
+
+	mQuery.OrgId = myid
+	mQuery.PqlQueryType = pqlQuerytype
+
+	intervalSeconds, err := mresults.CalculateInterval(endTime - startTime)
+	if err != nil {
+		return []*structs.MetricsQueryRequest{}, "", []*structs.QueryArithmetic{}, err
+	}
+
+	metricQueryRequest := &structs.MetricsQueryRequest{
+		MetricsQuery: mQuery,
+		TimeRange:    *timeRange,
+	}
+
+	mQueryReqs := make([]*structs.MetricsQueryRequest, 0)
+	mQueryReqs = append(mQueryReqs, metricQueryRequest)
+
+	queryArithmetic := make([]*structs.QueryArithmetic, 0)
+	var exitFromInspect bool
+
+	parser.Inspect(expr, func(node parser.Node, path []parser.Node) error {
+		if !exitFromInspect {
+			mQueryReqs, queryArithmetic, exitFromInspect, err = parsePromQLExprNode(node, mQueryReqs, queryArithmetic, intervalSeconds)
+		}
+		return err
+	})
+
+	if len(mQueryReqs) == 0 {
+		return mQueryReqs, pqlQuerytype, queryArithmetic, nil
+	}
+
+	mQuery = mQueryReqs[0].MetricsQuery
+
+	tags := mQuery.TagsFilters
+	for idx, tag := range tags {
+		var hashedTagVal uint64
+		switch v := tag.RawTagValue.(type) {
+		case string:
+			hashedTagVal = xxhash.Sum64String(v)
+		case int64:
+			hashedTagVal = uint64(v)
+		case float64:
+			hashedTagVal = uint64(v)
+		case uint64:
+			hashedTagVal = v
+		default:
+			log.Errorf("ParseMetricsRequest: invalid tag value type")
+		}
+		tags[idx].HashTagValue = hashedTagVal
+	}
+
+	if mQuery.MQueryAggs == nil {
+		mQuery.MQueryAggs = &structs.MetricQueryAgg{
+			AggBlockType:    structs.AggregatorBlock,
+			AggregatorBlock: &structs.Aggregation{AggregatorFunction: segutils.Avg},
+		}
+		mQueryReqs[0].MetricsQuery = mQuery
+	}
+	// else {
+	// If the first aggregation is not an aggregator block, then add an aggregator block with avg function
+	// Should I do this way??
+
+	// if mQuery.MQueryAggs.AggBlockType != structs.AggregatorBlock {
+	// 	mQuery.MQueryAggs = &structs.MetricQueryAgg{
+	// 		AggBlockType:    structs.AggregatorBlock,
+	// 		AggregatorBlock: &structs.Aggregation{AggregatorFunction: segutils.Avg},
+	// 		Next:            mQuery.MQueryAggs,
+	// 	}
+	// 	mQueryReqs[0].MetricsQuery = mQuery
+	// }
+	// }
+
+	return mQueryReqs, pqlQuerytype, queryArithmetic, nil
+}
+
+func parsePromQLExprNode(node parser.Node, mQueryReqs []*structs.MetricsQueryRequest, queryArithmetic []*structs.QueryArithmetic,
+	intervalSeconds uint32) ([]*structs.MetricsQueryRequest, []*structs.QueryArithmetic, bool, error) {
+	var err error = nil
+	exit := false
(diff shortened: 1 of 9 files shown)


pkg/integrations/prometheus/promql/parser.go
```diff
@@ -0,0 +1,564 @@
...
+		// If the first Block in the MQueryAggs is not an aggregator block, then add an default aggregator block with avg function
+
+		if mQuery.MQueryAggs.AggBlockType != structs.AggregatorBlock {
+			mQuery.MQueryAggs = &structs.MetricQueryAgg{
+				AggBlockType:    structs.AggregatorBlock,
+				AggregatorBlock: &structs.Aggregation{AggregatorFunction: segutils.Avg},
+				Next:            mQuery.MQueryAggs,
+			}
+			mQueryReqs[0].MetricsQuery = mQuery
+		}
```
AndrewHess, line 158: I think you'd be writing over whatever was at position 0, right? Do you need to prepend this new item instead?

pkg/integrations/prometheus/promql/parser.go
```diff
@@ -0,0 +1,564 @@
...
+	}
+
+	return mQueryReqs, pqlQuerytype, queryArithmetic, nil
+}
+
+func parsePromQLExprNode(node parser.Node, mQueryReqs []*structs.MetricsQueryRequest, queryArithmetic []*structs.QueryArithmetic,
+	intervalSeconds uint32) ([]*structs.MetricsQueryRequest, []*structs.QueryArithmetic, bool, error) {
+	var err error = nil
+	exit := false
```
AndrewHess, line 167: What does `exit` mean? Why is it set to true after handling a `BinaryExpr`, sometimes true when handling `SubqueryExpr`, and false for handling other things?

pkg/integrations/prometheus/promql/parser.go
```diff
@@ -0,0 +1,564 @@
...
+		mQuery.Aggregator.AggregatorFunction = segutils.Sum
+	case "max":
+		mQuery.Aggregator.AggregatorFunction = segutils.Max
+	case "min":
+		mQuery.Aggregator.AggregatorFunction = segutils.Min
+	case "quantile":
+		mQuery.Aggregator.AggregatorFunction = segutils.Quantile
+	default:
+		log.Infof("handleAggregateExpr: using avg aggregator by default for AggregateExpr (got %v)", aggFunc)
+		mQuery.Aggregator = structs.Aggregation{AggregatorFunction: segutils.Avg}
```
AndrewHess, line 234: The purpose of this is to use `avg` when no aggregation was provided, right? If so, I think it would be better to move this to a different case (like `case "":`) and then have the `default:` give an error. That way if we get a truly unexpected value like `"hello world"` or a typo, it'll be easier to find the error

pkg/integrations/prometheus/promql/parser.go
```diff
@@ -0,0 +1,564 @@
...
+	}
+
+	// Handle grouping
+	for _, grp := range expr.Grouping {
```
AndrewHess, line 238: nit: rename `grp` to `group`

pkg/integrations/prometheus/promql/parser.go
```diff
@@ -0,0 +1,564 @@
...
+		}
+		if err != nil {
+			return fmt.Errorf("handleCallExpr: cannot parse Call Node: %v", err)
+		}
+
+		// Break if the case is not default. As Call Expr can have multiple arguments
+		// And they will be procesed based on the Node type.
+		if !defaultCase {
+			break
+		}
```
AndrewHess, line 296: Switch these `if`s. As they are, if the first loop iteration is non-default and gives an error and the second loop iteration is default, then the error will get printed in both iterations

Alternatively, you may be able to label this loop and do something like `break ArgsLoop` inside the `default` case—I'm not sure

pkg/integrations/prometheus/promql/parser.go
```diff
@@ -0,0 +1,564 @@
...
+		// Otherwise, set the new MetricQueryAgg to the head of the chain
+		// and set the current head as the next of the new MetricQueryAgg
+		mQueryAgg.Next = mQuery.MQueryAggs
+		mQuery.MQueryAggs = mQueryAgg
+	}
+
+	// Reset the Function to handle the next function call correctly
+	mQuery.Function = structs.Function{}
+
+	return nil
```
AndrewHess, line 331: Why is this function (`handleCallExpr()`) responsible for updating the `mQuery` parameter like this? I'd recommend moving this logic to a new function. Then probably refactor so you call that new function outside of `handleCallExpr()`, but if you need to call it within `handleCallExpr()` you'll probably want to add a comment to the definition of `handleCallExpr()` saying how it updates `mQuery`, or return the new `mQuery` instead of just updating the one sent as a parameter

pkg/integrations/prometheus/promql/parser.go
```diff
@@ -0,0 +1,564 @@
...
+			currentAggs = currentAggs.Next
+		}
+		currentAggs.Next = mQuery.MQueryAggs
+
+		mQueryReq.MetricsQuery.Groupby = mQuery.Groupby
+		mQueryReq.MetricsQuery.SelectAllSeries = mQuery.SelectAllSeries
+		mQueryReq.MetricsQuery.TagsFilters = mQuery.TagsFilters
+	}
+	return mQueryReqs
+}
```
AndrewHess, line 553: Correct me if I'm wrong, but I think this function name is misleading. It sounds like it would update `mQuery` with aggs from `mQueryReqs`. But the code looks like it appends the `mQuery` aggs to the aggs of each element of `mQueryReqs`

pkg/integrations/prometheus/promql/parser.go
```diff
@@ -0,0 +1,564 @@
...
+func HandleNumberLiteral(literal *parser.NumberLiteral, mQuery *structs.MetricsQuery) {
+	// Check if the ValueList is nil, if so, initialize it
+	if mQuery.Function.ValueList == nil {
+		mQuery.Function = structs.Function{
+			ValueList: []string{},
+		}
+	}
+	// Append the number literal to the ValueList
+	mQuery.Function.ValueList = append(mQuery.Function.ValueList, literal.String())
+}
```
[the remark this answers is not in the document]
Macbeth98: No it does not need to be public, but I thought I would need this function, 
but as of now we are not using this...
But I kept it because for the queries such as `1 + 1` or `vector(1) + vector(1)` something like this... 
I wanted to deal with these queries when I am actually working on Formulas. So for now I wanted to leave it like that.

But I can remove this for this PR and add the function in the Formulas function. So shall I remove it now?

pkg/segment/structs/metricsstructs.go
```diff
@@ -369,3 +388,20 @@ func (mbs *MBlockSummary) UpdateTimeRange(ts uint32) {
...
+		Step:          metricFunc.Step,
+	}
+}
+
+func (agg Aggregation) Clone() *Aggregation {
+	return &Aggregation{
+		AggregatorFunction: agg.AggregatorFunction,
+		FuncConstant:       agg.FuncConstant,
+	}
+}
```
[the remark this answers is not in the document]
Macbeth98: Yes, I am using the clone function in this case here:

```
mQueryAgg := &structs.MetricQueryAgg{
		AggBlockType:    structs.AggregatorBlock,
		AggregatorBlock: mQuery.Aggregator.Clone(),
	}
```

I wanted to do clone because this AggregatorBlock is a pointer type but the Aggregator is not a pointer type but the direct value. 
So when I do `&Aggregation{}` and assigning these fields won't it automatically do something similar to deep clones, as all these are value types?

And I wanted it here because I will be giving a reference, in the next iteration if there is a change and that change is being applied here!! 

But are you saying instead of assigning them one by one... can I do something like this?
`return &Function{...metricFunc}`

pkg/integrations/prometheus/promql/parser.go
```diff
@@ -0,0 +1,564 @@
...
+	//go through labels
+	for _, lblEntry := range selectors {
+		for _, entry := range lblEntry {
+			if entry.Name != "__name__" {
+				tagFilter := &structs.TagsFilter{
+					TagKey:          entry.Name,
+					RawTagValue:     entry.Value,
+					HashTagValue:    xxhash.Sum64String(entry.Value),
+					TagOperator:     segutils.TagOperator(entry.Type),
+					LogicalOperator: segutils.And,
```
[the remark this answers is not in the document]
Macbeth98: Yes, I think for the TagFilters, it is always `And`. Although I am not exactly sure, but this is how our we are doing it currently.

pkg/integrations/prometheus/promql/parser.go
Macbeth98: Are you talking about line 157 or line 154. If it is line 154, then I am prepending it.
For line 157, yes I want to overwrite the first element and there will only be one element. This is applied in the case where each individual sub query in  a multi queries will be parsed once at a time.

pkg/integrations/prometheus/promql/parser.go
Macbeth98: So the way `BinaryExpr` works... this is how that we are currently doing and I used the same here.

It can recursively calls the sub queries to be parsed inside a `BinaryExpr`. So it would mean, it would already be parsing all the components of Binary Expr `LHS` and `RHS`.

So there is no reason for us to go and parse the queries again.

The `pql.Inspect` will walk through each node in the `AST`. Which would mean it would go to first sub query in the Binary Expr, which we have already parsed in the `handleBinaryExpr` call. So I want to exit when the BinaryExpr call is returned.

pkg/integrations/prometheus/promql/parser.go
Macbeth98: ohh Yup, I will do that.

pkg/integrations/prometheus/promql/parser.go
Macbeth98: I want this loop to exit as soon as it went to a case that is not default. So we want to return the error that comes from a non-default case.

pkg/integrations/prometheus/promql/parser.go
Macbeth98: ahh yes, it appends the `MQuery` to each of the `mQueryReqs`.
I will change the function name to `appendMetricAggsToMQueryRequests()`?

pkg/integrations/prometheus/promql/parser.go
Macbeth98: I will move this logic to a new function. As I am doing something similar in the other function `handleAggregateExpr`.

pkg/integrations/prometheus/promql/parser.go
AndrewHess: Sure, that's a better name

pkg/integrations/prometheus/promql/parser.go
AndrewHess: Oh my bad. Your code looks correct; I must have read that as `if defaultCase`

pkg/integrations/prometheus/promql/parser.go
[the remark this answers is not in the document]
AndrewHess: I see. Probably remove it here if it won't be used in this PR, but it's not a big deal either way

pkg/integrations/prometheus/promql/parser.go
AndrewHess: I'm talking about line 157, but I'm not understanding what you're saying.

Since we got here, `mQuery.MQueryAggs.AggBlockType != structs.AggregatorBlock`, so `mQuery.MQueryAggs.AggBlockType == FunctionBlock`. Lines 152-157 would write over this so you lose that `FunctionBlock` agg. Why is this the correct thing to do?

pkg/segment/structs/metricsstructs.go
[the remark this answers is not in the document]
AndrewHess: I'm not quite following. For `Aggregation` all the fields have primitive types, so a shallow copy and deep copy should be the same. For `Function`, I think the only field with a non-primitive type is `ValueList`, but you're doing a shallow copy of it anyways in your `Clone()` function. So I don't think there's any difference between:
```
copy1 := myFunction.Clone()
copy2 := myFunction
```
I'm pretty sure they'd both do a shallow copy. Maybe you meant to deep copy it (i.e., copy `ValuesList`)?

I don't think having `myFunction` as a pointer will change anything. See https://go.dev/play/p/B8xeDAXws72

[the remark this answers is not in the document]
AndrewHess: Actually I was wrong about the pointer bit, as that playground shows: https://go.dev/play/p/B8xeDAXws72. If that's the purpose of these `Clone()` methods, I think that's ok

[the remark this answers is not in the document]
Macbeth98: I think I get what you are trying to say, As you said I was not doing a deep copy. But that is not an issue within our code. As I am assigning the main `Aggregation` or `Function` to a new struct initialization. 

But as you said, we don't want to keep adding the new fields in the clone functions every time we update any of these structs. 

So I will keep the clone functions. But I will change them to this: 

```
func (metricFunc Function) Clone() *Function {
	functionCopy := metricFunc
	return &functionCopy
}

func (agg Aggregation) Clone() *Aggregation {
	aggCopy := agg
	return &aggCopy
}
```


Check this playground code once: https://go.dev/play/p/dyvSaF6fRuJ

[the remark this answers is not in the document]
Macbeth98: This code works, I have tested by running the test cases.

pkg/integrations/prometheus/promql/parser.go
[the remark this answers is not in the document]
Macbeth98: I guess, I will remove it. Makes it more clean.

pkg/integrations/prometheus/promql/parser.go
Macbeth98: At line 122: `mQuery = mQueryReqs[0].MetricsQuery`, I am taking the value of mQuery and it is a value type, so if I change here, it won't reflect in the `mQueryReqs` array.
So I am reassigning it. 

Also, the `FunctionBlock` won't be lost it will be in the next. We need the first block to be an `AggregatorBlock`, So when the first block is not an `AggregatorBlock`, we need to add the default aggregation Block which is what I was doing in the Lines 152 - 154 and at Line 155 which `Next`, I am assigning the current `mQueryAggs` by doing this `Next: mQuery.MQueryAggs`. 

So the first block is Aggregator block and the second will have the Function Block.

pkg/segment/structs/metricsstructs.go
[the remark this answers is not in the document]
AndrewHess: That's a great idea. Also renaming these to `ShallowClone()` could add clarity. Or if `Clone()` is typically understood to mean a shallow copy, keeping it as `Clone()` works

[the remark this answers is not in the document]
Macbeth98: Yeah, I think ShallowClone makes it better. So in future it helps.

pkg/integrations/prometheus/promql/parser.go
AndrewHess: Ok, makes sense

AndrewHess (approved):

PR #961, opened 2024-05-28
fix: time picker dropdown metrics explorer

### LINKED ISSUE #884: [BUG] Metric explorer page time picker does not work
# Description
The time picker on the 'metrics-explorer' page fails to update the metric names when a new time range is selected.

# Steps to Reproduce
1. Navigate to the 'metrics-explorer' page at `metrics-explorer.html`.
2. Note that upon initial page load, an API call is made to `/metric_names` with a default time range of now-1h.
3. Change the time range using the time picker. Observe that no new API call is made to update the metric names for the selected time range, and the metric names dropdown remains unpopulated.

# Expected Behavior
Upon selecting a new time range in the time picker, the metric names and other fields should update to reflect data over the new time range.

navyayy:
# Description
This change fixes the issue with the time picker of metrics explorer. Now, when a new time is selected from time picker it calls the metrics names, functions and timeseries APIs again with new endTime value.

Fixes #884 

# Testing
I have tested by selecting different values from the time picker.

# Checklist:


static/js/metrics-explorer.js
```diff
@@ -52,17 +52,23 @@ $(document).ready(function() {
...
     });
+    var selectedId = $(evt.currentTarget).attr("id");
     $(evt.currentTarget).addClass('active');
-    datePickerHandler($(this).attr('id'), "now", $(this).attr('id'))
+    datePickerHandler(selectedId, "now", selectedId);
+    const queryElement = $('#metrics-queries')
+    const metricNames = await getMetricNames();
+    metricNames.metricNames.sort();
+    queryElement.find('.metrics').val(metricNames.metricNames[0]); 
+    getFunctions();
```
sonamgupta21, line 67: We don't need to call the getFunctions every time we change the time range since it will always be the same.

static/js/metrics-explorer.js
```diff
@@ -52,17 +52,23 @@ $(document).ready(function() {
...
         $(this).removeClass('active');
     });
+    var selectedId = $(evt.currentTarget).attr("id");
     $(evt.currentTarget).addClass('active');
-    datePickerHandler($(this).attr('id'), "now", $(this).attr('id'))
+    datePickerHandler(selectedId, "now", selectedId);
+    const queryElement = $('#metrics-queries')
+    const metricNames = await getMetricNames();
+    metricNames.metricNames.sort();
+    queryElement.find('.metrics').val(metricNames.metricNames[0]);
```
sonamgupta21, line 66: We don't need to update the metricsName to the firstvalue. It should remain as it was. We should only update the options for metricsName in the dropdown and also update the tags for the value which is present as the metrics name.

navyayy: @sonamgupta21 if the `metricNames` array is coming empty from backend on a specific time range, then should we remove the existing value of the metrics input?

image

image

sonamgupta21: No, we don't need to remove the input value. We will just update the graph and the dropdown options.

static/js/metrics-explorer.js
```diff
@@ -52,18 +52,36 @@ $(document).ready(function() {
...
-        await getQueryDetails(queryName,queryDetails)
+        await getQueryDetails(queryName, queryDetails);
+        const tagsAndValue = await getTagKeyValue(queryDetails.metrics);
+        
+        queryDetails.everywhere = [];
+        queryDetails.everything = [];
+        availableEverywhere = tagsAndValue.availableEverywhere;
+        availableEverything = tagsAndValue.availableEverything[0];
+        const queryElement = $(`.metrics-query .query-name:contains(${queryName})`).closest('.metrics-query');
+        queryElement.find('.everywhere').autocomplete('option', 'source', availableEverywhere);
```
sonamgupta21, line 81: Looks good! 
Just one thing: sort availableEverywhere and availableEverything before updating the options.

sonamgupta21 (approved):

PR #966, opened 2024-05-28
Feat: Allow using bool modifier with comparison operators

Jiale-Fang:
# Description
- Allow using bool modifier with comparison operators (E.g.: `(testmetric0) > bool 132`) https://prometheus.io/docs/prometheus/latest/querying/operators/#comparison-binary-operators
- Add test cases for `==`

# Testing

# Checklist:


pkg/segment/segexecution.go
```diff
@@ -209,6 +185,17 @@ func HelperQueryArithmeticAndLogical(queryOps []structs.QueryArithmetic, resMap
...
+func setFinalRes(finalResult map[string]map[uint32]float64, groupID string, timestamp uint32, returnBool bool, comparisonBool bool, val float64) {
+	// If a comparison operator, return 0/1 rather than filtering.
+	if returnBool {
+		finalResult[groupID][timestamp] = 0
+		val = 1
+	}
+	if comparisonBool {
+		finalResult[groupID][timestamp] = val
+	}
+}
```
AndrewHess, line 197: This looks correct to me, but I think you can simplify the logic:
```
if returnBool {
    finalResult[groupID][timestamp] = comparisonBool
    return
}

// Filter based on the comparison.
if !comparisonBool {
    finalResult[groupID][timestamp] = 0
}
```

But if you think your version is easier to understand, you can keep your version

AndrewHess (commented):
Looks pretty good. In your unit tests you just test > and ==. You should add tests for the other operators too

AndrewHess (dismissed):

AndrewHess (approved):

PR #970, opened 2024-05-29
Feat: Add promql arithmetic ops mod/power

Jiale-Fang:
# Description
Add promql arithmetic ops mod/power

# Testing

# Checklist:


pkg/segment/results/mresults/tests/metricsresults_test.go
```diff
@@ -407,6 +407,62 @@ func TestCalculateInterval(t *testing.T) {
...
+			7200: 0,
+		},
+		[]structs.QueryArithmetic{
+			{
+				LHS:        1,
+				ConstantOp: true,
+				Operation:  utils.LetModulo,
+				Constant:   5,
+			},
+		},
```
AndrewHess, line 429: Nice test with a negative number. You should add another case to do a negative number not divisible by 5. So maybe `-7 % 5`. Some languages will give 3 and others -2. What does Go give? What should we give?

pkg/segment/segexecution.go
```diff
@@ -112,6 +112,18 @@ func HelperQueryArithmeticAndLogical(queryOps []structs.QueryArithmetic, resMap
 							val = val * -1
 						}
 						finalResult[groupID][timestamp] = val
+					case utils.LetModulo:
+						if swapped {
+							finalResult[groupID][timestamp] = math.Mod(valueRHS, valueLHS)
+						} else {
+							finalResult[groupID][timestamp] = math.Mod(valueLHS, valueRHS)
+						}
```
AndrewHess, line 120: This is fine (this PR can get merged like this)

Sometime later, you might want to refactor this since you have this swapping logic in every `case` of this `switch`. So before this `switch`, you could handle the swapping logic once and then you don't have to do it in each `case`. After that, you could move this whole `switch` to a separate function, since you have the same `switch` (except for the swapping logic) a little lower. That way there's less duplicate code; or you could keep it how it is if you prefer

pkg/segment/results/mresults/tests/metricsresults_test.go
Jiale-Fang: I tried that in promql ui, if the equation is like you said is: -7 % 5, it will return -2

pkg/segment/segexecution.go
Jiale-Fang: Yeah, right now we only support using one operator in an expression. I think I will modify that when I add support for using multiple operators at a time

AndrewHess (approved):

PR #971, opened 2024-05-29
Assign metric_name for target_info metrics

venkateshamatam:
# Description
Assign metric_name for target_info metrics

# Testing
- Tested by checking that the target_info metric is processed like other metrics and no error on the log for any metric being processed.

# Checklist:

### DIFF
--- a/pkg/segment/writer/metrics/metricssegment.go
+++ b/pkg/segment/writer/metrics/metricssegment.go
@@ -527,6 +527,7 @@ func ExtractOTSDBPayload(rawJson []byte, tags *TagsHolder) ([]byte, float64, uin
 				_, err := jp.ParseString(value)
 				if err != nil {
 					log.Errorf("failed to extract tags %+v", err)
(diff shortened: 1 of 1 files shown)


pkg/segment/writer/metrics/metricssegment.go
```diff
@@ -524,15 +524,11 @@ func ExtractOTSDBPayload(rawJson []byte, tags *TagsHolder) ([]byte, float64, uin
...
+				_, err := jp.ParseString(value)
 				if err != nil {
 					log.Errorf("failed to extract tags %+v", err)
 				}
-				if temp != "target_info" {
-					mName = value
-				} else {
-					return nil
-				}
+				mName = value
```
nkunal, line 531: this is wrong, if we have an error to parsestring, then whats the point to assign the value to mname, you should just return err and not continue processing this metric

nkunal (approved):

PR #979, opened 2024-05-29
Refac: Loading of Tags Tree File + Metrics Benchmark test added for PromQL metrics flow

Macbeth98:
# Description
- Added a new metrics benchmark test for PromQl.
-  Removed the unnecessary passing of params in loading of Tag Tree files.

# Testing
- Tested by running the benchmarks.

# Checklist:

### DIFF
--- a/metricsbenchmark_test.go
+++ b/metricsbenchmark_test.go
@@ -150,10 +150,11 @@ func Benchmark_MetricsEndToEnd_OTSDB(b *testing.B) {
 	if err != nil {
 		b.Fatalf("failed to initialize metrics meta")
 	}
-	startTime := "1678060800"
-	endTime := "1678147140"
+	endTime := uint32(time.Now().Unix())
+	startTime := endTime - 24*60*60
+
 	m := "max:1h-sum:cpu.usage_guest_nice{region=*}"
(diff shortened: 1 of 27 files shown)


nkunal (commented):
lgtm
@AndrewHess please review

metricsbenchmark_test.go
```diff
@@ -172,3 +173,81 @@ func Benchmark_MetricsEndToEnd(b *testing.B) {
...
+	if err != nil {
+		b.Fatalf("failed to populate metrics meta")
+	}
+
+	err = query.InitQueryNode(getMyIds, serverutils.ExtractKibanaRequests)
+	if err != nil {
+		b.Fatalf("failed to initialize metrics meta")
+	}
+	startTime := uint32(1716997213)
+	endTime := uint32(1717008013)
```
AndrewHess, line 213: Hardcoding unix times seems like a bad idea. What happens when this is no longer a recent time range? Maybe you should set these to be the most recent 24 hours or something, relative to when the benchmark is run

AndrewHess (commented):
You could move the shared setup code of `Benchmark_MetricsEndToEnd_OTSDB()` and `Benchmark_MetricsEndToEnd_PROMQL()` to a different function so you just have that once, but we can keep it this way if we want to get this PR merged quickly

PR #982, opened 2024-05-30
Feat: Add promql logical ops

Jiale-Fang:
# Description
This is a draft pr

# Testing

# Checklist:

### DIFF
--- a/pkg/segment/segexecution.go
+++ b/pkg/segment/segexecution.go
@@ -157,8 +157,22 @@ func HelperQueryArithmeticAndLogical(queryOps []structs.QueryArithmetic, resMap 
 			}
 
 		} else {
+			labelStrSet := make(map[string]struct{})
 			for groupID, tsLHS := range resultLHS.Results {
-				if _, ok := resultRHS.Results[groupID]; !ok {
+				// groupId is like: metricName{key:value,...
+				// So, if we want to determine whether there are elements with the same labels in another metric, we need to appropriately modify the group ID.
+				rGroupID := ""
+				labelStr := ""
+				if len(groupID) >= len(resultLHS.MetricName) {
+					labelStr = groupID[len(resultLHS.MetricName):]
+					rGroupID = resultRHS.MetricName + labelStr
+				}
+
+				if queryOp.Operation == utils.LetOr || queryOp.Operation == utils.LetUnless {
+					labelStrSet[labelStr] = struct{}{}
+				}
+
+				if _, ok := resultRHS.Results[rGroupID]; !ok && queryOp.Operation != utils.LetOr && queryOp.Operation != utils.LetUnless {
 					continue
 				} //Entries for which no matching entry in the right-hand vector are dropped
 				finalResult[groupID] = make(map[uint32]float64)
@@ -200,9 +214,42 @@ func HelperQueryArithmeticAndLogical(queryOps []structs.QueryArithmetic, resMap 
 						if valueLHS != valueRHS {
 							finalResult[groupID][timestamp] = valueLHS
 						}
+					case utils.LetAnd:
+						finalResult[groupID][timestamp] = valueLHS
+					case utils.LetOr:
+						finalResult[groupID][timestamp] = valueLHS
+					case utils.LetUnless:
+						finalResult[groupID][timestamp] = valueLHS
 					}
 				}
 			}
+			if queryOp.Operation == utils.LetOr || queryOp.Operation == utils.LetUnless {
+				for groupID, tsRHS := range resultRHS.Results {
+					labelStr := ""
+					if len(groupID) >= len(resultLHS.MetricName) {
+						labelStr = groupID[len(resultLHS.MetricName):]
+					}
+
+					// For unless op, all matching elements in both vectors are dropped
+					if queryOp.Operation == utils.LetUnless {
+						lGroupID := resultLHS.MetricName + labelStr
+						delete(finalResult, lGroupID)
+						continue
+					}
+
+					_, exists := labelStrSet[labelStr]
+					// If exists, which means we already add that label set when traversing the resultLHS
+					if exists {
+						continue
+					}
+
+					finalResult[groupID] = make(map[uint32]float64)
+					for timestamp, valueRHS := range tsRHS {
+						finalResult[groupID][timestamp] = valueRHS
+					}
+				}
+			}
+
 		}
 	}
 
--- a/pkg/integrations/prometheus/promql/metricsSearchHandler.go
+++ b/pkg/integrations/prometheus/promql/metricsSearchHandler.go
@@ -1206,7 +1206,13 @@ func convertPqlToMetricsQuery(searchText string, startTime, endTime uint32, myid
 		if rhsValType == parser.ValueTypeVector {
 			lhsValType = parser.ValueTypeVector
 		}
-		return append(lhsRequest, rhsRequest...), lhsValType, []structs.QueryArithmetic{arithmeticOperation}, nil
+		req := append(lhsRequest, rhsRequest...)
+		if isLogicalOperator(expr.Op) {
+			for i := 0; i < len(req); i++ {
+				req[i].MetricsQuery.GetAllLabels = true
+			}
+		}
+		return req, lhsValType, []structs.QueryArithmetic{arithmeticOperation}, nil
 	case *parser.ParenExpr:
 		return convertPqlToMetricsQuery(expr.Expr.String(), startTime, endTime, myid)
 	default:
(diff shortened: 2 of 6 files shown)


pkg/segment/segexecution.go
```diff
@@ -157,52 +157,99 @@ func HelperQueryArithmeticAndLogical(queryOps []structs.QueryArithmetic, resMap
...
-							finalResult[groupID][timestamp] = valueLHS
+							finalResult[lGroupID][timestamp] = valueLHS
 						}
+					case utils.LetAnd:
+						finalResult[lGroupID][timestamp] = valueLHS
+					case utils.LetOr:
+						finalResult[lGroupID][timestamp] = valueLHS
+					case utils.LetUnless:
+						finalResult[lGroupID][timestamp] = valueLHS
+					}
```
AndrewHess, line 223: I was confused for a while with this part, until I found that:
1. The `And` operation doesn't have to check the RHS here because from lines 175-176, we only get here if the left and right sides have the same labels or the operation is `Or` or `Unless`
2. We do the extra handling for `Or` and `Unless` a bit lower on lines 226-251 

I think it would be good to either rearrange some code or at least add some comments about the flows for these operations, to make it more clear

pkg/segment/segexecution.go
```diff
@@ -157,52 +157,99 @@ func HelperQueryArithmeticAndLogical(queryOps []structs.QueryArithmetic, resMap
...
+					if exists {
+						continue
+					}
+
+					finalResult[rGroupID] = make(map[uint32]float64)
+					for timestamp, valueRHS := range tsRHS {
+						finalResult[rGroupID][timestamp] = valueRHS
 					}
 				}
 			}
```
AndrewHess, line 251: I think handling `LetOr` and `LetUnless` as two different `if`s will make the code more clear. Especially if we have to add more special cases later on, handling them all in one `if` is going to get confusing

pkg/segment/segexecution.go
```diff
@@ -157,52 +157,99 @@ func HelperQueryArithmeticAndLogical(queryOps []structs.QueryArithmetic, resMap
...
+						finalResult[lGroupID][timestamp] = valueLHS
+					}
+				}
+			}
+			if queryOp.Operation == utils.LetOr || queryOp.Operation == utils.LetUnless {
+				for rGroupID, tsRHS := range resultRHS.Results {
+					labelStr := ""
+					if len(rGroupID) >= len(resultLHS.MetricName) {
+						labelStr = rGroupID[len(resultLHS.MetricName):]
+					}
```
AndrewHess, line 231: This seems wrong? Why are you doing `rGroupID[len(resultLHS.MetricName):]`? Won't that give bad results if the LHS and RHS have different-length metric names?

pkg/segment/results/mresults/tests/metricsresults_test.go
```diff
@@ -617,3 +617,208 @@ func test_GetResults_Ops(t *testing.T, initialEntries map[uint32]float64, ansMap
 		}
 	}
 }
+
+func Test_GetResults_And(t *testing.T) {
+	results := make(map[string]map[uint32]float64)
+	results["test.metric.1{color:red,type:compact,"] = map[uint32]float64{
```
AndrewHess, line 623: Is `test.metric.1` supposed to have a closing `}`?

pkg/segment/results/mresults/tests/metricsresults_test.go
```diff
@@ -617,3 +617,208 @@ func test_GetResults_Ops(t *testing.T, initialEntries map[uint32]float64, ansMap
...
+
+	for timestamp, val := range initialEntries1 {
+		series.AddEntry(timestamp, val)
+	}
+
+	metricsResults1.AddSeries(series, uint64(101), tsGroupId1)
+
+	var tsGroupId2 *bytebufferpool.ByteBuffer = bytebufferpool.Get()
+	defer bytebufferpool.Put(tsGroupId2)
+	_, err = tsGroupId2.Write([]byte(mQuery1.MetricName + "{color:red,type:compact,"))
```
AndrewHess, line 752: These spots with hardcoded strings: lines 728, 739, 752. Do you have those because in your tests (e.g., line 623) you have
```
results["test.metric.1{color:yellow,type:compact,"] = ...
```
I think that's going to cause problems in the future. If someone wants to write a test with different tags, how would that work?

pkg/segment/results/mresults/tests/metricsresults_test.go
AndrewHess, line 1: I'm not quite sure what's going on in the `And`, `Or`, and `Unless` tests. Can you make the tests more clear?

AndrewHess (commented):
I'm confused by quite a bit of this PR. The logic might be correct, but I think putting in some more effort to make the code more clear will pay off in the long run as we'll have to read and update this code in the future to add more features

pkg/segment/segexecution.go
```diff
@@ -157,52 +157,99 @@ func HelperQueryArithmeticAndLogical(queryOps []structs.QueryArithmetic, resMap
...
+			labelStrSet := make(map[string]struct{})
+			for lGroupID, tsLHS := range resultLHS.Results {
+				// groupId is like: metricName{key:value,...
+				// So, if we want to determine whether there are elements with the same labels in another metric, we need to appropriately modify the group ID.
+				rGroupID := ""
+				labelStr := ""
+				if len(lGroupID) >= len(resultLHS.MetricName) {
+					labelStr = lGroupID[len(resultLHS.MetricName):]
+					rGroupID = resultRHS.MetricName + labelStr
+				}
```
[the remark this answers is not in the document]
Jiale-Fang: If there are 2 lGroupID in map `resultLHS.Results map[string]map[uint32]float64`: `metric1{color:red}`, `metric1{color:blue}`
and 1 rGroupID in map: `metric2{color:red}`

We want to check if these two metrics have same label set. So for this time series: `metric1{color:red}`, i replace the metricName at the beginning. It become `metric2{color:red}`. So i can use this key to check if this key exists in `resultRHS.Results` map

pkg/segment/segexecution.go
```diff
@@ -157,52 +157,99 @@ func HelperQueryArithmeticAndLogical(queryOps []structs.QueryArithmetic, resMap
 			}
 
 		} else {
-			for groupID, tsLHS := range resultLHS.Results {
-				if _, ok := resultRHS.Results[groupID]; !ok {
+			labelStrSet := make(map[string]struct{})
+			for lGroupID, tsLHS := range resultLHS.Results {
+				// groupId is like: metricName{key:value,...
+				// So, if we want to determine whether there are elements with the same labels in another metric, we need to appropriately modify the group ID.
+				rGroupID := ""
```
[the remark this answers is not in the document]
Jiale-Fang: Yes

pkg/segment/segexecution.go
Jiale-Fang: Yes, that's a bug. I just want to extract label set by remove the metricName at the beginning

pkg/segment/results/mresults/tests/metricsresults_test.go
Jiale-Fang: Actually, in this function the groupId is without `}`. There is somewhere else we will modify the grpID again and add `}` to it.

pkg/segment/results/mresults/tests/metricsresults_test.go
Jiale-Fang: I will extract the tags

pkg/segment/results/mresults/tests/metricsresults_test.go
```diff
@@ -959,3 +959,211 @@ func test_GetResults_Ops(t *testing.T, initialEntries map[uint32]float64, ansMap
...
+			},
+		},
+		structs.Downsampler{
+			Interval:   2,
+			Unit:       "h",
+			CFlag:      false,
+			Aggregator: structs.Aggregation{AggregatorFunction: utils.Avg},
+		},
+	)
+}
```
AndrewHess, line 1039: I think this PR can get merged like this (just address my other nit first)

But later in a different PR it might be good to refactor this to something like the following (not exactly what I have though)
```
func Test_GetResults_Or(t *testing.T) {
	metric1Series1 := "test.metric.1{color:red,type:compact,"
	metric1Series2 := "test.metric.1{color:yellow,type:compact,"
	metric2Series1 := "test.metric.2{color:red,type:compact,"

	// The n-th element contains the metric values for the n-th time-window.
	metric1Series1Values := [][]float64{
		{45},
		{1045},
	}
	metric1Series2Values := [][]float64{
		{2},
		{8, 10},
	}
	metric2Series1Values := [][]float64{
		{},
		{4, 6},
		{17},
	}


	expectedResults := make(map[string][][]float64)
	expectedResults[metric1Series1] = [][]float64{
		{45},
		{1045},
		{17},
	}
	expectedResults[metric1Series2] = [][]float64{
		{2},
		{8, 10},
	}

	helperTestFunction(t, metric1Series1, metric1Series2, metric2Series1,
		metric1Series1Values, metric1Series2Values, metric2Series1Values,
		expectedResults, utils.LetOr)
}
```
(I'm not sure my expected results are correct)

This way I've abstracted away some things that don't really matter for what we're trying to test (e.g., the downsampler), though that would still have to be handled in the helper function. I think something like this will help clarify exactly what the unit test is testing, and make it easier for others to add similar unit tests for whatever features they add

AndrewHess (commented):
Just one nit to fix, then can be merged

pkg/segment/results/mresults/tests/metricsresults_test.go
Jiale-Fang: Yes, i will refactor that test.

AndrewHess (approved):

Issue #50, opened 2023-11-08
`SELECT 1` does not work


alexey-milovidov:
Playground (http://playground.sigscalr.io/index.html) shows this:

![Screenshot_20231108_021559](https://github.com/siglens/siglens/assets/18581488/3a3b1a3f-ea84-41a7-99d9-43add13c4ccd)

alexey-milovidov:
The gray color of the text has low contrast, which makes it unreadable if you are located on the beach in sunny weather.

alexey-milovidov:
The text has inconsistent punctuation: there is a whitespace after the first colon but not after the second.

alexey-milovidov:
It pronounces SQL in an inconsistent way: first it says "SQL", then "sql".

Pranav014:
I want to work on it

AndrewHess:
Sure, go ahead. Are you going to work on the `SELECT 1` issue or the other things mentioned?

venkateshamatam:
@AndrewHess It looks like select 1 works now without an error, but it always returns 1 row with 1 column and not all the matched rows

nkunal:
It is supposed to return just 1 row but all the columns

venkateshamatam:
I can work on this! Just confirming - SELECT 1 statement specifies returning the constant value 1. So this query will only return a single column - that column will contain the value 1 for each row that satisfies the condition, if there is any condition, right? @nkunal @AndrewHess

nkunal:
It is supposed to return just 1 row but all the columns

venkateshamatam:
Okay sure, so it is essentially the same as doing SELECT * from `*` LIMIT 1 ?

PR #991, opened 2024-05-31
Fix for Issue 50

vasukalariya:
# Description
Summarize the change.

**Root cause:**
SELECT 1 is supposed to return a single column having values 1 for all the rows (number of rows = rows of table).
In our current design it seems that for data having just the hardcoded columns, we are shrinking the record size to 1 (See rrcreader.go Line [137](https://github.com/siglens/siglens/blob/develop/pkg/segment/reader/record/rrcreader.go#L137)).
Moreover, we have missed to update the numProcessedRecords in case of only hardcoded column based record and the value being used to finalize records is the default value i.e. 0. Due to this our single record too was not being picked up in our final records (See rrcreader.go Line [176](https://github.com/siglens/siglens/blob/develop/pkg/segment/reader/record/rrcreader.go#L176))

**Proposed Fix:**
Thus, setting the numProcessedRecords to 1 in case of only harcoded column based record seems to fix this issue.

https://github.com/siglens/siglens/issues/50

# Testing
Tested this code by trying out queries such as 
- SELECT 1 (Fetched 1 record as expected)
- SELECT 1, 2 (Fetched 1 record as expected)
- SELECT 1, address ((Fetched all records as expected and column 1 having 1 values)
- SELECT 1, 2, address ((Fetched all records as expected and column 1 having 1 values and same goes for column 2)

# Checklist:


AndrewHess (approved):
Looks good

nkunal (approved):

PR #1003, opened 2024-05-31
Feat: Add vector matching ops: on, ignoring and group modifiers: group_left, group_right

Jiale-Fang:
# Description
- Add vector matching ops: on, ignoring
- Add group modifiers: group_left, group_right

https://prometheus.io/docs/prometheus/latest/querying/operators/#vector-matching

# Testing
Allow running queries like: 
- testmetric1{model=\"Altima\"} + on(color) group_right testmetric0
- testmetric1{model=\"Escape Fwd\"} + on(model)  testmetric0{model=\"Escape Fwd\"}

# Checklist:


pkg/integrations/prometheus/utils/operatorutils.go
```diff
@@ -0,0 +1,215 @@
...
+	}
+
+	matches := re.FindAllStringSubmatch(labelSetStr, -1)
+	for _, match := range matches {
+		if len(match) == 3 {
+			labelKey := match[1]
+			labelVal := match[2]
+			labelSetMap[labelKey] = labelVal
+		}
+	}
```
AndrewHess, line 184: We should always get `len(match) == 3`, right? Then you can add an error log if we don't get 3 items from the match, so later we can detect issues if we get something unexpected or we update the regex string but forget to update the label extraction code

pkg/integrations/prometheus/utils/operatorutils.go
```diff
@@ -0,0 +1,215 @@
...
+
+	matchingLabelValStr := ""
+	if on {
+		for _, label := range matchingLabels {
+			val, exists := labelSetMap[label]
+			if exists {
+				matchingLabelValStr += val
+			}
+		}
+	} else { // exclude matchingLabels
```
AndrewHess, line 194: The meaning of this `on` variable isn't clear to me. Perhaps it should be renamed

pkg/integrations/prometheus/utils/operatorutils.go
```diff
@@ -0,0 +1,215 @@
...
+	}
+
+	matchingLabelValStr := ""
+	if on {
+		for _, label := range matchingLabels {
+			val, exists := labelSetMap[label]
+			if exists {
+				matchingLabelValStr += val
+			}
+		}
```
AndrewHess, line 193: Do you need some delimiter between each `val` as you write to `matchingLabelValStr`?

pkg/integrations/prometheus/utils/operatorutils.go
```diff
@@ -0,0 +1,215 @@
...
+		for _, labelKeyToRemove := range matchingLabels {
+			delete(labelSetMap, labelKeyToRemove)
+		}
+
+		labelKeyStrs := make([]string, 0)
+		for labelKey := range labelSetMap {
+			labelKeyStrs = append(labelKeyStrs, labelKey)
+		}
+
+		sort.Strings(labelKeyStrs)
```
AndrewHess, line 204: Why is it important to sort here, but not in the `if on { ...` branch?

pkg/integrations/prometheus/utils/operatorutils.go
```diff
@@ -0,0 +1,215 @@
...
+			finalResult[groupID][timestamp] = valueLHS
+		case utils.LetOr:
+			finalResult[groupID][timestamp] = valueLHS
+		case utils.LetUnless:
+			finalResult[groupID][timestamp] = valueLHS
+		}
+	}
+}
+
+func ExtractMatchingLabelSet(groupIDStr string, matchingLabels []string, on bool) string {
```
AndrewHess, line 155: Can you add a comment above this function giving a simple example of how it's supposed to work? Later if someone else wants to use this function (or quickly check if this function can be used for what they want to do), it would be great if they can understand how it works from a simple example instead of having to read quite a bit of code

pkg/segment/segexecution.go
```diff
@@ -93,136 +101,120 @@ func HelperQueryArithmeticAndLogical(queryOps []structs.QueryArithmetic, resMap
...
+					matchingLabelValStr := putils.ExtractMatchingLabelSet(rGroupID, queryOp.VectorMatching.MatchingLabels, queryOp.VectorMatching.On)
+					// Right vector is always the 'One' (vector with lower cardinality), it should not have repeated MatchingLabels.
+					_, exists := matchingLabelsComb[matchingLabelValStr]
+					if exists {
+						return nil, fmt.Errorf("HelperQueryArithmeticAndLogical: many-to-many matching not allowed: matching labels must be unique on one side")
 					}
+					matchingLabelsComb[matchingLabelValStr] = struct{}{}
+
+					matchingLabelValTorightGroupID[matchingLabelValStr] = rGroupID
 				}
```
AndrewHess, line 148: I'm confused. Line 140 says `Right vector is always the 'One'`, but line 125 says `If the left vector is the 'One' ...`. Can they both be the One? What is the `One`?

pkg/segment/structs/metricsstructs.go
```diff
@@ -141,10 +143,35 @@ type QueryArithmetic struct {
...
+	Results        map[string]map[uint32]float64
+	OperatedState  bool //true if operation has been executed
+	VectorMatching *VectorMatching
 }
 
+// VectorMatching describes how elements from two Vectors in a binary
+// operation are supposed to be matched.
+type VectorMatching struct {
+	// The cardinality of the two Vectors.
+	Card VectorMatchCardinality
```
AndrewHess, line 155: nit: rename `Card` to `Cardinality`; this makes the meaning much more clear, for only 7 extra characters

pkg/integrations/prometheus/utils/operatorutils.go
```diff
@@ -0,0 +1,215 @@
...
+		if swapped {
+			val = val * -1
+		}
+		finalResult[groupID][timestamp] = val
+	case utils.LetModulo:
+		if swapped {
+			finalResult[groupID][timestamp] = math.Mod(valueRHS, valueLHS)
+		} else {
+			finalResult[groupID][timestamp] = math.Mod(valueLHS, valueRHS)
+		}
```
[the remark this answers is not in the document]
Jiale-Fang: Sorry, i did not get you. How to extract that when their logic are different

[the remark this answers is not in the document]
AndrewHess: Something like this:
```
if swapped {
    tmp := valueLHS
    valueLHS = valueRHS
    valueRHS = tmp
}

switch ... {
    case: utils.LetSubtract:
        finalResult[groupID][timestamp] = valueLHS - valueRHS
    case: utils.LetModulo
        finalResult[groupID][timestmap] = math.Mod(valueLHS, valueRHS)
    ...
}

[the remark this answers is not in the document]
Comment (Jiale-Fang) [reply pkg/integrations/prometheus/utils/operatorutils.go:93]: Yes, make sense

File: pkg/integrations/prometheus/utils/operatorutils.go (hunk 4 above)
Comment (Jiale-Fang) [reply pkg/integrations/prometheus/utils/operatorutils.go:204]: Because `labelKeyStrs` is gennerated by a map, and map is disorder. We want to use label set as a unique identifier. So the order is matter
```
labelKeyStrs := make([]string, 0)
for labelKey := range labelSetMap {
	labelKeyStrs = append(labelKeyStrs, labelKey)
}
```

File: pkg/segment/segexecution.go (hunk 6 above)
Comment (Jiale-Fang) [reply pkg/segment/segexecution.go:148]: It describes the cardinality relationship of two Vectors in a binary operation. So it has `One-to-one vector matches` and `Many-to-one and one-to-many vector matches`

Comment (AndrewHess) [reply pkg/segment/segexecution.go]: Ok, so on line 143 we'd change that later if we add many-to-many operators? And for one-to-many and many-to-one we'll swap left and right if necessary so that the right is the `one` in those cases?

File: pkg/integrations/prometheus/utils/operatorutils.go (hunk 4 above)
Comment (Jiale-Fang) [reply pkg/integrations/prometheus/utils/operatorutils.go:235]: I need to do some preprocessing like this. So i think maybe ues map will be better
```
for _, labelKeyToRemove := range matchingLabels {
			delete(labelSetMap, labelKeyToRemove)
		}
```

File: pkg/segment/segexecution.go (hunk 6 above)
Comment (Jiale-Fang) [reply pkg/segment/segexecution.go]: Yes, right will also be the `one` in this case. i will add a comment above to mark that. If we add `many-to-many operators` in the future, we should change line 143.

File: pkg/integrations/prometheus/utils/operatorutils.go [hunk 9]
```diff
@@ -0,0 +1,246 @@
...
+		finalResult[groupID][timestamp] = valueLHS / valueRHS
+	case utils.LetMultiply:
+		finalResult[groupID][timestamp] = valueLHS * valueRHS
+	case utils.LetSubtract:
+		val := valueLHS - valueRHS
+		if swapped {
+			val = val * -1
+		}
+		finalResult[groupID][timestamp] = val
```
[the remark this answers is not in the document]
Jiale-Fang: Yes, i forget to modify this one

pkg/integrations/prometheus/utils/operatorutils.go
```diff
@@ -0,0 +1,246 @@
...
+	matchingLabelValStr := ""
+	// When includeColumns is true, use matchingLabels to create a combination.
+	// Otherwise, use all other labels except matchingLabels to create a combination.
+	if includeColumns {
+		for _, label := range matchingLabels {
+			val, exists := labelSetMap[label]
+			if exists {
+				matchingLabelValStr += (val + ",")
+			}
+		}
```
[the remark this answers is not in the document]
Jiale-Fang: Yes, before we return the result. There is a func called removeTrailingComma to do post processing for the grpID str.

pkg/integrations/prometheus/utils/operatorutils.go
```diff
@@ -0,0 +1,243 @@
...
+			if queryOp.ReturnBool {
+				finalResult[groupID][timestamp] = 1
+				return
+			}
+			if swapped {
+				finalResult[groupID][timestamp] = valueRHS
+			} else {
+				finalResult[groupID][timestamp] = valueLHS
+			}
+		}
```
AndrewHess, line 108: You're still doing some swapping logic inside `case`s. Since you've moved the swapping logic above the `switch`, you need to remove it from all of the `case`s. Please go through this whole switch statement again and remove all swapping logic.

Also, why are the PR tests passing with this bug? You should probably add more unit tests that will test this issue of incorrectly swapping twice instead of once.

pkg/integrations/prometheus/utils/operatorutils.go
```diff
@@ -0,0 +1,243 @@
...
+			log.Errorf("SetFinalResult: does not support using this operator: %v", queryOp.Operation)
+		}
+	}
+}
+
+// This method will return uniuqe tag key value pair combination str, we refer it as labelStr.
+// If matchingLabels = [tag1], groupIDStr = "metricName{tag1:val1,tag2:val2,tag3:val3}""
+// When includeColumns is true, extract labels and generate labelStr from matchingLabels based on the previous groupIDStr. The result is {tag1:val1}
+// Otherwise, exclude the labels, so the result is {tag2:val2,tag3:val3}
+func ExtractMatchingLabelSet(groupIDStr string, matchingLabels []string, includeColumns bool) string {
```
AndrewHess, line 187: I like that this comment describes what is supposed to happen pretty clearly. But I don't think this is what the function will actually do. You should add some unit tests. I'm pretty sure that instead of returning something like `" {tag2:val2,tag3:val3}"`, it would return something like `"{tag2:val2,tag3:val3,"`

pkg/integrations/prometheus/utils/operatorutils.go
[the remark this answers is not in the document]
AndrewHess: Ok. But `removeTrailingComma()` isn't called in this function, so then your comment above the function describing what it should return is not correct.

pkg/integrations/prometheus/utils/operatorutils.go
AndrewHess: You can get a similar result using a slice too; the runtime would be a little worse, but probably still fine. Is it important to preserve the order of the tags?

pkg/integrations/prometheus/utils/operatorutils.go
Jiale-Fang: I think we still need the swapped logic for some operators. For example, for comparison operators, if the equation evaluates to true, we should preserve the value of the vector `on the left`. However, since we moved the swap logic out of the switch case, we want to ensure that the preserved value is from `the left vector`.

pkg/integrations/prometheus/utils/operatorutils.go
Jiale-Fang: Yeah, i will update the comments to make it more accurate

pkg/integrations/prometheus/utils/operatorutils.go
Jiale-Fang: Yes, the order is important because we concatenate the tag key-value pairs as a unique identifier. We use that to check if another vector has the exact same label set or not.

pkg/integrations/prometheus/utils/operatorutils.go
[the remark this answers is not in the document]
Jiale-Fang: Yes, i will modify the comment

AndrewHess (approved):

PR #1019, opened 2024-06-03
Cardinality support for sigclient

venkateshamatam:
# Description
Cardinality support for sigclient

# Testing
- Tested default cardinality without specifying the -u flag.
- Tested custom cardinality using the -u flag.
- Tested that the payload being sent actually loops through all the pregenerated series.
- Tested log ingestion to make sure it is not affected and -u flag has no effect on log ingestion.

Tested using the below command - 

```
go run main.go ingest metrics -d http://localhost:8081/otsdb -t 200000 -m 100 -p 1 -b 10_000 -g benchmark -u 5000000

INFO[0000] processCount : 1                             
INFO[0000] dest : http://localhost:8081/otsdb           
INFO[0000] totalEvents : 200000. Continuous: false      
INFO[0000] batchSize : 10000. Num metrics: 100          
INFO[0000] bearerToken :                                
INFO[0000] generatorType : benchmark.                   
INFO[0000] cardinality : 5000000.                       
INFO[0000] Starting ingestion at http://localhost:8081/otsdb for OTSDB 
INFO[0000] Initializing benchmark reader                
INFO[0008] Generated 5000000 unique series in 8.989280292s 
INFO[0008] Initializing benchmark reader                
INFO[0009] Total events ingested:+200000. Event type: OTSDB 
INFO[0009] Total Time Taken for ingestion 973.615959ms. Average events per second=205,419 
INFO[0009] Total HLL Approx of unique timeseries:4,747,032
```
# Checklist:


tools/sigclient/pkg/ingest/ingest.go
```diff
@@ -151,10 +152,60 @@ func generateOpenTSDBBody(recs int, rdr utils.Generator) ([]byte, error) {
...
+	var builder strings.Builder
+	builder.WriteString(metricName)
+	for key, value := range tags {
+		builder.WriteString(key)
+		builder.WriteString(fmt.Sprintf("%v", value))
+	}
+	return builder.String()
+}
+
+func generatePredefinedSeries(nMetrics int, cardinality int, gentype string) error {
```
nkunal, line 169: don't use `int` for cardinality, int will have a max value of `2 billion`, at some point we will test cardinality in trillions. so use a uint64

tools/sigclient/pkg/ingest/ingest.go
```diff
@@ -151,10 +152,60 @@ func generateOpenTSDBBody(recs int, rdr utils.Generator) ([]byte, error) {
...
+	preGeneratedSeries = make([]map[string]interface{}, 0, cardinality)
+	rdr, err := utils.InitMetricsGenerator(nMetrics, gentype)
+	if err != nil {
+		return err
```
nkunal, line 173: log a error here

tools/sigclient/pkg/ingest/ingest.go
```diff
@@ -151,10 +152,60 @@ func generateOpenTSDBBody(recs int, rdr utils.Generator) ([]byte, error) {
...
+	}
+	start := time.Now()
+	uniqueSeriesMap := make(map[string]struct{})
+	for len(preGeneratedSeries) < cardinality {
+		currPayload, err := rdr.GetRawLog()
+		if err != nil {
+			return err
```
nkunal, line 180: log error here

tools/sigclient/pkg/ingest/ingest.go
```diff
@@ -173,18 +224,24 @@ func runIngestion(iType IngestType, rdr utils.Generator, wg *sync.WaitGroup, url
...
+		if !continuous && eventCounter+batchSize > totalEvents {
 			recsInBatch = totalEvents - eventCounter
 		}
 		i++
 		if iType == ESBulk {
 			bb = bytebufferpool.Get()
 		}
-		payload, err := generateBody(iType, recsInBatch, i, rdr, actLines, bb)
+		if iType == OpenTSDB {
+			payload, err = generateBodyFromPredefinedSeries(recsInBatch)
```
nkunal, line 241: you need to keep track of which series, you have gone through in the previous iteration, else this for loop is only going to send data for the first x series

tools/sigclient/pkg/ingest/ingest.go
```diff
@@ -151,10 +152,60 @@ func generateOpenTSDBBody(recs int, rdr utils.Generator) ([]byte, error) {
...
+	}
+	elapsed := time.Since(start)
+	log.Infof("Generated %d unique series in %v", len(uniqueSeriesMap), elapsed)
+	return nil
+}
+
+func generateBodyFromPredefinedSeries(recs int) ([]byte, error) {
+	finalPayLoad := make([]interface{}, recs)
+	for i := 0; i < recs; i++ {
+		series := preGeneratedSeries[i%len(preGeneratedSeries)]
```
nkunal, line 196: you should not use the `i` for both for loop variable and to find the seriesId. Rather this series id var should be one continuous var, that keeps track of which series have been used so far

tools/sigclient/pkg/ingest/ingest.go
```diff
@@ -151,16 +151,68 @@ func generateOpenTSDBBody(recs int, rdr utils.Generator) ([]byte, error) {
...
+func generatePredefinedSeries(nMetrics int, cardinality uint64, gentype string) error {
+	preGeneratedSeries = make([]map[string]interface{}, 0, cardinality)
+	rdr, err := utils.InitMetricsGenerator(nMetrics, gentype)
+	if err != nil {
+		log.Errorf("generatePredefinedSeries: failed to initalize metrics generator: %+v", err)
+		return err
+	}
+	start := time.Now()
+	uniqueSeriesMap := make(map[string]struct{})
+	for uint64(len(preGeneratedSeries)) < cardinality {
```
nkunal, line 177: we know how much cardinality we want and what should be the length of `preGeneratedSeries`. 
so:

1. pre create the `preGeneratedSeries` array of `cardinality` length
2. change the for loop to be counter that we would run `cardinality` times
3. don;t do append (its mem expensive) , rather just insert into the `preGeneratedSeries` array at the for loop var index

tools/sigclient/pkg/ingest/ingest.go
```diff
@@ -151,16 +151,68 @@ func generateOpenTSDBBody(recs int, rdr utils.Generator) ([]byte, error) {
...
+		currPayload, err := rdr.GetRawLog()
+		if err != nil {
+			log.Errorf("generatePredefinedSeries: failed to get raw log: %+v", err)
+			return err
+		}
+		uniqueID := generateUniqueID(currPayload)
+		if _, exists := uniqueSeriesMap[uniqueID]; !exists {
```
nkunal, line 184: you should not be doing this existence check here, the `generateUniqueID` function is suppose to generate a unique id, why don't you just do this check there ?

nkunal: so then this for loop becomes easy, it always knows it is going to get a uniqye id and all it has to do is insert in the array

tools/sigclient/pkg/ingest/ingest.go
```diff
@@ -151,16 +151,68 @@ func generateOpenTSDBBody(recs int, rdr utils.Generator) ([]byte, error) {
...
+	}
+	elapsed := time.Since(start)
+	log.Infof("Generated %d unique series in %v", len(uniqueSeriesMap), elapsed)
+	return nil
+}
+
+func generateBodyFromPredefinedSeries(recs int, seriesId *uint64) ([]byte, error) {
+	finalPayLoad := make([]interface{}, recs)
+	for i := 0; i < recs; i++ {
+		series := preGeneratedSeries[*seriesId%uint64(len(preGeneratedSeries))]
```
nkunal, line 197: we don;t need to calculate length of `preGeneratedSeries` everytime. just pass it as a var to this func and calculate it once

tools/sigclient/pkg/ingest/ingest.go
```diff
@@ -173,18 +225,24 @@ func runIngestion(iType IngestType, rdr utils.Generator, wg *sync.WaitGroup, url
...
+		if !continuous && eventCounter+batchSize > totalEvents {
 			recsInBatch = totalEvents - eventCounter
 		}
 		i++
 		if iType == ESBulk {
 			bb = bytebufferpool.Get()
 		}
-		payload, err := generateBody(iType, recsInBatch, i, rdr, actLines, bb)
+		if iType == OpenTSDB {
+			payload, err = generateBodyFromPredefinedSeries(recsInBatch, &seriesId)
```
nkunal, line 242: you dont need to pass `seriesID` as a var, just add the `recsInBatch` to it after this func call. It also saves the additions you do everytime in the for loop in that function.

tools/sigclient/pkg/ingest/ingest.go
```diff
@@ -292,8 +356,12 @@ func StartIngestion(iType IngestType, generatorType, dataFile string, totalEvent
 		if err != nil {
 			log.Fatalf("StartIngestion: failed to initalize reader! %+v", err)
 		}
-		go runIngestion(iType, reader, &wg, url, totalEventsPerProcess, continuous, batchSize, i+1, indexPrefix,
-			&totalSent, bearerToken, indexName, numIndices)
+		if iType == OpenTSDB {
+			go runIngestion(iType, reader, &wg, url, totalEventsPerProcess, continuous, batchSize, i+1, indexPrefix, &totalSent, bearerToken, indexName, numIndices)
+		} else {
+			go runIngestion(iType, reader, &wg, url, totalEventsPerProcess, continuous, batchSize, i+1, indexPrefix,
```
nkunal, line 362: I don't see any difference between these go routine calls. what am I missing ?

AndrewHess (commented):
You should also add this cardinality option to the readme

tools/sigclient/pkg/ingest/ingest.go
```diff
@@ -224,7 +223,7 @@ func runIngestion(iType IngestType, rdr utils.Generator, wg *sync.WaitGroup, url
 	t.MaxIdleConns = 500
 	t.MaxConnsPerHost = 100
 	t.MaxIdleConnsPerHost = 100
-	preGeneratedSeriesLength := uint64(len(preGeneratedSeries))
+	seriesId := uint64(0)
```
AndrewHess, line 226: Are you trying to set the global `seriesId` here? Using `:=` will make a new local variable called `seriesId`

venkateshamatam: I removed the local var, it was unnecessary since we only need the global variable and the default value is 0.

AndrewHess (approved):

PR #1020, opened 2024-06-03
Check if file is empty before trying to unmarshal and update logs

venkateshamatam:
# Description
Check if file is empty before trying to unmarshal and update logs. Walked through methods called during startup and checked cases where json unmarshal could fail due to empty files, either updated the logs for better debugging or skipped unmarshal step for those files.
# Testing
- Tested startup with pre-existing data dir 
- Tested startup as a fresh install.
# Checklist:

### DIFF
--- a/pkg/blob/local/localstorage.go
+++ b/pkg/blob/local/localstorage.go
@@ -47,7 +47,7 @@ func InitLocalStorage() error {
 			log.Errorf("InitLocalStorage: Error reading SegSetKeys file %s: %v", filePath, err)
 			return err
 		}
-	} else if len(file) == 0 {
+	} else if len(strings.TrimSpace(string(file))) == 0 {
 		log.Warnf("InitLocalStorage: Empty file found at %s", filePath)
 	} else {
 		err := json.Unmarshal(file, &segSetKeys)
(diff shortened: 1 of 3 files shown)


pkg/blob/local/localstorage.go
```diff
@@ -47,8 +47,11 @@ func InitLocalStorage() error {
 			log.Errorf("InitLocalStorage: Error reading SegSetKeys file %s: %v", filePath, err)
 			return err
 		}
+	} else if len(file) == 0 {
```
nkunal: thsi code could be formatted a bit better:
`if err != nil {
		... blah blha
	} 

if len(file) == 0 {
 log error and return
}
err := json.Unmarshal(file, &segSetKeys)
.....
`
this way it reduces the deep if-els-eif-else...

pkg/config/config.go
```diff
@@ -529,6 +529,10 @@ func ReadRunModConfig(fileName string) (common.RunModConfig, error) {
 
 func ExtractReadRunModConfig(jsonData []byte) (common.RunModConfig, error) {
 	var runModConfig common.RunModConfig
+	if len(jsonData) == 0 {
```
nkunal: can you make sure, when there is no running config, we don;t error out or cause any othe issues. It is fine that the runMod config  may not have any data

venkateshamatam: we log an error message but we dont stop server startup if there's an error related to runMod config.

nkunal (approved):

PR #1031, opened 2024-06-04
Feat: Add promql aggregation ops (stddev, stdvar, group)

Jiale-Fang:
# Description
- This is a draft pr, will wait for `vector matching ops` pr get merged and then add unit test
Add promql aggregation ops (stddev, stdvar, group)
> Will open another pr to finish the `topk` and `bottomk`

# Testing

# Checklist:


pkg/segment/results/mresults/seriesresult.go
```diff
@@ -777,6 +777,16 @@ func reduceEntries(entries []Entry, fn utils.AggregateFunctions, fnConstant floa
...
 		}
+	case utils.Stddev:
+		fallthrough
+	case utils.Stdvar:
+		sum := 0.0
+		for i := range entries {
+			sum += entries[i].dpVal
+		}
+		ret = sum / float64(len(entries))
+	case utils.Group:
```
AndrewHess, line 788: How is the `utils.Stddev` case correct? Don't you have to take the square root?

AndrewHess (commented):
Looks pretty good so far. Across parser.go and metricsSearchHandler.go you have three identical `switch` statements for setting `mquery.Aggregator.AggregatorFunction` based on an `aggFunc`. Can you move this `switch` to a separate function and then call it three times instead of having three `switch`es? Either in this PR or a different PR would be ok

pkg/segment/results/mresults/seriesresult.go
Jiale-Fang: This reduce logic is the downsample logic not the aggregation logic. We downsample data points to each timestamp. 
For example: 

```
1. seriesId: {color: red}, timestampToVals: {1: 3, 2: 4, 5: 5...}
2. seriesId: {color: yellow}, timestampToVals: {1: 4, 2: 5, 5: 6...}

Downsample interval is 1 h.
So After Downsample, they will be:
1. seriesId: {color: red}, timestampToVals: {0: 4}
2. seriesId: {color: yellow}, timestampToVals: {0: 5}

Then do aggregation ...

```

pkg/integrations/prometheus/utils/operatorutils.go
```diff
@@ -260,3 +260,17 @@ func ExtractMatchingLabelSet(groupIDStr string, matchingLabels []string, include
...
+	for _, val := range values {
+		sum += val
+	}
+	avg := sum / length
+	sumValSquare := 0.0
+	for _, val := range values {
+		sumValSquare += (val - avg) * (val - avg)
+	}
+	return sumValSquare / length
+}
```
AndrewHess, line 276: nit: Perhaps a better place for this function would be in the main `utils` package. Something like `/pkg/utils/mathutils.go`. What do you think? Then later if we wanted to for example find the variance of latencies for traces, we could get it from the main utils package instead of this prometheus utils package

pkg/segment/results/mresults/metricresults.go
```diff
@@ -752,3 +758,48 @@ func (r *MetricsResult) computeAggCount(aggregation structs.Aggregation) {
...
 }
+
+// Retrieve all series values at each timestamp and calculate the results based on those values.
+func (r *MetricsResult) computeAggStdvarOrStddev(aggregation structs.Aggregation) {
+	timestampToVals := make(map[uint32][]float64)
+	r.Results = make(map[string]map[uint32]float64)
+
+	// If we use group by for this vector, we will only perform aggregation for each time series at each timestamp.
+	// Since there is only one value, the standard variance will be 0.
+	if len(aggregation.GroupByFields) > 0 {
```
AndrewHess, line 769: I don't quite understand this comment. I do understand that if we're computing the variance for just one vector, then we compute the variance at each time window and since there's only one point the variance will be 0.

But isn't it possible that we group by something and then each group has more than one series? So then the variance could be non-zero? I think I'm missing something

pkg/integrations/prometheus/utils/operatorutils.go
Jiale-Fang: Yes, i agree

pkg/segment/results/mresults/metricresults.go
Jiale-Fang: Yes, 1 group may have multiple series, i will modify that

AndrewHess (approved):

PR #1035, opened 2024-06-05
fix: sending correct queries when timepicker is changed

### LINKED ISSUE #1002: [BUG] The Metrics Explorer is not sending correct queries when the TimePicker is Changed.
# Description
- On the Metrics Explorer, there is a query entered with metric name, from (tag Filters) and group by fields selected. 
- But when I changed the time picker to for example `Last 24 Hours` from `Last 1 Hour`, the queries that sent to the backend through the Timeseries API Payload only has metric name and the tag Filters and group by fields are being ignored.
- It is working the first time you change, but if you change it again then this bug will appear.

# Steps to Reproduce
1. Go To Metrics Explorer Page and run two Queries (a, b) with the default time and include tag Filters and also Group by Fields for each Query.
2. Change the Time picker two times to some other time and now open the Network and check the payload the query being sent contains only the Metric Name. 
3. See the image below

# Expected Behavior
- It is supposed to sent the same exact queries that is being shown on the UI.

# Additional Context
image

navyayy:
# Description
When the `getTagKeyValue` function is called and the new values are assigned to `everywhere` and `everything` sources, which resets the global `queries` object. So, in this change, we are keeping the queries as it is.

Fixes #1002 

# Testing
I tested this by running two queries (a, b) with the default time and included tag filters and also group by fields for each query and then checking the time picker two times to some other time.

# Checklist:


nkunal:
this does not work for the following scenario:

select some metricname, select some from, select `avg by` `car_type`, change the timewindow, then select another `avg by` of `color`. In the below screenshot you can see it only picks the color as the avg by but not the car_type.
 

image

navyayy:
> this does not work for the following scenario:
> 
> select some metricname, select some from, select `avg by` `car_type`, change the timewindow, then select another `avg by` of `color`. In the below screenshot you can see it only picks the color as the avg by but not the car_type.
> 
> image

@nkunal I tested the same way you wrote, I am getting both `car_type` and `color`, please review it and let me know if the issue still persists
image

sonamgupta21 (approved):

static/js/metrics-explorer.js
```diff
@@ -70,16 +69,17 @@ async function metricsExplorerDatePickerHandler(evt) {
...
-        
-        queryDetails.everywhere = [];
-        queryDetails.everything = [];
         availableEverywhere = tagsAndValue.availableEverywhere.sort();
         availableEverything = tagsAndValue.availableEverything[0].sort();
         const queryElement = $(`.metrics-query .query-name:contains(${queryName})`).closest('.metrics-query');
         queryElement.find('.everywhere').autocomplete('option', 'source', availableEverywhere);
         queryElement.find('.everything').autocomplete('option', 'source', availableEverything);
+
+        queries[queryName] = queryDetails;
```
sonamgupta21, line 80: This line is not needed, as in the code above, we are not changing the queryDetails object.

sonamgupta21 (approved):

PR #1043, opened 2024-06-06
Fix: Dual index bug

Jiale-Fang:
# Description
If we have two indices, `ind-0` and `ind-1`, `ind-0` index contains `http_status` while `ind-1` index contains `http_method`

In the past, If the filter query uses regex(E.g. http_status=2*), it will give us an error as shown below, since `ind-1` does not contain field `http_status`
image

After this pr, query should just return rows meet the condition without giving error

# Testing

# Checklist:

### DIFF
--- a/pkg/segment/search/conditioncheck.go
+++ b/pkg/segment/search/conditioncheck.go
@@ -83,7 +83,8 @@ func ApplyColumnarSearchQuery(query *SearchQuery, multiColReader *segread.MultiC
 	case RegexExpression:
 		rawColVal, err := multiColReader.ReadRawRecordFromColumnFile(query.QueryInfo.ColName, blockNum, recordNum, qid)
 		if err != nil {
-			return false, err
+			log.Errorf("ApplyColumnarSearchQuery: %v", err)
+			return false, nil
 		}
 		return writer.ApplySearchToExpressionFilterSimpleCsg(query.QueryInfo.QValDte, query.ExpressionFilter.FilterOp, rawColVal, true, holderDte)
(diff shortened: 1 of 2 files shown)


pkg/segment/search/conditioncheck.go
```diff
@@ -83,7 +83,8 @@ func ApplyColumnarSearchQuery(query *SearchQuery, multiColReader *segread.MultiC
 	case RegexExpression:
 		rawColVal, err := multiColReader.ReadRawRecordFromColumnFile(query.QueryInfo.ColName, blockNum, recordNum, qid)
 		if err != nil {
-			return false, err
+			log.Errorf("ApplyColumnarSearchQuery: %v", err)
```
nkunal, line 86: on a failure scenario , logging this error will cause a huge flood of the logs lines.

pkg/segment/search/conditioncheck.go
```diff
@@ -83,7 +83,8 @@ func ApplyColumnarSearchQuery(query *SearchQuery, multiColReader *segread.MultiC
 	case RegexExpression:
 		rawColVal, err := multiColReader.ReadRawRecordFromColumnFile(query.QueryInfo.ColName, blockNum, recordNum, qid)
 		if err != nil {
-			return false, err
+			log.Debugf("ApplyColumnarSearchQuery: %v", err)
```
nkunal, line 86: the log should be a bit clear. something like `Apply... failed to read rec from column file. qid: %v ...`

nkunal (approved):

PR #1082, opened 2024-06-08
Feat: Add Spl stats subcmd parsing logic

Jiale-Fang:
# Description
- Add stats cmd options parsing logic (https://docs.splunk.com/Documentation/Splunk/9.2.1/SearchReference/stats#Optional_arguments)
- Add stats cmd functions parsing logic (https://docs.splunk.com/Documentation/Splunk/9.2.1/SearchReference/stats#Stats_function_options)

# Testing
1. Optional Arguments
```
city=Boston | stats count By city dedup_splitvals=true
city=Boston | stats count By city allnum=true
city=Boston | stats count By city partitions=2
city=Boston | stats count By city delim="a"
city=Boston | stats count delim="a" partitions=2 allnum=true
```

2. Stats Functions
```
* | stats estdc(latency) 
* | stats estdc_error(latency) 
* | stats exactperc99(latency) as "99th Percentile"
* | stats perc66.6(latency)
* | stats median(latency) 
* | stats stdev(latency) 
* | stats stdevp(latency) 
* | stats sumsq(latency) 
* | stats upperperc6.6(latency) 
* | stats var(latency) 
* | stats varp(latency) 
* | stats first(latency) 
* | stats last(latency) 
* | stats list(latency) 
* | stats earliest(latency) 
* | stats earliest_time(latency) 
* | stats latest(latency) 
* | stats latest_time(latency) 
* | stats rate(latency) 
```
Using eval statements inside stats funcs
```
* | stats estdc(eval(http_method="GET")) as newField
* | stats estdc_error(eval(http_method="GET"))
* | stats exactperc9.8(eval(http_method="GET"))
* | stats perc66(eval(http_method="GET"))
* | stats upperperc6.6(eval(http_method="GET"))
* | stats median(eval(http_method="POST"))
* | stats mode(eval(http_method="POST"))
* | stats stdev(eval(http_method="POST"))
* | stats stdevp(eval(http_status>300))
* | stats sumsq(eval(http_status>300))
* | stats var(eval(http_status>300))
* | stats varp(eval(http_status>300))
* | stats first(eval(http_status>300))
* | stats last(eval(http_status>300))
* | stats list(eval(weekday))
* | stats earliest(eval(http_status>300))
* | stats earliest_time(eval(http_status>300))
* | stats latest(eval(http_status>300))
* | stats latest_time(eval(http_status>300))
* | stats rate(eval(http_status>300))
```

# Checklist:


pkg/ast/spl/spl.peg
```diff
@@ -320,9 +320,78 @@ AggregatorBlock <- PIPE CMD_STATS aggs:AggregationList byFields:(GroupbyBlock)?
...
+// Optional arguments for stats block
+StatsOptions <- option:(StatsOption)*
+{
+    //Default value
+    options := &structs.StatsOptions {
+        Delim: " ",
+        Partitions: 1,
+        DedupSplitvals: false,
+        Allnum: false,
+    }
```
AndrewHess, line 347: nit: maybe you should have a function that returns a default `StatsOptions`, so you don't have to define the default in both of these places

pkg/ast/spl/spl.peg
```diff
@@ -2157,6 +2227,507 @@ AggSum <- "sum" L_PAREN "eval" L_PAREN boolComparisonExpr:BoolComparisonExpr R_P
...
+    agg := &structs.MeasureAggregator {
+        MeasureCol: field.(string),
+        MeasureFunc: utils.ExactPerc,
+        Param: numStr.(string),
+    }
+
+    return agg, nil
+}
+
+// Todo: implement the processing logic
```
AndrewHess, line 2307: For these `Todo`s, you mean implement the logic in the backend to actually support these commands, right? If so, I don't think you need these TODOs in this file, since the thing to do wouldn't be implemented in this file. A better place would be in segaggs.go where we do the other processing, we can check the `MeasureFunc` and add some sort of `Not implemented` message

pkg/ast/spl/spl.peg
AndrewHess, line 1: Almost all of these new functions you're parsing have the same structure (e.g., median, mode, stddev, etc.). Do you think you could restructure the parsing so there's not so much repeated code? If not, what you have should be fine

pkg/ast/spl/spl.peg
Jiale-Fang: Yes, but I'm actually not sure where exactly to implement these methods because they could be in `segaggs.go, segresults.go, evalaggs.go`, or possibly other files. What should I do in this situation?

pkg/ast/spl/spl.peg
Jiale-Fang: Yes, and for the existing agg like sum, min, max ... They all share some similarities. So probably refactoring them in another prs will be better. I will create a github issue for that.

pkg/ast/spl/spl.peg
AndrewHess: Hmm good point. I think it should flow through segaggs.go anyways though, even if it's not implemented there. So could you add a `Not implemented` message there and later when we actually implement these we would remove that message but can implement it wherever it actually should be implemented? Do you think that would work?

Right now it I think these queries are silently failing because they're ignoring the new data being parsed. This is going to lead to a lot of customer confusion because running the query doesn't give any indication that we don't support those options/functions, but the results won't be correct in many cases.

AndrewHess:
You should also add some unit tests to splParser_test.go

pkg/segment/aggregations/segaggs.go
```diff
@@ -121,6 +121,12 @@ func PostQueryBucketCleaning(nodeResult *structs.NodeResult, post *structs.Query
 	return nodeResult
 }
 
+/*
+* TODO: The processing logic for SPL commands that need to be implemented is outlined below. These commands may need to be implemented elsewhere.
+ 1. Stats cmd:
+    1.1 stats options: dedup_splitvals, allnum, partitions, delim
+    1.2 stats functions: estdc, estdc_error, exactperc99, perc66.6, median, stdev, stdevp, sumsq, upperperc6.6, var, varp, first, last, list, earliest, earliest_time, latest, latest_time, rate
+*/
 func performAggOnResult(nodeResult *structs.NodeResult, agg *structs.QueryAggregators, recs map[string]map[string]interface{},
```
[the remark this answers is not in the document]
Jiale-Fang: I added the error handling logic after the parser finished parsing, because there might be some issues with `nodeResult.ErrList`. Even when I arbitrarily added an error to that list, the frontend did not display anything, and I could not find where `ErrList` is returned. I have created a GitHub issue for that. https://github.com/siglens/siglens/issues/1101

pkg/segment/structs/segstructs.go
```diff
@@ -672,3 +681,40 @@ func (qtype QueryType) String() string {
...
+
+func CheckUnsupportedFunctions(agg *QueryAggregators) error {
+	if agg == nil {
+		return nil
+	}
+
+	if agg.GroupByRequest != nil {
+		for _, measureAgg := range agg.GroupByRequest.MeasureOperations {
+			if isUnsupported(measureAgg.MeasureFunc) {
+				return fmt.Errorf("checkUnsupportedFunctions: does not support using this stats function: %v", measureAgg.MeasureFunc.String())
```
AndrewHess, line 709: nit: change `does not support using this stats function: %v` to `using %v in stats is not yet supported`. Since this will be shown to users, we want to be clear that it's temporarily unsupported but will be added later

Same comment for line 716

AndrewHess (commented):
These changes look good, I really like how you cleaned up the spl parsing tests.

When I run a query on the UI using one of the unsupported functions, I correctly get a message saying it's not supported. But when I run a query with one of the unsupported options
```
city=Boston | stats count By city allnum=true
```
the query runs the same as without the `allnum=true` option. This query should also fail and show a message that `allnum` is not yet supported

PR #1083, opened 2024-06-09
Exclude '_index' from columnsOrder when 'fields' command is entered

### LINKED ISSUE #965: [BUG] Exclude '_index' from columnsOrder when 'fields' command is entered
# Description
When the fields command is entered, the _index field should not be included in the columnsOrder. This is to ensure that the output only contains the relevant fields specified by the user.

# Steps to Reproduce
1. On logs search screen run a query including fields. ex `* | fields city, user_color, app_name`
2. See the backend WS response it contains `_index` in columnsOrder. 
image

# Expected Behavior
Should only include the fields user has entered and timestamp.

# Environment
- Commit Hash: [e.g. 9ed5f0a]
- OS: [e.g. MacOS 14.1.1 on apple silicon]
- Browser and Version (if applicable): [e.g. chrome 119.0.6045.159 arm64, safari 19616.2.9.11.7]

ahegde3:
# Description
Summarize the change.
- Adding logic to exclude **_index** based on the length of allMatchedColumns.
- Changes unit test to handle exclusion of index.

Fixes #965  (link all the GitHub issues this addresses)

# Testing
- Running query on frontend for different combination of fields.
- Running all unit tests using command  ``` make all ```

# Checklist:

### DIFF
--- a/pkg/segment/reader/record/recordreader.go
+++ b/pkg/segment/reader/record/recordreader.go
@@ -58,6 +58,10 @@ func GetRecordsFromSegment(segKey string, vTable string, blkRecIndexes map[uint1
 			return nil, allCols, errors.New("failed to get column names for segkey in rotated and unrotated files")
 		}
 	}
+
+	//keep track of total column count before applying filter.
+	totalColumnCount := len(allCols)
+
 	allCols = applyColNameTransform(allCols, aggs, colsIndexMap, qid)
 	numOpenFds := int64(len(allCols))
 	err = fileutils.GLOBAL_FD_LIMITER.TryAcquireWithBackoff(numOpenFds, 10, fmt.Sprintf("GetRecordsFromSegment.qid=%d", qid))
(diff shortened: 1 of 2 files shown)


pkg/segment/reader/record/recordreader.go
```diff
@@ -166,7 +170,12 @@ func GetRecordsFromSegment(segKey string, vTable string, blkRecIndexes map[uint1
 
 		for r := range resultAllRawRecs {
 			resultAllRawRecs[r][config.GetTimeStampKey()] = recordIdxTSMap[r]
-			resultAllRawRecs[r]["_index"] = vTable
+
+			// if the total column count is less than or equal to the number of columns we have read, it indicates there is no field operations.
+			if totalColumnCount <= len(allMatchedColumns) {
+				resultAllRawRecs[r]["_index"] = vTable
+				addedExtraFields = true
+			}
```
AndrewHess, line 178: The truth of this comment isn't obvious to me. Can you explain why this is true?

ahegde3: **totalColumnCount** has count of all the columns before applying the filter. The condition checks whether the **allMatchedColumns** has the same number or more columns. If its **allMatchedColumns** has less number of columns than **totalColumnCount**, we can safely say **fields** query was used.

AndrewHess: I understand what the code and comment say, I just don't understand why it's correct. _Why_ can we infer `fields` was not used if `totalColumnCount <= len(allMatchedColumns)`?

What if `fields` was used to select every single column, or if `eval` (see https://docs.splunk.com/Documentation/Splunk/9.1.1/SearchReference/Eval) was used to create multiple columns, and then `fields` was used to select any number of columns (either more or fewer columns that the number of raw columns in the segment)?

AndrewHess:
Closing due to inactivity. @ahegde3 you can reopen this if you still want to work on it

PR #1086, opened 2024-06-10
Index dropdown changes , allow regex string, add * if all index names are removed

sunitakawane:
# Description
Summarize the change.

Added a '+' icon to the index dropdown.
Removed the logic that removes all options if * is selected
Also if a wildcard is selected, do not remove the options that match it.
If the user tries to remove the last option, add * as the selected index.
Add regex support for index e.g  ind*-8 

# Testing

# Checklist:


static/js/index.js
```diff
@@ -157,90 +142,73 @@ async function initializeIndexAutocomplete() {
...
-                }
             } else {
+                var matcher = new RegExp( typedValue.replace("*",".*") ,"gi" );
+                function hasMatchingString(arr, regex) {
+                    return arr.some((element) => regex.test(element.toLowerCase()));
+                }
+                const matchesIndex =  hasMatchingString(indexValues, matcher);
+                // If the typed value matches any index, add the option to index list
+                if (matchesIndex) {
+                    indexValues.unshift(typedValue);
```
sonamgupta21, line 172: We can remove line 172, so the wildcard option will not appear in the options after it is selected.
image

static/js/index.js
```diff
@@ -157,90 +142,73 @@ async function initializeIndexAutocomplete() {
...
-            return !option.startsWith(prefix);
-        });
-        
-        // Add back only the non-wildcard options that match the prefix
-        const matchingOptions = originalIndexValues.filter(function(option) {
-            return option.startsWith(prefix) && !selectedSearchIndex.includes(option);
-        });
-        
-        indexValues.push(...matchingOptions);
+    if (!removedValue.endsWith('*')) {
```
sonamgupta21, line 207: Here, we can make it as 
`if (!removedValue.includes('*'))`  , so that when we remove the wildcard, it does not get added to the options."

sonamgupta21 (approved):

PR #1092, opened 2024-06-10
feat: Added Metrics e2e tests

Macbeth98:
# Description
- Created the setup for Metrics E2E tests.
- Added some tests on Metrics Query including basic Regex queries on Metric name and group by metric name queries.
- Fixed a bug with Regex Metric search.
- Moved the promQl parser function for PromQL API Handlers from `ConvertPqlToMetricsQuery` to the new `ConvertPromQLToMetricsQuery`.

# Checklist:


pkg/e2etests/metrics_e2e_test.go
```diff
@@ -0,0 +1,1186 @@
...
+*/
+
+type timeSeries struct {
+	Metric    string            `json:"metric"`
+	Tags      map[string]string `json:"tags"`
+	Timestamp uint32            `json:"timestamp"`
+	Value     int               `json:"value"`
+}
+
+const dataStartTimestamp uint32 = 1718052279
```
AndrewHess, line 72: Is this arbitrary, or should it be a recent unix epoch? If it is arbitrary I'd recommend saying so in a comment and optionally setting it to a different value (e.g., 42, or 1 billion) to make it more obvious

pkg/e2etests/metrics_e2e_test.go
```diff
@@ -0,0 +1,1186 @@
...
+
+	/*
+		Expected Results:
+		At T1: testmetric0: 10, testmetric1: 20, testmetric2: 30
+		At T2: testmetric0: 40, testmetric1: 80, testmetric2: 120
+		At T3: testmetric0: 50, testmetric1: 90, testmetric2: 130
+		At T4: testmetric0: 60, testmetric1: 100, testmetric2: 140
+		At T5: testmetric0: 70, testmetric1: 110, testmetric2: 150
+
+		The time diff between T1 and T2 is less than 6 seconds, so the values at T1 and T2 will be aggregated to the same bucket.
```
AndrewHess, line 564: Why is 6 seconds important? Where is that coming from?

pkg/e2etests/metrics_e2e_test.go
```diff
@@ -0,0 +1,1186 @@
...
+
+	/*
+		Expected Results:
+		At T1: testmetric0: 10, testmetric1: 20
+		At T2: testmetric0: 40, testmetric1: 80
+		At T3: testmetric0: 50, testmetric1: 90
+		At T4: testmetric0: 60, testmetric1: 100
+		At T5: testmetric0: 70, testmetric1: 110
+
+		The time diff between T1 and T2 is less than 5 seconds, so the values at T1 and T2 will be aggregated to the same bucket.
```
AndrewHess, line 632: Why is it 5 seconds in this test but 6 seconds in the above test? (I'm just missing where these numbers are coming from)

pkg/e2etests/metrics_e2e_test.go
AndrewHess, line 1: Nice usage comment near the top of this file

pkg/integrations/prometheus/promql/metricsSearchHandler.go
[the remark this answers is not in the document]
Macbeth98: Yes, I will update this for `ProcessGetAllMetricTagsRequest`. But I will not update it for the other one, as we are still verifying and testing it. So it would be better to keep it so that we can compare it. 
Once the testing is done fully done, then I will remove it.

This `ProcessUiMetricsSearchRequest` is the old Metrics Query Handler and is currently not being used. But I just kept it, so we can quickly compare the results by calling corresponding api with the same query.

pkg/segment/query/queryrefresh.go
```diff
@@ -72,6 +72,9 @@ func initMetricsMetaRefresh() {
 }
 
 func PopulateMetricsMetadataForTheFile_TestOnly(mFileName string) error {
+	metaFileLastModifiedLock.Lock()
+	metaFileLastModified[mFileName] = 0
+	metaFileLastModifiedLock.Unlock()
 	return populateMetricsMetadata(mFileName)
 }
```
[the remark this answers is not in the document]
Macbeth98: Yes, it was not working. 
So if I run each test individually there are no issues and I don't need this reset here.
But if I run all the tests in this file/package at the same time.. if I run it through `make test/make pr`, then this is causing an error, as this function is not loading the newly ingested metrics. So to make sure that this function will load the metrics, I need to reset the time.

pkg/e2etests/metrics_e2e_test.go
Macbeth98: Yes, it is kind of arbitrary. At first, I kept it dynamic, setting it to `24h` before, but that sometimes is failing making the test to be flaky. 
You might have seen some comments where I have that at T1 and T2, the time diff is less than 5 seconds, so they will aggregated into the same bucket. And sometimes that downsampling is not happening, when I set it to dynamic.

As our downsampling formula = `downsampledTime = (ts / s.dsSeconds) * s.dsSeconds` , So sometimes they are not ending up under the same bucket and causing my test to be failed. 

So then I changed it to be static. 
And actually this epoch timestamp does not matter, it is just so that the data generation will use this as start timestamp and the queries will also use this as start timestamp. So as long as both of these use the same timestamp it should be fine, given that T1 and T2 will end up in the same bucket. 

I will write some comment, explaining this.

pkg/e2etests/metrics_e2e_test.go
Macbeth98: Actually I need to update this comment, I will update these comments.

pkg/e2etests/metrics_e2e_test.go
Macbeth98: These numbers are from data generation part above from the function `GetTestMetricsData`. I will update these comments. 

And I have written these comments as to explain myself and others how I have arrived at these expected results.

pkg/e2etests/metrics_e2e_test.go
AndrewHess: Ok, and where is the downsampling interval getting set? is it by `ConvertPromQLToMetricsQuery()`? What does it get set to for your tests? You want `dataStartTimestamp` to be set so that items at `+1` and `+2` from that get put into the same bucket, right? This info might be useful in the comment you've added

Macbeth98: Yes, it is being set in `ConvertPromQLToMetricsQuery`. And the time range I am setting when making the query is 1 hour. As the max data ingested is at `startTimestamp + 3600`. 

And Yes, I have added this information in the comment above `dataStartTimestamp` variable.

AndrewHess (approved):

PR #1106, opened 2024-06-12
feat: Formula Operations on timeseries API

Macbeth98:
# Description
- For the Metrics API that uses `Timeseries` API endpoint, the operation between two vectors should happen regardless of  whether there is a match or not given that it is a one to one or one to Many.
- If the Query given has `VectorMatchingOps` such as on and ignoring, then that will take precedence.

# Testing
- Tested by writing the test cases and updating the test cases.

# Checklist:


pkg/segment/segexecution.go
```diff
@@ -156,10 +156,10 @@ func processQueryArithmeticNodeOp(queryOp *structs.QueryArithmetic, resMap map[u
 
 	*operationCounter++
 
-	return HelperQueryArithmeticAndLogical(queryOp, resMap)
+	return HelperQueryArithmeticAndLogical(queryOp, resMap, timeseriesApi)
 }
 
-func HelperQueryArithmeticAndLogical(queryOp *structs.QueryArithmetic, resMap map[uint64]*mresults.MetricsResult) (map[string]map[uint32]float64, error) {
+func HelperQueryArithmeticAndLogical(queryOp *structs.QueryArithmetic, resMap map[uint64]*mresults.MetricsResult, timeseriesApi bool) (map[string]map[uint32]float64, error) {
```
AndrewHess, line 162: nit: here and elsewhere, what do you think about renaming `timeseriesApi` to `labelsMustMatch` or similar?

pkg/segment/segexecution.go
```diff
@@ -208,6 +208,8 @@ func HelperQueryArithmeticAndLogical(queryOp *structs.QueryArithmetic, resMap ma
 				resultRHS = resMap[queryOp.LHS]
 			}
 
+			// If the query has vector matching operation, then we need to perform operations on that matching label set.
+
```
AndrewHess, line 212: Just to clarify: is this comment correct even if `timeseriesApi == true`? Or is it not possible for both `hasVectorMatchingOp` and `timeseriesApi` to be `true` at the same time?

AndrewHess (commented):
Looks pretty good

pkg/segment/segexecution.go
Macbeth98: Yes, so for example, when I say `timeseriesApi`, it is the metric query request that comes through our UI. 
Right now we are not allowing to modify the query and users can only make single queries (a, b, c..) and then use formula such as `(a + b + c)` to do some operations. 

but if we let the users to edit these queries and they an write proper promql queries where a single query (a) can include multiple queries.. and in that query, the user can have `on, ignoring` matching label set. 

So in that case if `hasVectorMatchingOp` is true, that should always take priority.

pkg/segment/segexecution.go
Macbeth98: Hmm, yes I think I will refactor it to that. Seems generalized and makes more sense.

Macbeth98: I will rename it to `opLabelsMayNotMatch`, as for timeseries API, the operation labels does not need to match.

AndrewHess: I think that's still going to be kind of confusing, because sometimes "may not" means "must not" and in other contexts "may not" means "isn't required"

Macbeth98: My context here is convey that the operation labels does not need to match to perform the operation.

AndrewHess (approved):

PR #1117, opened 2024-06-12
feat: Added support for Metrics regex query verification

Macbeth98:
# Description
- Updated `otsbdquery.go` with the support to verify regex queries.
- Updated `metrics_test.csv` with some of the regex queries.

# Testing
- Tested by running ingestion command in the circle ci config.
- And run the `metric_test.csv` query command.

# Checklist:

### DIFF
--- a/pkg/segment/query/metricsquery.go
+++ b/pkg/segment/query/metricsquery.go
@@ -385,16 +385,15 @@ func applyMetricsOperatorOnSegments(mQuery *structs.MetricsQuery, allSearchReqes
 }
 
 func getRegexMatchedMetricNames(mSegSearchReq *structs.MetricsSearchRequest, regexPattern string, operator utils.TagOperator) ([]string, error) {
-	mNamesMap, err := series.GetAllMetricNames(mSegSearchReq.MetricsKeyBaseDir)
-	if err != nil {
-		return nil, err
-	}
+	var mNamesMap map[string]bool
+	var err error
 
 	if len(mSegSearchReq.UnrotatedMetricNames) > 0 {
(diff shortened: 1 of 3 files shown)


pkg/segment/query/metricsquery.go
```diff
@@ -385,16 +385,15 @@ func applyMetricsOperatorOnSegments(mQuery *structs.MetricsQuery, allSearchReqes
...
-			if _, ok := mNamesMap[mName]; !ok {
-				mNamesMap[mName] = true
-			}
+		mNamesMap = mSegSearchReq.UnrotatedMetricNames
+	} else {
+		mNamesMap, err = series.GetAllMetricNames(mSegSearchReq.MetricsKeyBaseDir)
+		if err != nil {
+			return nil, err
 		}
 	}
```
AndrewHess, line 398: I don't quite understand this change. It looks like the old version would combine all metric names from rotated and unrotated data, but the new version will only get the metric names from one source (unrotated or rotated) without combining them. Why is that what we want to do?

Macbeth98: Yeah a SearchRequest is basically formed from getting the Rotated and Unrotated Segments.
So a single Search Request can only be either Rotated or Unrotated search request.  

For Unrotated search request when I call `GetAllMetricNames()`, since the segment is not rotated yet, there will not be any file with ext `.mnm` (file where we store all metric names for this segment) will not created. And this will throw file not found error.

Since this is a definitive error thrown case, i decided to change the code to call the function only for the rotated segements.

And the before code, for unrotated segments it will throw error causing it exit the function, without checking for unrotated metric names that is extracted from the search request.

AndrewHess (approved):

PR #1126, opened 2024-06-14
feat: Added parsing logic for SPL command: spath

Macbeth98:
# Description
- Added parsing for SPL command: `spath` following the command reference: https://docs.splunk.com/Documentation/SplunkCloud/9.1.2312/SearchReference/Spath
- Added test cases with various `spath` query patterns.

# Testing
- Added test cases, testing the `spath` queries.

# Checklist:


pkg/ast/spl/tests/splParser_test.go
```diff
@@ -7025,3 +7027,291 @@ func Test_MakeMVWithFieldAndDelimiterPlusAllowEmpty(t *testing.T) {
...
+	assert.NotNil(t, aggregator.OutputTransforms.LetColumns.ValueColRequest)
+	assert.Equal(t, structs.ValueExprMode(structs.VEMStringExpr), aggregator.OutputTransforms.LetColumns.ValueColRequest.ValueExprMode)
+	assert.NotNil(t, aggregator.OutputTransforms.LetColumns.ValueColRequest.StringExpr)
+	assert.Equal(t, structs.StringExprMode(structs.SEMTextExpr), aggregator.OutputTransforms.LetColumns.ValueColRequest.StringExpr.StringExprMode)
+	assert.NotNil(t, aggregator.OutputTransforms.LetColumns.ValueColRequest.StringExpr.TextExpr)
+	assert.Equal(t, "spath", aggregator.OutputTransforms.LetColumns.ValueColRequest.StringExpr.TextExpr.Op)
+	assert.NotNil(t, aggregator.OutputTransforms.LetColumns.ValueColRequest.StringExpr.TextExpr.SPathExpr)
+	assert.Equal(t, "_raw", aggregator.OutputTransforms.LetColumns.ValueColRequest.StringExpr.TextExpr.SPathExpr.InputColName)
+	assert.Equal(t, "myfield", aggregator.OutputTransforms.LetColumns.ValueColRequest.StringExpr.TextExpr.SPathExpr.OutputColName)
+	assert.Equal(t, "vendorProductSet.product{@year}.locDesc", aggregator.OutputTransforms.LetColumns.ValueColRequest.StringExpr.TextExpr.SPathExpr.Path)
```
AndrewHess, line 7261: We might want to do more parsing on the path part like `vendorProductSet.product{@year}.locDesc`. Or we could do the extra parsing in Go when we implement the logic for spath. What do you think?

For example, for implementing the logic, it will probably be more useful to have something like:
```
[]myStruct{
    myStruct{field: "vendorProductSet"},
    myStruct{field: "product", section: "year"},
    myStruct{field: "locDesc"},
}
```

pkg/ast/spl/spl.peg
```diff
@@ -3528,9 +3594,73 @@ SetSvOption <- "setsv" EQUAL boolVal:Boolean {
...
+}
+PathField <- ("path" EQUAL SPathFieldString / SPathFieldString ) {
+     pathField := string(c.text)
+    if strings.HasPrefix(pathField, "path") {
+        pathField = strings.TrimSpace(strings.TrimPrefix(pathField, "path"))
+        pathField = strings.TrimSpace(strings.TrimPrefix(pathField, "="))
+        if pathField != "" && pathField[0] == '"' {
+            pathField = removeQuotes(pathField)
+        }
+    }
```
AndrewHess, line 3655: I think instead of
```
    if strings.HasPrefix(pathField, "path") {
        pathField = strings.TrimSpace(strings.TrimPrefix(pathField, "path"))
        pathField = strings.TrimSpace(strings.TrimPrefix(pathField, "="))
```
You want
```
    if strings.HasPrefix(pathField, "path=") {
        pathField = strings.TrimSpace(strings.TrimPrefix(pathField, "path="))
```
That way if you parse something where `path=` was not given and the path starts with `"path"`, you don't get a bug. For example, `spath pathApple` should give the path as `pathApple` rather than just `Apple`

pkg/ast/spl/tests/splParser_test.go
Macbeth98: Yes, I am thinking of parsing this at the implementation side, as it has multiple formats:
1. Where we need to fetch every value that is matching the given field in the Path Value (Instead of Path, they can give a field Value: example: `username` and we should fetch values for every field/nested field with name = `username`.
2. And there is one way of handling for array and nest array.
3. One way for XML fields.
4. And the normal Path.

So I think it is better to do this parsing at the implementation point.

pkg/ast/spl/spl.peg
Macbeth98: yeah I did not think of this case. But also, I am thinking this will not handle the case where path is given like this:
`path = pathApple`. 
if there is space between path and `=`

Macbeth98: I could probably do this:

```go
        str = `path=  "pathval here"`
	str = strings.TrimSpace(strings.TrimPrefix(str, "path="))
	str = strings.TrimSpace(strings.TrimPrefix(str, "path ="))
	fmt.Println(str)
```

check for both the cases. Check the go playground: https://go.dev/play/p/iz8TfRN3LGi

pkg/ast/spl/tests/splParser_test.go
```diff
@@ -7025,3 +7027,324 @@ func Test_MakeMVWithFieldAndDelimiterPlusAllowEmpty(t *testing.T) {
...
+	assert.Equal(t, "spath", aggregator.OutputTransforms.LetColumns.ValueColRequest.StringExpr.TextExpr.Op)
+	assert.NotNil(t, aggregator.OutputTransforms.LetColumns.ValueColRequest.StringExpr.TextExpr.SPathExpr)
+	assert.Equal(t, "_raw", aggregator.OutputTransforms.LetColumns.ValueColRequest.StringExpr.TextExpr.SPathExpr.InputColName)
+	assert.Equal(t, "entities.hashtags", aggregator.OutputTransforms.LetColumns.ValueColRequest.StringExpr.TextExpr.SPathExpr.Path)
+	assert.False(t, aggregator.OutputTransforms.LetColumns.ValueColRequest.StringExpr.TextExpr.SPathExpr.IsPathFieldName)
+	assert.Equal(t, "hashtags", aggregator.OutputTransforms.LetColumns.NewColName)
+}
+
+func Test_Spath_Extract_PathTest(t *testing.T) {
+	queries := []string{
```
Macbeth98, line 7320: I have added this test case to verify the parsing of path argument.

pkg/ast/spl/spl.peg
AndrewHess: Is that a valid query? I would assume `spath path = myPath` is invalid, but do you see that in the documentation/examples/etc.?

Macbeth98: Hmm, as far as I see in the documentation it is mentioned as `path=` only. But I just thought users might type with space. I am not sure whether Splunk will accept this or throws an error.

But we defined `=` as this: `EQUAL <- EMPTY_OR_SPACE "=" EMPTY_OR_SPACE` 

And In many cases we use it like including for `eval`. so I assume even here this: `spath path = myPath` might be a valid case.

This is how eval syntax is mentioned.

```
eval
<field>=<expression> ["," <field>=<expression> ].
```

For Eval command in the splunk doc, they have given both kind of examples:

```
1. eval error = if(status == 200, "OK", "Problem")
2. eval velocity=distance/time
```
But for spath it was always `path=`

So should I change this to accept only `path=` format?

AndrewHess: Hmm ok then let's keep what you have so we can handle it with spaces

AndrewHess: Or actually, do something similar to what you suggested so you also handle the `pathApple` case

Macbeth98: Yes all the cases are being handled. All the below cases are working fine.

```
queries := []string{
		`* | spath pathApple`,
		`* | spath path= pathApple`,
		`* | spath path = pathApple`,
		`* | spath path =pathApple`,
	}
```

AndrewHess (approved):

PR #1127, opened 2024-06-14
feat: Implementation for SPL command: makemv

Macbeth98:
# Description
- Added implementation for SPL command: `makemv`
The below queries will work now.

```
1. * | stats count(*) by city, country | makemv setsv=true country
2. * | makemv delim="a" country | fields country
```

# Testing
- Will be writing tests now.

# Checklist:


pkg/segment/structs/segstructs.go
```diff
@@ -819,3 +821,113 @@ func CheckUnsupportedFunctions(post *QueryAggregators) error {
...
+// GetBucketValueForGivenField returns the value of the bucket for the given field name.
+// It checks for the field in the statistic results and group by keys.
+// If the field is not found, it returns nil.
+// The first return value is the value of the field.
+// The second return value is the index of the field in the group by keys.
+// The third return value is a boolean indicating if the field was found in the Statistic results.
+// For Statistic results, the index is -1. and Bool is true.
+// For GroupBy keys, the index is the index of the field in the group by keys; given that the Bucket key is a List Type. and Bool is false.
+// Otherwise, the index is -1 and Bool is false.
+func (br *BucketResult) GetBucketValueForGivenField(fieldName string) (interface{}, int, bool) {
```
Macbeth98, line 834: @AndrewHess I would like to get your thoughts on this method and the below Set Method.

Macbeth98: Based on your feedback I will write the test cases for both of the functions and then we can merge this.

AndrewHess: Those functions look good to me

pkg/segment/aggregations/segaggs.go
```diff
@@ -1482,6 +1487,124 @@ func performSortColRequestOnMeasureResults(nodeResult *structs.NodeResult, letCo
...
+		default:
+			return fmt.Errorf("performMultiValueColRequestWithoutGroupby: unknown command %s", mvColReq.Command)
+		}
+	}
+
+	return nil
+}
+
+func performMultiValueColRequestOnHistogram(nodeResult *structs.NodeResult, letColReq *structs.LetColumnsRequest) error {
+	// mvColReq := letColReq.MultiValueColRequest
```
AndrewHess, line 1534: nit: delete this comment or uncomment it and use the variable

pkg/segment/aggregations/segaggs.go
```diff
@@ -1482,6 +1487,124 @@ func performSortColRequestOnMeasureResults(nodeResult *structs.NodeResult, letCo
...
+}
+
+func performMakeMV(strVal string, mvColReq *structs.MultiValueColLetRequest) interface{} {
+	if strVal == "" {
+		return ""
+	}
+
+	var values []string
+	if mvColReq.IsRegex {
+		re := regexp.MustCompile(mvColReq.DelimiterString)
```
AndrewHess, line 1577: I think `DelimiterString` is provided by the user. If it is, you shouldn't use `MustCompile()` because bad user input will cause a panic. Instead use `Compile()` and check for an error

Macbeth98: In the parser, while parsing the regex, I am checking with `regexp.Complie` and it will throw an error if regex is not valid. So I believe this should be fine right? As we are sure it will only come here, if the regex is valid.

image

pkg/segment/structs/segstructs.go
Macbeth98: Got it, thank you!

AndrewHess (approved):

AndrewHess (approved):

PR #1133, opened 2024-06-17
Add SPL ipmask, replace and isnotnull commands implementation

venkateshamatam:
# Description
Add SPL ipmask, replace and isnotnull commands implementation

# Testing
- Will add test cases
# Checklist:


pkg/segment/structs/evaluationstructs.go
```diff
@@ -366,7 +366,19 @@ func (self *BoolExpr) Evaluate(fieldToValue map[string]utils.CValueEnclosure) (b
...
+		if !exists {
+			return false, fmt.Errorf("BoolExpr.Evaluate: Field '%s' not found in data", fields[0])
+		}
+		if value.Dtype != utils.SS_DT_BACKFILL {
+			return true, nil
+		}
+		return false, nil
 	case "isnum":
 		return false, fmt.Errorf("BoolExpr.Evaluate: does support using this operator: %v", self.ValueOp)
 	}
```
AndrewHess, line 384: I think you want to move the `"searchmatch"` case down so the `fallthrough` will hit the `"isnum"` case. Or you can have the `isnum` branch match multiple cases.

Also, you should add a `default` branch here that gives a slightly different error message, like `unknown operator: %v`

pkg/segment/structs/evaluationstructs.go
```diff
@@ -1338,6 +1350,42 @@ func (self *NumericExpr) Evaluate(fieldToValue map[string]utils.CValueEnclosure)
...
+		}
+		replacementStr, err := self.ValueList[1].Evaluate(fieldToValue)
+		if err != nil {
+			return "", fmt.Errorf("TextExpr.EvaluateText: cannot evaluate replacement as a string: %v", err)
+		}
+
+		baseString, exists := fieldToValue[self.Val.NumericExpr.Value]
+		if !exists {
+			return "", fmt.Errorf("TextExpr.EvaluateText: field '%s' not found in data", self.Param.RawString)
+		}
```
AndrewHess, line 1383: Is this correct? You're looking up `self.Val.NumericExpr.Value` but if it doesn't exist in the map you're printing `self.Param.RawString`

pkg/segment/structs/evaluationstructs.go
```diff
@@ -1338,6 +1350,42 @@ func (self *NumericExpr) Evaluate(fieldToValue map[string]utils.CValueEnclosure)
...
+		regex, err := regexp.Compile(regexStr)
+		if err != nil {
+			return "", fmt.Errorf("TextExpr.EvaluateText: failed to compile regex: %v", err)
```
AndrewHess, line 1386: nit: in the error message, include the regex you're trying to compile

pkg/segment/structs/evaluationstructs.go
```diff
@@ -1338,6 +1350,42 @@ func (self *NumericExpr) Evaluate(fieldToValue map[string]utils.CValueEnclosure)
...
+		}
+		return regex.ReplaceAllString(baseString.CVal.(string), replacementStr), nil
```
AndrewHess, line 1388: You should probably check this type assertion `baseString.CVal.(string)` so we don't panic if it errors

pkg/segment/structs/evaluationstructs.go
```diff
@@ -363,12 +363,22 @@ func (self *SortExpr) ReleaseProcessedSegmentsLock() {
...
+		if !exists {
+			return false, fmt.Errorf("BoolExpr.Evaluate: Field '%s' not found in data", fields[0])
+		}
+		if value.Dtype != utils.SS_DT_BACKFILL {
+			return true, nil
+		}
+		return false, nil
+	case "searchmatch", "isnum":
+		return false, fmt.Errorf("BoolExpr.Evaluate: does not support using this operator: %v", self.ValueOp)
 	}
```
AndrewHess, line 382: Add a `default` branch to catch anything we don't expect. And the error on line 381 should clarify that these are valid operators that we don't yet support, while the error in the `default` branch should indicate the operator is invalid

AndrewHess:
After addressing my comment, is this PR ready to merge, or do you still want it as Draft?

pkg/segment/structs/evaluationstructs.go
venkateshamatam: Adding a default check here would make the code below unreachable since there are more commands for both `boolOp` and `valueOp` that we support that are not part of this switch statement. I can refactor the method and push it to the same PR so we can also have a default case.

venkateshamatam:
I will refactor the method so we can also have a default case and resolve the two conflicts and then it will be ready to merge.

pkg/segment/structs/evaluationstructs.go
AndrewHess: You're right, I didn't realize that. Sure, as long as we have some of detecting when an invalid operation is specified it should be good

venkateshamatam: I have cleaned up some code and added a switch case statement to handle all implemented commands, and it looks like we are already handling the default case in https://github.com/siglens/siglens/blob/6884e6a330476401fbcec23e45120636d83dd3af/pkg/segment/structs/evaluationstructs.go#L548
https://github.com/siglens/siglens/blob/6884e6a330476401fbcec23e45120636d83dd3af/pkg/segment/structs/evaluationstructs.go#L558  https://github.com/siglens/siglens/blob/6884e6a330476401fbcec23e45120636d83dd3af/pkg/segment/structs/evaluationstructs.go#L568

AndrewHess (approved):

PR #1140, opened 2024-06-17
Add SPL trig/hyperbolic commands

venkateshamatam:
# Description
Add SPL trig/hyperbolic commands

# Testing
`* | eval n=acosh(2)`
`* | eval n=asin(0.5)`
`* | eval n=tan(0.75)`
`* | eval n1=acosh(2), n2=asin(0.5), n3=tan(0.75)`
`* | eval n=tan(1.5708)`
`* | eval n=acosh(2)`
`* | eval n=atan2(1, 1)`
`* | eval n=atan2(0, 0)`
`* | eval result=hypot(0, 0)`
`* | eval result=hypot(3, 4)`
# Checklist:


pkg/segment/structs/evaluationstructs.go
```diff
@@ -1285,6 +1285,60 @@ func (self *NumericExpr) Evaluate(fieldToValue map[string]utils.CValueEnclosure)
...
+		case "tan":
+			// Check for points where cos(x) = 0, which would cause tan to be undefined.
+			// These are points (pi/2 + k*pi) where k is an integer.
+			// To check for this, see if left modulo pi is pi/2 (or very close due to floating point precision).
+			halfPi := math.Pi / 2
+			mod := math.Mod(left, math.Pi)
+			if math.Abs(mod-halfPi) < 0.0000001 || math.Abs(mod+halfPi) < 0.0000001 {
+				return -1, fmt.Errorf("NumericExpr.Evaluate: tan is undefined at pi/2 + k*pi, got: %v", left)
+			}
+			return math.Tan(left), nil
```
AndrewHess, line 1329: I don't think the error handling will work for some negative values. For example `-3pi/2`.

From https://pkg.go.dev/math#Mod
> Mod returns the floating-point remainder of x/y. The magnitude of the result is less than y and its sign agrees with that of x.

So your `mod` would be `-pi/2` and `mod-halfPi` would be `-pi` instead of `0`

venkateshamatam: In cases where we have a negative value like `-3pi/2`, `math.Abs(mod+halfPi)` will be very close to zero, so the condition will be true, and the function will correctly return an error. Am I missing something here?

AndrewHess: `math.Mod` sometimes returns negative numbers. https://go.dev/play/p/X2ZNWbyOemt

AndrewHess (dismissed):
Good to merge after the conflicts are resolved

AndrewHess (approved):

PR #1141, opened 2024-06-17
Spl funcs 2

vasukalariya:
# Description
Summarize the change.
Case and Coalesce functions. For now, we are skipping if the condition in case if it gives an error. We are assuming it as false.
Important change is DTypeEnclosure was earlier set to SS_DT_STRING in (getRecordFieldValues), changed that to SS_DT_BACKFILL to indicate NULL value. With this, isnull should also work properly as expected.

# Testing

Create data
```
Insert data using vector insert
​​curl -X POST "http://localhost:8081/elastic/_bulk" \
-H 'Content-Type: application/json' \
-d '{ "index" : { "_index" : "null_test_data" } }
{ "name" : "john", "age":40}'

Insert data { "name" : null, "age":null}
Insert data { "name" : “mary”, "age":20}
Insert data { "name" : “jack”, "age":null}
Insert data { "name" : null, "age":19}
Insert data { "name" : “jill”, "age":30}

```

Tests
```
Case
* | eval n = case(age> 20, "older than 20", name = "mary", "name is mary")
* | eval n = case(age> 20, "older than 20", name = "mary", 50)
city=Boston | stats count AS Count BY http_status | eval n = case(http_status = 200, "OK", Count > 170, "check")
* | eval n = case(http_status = 200, "OK")


Coalesce
* | eval n = coalesce(age, name, "jerry")
* | eval n = coalesce(name, age, "tom")
* | eval n = coalesce(age, name, 25)
city=Boston | stats count AS Count BY http_status | eval n = coalesce(http_status, Count)
* | eval n = coalesce(http_status, address)
```

# Checklist:


pkg/segment/structs/evaluationstructs.go
```diff
@@ -1533,6 +1533,52 @@ func (self *ValueExpr) EvaluateValueExprAsString(fieldToValue map[string]utils.C
...
+			val, ok := fieldToValue[field]
+			// if a field has a NIL value in the expression we do not want that expression to be considered
+			if !ok || val.Dtype == utils.SS_DT_BACKFILL {
+				skip = true
+				break
+			}
+		}
+		if skip {
+			continue
+		}
```
AndrewHess, line 1570: FYI: in Go you can label loops and then `continue` or `break` those loops even while you're inside an inner loop. See https://go.dev/play/p/HkQCdidkgcw for an example

But what you have with the `skip` variable is also fine

pkg/segment/structs/evaluationstructs.go
```diff
@@ -1533,6 +1533,52 @@ func (self *ValueExpr) EvaluateValueExprAsString(fieldToValue map[string]utils.C
 	return str, nil
 }
 
+func handleCaseFunction(self *ConditionExpr, fieldToValue map[string]utils.CValueEnclosure) (string, error) {
+
+	for _, cvPair := range self.ConditionValuePairs {
+		res, err := cvPair.Condition.Evaluate(fieldToValue)
+		if err != nil {
+			continue
+		}
```
AndrewHess, line 1542: Do you `continue` on line 1541 because `Evaluate()` may have failed because a required field didn't have a value? I think we'll want to detect that separately, since we do want to `continue` in that case, but other times we might get a different kind of error during `Evaluate()` and in that case we would want to log and return the error

pkg/segment/structs/evaluationstructs.go
```diff
@@ -502,6 +502,18 @@ func (self *BoolExpr) Evaluate(fieldToValue map[string]utils.CValueEnclosure) (b
...
+						return false, fmt.Errorf("BoolExpr.Evaluate: Field '%s' not found in data", leftField[0])
+					}
+					// Check if the value's Dtype is SS_DT_BACKFILL
+					if value.Dtype == utils.SS_DT_BACKFILL {
+						return false, nil
+					}
+				}
 				return false, fmt.Errorf("BoolExpr.Evaluate: left cannot be evaluated to a string or float")
 			}
 			if errRightStr != nil && errRightFloat != nil {
```
AndrewHess, line 519: I don't think this change is going to handle all the cases you want it to handle. Some notes:
1. Why did you add thisi logic for the `if` at 504 but not also add similar logic for the `if` at 519?
2. I think your goal is: "when I try to evaluate a boolean expression but some of the fields don't exist or have `nil` values, that expression should evaluate to `false`". Is that your goal? Then I think you'll still miss some types of boolean expression (e.g., when the `ValueOp` is `isbool, in, like, etc.`).
3. It might make more sense to do a check similar to this (but iterating over all the fields in `GetFields()` for the boolean expression, and if any field is not in `fieldToValue`, then return a bool indicating that; so then the return type of this function would be `(bool, bool, error)`. What do you think of that?

vasukalariya: There are some functions like is null etc who operate on NULL values. As per our discussion, I have created helper method to get the null fields. Now, the calling methods have to make the decision on what they want to do if the evaluation fails because of a NULL value.

pkg/segment/structs/evaluationstructs.go
```diff
@@ -357,6 +357,57 @@ func (self *SortExpr) ReleaseProcessedSegmentsLock() {
...
+func findNullFields(fields []string, fieldToValue map[string]utils.CValueEnclosure) ([]string, error) {
+	nullFields := []string{}
+	for _, field := range fields {
+		val, exists := fieldToValue[field]
+		if !exists {
+			return []string{}, fmt.Errorf("findNullFields: Expression has a field for which value is not present")
+		}
+		if val.Dtype == utils.SS_DT_BACKFILL {
+			nullFields = append(nullFields, field)
+		}
```
AndrewHess, line 369: Why do you return an error on line 365 instead of just adding the field to the list of `nullFields`?

vasukalariya: Actually !exists is different from NULL field. !exists mean that field was present in the query but we did not fetched the value for it. Even if it is NULL it would still show up as exists. In most of the methods like isnull etc. We are using this check and returning error cause the fieldToValue should have this entry but it does not have it right now so it is an error.

AndrewHess (approved):

PR #1142, opened 2024-06-18
Add SPL validate command

venkateshamatam:
# Description
Add SPL validate command

# Testing

# Checklist:


pkg/segment/structs/evaluationstructs.go
```diff
@@ -1213,6 +1213,28 @@ func handleNoArgFunction(op string) (float64, error) {
...
 }
 
+func handleComparisonAndConditionalFunctions(self *ConditionExpr, fieldToValue map[string]utils.CValueEnclosure, functionName string) (string, error) {
+	switch functionName {
+	case "validate":
+		for _, cvPair := range self.ConditionValuePairs {
+			res, err := cvPair.Condition.Evaluate(fieldToValue)
+			if err != nil {
+				continue
+			}
```
AndrewHess, line 1223: Why do you `continue` on line 1222?

venkateshamatam: If we have a null value in a field then we want to continue to check if any of the subsequent conditions evaluate to false(returning an error in this case would return null) and we want to show the value that corresponds to the first condition that evaluates to false. But `continue` skips any other possible errors returned by the `Evaluate` function so I have pushed a commit to return an err from here but we might have to handle the null case in the `Evaluate` method itself.

AndrewHess: Ok yeah I think we'll want to handle null cases differently than actual errors, as you said. You should check how Splunk handles cases where a condition can't be evaluated because one of the fields is null. My guess would be in that case we'd treat it as false and so return the value part of that pair. But if we do get an actual error during evaluation we should return the error.

AndrewHess (approved):

PR #1146, opened 2024-06-18
Fix: Proper Rendering of Line Breaks in Wrap View

sonamgupta21:
# Description

1. In wrap line view, \n now correctly moves to the next line.
2. Fixed an issue where a single row result did not display correctly when switching from wrap to single view.
image
image

Previously - 
image

### DIFF
--- a/static/js/event-handlers.js
+++ b/static/js/event-handlers.js
@@ -589,6 +589,20 @@ function logOptionSingleHandler() {
         if (colDef.field === "logs"){
             colDef.cellStyle = null;
             colDef.autoHeight = null;
+            colDef.cellRenderer = function(params) {
+                const data = params.data || {};
+                let logString = '';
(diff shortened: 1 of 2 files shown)


static/js/event-handlers.js
```diff
@@ -589,6 +589,20 @@ function logOptionSingleHandler() {
         if (colDef.field === "logs"){
             colDef.cellStyle = null;
             colDef.autoHeight = null;
+            colDef.cellRenderer = function(params) {
+                const data = params.data || {};
+                let logString = '';
+                let counter = 0;
```
sunitakawane, line 595: can this be boolean flag instead of counter?

sonamgupta21: Done.

sunitakawane (approved):

PR #1151, opened 2024-06-19
Add Test contact point API

venkateshamatam:
# Description
Add Test contact point API

# Testing
- Tested the below cases: 

Correct slack token and channel ID
Correct slack token and wrong channel ID
Wrong slack token and correct channel ID
Missing slack token and missing channel ID when type is `slack`
Missing slack token when type is `slack`
Missing channel ID when type is `slack`
Correct webhook URL and verified the request details come through on webhook endpoint.
Missing webhook URL when type is `webhook`
Incorrect webhook URL when type is `webhook`
Missing type
Invalid/unsupported type.

# Checklist:

### DIFF
--- a/pkg/alerts/alertsHandler/alertsHandler.go
+++ b/pkg/alerts/alertsHandler/alertsHandler.go
@@ -18,11 +18,9 @@
 package alertsHandler
 
 import (
-	"bytes"
 	"encoding/json"
 	"errors"
 	"fmt"
-	"io"
 	"net/http"
 	"time"
 
@@ -184,7 +182,7 @@ func ProcessSilenceAlertRequest(ctx *fasthttp.RequestCtx) {
 func ProcessTestContactPointRequest(ctx *fasthttp.RequestCtx) {
 	var testContactRequest TestContactPointRequest
 	if err := json.Unmarshal(ctx.PostBody(), &testContactRequest); err != nil {
-		utils.SendError(ctx, "Failed to unmarshal json", "ProcessTestContactPointRequest: Request Body: "+string(ctx.PostBody()), err)
+		utils.SendError(ctx, "Failed to unmarshal json", "Request Body: "+string(ctx.PostBody()), err)
 		return
 	}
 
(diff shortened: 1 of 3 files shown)


pkg/alerts/alertsHandler/alertsHandler.go
```diff
@@ -173,6 +181,115 @@ func ProcessSilenceAlertRequest(ctx *fasthttp.RequestCtx) {
 	utils.WriteJsonResponse(ctx, responseBody)
 }
 
+func ProcessTestContactPointRequest(ctx *fasthttp.RequestCtx) {
+	var testContactRequest TestContactPointRequest
+	if err := json.Unmarshal(ctx.PostBody(), &testContactRequest); err != nil {
```
nkunal, line 186: you should check if the PostBody is not nil/empty, else it might panic (verify via postman what happens)

venkateshamatam: if the postbody is nil, `json.Unmarshal` returns an error and the API returns 

```
{
    "error": "Failed to unmarshal json"
}
```
Just verified this on postman

pkg/alerts/alertsHandler/alertsHandler.go
```diff
@@ -173,6 +181,115 @@ func ProcessSilenceAlertRequest(ctx *fasthttp.RequestCtx) {
...
+		utils.SendError(ctx, "Failed to unmarshal json", "ProcessTestContactPointRequest: Request Body: "+string(ctx.PostBody()), err)
```
AndrewHess, line 187: nit: you don't need `ProcessTestContactPointRequest:` (but keep the rest of that string). This `SendError()` function automatically adds the function name

Same comment many other places you're calling `SendError()`

Just for your own benefit, you might want to make `utils.SendError()` execute once and then look at the log file for what the output looks like

pkg/alerts/alertsHandler/alertsHandler.go
```diff
@@ -173,6 +181,115 @@ func ProcessSilenceAlertRequest(ctx *fasthttp.RequestCtx) {
...
+		return
+	}
+
+	switch testContactRequest.Type {
+	case "slack":
+		channelID, ok := testContactRequest.Settings["channel_id"].(string)
+		if !ok {
+			utils.SendError(ctx, "channel_id is required but is missing", "ProcessTestContactPointRequest: Slack Channel ID is missing. Request Body: "+string(ctx.PostBody()), nil)
```
AndrewHess, line 195: nit: In `SendError()` you also don't have to repeat `"Slack Channel ID is missing"` in the third parameter because you've included it in the second parameter as `"channel_id is required but is missing"`. You can look at the definition of `SendError` in httpserverutils.go

pkg/alerts/alertsHandler/alertsHandler.go
```diff
@@ -173,6 +181,115 @@ func ProcessSilenceAlertRequest(ctx *fasthttp.RequestCtx) {
...
+	}
+
+	var result map[string]interface{}
+	err = json.Unmarshal(body, &result)
+	if err != nil {
+		log.Errorf("sendTestMessageToSlack: failed to unmarshal response body. Body: %v err:%v", body, err)
+		return err
+	}
+
+	if !result["ok"].(bool) {
```
AndrewHess, line 271: Check the type assertion so we don't panic if it fails

pkg/alerts/alertsHandler/alertsHandler.go
```diff
@@ -173,6 +181,115 @@ func ProcessSilenceAlertRequest(ctx *fasthttp.RequestCtx) {
...
+	default:
+		utils.SendError(ctx, "Invalid type", "ProcessTestContactPointRequest: Received unsupported test contact request type Request Body:"+string(ctx.PostBody()), nil)
+		return
+	}
+
+	ctx.SetStatusCode(fasthttp.StatusOK)
+	utils.WriteJsonResponse(ctx, map[string]interface{}{"message": "Successfully verified contact point"})
+}
+
+func sendTestMessageToSlack(channelID, slackToken string) error {
```
Macbeth98, line 227: Also, I think it is better if we use the existing `sendSlack` function or the `slack` pkg, instead of sending this through this http post message, as that pkg is what we are using to actually send the Alert messages.

You could probably use the `sendSlack` method or mimic something similar but use the `slack` pkg. Check these lines of code: https://github.com/siglens/siglens/blob/develop/pkg/alerts/alertsHandler/notificationHandler.go#L195-L219

@AndrewHess @nkunal Thoughts on this?

AndrewHess: I agree. I think you can just call the `sendSlack` function. But if there's a reason not to, then you could use the `slack` package like Mani said and it would probably help clean up the logic

nkunal (approved):

PR #1161, opened 2024-06-20
Feat: Parsing and Implementation for Relative time modifiers for earliest and latest

vasukalariya:
# Description
Summarize the change.
Created parser for relative time modifier and implemented the functionality.

# Testing
Checked the final start and time stamps for all of these queries.

```

address = "4852 Lake Ridge port, Santa Ana, Nebraska 13548" | earliest=@w0

address = "4852 Lake Ridge port, Santa Ana, Nebraska 13548" | earliest=-1h@h


address = "4852 Lake Ridge port, Santa Ana, Nebraska 13548" | earliest=@-1d@d


address = "4852 Lake Ridge port, Santa Ana, Nebraska 13548" | earliest=@-24hr


address = "4852 Lake Ridge port, Santa Ana, Nebraska 13548" | earliest=@-7d@d


address = "4852 Lake Ridge port, Santa Ana, Nebraska 13548" | earliest=@-7d@m

address = "4852 Lake Ridge port, Santa Ana, Nebraska 13548" | earliest=+1d@d

address = "4852 Lake Ridge port, Santa Ana, Nebraska 13548" | earliest=+1d@d

address = "4852 Lake Ridge port, Santa Ana, Nebraska 13548" | earliest=+24h

address = "4852 Lake Ridge port, Santa Ana, Nebraska 13548" | earliest=-y@w0 latest=-1d@d

address = "4852 Lake Ridge port, Santa Ana, Nebraska 13548" | earliest=@w3 latest=-24hr

address = "4852 Lake Ridge port, Santa Ana, Nebraska 13548" | earliest=@w2 latest=-7d@d

address = "4852 Lake Ridge port, Santa Ana, Nebraska 13548" | earliest=@w6 latest=-7d@m

address = "4852 Lake Ridge port, Santa Ana, Nebraska 13548" | earliest=@w5 latest=@w0

address = "4852 Lake Ridge port, Santa Ana, Nebraska 13548" | earliest=+1d@w3 latest=+1d@d

address = "4852 Lake Ridge port, Santa Ana, Nebraska 13548" | earliest=-1d@w4 latest=+24h

address = "4852 Lake Ridge port, Santa Ana, Nebraska 13548" | earliest=-60m

address = "4852 Lake Ridge port, Santa Ana, Nebraska 13548" | earliest=@quarter latest=@d

address = "4852 Lake Ridge port, Santa Ana, Nebraska 13548" | earliest=-years@year latest=@year

address = "4852 Lake Ridge port, Santa Ana, Nebraska 13548" | earliest=-week@m latest=@s

address = "4852 Lake Ridge port, Santa Ana, Nebraska 13548" | earliest=-week@mon latest=@s

address = "4852 Lake Ridge port, Santa Ana, Nebraska 13548" | earliest=-month@mon latest=@minute

address = "4852 Lake Ridge port, Santa Ana, Nebraska 13548" | earliest=-month@year latest=-2days@minute

address = "4852 Lake Ridge port, Santa Ana, Nebraska 13548" | earliest=-qtr@month latest=-week@hr

address = "4852 Lake Ridge port, Santa Ana, Nebraska 13548" | earliest=06/19/2024:18:55:00 latest=-3m

* | earliest=06/19/2024:18:55:00 latest=-5minutes

```

# Checklist:


pkg/ast/spl/spl.peg
```diff
@@ -3255,16 +3345,65 @@ FieldNameList <- first:FieldName rest:(COMMA FieldName)* {
...
+        t := currTime
+        if offsetNum != 0 {
+            t, err = applyOffset(offsetNum, offsetTimeUnit, t)
+            if err != nil {
+                return nil, err
+            }
+        }
+        if snap != "" {
+            t, err = applySnap(snap, t)
+            if err != nil {
```
Macbeth98, line 3375: Can both `snap` and `offsetNum` be given at the same time in the query? If yes, then you are overwriting the `offsetNum` value with `snap` value.

pkg/ast/spl/spl.peg
```diff
@@ -3277,10 +3416,29 @@ EarliestAndLatest <- CMD_EARLIEST EQUAL earliestTime:AbsoluteTimestamp SPACE CMD
...
+    earliestTimeModifier := earliestTime.(ast.RelativeTimeModifier)
+
+    if earliestTimeModifier.AbsoluteTime != "" {
+        startEpoch, err = ConvertCustomDateTimeFormatToEpochMs(earliestTimeModifier.AbsoluteTime)
+        if err != nil {
+            return nil, err
+        }
+    } else {
+        t := time.Now()
+        offsetNum := earliestTimeModifier.RelativeTimeOffset.Offset
```
Macbeth98, line 3432: I would suggest to refactor this part of the else code block into a separate function and call it wherever it is required. As I can see the similar else block is there in three places.

pkg/ast/spl/spl.peg
```diff
@@ -3295,10 +3453,92 @@ EarliestOnly <- CMD_EARLIEST EQUAL earliestTime:AbsoluteTimestamp !. {
...
+
+AbsoluteTimestamp <- [0-9][0-9] '/' [0-9][0-9] '/' [0-9][0-9][0-9][0-9] ':' [0-9][0-9] ':' [0-9][0-9] ':' [0-9][0-9] {
+    relTimeOffset := ast.RelativeTimeOffset{
+        Offset: 0,
+        TimeUnit: 0,
+    }
+    relTimeModifier := ast.RelativeTimeModifier{
+        RelativeTimeOffset: relTimeOffset,
+        Snap: "",
+        AbsoluteTime: string(c.text),
```
Macbeth98, line 3537: You do not need to create default values like this. Go automatically creates the defaults. So you can directly do this:

```
relTimeModifier := ast.RelativeTimeModifier{
       AbsoluteTime: string(c.text)
}
// You can ignore any or all other fields, if you are going to assign some default values.
```

Macbeth98: Same comment for the above `RelativeTimestamp` definition.
You can just do this: `relTimeOffset := ast.RelativeTimeOffset`

pkg/ast/spl/spl.peg
Macbeth98: From the below Parsing definition at #L3500:
`RelativeTimestamp <- offset:(Offset)? snapParam:(Snap)? { ... }`
I believe both of them can be set. 

If that is not the case, then you need to change the parser to reflect that.

Macbeth98 (changes requested):
The Parsing looks pretty good, just some minor nits and a refactoring. And one possible bug. 

Also, please write test cases in `splParser_test.go`. Write test cases for all the queries that you have listed in the PR. 
This is important, so that any future changes to `peg` may break this `earliest`, `latest` or any other commands grammar. So we need these test cases to instantly identify these type of breaking changing.

pkg/ast/spl/spl.peg
Macbeth98: From the below Parsing definition at #L3500:
`RelativeTimestamp <- offset:(Offset)? snapParam:(Snap)? { ... }`
I believe both of them can be set. 

If that is not the case, then you need to change the parser to reflect that.

pkg/ast/spl/spl.peg
```diff
@@ -40,6 +40,96 @@ func ConvertCustomDateTimeFormatToEpochMs(dateStr string) (int64, error) {
...
+    hour := t.Hour()
+    day := t.Day()
+    week := t.Weekday()
+    mon := t.Month()
+    year := t.Year()
+    
+    if snap[0] != 'w' {
+        tunit, err := strconv.Atoi(snap)
+        if err != nil {
+            return t, fmt.Errorf("Error while converting the snap to integer, err: %v", err)
```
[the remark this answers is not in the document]

pkg/ast/spl/spl.peg
```diff
@@ -40,6 +40,96 @@ func ConvertCustomDateTimeFormatToEpochMs(dateStr string) (int64, error) {
...
+                }
+                return time.Date(year, mon, 1, 0, 0, 0, 0, time.Local), nil
+            case utils.TMYear:
+                return time.Date(year, 1, 1, 0, 0, 0, 0, time.Local), nil
+            default:
+                return t, fmt.Errorf("Unsupported time unit for relative timestamp: %v", tunit)
+        }
+    } else {
+        if len(snap) != 2 {
+            return t, fmt.Errorf("Error for special week snap, should be of type w[0-7]")
```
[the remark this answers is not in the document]

pkg/ast/spl/spl.peg
vasukalariya: First offset is applied and then snap is applied when both are there. If offset is not there then snap is only applied and similar is the case if offset is there and snap is not.

pkg/ast/spl/spl.peg
vasukalariya: addressed in the latest commit.

pkg/ast/spl/spl.peg
```diff
@@ -3255,16 +3379,25 @@ FieldNameList <- first:FieldName rest:(COMMA FieldName)* {
 
 TimeModifiers <- EarliestAndLatest / EarliestOnly
 
-EarliestAndLatest <- CMD_EARLIEST EQUAL earliestTime:AbsoluteTimestamp SPACE CMD_LATEST EQUAL latestTime:AbsoluteTimestamp !. {
-    startEpoch, err := ConvertCustomDateTimeFormatToEpochMs(earliestTime.(string))
+EarliestAndLatest <- CMD_EARLIEST EQUAL earliestTime:(AbsoluteTimestamp / RelativeTimestamp) SPACE CMD_LATEST EQUAL latestTime:(AbsoluteTimestamp / RelativeTimestamp) !. {
+     var startEpoch int64 = 0
+     var endEpoch int64 = 0
+    var err error
```
AndrewHess, line 3385: nit: something's wrong with the indentation. Maybe you're using a combination of tabs and spaces? Unlike the .go files, for this .peg file we use spaces to indent

pkg/ast/spl/spl.peg
```diff
@@ -3295,10 +3433,102 @@ EarliestOnly <- CMD_EARLIEST EQUAL earliestTime:AbsoluteTimestamp !. {
     return node, nil
 }
 
-AbsoluteTimestamp <- [0-9][0-9] '/' [0-9][0-9] '/' [0-9][0-9][0-9][0-9] ':' [0-9][0-9] ':' [0-9][0-9] ':' [0-9][0-9] {
+RelIntegerAsString <- [-+][0-9]* {
+    return string(c.text), nil
+}
```
AndrewHess, line 3438: Did you make this new rule instead of reusing `IntegerAsString` because the `-` or `+` is required? Also do you need to change the `*` to `+` so that at least one digit is present?

pkg/ast/spl/spl.peg
```diff
@@ -3295,10 +3433,102 @@ EarliestOnly <- CMD_EARLIEST EQUAL earliestTime:AbsoluteTimestamp !. {
...
+
+WeekSnap <- ('w'[0-7]) {
```
AndrewHess, line 3440: Just double checking: does splunk allow 0-7 for the weekday, or is it 0-6 or 1-7?

pkg/ast/spl/spl.peg
```diff
@@ -3295,10 +3433,102 @@ EarliestOnly <- CMD_EARLIEST EQUAL earliestTime:AbsoluteTimestamp !. {
...
+        TimeUnit: utils.TimeUnit(timeUnit),
+    }
+
+    return relTimeOffset, nil
+}
+
+
+
+// TODO: Chained relative modifiers
+RelativeTimestamp <- defaultTime:("now" / "1")? offset:(Offset)? snapParam:(Snap)? {
```
AndrewHess, line 3483: Why are `"now"` and `"1"` the only valid `defaultTime` values, but not other digits? Do you have a splunk documentation link? It could be helpful to add that as a comment here

pkg/ast/spl/spl.peg
```diff
@@ -40,6 +40,130 @@ func ConvertCustomDateTimeFormatToEpochMs(dateStr string) (int64, error) {
...
+                }
+                return time.Date(year, mon, 1, 0, 0, 0, 0, time.Local), nil
+            case utils.TMYear:
+                return time.Date(year, 1, 1, 0, 0, 0, 0, time.Local), nil
+            default:
+                return t, fmt.Errorf("Unsupported time unit for relative timestamp: %v", tunit)
+        }
+    } else {
+        if len(snap) != 2 {
+            return t, fmt.Errorf("Error for special week snap, should be of type w[0-7] got: %v", snap)
```
AndrewHess, line 91: nit: I think `should be of type w[0-7]` is a bit unclear; I think what you mean is `should follow the regex "w[0-7]"`, right?

pkg/ast/spl/spl.peg
```diff
@@ -40,6 +40,130 @@ func ConvertCustomDateTimeFormatToEpochMs(dateStr string) (int64, error) {
     return t.UnixMilli(), nil
 }
 
+
+func applySnap(snap string, t time.Time) (time.Time, error) {
```
AndrewHess, line 44: It's pretty confusing reading this function unless you already know what `snap` should look like. Can you add a comment above this function to describe or give examples of what `snap` should contain?

pkg/ast/spl/spl.peg
```diff
@@ -40,6 +40,130 @@ func ConvertCustomDateTimeFormatToEpochMs(dateStr string) (int64, error) {
...
+            diff += 7
+        }
+        return time.Date(year, mon, day, 0, 0, 0, 0, time.Local).AddDate(0, 0, -diff), nil
+    }
+}
+
+
+func applyOffset(num int64, unit utils.TimeUnit, t time.Time) (time.Time, error) {
+
+    durNum := time.Duration(num)
```
AndrewHess, line 108: I think renaming `num` will help make this function more clear. Perhaps `nanoseconds` would be a better choice. What do you think?

pkg/ast/spl/spl.peg
```diff
@@ -40,6 +40,130 @@ func ConvertCustomDateTimeFormatToEpochMs(dateStr string) (int64, error) {
...
+    } else if relTimeModifier.AbsoluteTime != "" {
+        epoch, err = ConvertCustomDateTimeFormatToEpochMs(relTimeModifier.AbsoluteTime)
+        if err != nil {
+            return 0, err
+        }
+    } else {
+        offsetNum := relTimeModifier.RelativeTimeOffset.Offset
+        offsetTimeUnit := relTimeModifier.RelativeTimeOffset.TimeUnit
+        snap := relTimeModifier.Snap
+        t := currTime
```
AndrewHess, line 148: Why are you renaming `currTime` to `t`? It seems like this only makes the code less clear

pkg/ast/spl/spl.peg
vasukalariya: Yes -/+ is required. No, + would not be there cause you can just pass +/-.

pkg/ast/spl/spl.peg
vasukalariya: w0 and w7 would be treated same. I just followed what they have.

pkg/ast/spl/spl.peg
vasukalariya: you can pass now to set to the currenTime and 1 is the unix epoch 1. These are the only constants that are there on [doc](https://docs.splunk.com/Documentation/Splunk/9.1.2/Search/Specifytimemodifiersinyoursearch)

pkg/ast/spl/spl.peg
vasukalariya: yes basically they could be from w0, w1, ... to w7

pkg/ast/spl/spl.peg
vasukalariya: num is actually the number of times you want to apply the offset so basicall the offset number.

pkg/ast/spl/spl.peg
vasukalariya: We are operating on t to apply offset and then whatever we get we would apply snap if present on that time. setting the currTime to offsetApplied currTime would be a bit unclear so just used a variable to represent the currTime.

pkg/ast/spl/spl.peg
AndrewHess: Ah ok, that makes sense. Then why do you have `durNum := time.Duration(num)`? I see later you do `t.Add(durNum * time.Second)` but couldn't you do `t.Add(num * time.Second)` instead and then there's no point in having `durNum`? Or will that not work?

pkg/ast/spl/spl.peg
AndrewHess: Ok. Can you add that link as a comment in the peg file?

pkg/ast/spl/spl.peg
AndrewHess: Ok, then maybe naming it something like `finalTime` would be more clear?

Aside from loop indexes, I generally find that 1-letter names are not very clear, and once they're in the codebase it makes searching for them a little more annoying because you have to do a regex search like `[^a-z]t[^a-z]` instead of just `myVariable` or else you'll get a lot of matches in the middle of other words

pkg/ast/structs.go
```diff
@@ -80,6 +80,17 @@ type ParseError struct {
 	Expected []string `json:"expected"`
 }
 
+type RelativeTimeModifier struct {
+	RelativeTimeOffset RelativeTimeOffset
+	Snap               string
+	AbsoluteTime       string
```
venkateshamatam, line 86: Maybe rename the struct to TimeModifier because it has both Absolute and RelativeTimeModifier variables. And update it in a way that it is easy to differentiate between the two. Maybe like this 

```
type TimeModifier struct  {
	RelativeTime       RelativeTimeModifier
	AbsoluteTime       string
}

type RelativeTimeModifier struct {
	RelativeTimeOffset RelativeTimeOffset
	Snap               string
}
```

Update the variable names in the PEG file accordingly

vasukalariya: The issue is in future it could be the case that we need to support absolute time with relative like now-5d etc. That time it would be a relative time modifier but along with absolute time and relative time as well. Moreover, name TimeModifier is already taken for a struct so this was the only plausible name I could think of.

pkg/ast/spl/spl.peg

pkg/ast/spl/spl.peg
vasukalariya: addressed in the latest commit

pkg/ast/spl/spl.peg
vasukalariya: addressed in the latest commit

pkg/ast/spl/spl.peg

pkg/ast/structs.go
Macbeth98: I do not understand what you mean by this `we need to support absolute time with relative like now-5d`?

In that case, wouldn't it be a relative case? 

Also, for the other struct that is using the type `TimeModifier` , I think you are referring to `TimeModifiers` struct in `structs.go`. 

```
// Node is a node in the query parse tree
type Node struct {
	NodeType      NodeType
	Comparison    Comparison
	Left          *Node
	Right         *Node
	TimeModifiers *TimeModifiers
}

// TimeModifiers is a struct that holds the time modifiers for a query
type TimeModifiers struct {
	StartEpoch uint64
	EndEpoch   uint64
}
```

And based on the struct and it's use case, I would say refactor/rename this struct to `TimeRangeModifier`

pkg/ast/spl/tests/splParser_test.go
```diff
@@ -7793,6 +7795,306 @@ func Test_Format_cmd_Incomplete_RowCol_Options(t *testing.T) {
...
+	time := time.Date(2024, time.June, 5, 13, 37, 5, 0, time.Local)
+	// w0
+	rtm := ast.RelativeTimeModifier{
+		Snap: "w0",
+	}
+	epoch, err := spl.CalculateRelativeTime(rtm, time)
+	assert.Nil(t, err)
+	// Sunday, June 2, 2024 12:00:00 AM UTC-4
+	assert.Equal(t, int64(1717300800000)+getTimeDiff(false), epoch)
+}
```
AndrewHess, line 7824: nit: you don't need the comment on line 7816 because it's clear from the code right after it, so the comment doesn't add any information. For lines 7822-7823, it would probably be a little better to remove the comment (except for the weekday, since that's important for this particular test) and make a `time.Date` directly using that info and then call `Unix()` to get the unix epoch for that date.

You can keep 7822-7823 how you have it if you want, but you should remove the comment on 7816

Same comment for several of your other tests

pkg/ast/spl/tests/splParser_test.go
```diff
@@ -7793,6 +7795,306 @@ func Test_Format_cmd_Incomplete_RowCol_Options(t *testing.T) {
 	assert.NotNil(t, err)
 }
 
+func getTimeDiff(daylight bool) int64 {
+	localTime := time.Date(2024, time.June, 5, 13, 37, 5, 0, time.Local)
+	if daylight {
+		localTime = time.Date(2024, time.January, 1, 0, 0, 0, 0, time.Local)
+	}
```
AndrewHess, line 7802: Isn't daylight saving time in the summer?

vasukalariya: removed this to use fixed dates for clean code. Addressed in the new commit.

pkg/ast/spl/tests/splParser_test.go
vasukalariya: addressed in the new commit.

pkg/ast/structs.go
vasukalariya: addressed in the new commit.

AndrewHess (approved):

Macbeth98 (approved):

PR #1164, opened 2024-06-21
Add Parsing for eventcount index command

venkateshamatam:
# Description
https://docs.splunk.com/Documentation/SplunkCloud/9.1.2308/SearchReference/Eventcount

# Testing
- Added testcases for parser.
# Checklist:

### DIFF
--- a/pkg/ast/spl/tests/splParser_test.go
+++ b/pkg/ast/spl/tests/splParser_test.go
@@ -7795,41 +7795,66 @@ func Test_Format_cmd_Incomplete_RowCol_Options(t *testing.T) {
 
 func Test_EventCount_Defaults(t *testing.T) {
 	query := `* | eventcount`
-	performEventCountTest(t, query, "*", true, false, true)
+	performEventCountTest(t, query, []string{"*"}, true, false, true)
 }
 
 func Test_EventCount_IndexSpecified(t *testing.T) {
 	query := `* | eventcount index=my_index`
-	performEventCountTest(t, query, "my_index", true, false, true)
+	performEventCountTest(t, query, []string{"my_index"}, true, false, true)
 }
 
 func Test_EventCount_SummarizeFalse(t *testing.T) {
 	query := `* | eventcount summarize=false`
-	performEventCountTest(t, query, "*", false, false, true)
+	performEventCountTest(t, query, []string{"*"}, false, false, true)
 }
 
 func Test_EventCount_ReportSizeTrue(t *testing.T) {
 	query := `* | eventcount report_size=true`
-	performEventCountTest(t, query, "*", true, true, true)
+	performEventCountTest(t, query, []string{"*"}, true, true, true)
 }
 
 func Test_EventCount_ListVixFalse(t *testing.T) {
 	query := `* | eventcount list_vix=false`
-	performEventCountTest(t, query, "*", true, false, false)
+	performEventCountTest(t, query, []string{"*"}, true, false, false)
 }
 
 func Test_EventCount_Combination1(t *testing.T) {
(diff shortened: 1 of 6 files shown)


pkg/ast/spl/tests/splParser_test.go
```diff
@@ -7792,3 +7792,62 @@ func Test_Format_cmd_Incomplete_RowCol_Options(t *testing.T) {
...
+
+func Test_EventCount_Combination1(t *testing.T) {
+	query := `* | eventcount index=my_index summarize=false report_size=true`
+	performEventCountTest(t, query, "my_index", false, true, true)
+}
+
+func Test_EventCount_Combination2(t *testing.T) {
+	query := `* | eventcount index=my_index summarize=true report_size=false list_vix=false`
+	performEventCountTest(t, query, "my_index", true, false, false)
+}
```
AndrewHess, line 7829: You should also add a unit test where the optional parameters are in a different order. (I'm pretty sure your existing parsing solution will handle this correctly)

pkg/segment/structs/evaluationstructs.go
```diff
@@ -181,20 +181,21 @@ type StringExpr struct {
...
+	ValueList      []*StringExpr
+	StartIndex     *NumericExpr
+	EndIndex       *NumericExpr
+	LengthExpr     *NumericExpr
+	Val            *ValueExpr
+	Condition      *BoolExpr       // To filter out values that do not meet the criteria within a multivalue field
+	InferTypes     bool            // To specify that the mv_to_json_array function should attempt to infer JSON data types when it converts field values into array elements.
+	Cluster        *Cluster        // generates a cluster label
+	SPathExpr      *SPathExpr      // To extract information from the structured data formats XML and JSON.
+	EventCountExpr *EventCountExpr // To count the number of events in an index
```
AndrewHess, line 198: Does `EventCountExpr` belong in the `TextExpr` like you have it or in the `NumericExpr`? I forget if the type refers to the type that expression returns or the type that expression takes as inputs

AndrewHess (commented):
Looks pretty good

pkg/segment/structs/evaluationstructs.go
venkateshamatam: I think it doesn't belong to `TextExpr` or `NumericExpr` since we're not doing any text or numeric manipulation on the results. I created another Request type called `EventCountRequest` in `LetColumnsRequest`, similar to what we have for `FormatResults` in `LetColumnRequest`

pkg/ast/spl/tests/splParser_test.go
venkateshamatam: Added more testcases for parser, and also updated the parser to handle multiple index names in the query like 

`eventcount summarize=false index=_audit index=main` and added test cases for this as well in multiple combinations

pkg/segment/structs/segstructs.go
```diff
@@ -841,15 +850,35 @@ func CheckUnsupportedFunctions(post *QueryAggregators) error {
...
+			return err
 		}
 	}
 
 	return nil
 }
+func checkUnsupportedLetColumnCommand(agg *QueryAggregators, command string) error {
+	if agg.hasLetColumnsRequest() {
+		letColumns := agg.OutputTransforms.LetColumns
+		switch command {
```
Macbeth98, line 869: I think it is better, to keep a global slice or map of `LetColumnsCommand` that we do not support and then loop through that slice or map in the `if` block. 

The code in the `if` block will remain the same and you don't need to pass the command as an argument. 
Through this you will only call the `checkUnsupportedLetColumnCommand` only once in `CheckUnsupportedFunctions` definition.

Macbeth98 (approved):

PR #1165, opened 2024-06-21
feat: Updated the logic for Alert State Transition during Alert Evaluation based on Alert Eval Window

Macbeth98:
# Description
- Updated the Alert state transition logic based on the Evaluations across the Eval window.
- State Transition Rules:
   - `Normal` to `Pending`: When an alert condition is matched.
   - `Pending` to `Firing`: When an alert condition is matched for all the last (Interval Count) Evaluations.
   - `Firing` to `Normal` / `Pending` to `Normal`: When an alert condition is not matched.

- `NotificationHandler` will be called to send an alert notification when:
     - The Alert is in `Firing` state or transitioned from `Pending` to `Firing` state.
     - ~~The Alert is Transitioned from `Firing` to `Normal` state (to let user know that the alert state reverted to Normal).~~
     - **UPDATE:** The Alert is in `Normal` state.

**NOTE**: The implementation of cooldown period on the `NotificationHandler` is expected to decide whether to accept or reject a notification request from Alert System. 
**UPDATE:** And the Notification System is expected to send the `Normal` state notification only if the previous Alert notification sent was `Firing`.

### Transition from `Pending` to `Firing`:
- Every time when the alert condition is matched, the Alert system will try to get the last (Interval Count -1) Alert History records from the DB. 
- Interval Count - 1. because, including the current matched one => Interval Count -1 + 1 = Interval Count.
- Once the records are fetched and the size matches (Interval Count - 1), then will check if for all these records, the alert state is in either `Pending` or `Firing`. 
- If `true`, then the state will be transitioned to `Firing`, otherwise, the state will remain in `Pending`.

### Updated the `GetAlertHistory` API
- To fetch the records by sending optional parameters like `offset`, `limit`, and `sort_order`.

### Refactoring:
- Refactored the code after an Alert Condition is evaluated.
- In `server.go` pushed the initiation the Alerting service to last (After `InitQueryMetrics` and `InitQueryNode`). This is because, the alerting service is getting initialized and started to fetch the data before the rotated segments are loaded into the memory buffer.

# Testing
- Tested by creating two test Alerts and verified the transitions from `Pending` to `Firing` and a Firing Alert Notification.
- Verified the Transition from `Firing` to `Normal` and a Normal update Alert Notification.

# Checklist:

### DIFF
--- a/pkg/alerts/alertsHandler/cronJobHandler.go
+++ b/pkg/alerts/alertsHandler/cronJobHandler.go
@@ -223,7 +223,7 @@ func evaluateLogAlert(alertToEvaluate *alertutils.AlertDetails, job gocron.Job) 
 
 	if isResultsEmpty {
 		// Should not return here, as this can mean, there are no valid logs that satisfies the Alert Query.
-		// This should be considered as a normal state. And we should update the alert state to Inactive.
+		// This should be considered as a normal state. And we should update the alert state to Normal.
 		log.Warnf("ALERTSERVICE: evaluateLogAlert: Empty response returned by server.")
 
 		// We should call the update here instead of letting the execution go to the evaluation of the conditions.
@@ -350,15 +350,15 @@ func evaluateMinionSearch(msToEvaluate *alertutils.MinionSearch, job gocron.Job)
 
 	if isResultsEmpty {
 		// Should not return here, as this can mean, there are no valid logs that satisfies the Alert Query.
-		// This should be considered as a normal state. And we should update the alert state to Inactive.
+		// This should be considered as a normal state. And we should update the alert state to Normal.
 		log.Warnf("MinionSearch: evaluate: Empty response returned by server.")
 
 		// We should call the update here instead of letting the execution go to the evaluation of the conditions.
 		// This is because, we return -1 as the result, when there are no logs that satisfies the query.
 		// And the condition in the evaluation can be looking for a value that is (>, <, =, !=) -1.
-		err := updateMinionSearchStateAndCreateAlertHistory(msToEvaluate, alertutils.Inactive, alertutils.AlertNormal)
+		err := updateMinionSearchStateAndCreateAlertHistory(msToEvaluate, alertutils.Normal, alertutils.AlertNormal)
 		if err != nil {
-			log.Errorf("ALERTSERVICE: evaluateMinionSearch: Error in updateMinionSearchStateAndCreateAlertHistory. AlertState=%v, Alert=%+v & err=%+v.", alertutils.Inactive, msToEvaluate.AlertName, err)
+			log.Errorf("ALERTSERVICE: evaluateMinionSearch: Error in updateMinionSearchStateAndCreateAlertHistory. AlertState=%v, Alert=%+v & err=%+v.", alertutils.Normal, msToEvaluate.AlertName, err)
 		}
 		return
 	}
@@ -375,9 +375,9 @@ func evaluateMinionSearch(msToEvaluate *alertutils.MinionSearch, job gocron.Job)
 			return
 		}
 	} else {
-		err := updateMinionSearchStateAndCreateAlertHistory(msToEvaluate, alertutils.Inactive, alertutils.AlertNormal)
+		err := updateMinionSearchStateAndCreateAlertHistory(msToEvaluate, alertutils.Normal, alertutils.AlertNormal)
 		if err != nil {
-			log.Errorf("ALERTSERVICE: evaluateMinionSearch: Error in updateMinionSearchStateAndCreateAlertHistory. AlertState=%v, Alert=%+v & err=%+v.", alertutils.Inactive, msToEvaluate.AlertName, err)
+			log.Errorf("ALERTSERVICE: evaluateMinionSearch: Error in updateMinionSearchStateAndCreateAlertHistory. AlertState=%v, Alert=%+v & err=%+v.", alertutils.Normal, msToEvaluate.AlertName, err)
 		}
 	}
 }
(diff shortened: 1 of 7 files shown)


nkunal:
@AndrewHess please review this on monday

pkg/alerts/alertsHandler/cronJobHandler.go
```diff
@@ -116,6 +117,109 @@ func updateAlertStateAndCreateAlertHistory(alertDetails *alertutils.AlertDetails
...
+		}
+
+		// If the previous state was Firing, then we should send the Alert Notification.
+		// The cooldown period on the Notification Handler will decide if the notification should be sent. So that false positives are avoided.
+		if lastAlertHistory != nil && lastAlertHistory.AlertState == alertutils.Firing {
+			err = NotifyAlertHandlerRequest(alertToEvaluate.AlertId, "The Alert State has been updated to Normal.")
+			if err != nil {
+				return fmt.Errorf("handleAlertCondition: Could not send Alert Notification. found error = %v", err)
+			}
+		}
```
AndrewHess, line 217: Do you also have to tell the Notification system the alert's state? I assume we'd want the cooldown to only apply to alerts in the Firing state so we still get Normal notifications immediately, but maybe I'm wrong

I think with the current `NotifyAlertHandlerRequest()` implementation, it's possible that we'd never get a Normal message sent: suppose an alert moves to Firing, so then we get a Firing notification, and then it moves back to Normal but the notification system doesn't send a message because the cooldown is active, then eventually the cooldown ends and even if we evaluate this alert again to Normal we won't send the Normal notification because now the previous state was Normal (not Firing)

Would it make sense to always send the new state to the notification system and make the notification system responsible for deciding when to send a notification?

pkg/alerts/alertsHandler/cronJobHandler.go
```diff
@@ -116,6 +117,109 @@ func updateAlertStateAndCreateAlertHistory(alertDetails *alertutils.AlertDetails
...
+	}
+
+	if len(alertHistoryList) == 0 {
+		return nil, nil
+	}
+
+	return alertHistoryList[0], nil
+}
+
+func handleAlertCondition(alertToEvaluate *alertutils.AlertDetails, isAlertConditionMatched bool, alertDataMessage string) error {
```
AndrewHess, line 175: The `AlertDetails` type has a `Message` field. Should you use that instead of passing this `alertDataMessage`?

static/js/all-alerts.js
```diff
@@ -23,9 +23,10 @@ let alertGridDiv = null;
 let alertRowData = [];
 
 let mapIndexToAlertState=new Map([
-    [0,"Normal"],
-    [1,"Pending"],
-    [2,"Firing"],
+    [0, "Inactive"],
+    [1,"Normal"],
+    [2,"Pending"],
+    [3,"Firing"],
```
AndrewHess, line 29: What does `Inactive` vs `Normal` mean?

Macbeth98: Inactive is the initial state when the alert was first created. It also implies, that the alert conditions evaluations has not been executed even once.

pkg/alerts/alertsHandler/cronJobHandler.go
Macbeth98: The `Message` in the `AlertDetails` struct is the Custom message that the user has set for this Alert. 
And the `alertDataMessage` contains the specific records or series that satisfied the Alert Condition.

pkg/alerts/alertsHandler/cronJobHandler.go
Macbeth98: I guess that makes sense, so every time when the state is in `Normal`, the alert system would call the `NotifyAlertHandlerRequest` to send the notification,  similar to `Firing` state. And then The `NotificationSystem` will decide whether to send this notification or not.

But also, the user would probably not like to receive `Normal` state alert notifications all the time. It would probably be okay, only once when the state is transitioned from `Firing` to `Normal`. 

So I believe that the Notification system should keep track of the notifications it sent or at least the last Notification Alert state it sent. 
- If the current state is in `Firing` depending on the cooldown period it will send the notification.
- If the current state is in `Normal`, then depending on the cooldown period plus the Alert state when the last Notification is sent, it should send the notification. It should only send if the Alert state of last notification that it sent was `Firing` and if it is `Normal`, it should not send.

static/js/all-alerts.js
AndrewHess: Ok and we want to distinguish between them so the user can see which alerts haven't run vs which alerts have run and are normal?

I see that `evaluateMinionSearch()` sometimes uses a combination like this:
```
err := updateMinionSearchStateAndCreateAlertHistory(msToEvaluate, alertutils.Inactive, alertutils.AlertNormal)
```
Do we need to change that so it's just using one of the them? (If we do need to change that, it can be a separate PR  if you want)

pkg/alerts/alertsHandler/cronJobHandler.go
AndrewHess: Yeah so do you need to also send the alert state, or will sending just the `alertID` and the message like you are doing be enough information for the notification system to figure out what to do?

Macbeth98: Yes, I think, I will work on that, when I am implementing the cooldown period and making fixes on Notification system.

static/js/all-alerts.js
Macbeth98: I have updated it in this PR, itself. Now Minion search also updates the state to `Normal` instead of `Inactive`.

AndrewHess (approved):

PR #1170, opened 2024-06-25
feat: Reject if Alert Logs Query is not a Stats Query + Update Notification Logic to send an alert based on AlertState

Macbeth98:
# Description
- Added a check to verify whether the given Logs query is a stats query or not and reject if it is not a stats query.
- Updated the Notification system logic to keep track of Last Alert state when the last notification is sent and based on that along with cooldown and silence minutes, a notification will be sent.
    - The Notification will not be sent if the last notification alert state is `Normal` and the current state is also `Normal`.

# Testing
- Tested by creating two notifications one for logs and other for Metrics.
- Verified that they are sending the notifications when the state is in `Firing` and when the state is transitioned from `Firing` to `Normal`.

# Checklist:

### DIFF
--- a/pkg/alerts/alertsHandler/alertsHandler.go
+++ b/pkg/alerts/alertsHandler/alertsHandler.go
@@ -58,9 +58,10 @@ type database interface {
 	GetAllContactPoints(orgId uint64) ([]alertutils.Contact, error)
 	UpdateContactPoint(contact *alertutils.Contact) error
 	GetCoolDownDetails(alert_id string) (uint64, time.Time, error)
+	GetAlertNotification(alert_id string) (*alertutils.Notification, error)
 	GetContactDetails(alert_id string) (string, string, string, error)
 	GetEmailAndChannelID(contact_id string) ([]string, []alertutils.SlackTokenConfig, []alertutils.WebHookConfig, error)
-	UpdateLastSentTime(alert_id string) error
+	UpdateLastSentTimeAndAlertState(alert_id string, alertState alertutils.AlertState) error
 	UpdateAlertStateByAlertID(alertId string, alertState alertutils.AlertState) error
 	DeleteContactPoint(contact_id string) error
 }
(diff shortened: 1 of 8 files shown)


AndrewHess (approved):
Looks good. One thing: if the user specifies the alert should run every 5 minutes and the cooldown period is an integer multiple of that (e.g., 10 minutes), then there could be some inconsistency because sometimes the alert may get run just under 10 minutes later or just over 10 minutes later, so then sometimes the de facto cooldown is 15 minutes and other times it's 10 minutes. If we want to solve this, one way is that when a notification would be sent but a cooldown or silence is still active, we have another process wait until it's no longer active and then immediately check again whether the notification should be sent. What do you think?

Macbeth98:
> Looks good. One thing: if the user specifies the alert should run every 5 minutes and the cooldown period is an integer multiple of that (e.g., 10 minutes), then there could be some inconsistency because sometimes the alert may get run just under 10 minutes later or just over 10 minutes later, so then sometimes the de facto cooldown is 15 minutes and other times it's 10 minutes. If we want to solve this, one way is that when a notification would be sent but a cooldown or silence is still active, we have another process wait until it's no longer active and then immediately check again whether the notification should be sent. What do you think?

So, if a cooldown period or silence minutes is active when a notification is being sent, then we will spawn a `go routine` or a function that will be called immediately after the cooldown period or silence minutes is done and checks whether to send the Notification or not? Is my understanding correct?

But I also believe these type of cases will be avoided when the next time the evaluation is executed right? Like the when the alert is in `Firing` or `Normal` state, we will always call the Notification handler, to send an alert and if the cooldown period is active, then even if we don't send for the current interval, we would definitely send the notification for the next evaluation I think. But we might skip sending a notification for which a notification should have been sent.

And If we do in the way as you have suggested, probably we should have some global maps, so we do not spawn these processes for an Alert for which there is already a process running. 
And if for an alert while a process is spawned to check and send a notification... and meanwhile, the normal alert flow process requested to send a Notification and the notification was sent, then we should probably disable the spawned process.

I think we should also come back to this discussion when we are actually doing the cooldown period.

PR #1176, opened 2024-06-26
Log page Builder, Code and Filtering

602dhruviii:
# Description
Maintain Query Tab on Logs Screen Refresh and Unselecting all fields in 'single line view' or 'wrap line view' changes the view to table while should maintain the user's selected view.

# Testing

# Checklist:


static/js/ag-grid-seg-stats.js
```diff
@@ -33,11 +33,22 @@
          let title =  colName;
          let resize = index + 1 == columnOrder.length ? false : true;
          let maxWidth = Math.max(displayTextWidth(colName, "italic 19pt  DINpro "), 200)         //200 is approx width of 1trillion number
+         // Define the cellRenderer only for myField column
```
sonamgupta21, line 36: You should remove this change from this PR.

static/js/saved-query.js
```diff
@@ -281,7 +282,7 @@ $(document).ready(function () {
    {
      field: "filterTab",
      headerName: "FilterTab",
-     hide: true,
+     hide: false,
```
sonamgupta21, line 285: 1. Don't need to show the filter tab on the saved query page.
2. The Filter tab is not getting maintained when you refresh.
For example, if the query mode (default) is set to code, and you run a query through the builder and then refresh the tab, it goes to code mode. It should remain on the builder tab.

static/js/saved-query.js
```diff
@@ -159,6 +159,7 @@
  class linkCellRenderer {
      // init method gets the details of the cell to be renderer
      init(params) {
+        Cookies.remove('queryMode');
```
sonamgupta21, line 162: You should not remove the cookie because if the user returns to the page again, their default choice gets removed.

PR #1177, opened 2024-06-26
feat: Add cicd/E2E tests for Alerting 

Macbeth98:
# Description
- Added E2E Tests for Alerting.
- Alerting E2E test can be run with command.
- Updated read me of SigClient with this command.

```
go run main.go alerts -d http://localhost:5122
```

# Testing
- Executed E2E tests and verified that the tests are working fine.

# Checklist:


AndrewHess (commented):
You're using `log.Fatal()` but you're also doing some cleanup after running tests. So if the tests fail you won't do some cleanup. Will that be ok? If that'll cause issues then instead of `log.Fatal()` you should probably throw an error and log it, stop the test, cleanup everything and exit with an error code

Macbeth98:
> You're using `log.Fatal()` but you're also doing some cleanup after running tests. So if the tests fail you won't do some cleanup. Will that be ok? If that'll cause issues then instead of `log.Fatal()` you should probably throw an error and log it, stop the test, cleanup everything and exit with an error code

I actually ignored it, because these tests will only be run inside `cicd` isolated environment. But I think it is better to clean up as you have suggested, so it can be used in any environment.

tools/sigclient/pkg/alerts/alerts.go
```diff
@@ -0,0 +1,633 @@
...
+		alert.QueryParams = alertutils.QueryParams{
+			DataSource:    LogsString,
+			QueryLanguage: LogsQueryLanguage,
+			QueryText:     LogsQueryText,
+			StartTime:     LogsStartTime,
+			EndTime:       LogsEndTime,
+		}
+	}
+
+	url := host + "/api/alerts/create" // Ensure the URL is correct
```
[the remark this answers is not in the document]
Macbeth98: Oh Yeah, I wanted to make a check to remove any trailing `/` for the host and I forgot about that. Thanks for pointing.

tools/sigclient/pkg/alerts/alerts.go
```diff
@@ -451,6 +388,13 @@ func verifyAlertHistory(host string, alerts []*alertutils.AlertDetails) error {
 	return nil
 }
 
+func removeTrailingSlashes(url string) string {
+	for strings.HasSuffix(url, "/") {
+		url = strings.TrimSuffix(url, "/")
+	}
+	return url
+}
```
AndrewHess, line 396: fyi: you can use `TrimRight` and then you don't need the loop. See https://go.dev/play/p/SwbhHJU0k2W
But I think the intent is more clear how you wrote it, so I'd suggest keeping what you have

tools/sigclient/pkg/alerts/alerts.go
```diff
@@ -464,6 +408,8 @@ func verifyAlertHistory(host string, alerts []*alertutils.AlertDetails) error {
 // Delete the Contact Points
 // Close the Webserver
 func RunAlertsTest(host string) {
+	// Remove Trailing Slashes from the Host
+	host = removeTrailingSlashes(host)
```
AndrewHess, line 412: nit: this comment doesn't add any info

AndrewHess (approved):
Looks good, you can merge as is if you want

tools/sigclient/pkg/alerts/alerts.go
Macbeth98: Ohh, I actually pushed the changes without reading your comment full after looking at the playground code. 
But I believe even `TrimRight` and `/` as the param will also be clear in the same way, especially with the function name.

AndrewHess (approved):

PR #1178, opened 2024-06-26
Tail command

vasukalariya:
# Description
Summarize the change.
Tail command for SPL

# Testing
```

app_name = "Bracecould" | tail 7

app_name = "Bracecould" | gender = male | tail 2

city=Boston | stats count, avg(latency) BY weekday, http_method | sort http_method | tail 3

* | tail 30000

app_name = "Bracecould" | tail 3000

app_name = "Bracecould" | sort batch | tail 5

city = Boston | group = "group 0" | tail

```

# Checklist:


pkg/ast/spl/spl.peg
```diff
@@ -2571,6 +2571,47 @@ HeadBlock <- PIPE CMD_HEAD ("limit" EQUAL)? intAsStr:IntegerAsString {
     return queryAgg, nil
 }
 
+
+// Returns *structs.QueryAggregators
+TailBlock <- PIPE CMD_TAIL ("limit" EQUAL)? intAsStr:IntegerAsString {
```
Macbeth98, line 2576: Is the `limit EQUAL` required? I don't see it on the SPL Command Tail reference page.

pkg/segment/aggregations/segaggs.go
```diff
@@ -196,6 +209,106 @@ func performAggOnResult(nodeResult *structs.NodeResult, agg *structs.QueryAggreg
...
+			if !exists {
+				continue
+			}
+			heap.Push(tailExpr.TailPQ, &utils.Item{
+				Priority: float64(timeVal.(uint64)),
+				Value:    k,
+			})
+			tailExpr.TailRecords[k] = v
+			if tailExpr.TailPQ.Len() > int(aggs.OutputTransforms.TailRows) {
+				item := heap.Pop(tailExpr.TailPQ).(*utils.Item)
```
Macbeth98, line 260: I think this will delete the oldest timestamp item right? Do we want to keep the oldest timestamp one or the latest timestamp one?

pkg/segment/structs/segstructs.go
```diff
@@ -187,6 +187,9 @@ type OutputTransforms struct {
 	FilterRows             *BoolExpr          // discard rows failing some condition
 	MaxRows                uint64             // if 0, get all results; else, get at most this many
 	RowsAdded              uint64             // number of rows added to the result. This is used in conjunction with MaxRows.
+	Tail                   bool
+	TailRows               uint64
+	TailRequest            *TailExpr
```
Macbeth98, line 192: From what I can see you are only using this, `Tail bool` value just to check if the Query Aggs has a Tail Block. 
I would suggest to refactor it this way so that it will be cleaner:

remove `Tail` and move `TailRows` as part of `TailRequest` meaning it will be inside `TailExpr` struct.

So the `TailExpr` struct will look like

```
type TailExpr struct {
        TailRows            uint64
	TailRecords          map[string]map[string]interface{}
	TailPQ               *sutils.PriorityQueue
	NumProcessedSegments uint64
}
```

And the output Transforms will only have `TailRequest` and you can remove `Tail` and `TailRows`.

And when checking whether Tail Block is there in `HasTail()` you can do something like this
```
if qa != nil && qa.OutputTransforms != nil && qa.OutputTransforms.TailRequest != nil {
		return true
	}
```

pkg/segment/aggregations/segaggs.go
```diff
@@ -196,6 +209,106 @@ func performAggOnResult(nodeResult *structs.NodeResult, agg *structs.QueryAggreg
...
+				delete(tailExpr.TailRecords, item.Value)
+			}
+			delete(recs, k)
+		}
+	}
+
+	if tailExpr.NumProcessedSegments < numTotalSegments {
+		return nil
```
Macbeth98, line 268: I am assuming you are not doing any storing of records like you do if there is no sort...
For a query that has sort, because sort will handle all the segments? So you are just returning if it is not final segment?

Macbeth98 (commented):
Looks Good. Please Add parser test cases for `Tail` command and also add some `tail` command queries on to `ingest.csv` and `restart.csv` as @AndrewHess mentioned.

pkg/ast/pipesearch/searchQueryParser.go
```diff
@@ -473,6 +473,11 @@ func parseColumnsCmd(node *structs.OutputTransforms, qid uint64) (*QueryAggregat
 	}
 
 	aggNode.OutputTransforms.MaxRows = node.MaxRows
+	aggNode.OutputTransforms.Tail = node.Tail
+	aggNode.OutputTransforms.TailRows = node.TailRows
+	if node.TailRequest != nil {
+		aggNode.OutputTransforms.TailRequest = node.TailRequest
+	}
```
AndrewHess, line 480: Why do you need this `if` for `TailRequest` and not need it for `Tail` and `TailRows`?

pkg/segment/aggregations/segaggs.go
```diff
@@ -196,6 +209,106 @@ func performAggOnResult(nodeResult *structs.NodeResult, agg *structs.QueryAggreg
...
+	}
+	if tailExpr.TailPQ == nil {
+		pq := make(utils.PriorityQueue, 0)
+		tailExpr.TailPQ = &pq
+		heap.Init(tailExpr.TailPQ)
+	}
+
+	if !hasSort {
+		for k, v := range recs {
+			timeVal, exists := v["timestamp"]
```
AndrewHess, line 250: nit: rename `k` and `v`. I'm not sure what `k` is, but it looks like `v` is a single record so maybe renaming it to `record` would make sense

pkg/segment/aggregations/segaggs.go
AndrewHess: I think that's correct. But Vasu, you should check if we get an unexpected state (like `sort` is true and this not the last segment but `recs` has some records) and log an error in that case so that we'll be able to detect and debug that quicker if it ever happens

pkg/segment/aggregations/segaggs.go
vasukalariya: We are using pkg/utils/PriorityQueue. It is a maxheap so removes the max timestamp first (i.e. the newest/latest one). So in the end we would only have the oldest ones in the heap/priority queue.

pkg/ast/spl/spl.peg

pkg/segment/structs/segstructs.go
vasukalariya: addressed in the latest commit

pkg/ast/pipesearch/searchQueryParser.go
vasukalariya: addressed in the latest commit

pkg/segment/aggregations/segaggs.go
vasukalariya: addressed in the latest commit. k refers to the string key use to map actual record in recs

pkg/segment/aggregations/segaggs.go
```diff
@@ -246,17 +246,17 @@ func performTail(nodeResult *structs.NodeResult, aggs *structs.QueryAggregators,
 	}
 
 	if !hasSort {
-		for k, v := range recs {
-			timeVal, exists := v["timestamp"]
+		for k, record := range recs {
+			timeVal, exists := record["timestamp"]
```
AndrewHess, line 250: nit: you should also rename `k`

pkg/segment/aggregations/segaggs.go
AndrewHess: You should probably also return an error when you log the error

pkg/segment/aggregations/segaggs.go
```diff
@@ -294,8 +297,8 @@ func performTail(nodeResult *structs.NodeResult, aggs *structs.QueryAggregators,
 			recordIndexInFinal[k] = idx
 		}
 	} else {
-		for k, v := range tailExpr.TailRecords {
-			recs[k] = v
+		for k, record := range tailExpr.TailRecords {
+			recs[k] = record
```
AndrewHess, line 301: nit: also rename `k` here

pkg/segment/aggregations/segaggs.go
```diff
@@ -196,6 +209,109 @@ func performAggOnResult(nodeResult *structs.NodeResult, agg *structs.QueryAggreg
...
+			}
+			delete(recs, k)
+		}
+	}
+
+	if tailExpr.NumProcessedSegments < numTotalSegments {
+		if hasSort && len(recs) > 0 {
+			log.Errorf("performTail: Sort was applied but still records are found in recs for a non-last segment")
+		}
+		return nil
```
AndrewHess, line 271: You should probably also return an error when you log the error

vasukalariya: addressed in the new commit. Just returning the error because we are logging that same error in postQueryBucketCleaning. So, no point in logging it twice.

pkg/segment/aggregations/segaggs.go
vasukalariya: addressed in the new commit.

pkg/segment/aggregations/segaggs.go
vasukalariya: addressed in the new commit.

AndrewHess (approved):
Looks good

PR #1187, opened 2024-06-27
Add support for SPL typeof

venkateshamatam:
# Description
Add support for SPL typeof. Works for field names and raw values, but the current logic converts the field to string and checks if the value is of a certain datatype.

# Testing
- Tested a few queries will add test cases for the same.
- queries like `* | eval res=typeof(12) + typeof("john") ` where we are performing an operation on eval doesnt work. Even queries like `* | eval ab=5+3` does not work
# Checklist:


pkg/segment/structs/evaluationstructs.go
```diff
@@ -1710,7 +1710,25 @@ func (self *TextExpr) EvaluateText(fieldToValue map[string]utils.CValueEnclosure
...
+			}
+			if isNull(valueExpr) {
+				return "Null", nil
+			}
+			return "Invalid", nil
+		}
+		if isNumber(valueExpr) {
+			return "Number", nil
+		}
+		return "String", nil
```
AndrewHess, line 1731: 1. I could be wrong, but I think if it fails to evaluate to a string, it will never succeed evaluating to a float
2. Don't you want the `isBoolean` and `IsNull` checks to be outside the `if err != nil` block?

venkateshamatam: 1. Yes, I think you're right but this comment on [L672](https://github.com/siglens/siglens/blob/3a413b9716e2a8365e8effb0f1d27643369fab9d/pkg/segment/structs/evaluationstructs.go#L672) says otherwise. However, I updated the logic to get the datatype from `utils.CValueEnclosure` instead of parsing it to find the datatype in case of field names like `eval res=typeof(city)`. If case the command has a raw value like `eval res=typeof(23) ` or `eval res =typeof("city")` then `CValueEnclosure` would have datatype as invalid and then we perform the parsing on the raw value to find the datatype.  Also, I have added testcases to `ingest.csv` for the strings and numbers, for null and boolean sigclient doesn't ingest these datatypes so I tested it manually. 
image

pkg/segment/structs/evaluationstructs.go
```diff
@@ -1710,7 +1710,44 @@ func (self *TextExpr) EvaluateText(fieldToValue map[string]utils.CValueEnclosure
...
+		var val utils.CValueEnclosure
+
+		if self.Val.NumericExpr == nil {
+			val.Dtype = utils.SS_INVALID
+		} else {
+			val, ok := fieldToValue[self.Val.NumericExpr.Value] // The parser always sets the fieldName in NumericExpr even if it's a string
+			if !ok {
+				val.Dtype = utils.SS_INVALID
+			}
+		}
```
AndrewHess, line 1722: I'm a little confused. The `Dtype` is `SS_INVALID` except for some cases when `NumericExpr != nil`, and in those cases wouldn't the `Dtype` be some numeric type? But later you have a `switch` that also checks for boolean and null types. When would those cases run?

pkg/segment/structs/evaluationstructs.go
```diff
@@ -1876,6 +1913,16 @@ func (self *ValueExpr) EvaluateValueExprAsString(fieldToValue map[string]utils.C
 	return str, nil
 }
 
+func isNumber(str string) bool {
+	_, err := strconv.ParseFloat(str, 64)
+	return err == nil
+}
```
Macbeth98, line 1919: I think it would be better to move this and the below functions into `utils` package, specifically to `pkg/segment/utils/segconsts.go` and make them public. This is the place where the above types are also defined. 

In that way in future we might use them at some other places.

pkg/segment/structs/evaluationstructs.go
venkateshamatam: The parser sets the fieldName in `NumericExpr` but the `Dtype` can be string or boolean or null types as well (Any of these [link](https://github.com/venkateshamatam/siglens/blob/21eb99b433a9808fa0553b155191fc6f1a6002e6/pkg/segment/utils/segconsts.go#L141)) The switch cases are executed when the query does not have a fieldName like `typeOf(city)` but it has a raw value like `typeOf(17)` or `typeOf("somestring")` then the `Dtype` would be set as invalid and we determine the datatype of the raw value in the query using the logic in the switch cases below.

AndrewHess (approved):

Issue #1196, opened 2024-06-27
[BUG] Alert detail page doesn't show full information for metrics alert type


bernatixer:
# Description
The alert details page doesn't show the full information for metrics type.

# Steps to Reproduce
1. Create a metrics alert
2. Go to the created alert details page

# Expected Behavior
The page shouls show all the details 

# Environment
- **SigLens Version:** 0.2.19
- Running on Linux

# Additional Context
Metric alert details page:
![image](https://github.com/siglens/siglens/assets/2202231/67420ee2-2538-4af7-8a4d-55a492927ddf)

Metric alert edit page:
![image](https://github.com/siglens/siglens/assets/2202231/57ba1fc1-603c-4ca0-9b49-ab3e9bfe6aab)

Example of a logs alert:
![image](https://github.com/siglens/siglens/assets/2202231/5692de08-9126-47cd-9fc2-9ab625b5c64c)

nkunal:
@Divyanshu11011 please add this to your list

Divyanshu11011:
@nkunal Will get Resolved !

PR #1200, opened 2024-06-28
fixed table view for metrics type alert , fixed label type for log type alert

Divyanshu11011:
# Description

Now table shows Required details for "Metrics" Alert type and also fixed the value format for label for "logs" type alert

Fixes [#1196 ](https://github.com/siglens/siglens/issues/1196)

image

image


Divyanshu11011:
I have fixed the word-wrap issue. If I make the first column narrower, it doesn't look good because there is a lot of space left on the right side. Should I still proceed with making it that way?
image

fixed word-wrap for longer queries :-
image

sonamgupta21:
> I have fixed the word-wrap issue. Making the first column narrower doesn't look good because there is a lot of space left on the right side. Should I still include that change with this pr? 

Second option is fine.

sonamgupta21 (approved):
Looks good.

PR #1201, opened 2024-06-28
Maintain Query Tab on Logs Screen Refresh SIG-1621

602dhruviii:
# Description
1. Saved queries will appear on the logs page with respect to the query preference they saved with —whether in the 'Builder' or 'Code' tab. 
2. Upon refreshing the page or navigating back to index.html for the logs page, the selected tab (either 'Builder' or 'Code') will be retained, ensuring continuity in query preferences

# Testing

# Checklist:


static/js/log-search.js
```diff
@@ -24,12 +24,67 @@ let indexValues = [];
...
+    // Call manageQueryModeCookie initially on document ready
+    manageQueryModeCookie();
+
+    // Call manageQueryModeCookie whenever the URL changes
+    $(window).on('hashchange', manageQueryModeCookie);
+    $(window).on('popstate', manageQueryModeCookie);
+
     let queryMode = Cookies.get('queryMode');
     if (queryMode !== undefined){
         if(queryMode === "Builder"){
```
sonamgupta21, line 90: We do not need to store it in session or remove the cookie. We can simply check if the filter tab is present and avoid triggering the tab in that case.

602dhruviii: I tried doing that but in that case, I am not able to get the original mode back, that's why I used session storage ...in this code, nothing is changing or creating problems to other tabs or URLs. I have thoroughly tested it by navigating to different tabs and trying out possible edge cases.

602dhruviii: Also, nothing is being changed here (neither session nor cookies),its just that I am removing cookies and getting it back as it would be stored in session when we are back on the logs tab.

sonamgupta21: Yes, I understand this works, but I think this can be done in a simpler way. 
Can you try 
```
    if (queryMode !== undefined){
        const searchParams = new URLSearchParams(window.location.search);

        // Check if the URL has the 'filterTab' parameter
        const hasFilterTab = searchParams.has('filterTab');

        if (!hasFilterTab){
           //If filter tab is not present then do trigger.
            if(queryMode === "Builder"){
                $('.custom-code-tab a:first').trigger('click');
            }else{
                $('.custom-code-tab a[href="#tabs-2"]').trigger('click');
            }
        }
        // Add active class to dropdown options based on the queryMode selected.
        if(queryMode === "Builder"){
            $(".query-mode-option").removeClass("active");
            $("#query-mode-options #mode-option-1").addClass("active");
            $('#query-mode-btn span').html("Builder");
        }else{
            $(".query-mode-option").removeClass("active");
            $("#query-mode-options #mode-option-2").addClass("active");
            $('#query-mode-btn span').html("Code");
        }
    }

sonamgupta21 (approved):

PR #1204, opened 2024-06-28
feat: Add load test for Alerting

Macbeth98:
# Description
- Alerting will create specified number of Alerts with 1 min eval interval and 1 min eval window. This ensures that every minute the specified number of Alert Evaluations will be done and alerts will be triggered.
- The script verifies that the Alerts are received per minute and logs the information.
- Clean exit.

# Testing
- Tested by running 10 Alerts.

# Checklist:


tools/sigclient/pkg/alerts/alerts_loadtest.go
```diff
@@ -0,0 +1,321 @@
...
+				if minuteData, exists := summary.MinuteData[minute]; exists {
+					minuteData.AverageDelay = minuteData.TotalDelay / time.Duration(minuteData.TotalAlertsReceived)
+					logMinuteSummary(minute, minuteData, numAlerts)
+					summary.MinuteData[minute] = minuteData
+				}
+			}
+		case <-timeout.C:
+			log.Errorf("Timed out waiting for alerts")
+			doneChan <- false
+			return
```
AndrewHess, line 188: nit: maybe rename `doneChan` to `successChan`, since I think you're actually sending the `success` state on it. The current version is a little confusing because here we timeout so we should be done but you send `false` on the channel

tools/sigclient/pkg/alerts/alerts_loadtest.go
```diff
@@ -0,0 +1,321 @@
...
+		}
+	}
+}
+
+func logMinuteSummary(minute int, minuteData MinuteSummaryData, numAlerts int) {
+	alertsCount := len(minuteData.Alerts)
+	log.Infof("Minute=%d,IsSummary=%v,UniqueAlertsCount=%d,Pass=%v,TotalAlertsCount=%v,AverageDelay=%v", minute, true, alertsCount, alertsCount >= numAlerts, minuteData.TotalAlertsReceived, minuteData.AverageDelay)
+	for title, data := range minuteData.Alerts {
+		averageDelay := data.TotalDelay / time.Duration(data.ReceivedCount)
+		log.Infof("Minute=%d,Alert=%s,ReceivedCount=%d,AverageDelay=%v", minute, title, data.ReceivedCount, averageDelay)
```
AndrewHess, line 213: nit: add a space after the commas in the strings, just so it's a little easier to read for someone looking at the output logs

Also for `AverageDelay` it would be good to indicate the units of time (or maybe that already gets printed?)

AndrewHess (commented):
Looks good, just some minor nits

tools/sigclient/pkg/alerts/alerts_loadtest.go
Macbeth98: Yes, I want to throw an Error, if we timed out. I will change it to `successChan`.

tools/sigclient/pkg/alerts/alerts_loadtest.go
Macbeth98: For `AverageDelay`, the unit is already printed. I did not added space specifically because, I am going to tunnel these logs to SigLens and parse them as `csv` through vector and then use `parse_key_value` so that they will be properly formatted into columns.

AndrewHess (approved):

PR #1208, opened 2024-07-01
Parser and Implementation of SPL Head with eval expressions

vasukalariya:
# Description
Summarize the change.
Head eval with condition

# Testing
Default ingest data
```
* | head 2

city=Boston | head limit=3

city=Boston | stats count, avg(latency) BY weekday, http_method | sort http_method | head

app_name = "Bracecould" | head gender="female"

app_name = "Bracecould" | head gender="female" keeplast=true

app_name = "Bracecould" | head gender="female" limit=2

app_name = "Bracecould" | head gender="female" AND gender="male" limit = 2

app_name = "Bracecould" | head gender="female" OR gender="male" limit = 2

app_name = "Bracecould" | head gender="female" OR gender="male" limit = 2 keeplast=true

app_name = "Bracecould" | sort http_status | head http_status < 400

app_name = "Bracecould" | sort http_status | head http_status < 400 keeplast=true

city=Boston | stats count as cnt, avg(latency) BY weekday, http_method | sort cnt | head cnt < 21

city=Boston | stats count as cnt, avg(latency) BY weekday, http_method | sort cnt | head cnt < 21 keeplast=true

city=Boston | stats count as cnt, avg(latency) BY weekday, http_method | sort http_method | head http_method="DELETE"

city=Boston | stats count as cnt, avg(latency) BY weekday, http_method | sort http_method | head http_method="DELETE" keeplast=true
```

Ingest new data for null checks
Below is the order of ingestion
```
{ "name" : "john", "age":"23" }'
{ "name" : "jack", "age":"24" }'
{ "name" : "jim", "age":null }'
{ "name" : "jill", "age":20 }'
{ "name" : "manny", "age":22 }'
```
```
* | head
* | head limit=3
* | head age < 25
* | head age < 25 limit=1
* | head age < 25 keeplast=true
* | head age < 25 null=true
* | head age < 24 null=true
* | head age < 24 null=true keeplast=true
```

# Checklist:


pkg/segment/aggregations/segaggs.go
```diff
@@ -196,26 +206,247 @@ func performAggOnResult(nodeResult *structs.NodeResult, agg *structs.QueryAggreg
...
+				} else if headExpr.Keeplast {
+					addRecordToHeadExpr(headExpr, rec, recordKey, hasSort)
+					headExpr.Done = true
+					break
+				} else {
+					headExpr.Done = true
+					break
+				}
+			} else {
+				return fmt.Errorf("processSegmentRecordsForHeadExpr: Error while evaluating expression, err: %v", err)
```
AndrewHess, line 322: On line 308 you make a new `err` variable, so on line 322 you're printing that new `err` instead of the `err` used on 307. I don't think that's what you intended

pkg/ast/spl/spl.peg
```diff
@@ -2552,8 +2557,127 @@ LenExpr <- str:QuotedString !(EVAL_CONCAT) {
...
+    queryAgg := &structs.QueryAggregators {
+        PipeCommandType: structs.OutputTransformType,
+        OutputTransforms: &structs.OutputTransforms {
+            HeadRequest: headExpr.(*structs.HeadExpr),
+        },
+    }
+
+    return queryAgg, nil
+}
+/ PIPE CMD_HEAD intAsStr:IntegerAsString {
```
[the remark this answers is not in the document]
AndrewHess: Nevermind. From the docs, what I suggested is not correct, so I think what you have is correct
https://docs.splunk.com/Documentation/Splunk/9.1.1/SearchReference/Head#Optional_arguments

pkg/ast/pipesearch/wsSearchHandler.go
```diff
@@ -374,7 +374,7 @@ func createRecsWsResp(qid uint64, sizeLimit uint64, searchPercent float64, scrol
 		}
 	case structs.RRCCmd:
 		useAnySegKey := false
-		if aggs.OutputTransforms != nil && aggs.OutputTransforms.MaxRows != 0 {
+		if aggs.OutputTransforms != nil && (aggs.OutputTransforms.HeadRequest != nil && (aggs.OutputTransforms.HeadRequest.BoolExpr == nil && aggs.OutputTransforms.HeadRequest.MaxRows != 0)) {
```
Macbeth98, line 377: I think can be refactored to 
```
if aggs.OutputTransforms != nil && aggs.HasHeadBlock() && aggs.OutputTransforms.HeadRequest.MaxRows != 0 {
  // If block code.
}
```

Since you already have that method `hasHeadBlock` make it public and you can it use it here.

pkg/ast/spl/tests/splParser_test.go
```diff
@@ -8240,3 +8188,432 @@ func performEventCountTest(t *testing.T, query string, expectedIndices []string,
...
+
+	astNode, aggregator, err := pipesearch.ParseQuery(string(query), 0, "Splunk QL")
+	assert.Nil(t, err)
+	assert.NotNil(t, astNode)
+	assert.NotNil(t, aggregator)
+	assert.Nil(t, aggregator.Next)
+
+	assert.Equal(t, aggregator.PipeCommandType, structs.OutputTransformType)
+	assert.NotNil(t, aggregator.OutputTransforms.HeadRequest)
+	assert.Equal(t, uint64(0), aggregator.OutputTransforms.HeadRequest.MaxRows)
```
Macbeth98, line 8251: When only condition is given, we won't apply the default limit of `10`?

pkg/segment/aggregations/segaggs.go
```diff
@@ -196,26 +206,247 @@ func performAggOnResult(nodeResult *structs.NodeResult, agg *structs.QueryAggreg
...
+				if err != nil {
+					return fmt.Errorf("performConditionalHeadOnHistogram: error while getting agg result fields values, err: %v", err)
+				}
+
+				// Evaluate the expression to a value.
+				result, err := headExpr.BoolExpr.Evaluate(fieldToValue)
+				if err != nil {
+					nullFields, err := headExpr.BoolExpr.GetNullFields(fieldToValue)
+					if err == nil && len(nullFields) > 0 {
+						// evaluation failed due to null fields
```
Macbeth98, line 230: Can there be a case where err is `nil` and `len(nullFields)` is also zero?

Then since the second condition is failed, it will go to else block and return with Error.

I believe it is better if you refactor it so that you check `err != null` condition first and return the error if it is not null and then check for `len(nullFields) == 0` and then return with appropriate message saying similar to `Null Fields is Empty`.

Also, generally it is always advised to check the `err != nil` check and then do the rest of the handling.

pkg/segment/aggregations/segaggs.go
```diff
@@ -196,26 +206,247 @@ func performAggOnResult(nodeResult *structs.NodeResult, agg *structs.QueryAggreg
...
+						if headExpr.Keeplast {
+							newResults = append(newResults, bucketResult)
+						}
+						headExpr.Done = true
+						break
+					}
+				}
+
+				if headExpr.MaxRows > 0 && headExpr.RowsAdded == headExpr.MaxRows {
+					headExpr.Done = true
```
Macbeth98, line 258: Where are you incrementing the `RowsAdded`? And why is it in the else condition only? Don't you need to handle this check for `NullFields` case also?

pkg/segment/structs/segstructs.go
```diff
@@ -543,9 +555,28 @@ func (qa *QueryAggregators) hasLetColumnsRequest() bool {
...
+		return false
+	}
+	if qa.OutputTransforms.HeadRequest.BoolExpr != nil {
+		return true
+	}
+	if qa.OutputTransforms.HeadRequest.MaxRows > qa.OutputTransforms.HeadRequest.RowsAdded {
+		return true
+	}
+	return false
+}
```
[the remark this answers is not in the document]
Macbeth98: I think he needs those conditions because `BoolExpr` can be nil but `MaxRows` can be greater than zero.
As you can just given `head 20` or just `head`.

Then BoolExpr` will be nil.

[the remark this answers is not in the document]
AndrewHess: You're right; I think I'm just confused what this function is supposed to do. With the name `hasHeadBlock()`, I would expect it to return `true` any time `OutputTransforms.HeadRequest` is not nil. I guess these other checks could be useful for ensuring the non-nil head block is valid though; in that case, why does it matter that `MaxRows > RowsAdded`?

pkg/ast/pipesearch/searchQueryParser.go
```diff
@@ -472,10 +472,10 @@ func parseColumnsCmd(node *structs.OutputTransforms, qid uint64) (*QueryAggregat
 		aggNode.OutputTransforms.FilterRows = node.FilterRows
 	}
 
-	aggNode.OutputTransforms.MaxRows = node.MaxRows
+	aggNode.OutputTransforms.HeadRequest = node.HeadRequest
```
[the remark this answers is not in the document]
vasukalariya: No, I did not find any other occurence except head that sets/uses maxrows.

pkg/segment/aggregations/segaggs.go
```diff
@@ -171,8 +175,14 @@ func performAggOnResult(nodeResult *structs.NodeResult, agg *structs.QueryAggreg
...
-		if agg.OutputTransforms.MaxRows > 0 {
-			err := performMaxRows(nodeResult, agg, agg.OutputTransforms.MaxRows, recs)
+		if agg.OutputTransforms.HeadRequest != nil {
+			headExpr := agg.OutputTransforms.HeadRequest
+			var err error
+			if headExpr.BoolExpr != nil {
+				err = performConditionalHead(nodeResult, headExpr, recs, recordIndexInFinal, numTotalSegments, finishesSegment, hasSort)
+			} else {
+				err = performMaxRows(nodeResult, headExpr, agg.OutputTransforms.HeadRequest.MaxRows, recs)
+			}
```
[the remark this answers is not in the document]
vasukalariya: Yes, that case is handled in processSegmentRecordsForHeadExpr.
```

if headExpr.MaxRows > 0 && headExpr.RowsAdded == headExpr.MaxRows {
	headExpr.Done = true
	break
}
```

pkg/segment/structs/segstructs.go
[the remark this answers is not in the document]
vasukalariya: This check was present already, my best guess is that once we have processed the required number of rows we would not want this aggregation to be processed even though it is present cause we have added the required number of rows.

pkg/segment/aggregations/segaggs.go
vasukalariya: If there is null the evaluation will fail and give an error. So null check is only required when we have thrown an error because we cannot evaluate due to null value.

pkg/ast/spl/spl.peg
```diff
@@ -2552,8 +2557,127 @@ LenExpr <- str:QuotedString !(EVAL_CONCAT) {
...
+NullOption <- "null" EQUAL boolVal:Boolean {
+    optionArg := &HeadOptionArgs{
+        argOption: "null",
+        headExpr: &structs.HeadExpr{
+            Null: boolVal.(bool),
+        },
+    }
+
+    return optionArg, nil
+}
```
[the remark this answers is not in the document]

pkg/ast/spl/spl.peg
```diff
@@ -2552,8 +2557,127 @@ LenExpr <- str:QuotedString !(EVAL_CONCAT) {
...
+        }
+    }
+
+    isBoolExpr, _ := mapOption["boolexpr"]
+
+    for option, exist := range mapOption {
+        if (option != "limit" && exist) && !isBoolExpr {
+            return nil, fmt.Errorf("Invalid Syntax %v can only be present with condition expression", option)
+        }
+    }
```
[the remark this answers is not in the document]

pkg/segment/aggregations/segaggs.go
```diff
@@ -196,26 +206,247 @@ func performAggOnResult(nodeResult *structs.NodeResult, agg *structs.QueryAggreg
...
+			return fmt.Errorf("processSegmentRecordsForHeadExpr: record %v not found in segment records", recordKey)
+		}
+
+		fieldToValue := make(map[string]segutils.CValueEnclosure, 0)
+		err := getRecordFieldValues(fieldToValue, fieldsInExpr, rec)
+		if err != nil {
+			return fmt.Errorf("processSegmentRecordsForHeadExpr: Error while retrieving values, err: %v", err)
+		}
+
+		result, err := headExpr.BoolExpr.Evaluate(fieldToValue)
```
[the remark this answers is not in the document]
vasukalariya: addressed in the latest commit

pkg/segment/aggregations/segaggs.go
vasukalariya: addressed in the latest commit

pkg/ast/pipesearch/wsSearchHandler.go
vasukalariya: this we cannot do since hasHeadBlock we can have a corner case like BoolExpr=true and maxrows is 3, in this case we would still need to continue processing the segments and not wait for search to be 100% finished (if we enter this if block it checks if search < 100 we break out).

pkg/ast/spl/tests/splParser_test.go
vasukalariya: I have not seen this limit for condition expression anywhere. Also, user might not know before hand how many records are there that evaluate to true for this condition. If they want the limit they can use the limit clause. The default limit for condition is not mentioned anywhere.

pkg/segment/aggregations/segaggs.go
vasukalariya: addressed in the latest commit. When err is nil and len(nullFields) == 0, it means that evaluation failed due to some other reason so we would want to return an error that we received before.

AndrewHess (approved):

PR #1212, opened 2024-07-01
feat: Added vector setup and config to tunnel logs to SigLens.

Macbeth98:
# Description
- The server will try to start the vector to tunnel the logs into SigLens.
- The server also handles the case of errors from vector and closing of the vector server at the end.

# Testing
- Tested by executing the scale test command with 10 Alerts.

# Checklist:

### DIFF
--- a/tools/sigclient/cmd/sigclient.go
+++ b/tools/sigclient/cmd/sigclient.go
@@ -179,7 +179,10 @@ var alertsLoadTestCmd = &cobra.Command{
 		}
 		if runVector > 1 {
 			runVector = 1
+		} else if runVector < -1 {
+			runVector = -1
 		}
+
 		alerts.RunAlertsLoadTest(destHost, numAlerts, runVector)
 	},
 }
--- a/tools/sigclient/cmd/sigclient.go
+++ b/tools/sigclient/cmd/sigclient.go
@@ -168,6 +168,7 @@ var alertsLoadTestCmd = &cobra.Command{
 	Run: func(cmd *cobra.Command, args []string) {
 		destHost, _ := cmd.Flags().GetString("dest")
 		numAlerts, _ := cmd.Flags().GetUint64("numAlerts")
+		runVector, _ := cmd.Flags().GetInt8("runVector")
 		log.Infof("destHost : %+v\n", destHost)
 		log.Infof("numAlerts : %+v\n", numAlerts)
 		if destHost == "" {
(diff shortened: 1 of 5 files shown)


tools/sigclient/pkg/alerts/alerts_loadtest.go
```diff
@@ -77,6 +78,83 @@ func setUpLoggingToFileAndStdOut() {
...
 
+	return logFile
+}
+
+// tunnel the logs into SigLens
+func startVector(host string, logFile string) (*exec.Cmd, chan error, error) {
+	os.Setenv("LOG_FILE_PATH", logFile)
+	os.Setenv("HOST", host)
+
+	dataDir := filepath.Join("./vector_log_lib/data")
```
AndrewHess, line 89: Where is this directory coming from?

tools/sigclient/pkg/alerts/alerts_loadtest.go
```diff
@@ -77,6 +78,83 @@ func setUpLoggingToFileAndStdOut() {
...
+		err := cmd.Wait()
+		errChan <- err
+		close(errChan)
+	}()
+
+	return cmd, errChan, nil
+}
+
+func listenForVectorErrors(vectorErrChan chan error, successChan chan bool, runVector int8) {
+	if runVector < 0 {
```
AndrewHess, line 129: nit: converting `runVector` to an enum would help make this more readable. Then you could do something like `if runVector == vectorModeDisabled {`

tools/sigclient/cmd/sigclient.go
```diff
@@ -326,6 +330,7 @@ func init() {
 	metricsBenchCmd.PersistentFlags().IntP("numQueryIterations", "n", 1, "Number of query iterations loop")
 
 	alertsLoadTestCmd.PersistentFlags().Uint64P("numAlerts", "n", 1, "Number of alerts to create")
+	alertsLoadTestCmd.PersistentFlags().Int8P("runVector", "v", 0, "Run vector: -1 - Explicitly Disable running of Vector, 0 - Optional State: Will try to run vector, but will not stop the test if encountered and issue with Vector, 1 - Explictly Enable running of Vector")
```
AndrewHess, line 333: Is there a reason we want to allow `runVector=0`?

tools/sigclient/pkg/alerts/alerts_loadtest.go
Macbeth98: So that is the directory, that vector uses internally to keep track of till where it has read the contents of a file or some other metadata information about a source. 

And I am just creating the folder in the root directory of SigClient. It will be created here:

image

tools/sigclient/cmd/sigclient.go
Macbeth98: Hmm no particular reason, but by default it will try to start the vector server to ingest these logs into SigLens. If it cannot, it won't stop the server and exit. 
But when enabled explicitly, it will stop the server and exit when vector throws an error.

tools/sigclient/pkg/alerts/alerts_loadtest.go
Macbeth98: Yes, I have updated the PR with this suggestion.

AndrewHess (approved):

PR #1216, opened 2024-07-02
Parser for SPL Bin command

vasukalariya:
# Description
Summarize the change.
Parser for SPL bin command

# Testing

# Checklist:


pkg/ast/spl/spl.peg
```diff
@@ -951,6 +956,301 @@ TcOptionCMD <- option:("usenull" / "useother" / "nullstr" / "otherstr") {
     return optionStr, nil
 }
 
+SpanLenOption <- number:(FloatAsString / IntegerAsString) timeScale:(TimeScale)? {
+    if timeScale != nil {
```
AndrewHess, line 960: nit: since this rule is specific to `bin` (it returns a `structs.BinSpanLength`), you should rename the rule to specify that (e.g., `BinSpanLenOption`)

pkg/ast/spl/spl.peg
```diff
@@ -951,6 +956,301 @@ TcOptionCMD <- option:("usenull" / "useother" / "nullstr" / "otherstr") {
...
+
+    num, err := strconv.ParseFloat(number.(string), 64)
+    if err != nil {
+        return nil, fmt.Errorf("SpanLength: Invalid num (%v): %v", number.(string), err)
+    }
+
+    spanLength := &structs.BinSpanLength {
+        Num: num,
+        TimeScale: utils.TimeUnit(0),
+    }
```
AndrewHess, line 997: Use enum names rather than their values. So do something like
```
TimeScale: utils.TMMicrosecond
```
That makes the code more readable and would also make it so we don't get bugs with this part of the code if we ever rearrange or renumber the enum values

pkg/ast/spl/spl.peg
```diff
@@ -951,6 +956,301 @@ TcOptionCMD <- option:("usenull" / "useother" / "nullstr" / "otherstr") {
     return optionStr, nil
 }
 
+SpanLenOption <- number:(FloatAsString / IntegerAsString) timeScale:(TimeScale)? {
+    if timeScale != nil {
+        num, err := strconv.ParseInt(number.(string), 10, 64)
+        if err != nil {
+            return nil, fmt.Errorf("SpanLength: Invalid num (%v): %v", number.(string), err)
```
AndrewHess, line 963: nit: in the log message change `num` to `number` so it's slightly more clear

pkg/ast/spl/spl.peg
```diff
@@ -951,6 +956,301 @@ TcOptionCMD <- option:("usenull" / "useother" / "nullstr" / "otherstr") {
...
+        binOptions: &structs.BinCmdOptions {
+            MaxBins: numBins,
+        },
+    }
+    return binOptionArgs, nil
+}
+
+PositiveIntegerAsString <- [0-9]+ {
+    return string(c.text), nil
+}
```
AndrewHess, line 1029: nit: move this near the `IntegerAsString` rule

pkg/ast/spl/spl.peg
```diff
@@ -951,6 +956,301 @@ TcOptionCMD <- option:("usenull" / "useother" / "nullstr" / "otherstr") {
...
+        base, err = strconv.ParseFloat(num2.(string), 64)
+        if err != nil || base <= 1.0 {
+            return nil, fmt.Errorf("Spl peg: Invalid base in log span, err: %v", err)
+        }
+    }
+    
+    if num1 != nil {
+        coeff, err = strconv.ParseFloat(num1.(string), 64)
+        if err != nil || coeff < 1.0 || coeff >= base {
+            return nil, fmt.Errorf("Spl peg: Invalid coefficient in log span, err: %v", err)
```
AndrewHess, line 1112: Split up these error cases so if one of them happens, it's more clear what the actual error is. For example, on 1104 if `base <= 1.0` the error we'd get is
```
Spl peg: Invalid base in log span, err: <nil>
```
which isn't very clear

pkg/ast/spl/spl.peg
```diff
@@ -951,6 +956,301 @@ TcOptionCMD <- option:("usenull" / "useother" / "nullstr" / "otherstr") {
...
+                binCmdOption.MaxBins = binOptArg.binOptions.MaxBins
+            case "start":
+                binCmdOption.Start = binOptArg.binOptions.Start
+            case "end":
+                binCmdOption.End = binOptArg.binOptions.End
+            case "aligntime":
+                binCmdOption.AlignTime = binOptArg.binOptions.AlignTime
+            default:
+                return nil, fmt.Errorf("Not a Valid Bin Command Option: %v", argOption)
+        }
```
AndrewHess, line 1198: I don't think you need the `mapOption`; the `default` case in your `switch` should catch that issue. If you have it to detect if an option is specified twice, you could instead have a set (a `map[string]struct{}`) of options that were set; that way you don't have to have the setup on lines 1153-1158

pkg/segment/structs/evaluationstructs.go
```diff
@@ -240,6 +240,26 @@ type TcOptions struct {
 	OtherStr   string
 }
 
+type BinCmdOptions struct {
+	BinSpanOptions *BinSpanOptions
+	MinSpan        *BinSpanLength
+	MaxBins        uint64
+	Start          *float64
+	End            *float64
+	AlignTime      *uint64
```
AndrewHess, line 249: Why are these pointers?

pkg/ast/spl/spl.peg

pkg/ast/spl/spl.peg
vasukalariya: Wouldn't this be dependent on the Microsecond. We were planning to use iota+1 for constants and 0 to specify if they are not set right? Or do we want this to be microsecond to denote not specified because siglens works on millisecond level?

pkg/ast/spl/spl.peg

pkg/ast/spl/spl.peg

pkg/ast/spl/spl.peg

pkg/ast/spl/spl.peg

pkg/segment/structs/evaluationstructs.go
vasukalariya: We would need a way to store start and end values and we cannot use default values i.e. 0 because the numerical values in the field could be 0 or negative. Pointers are just used as a concise way to represent if the value is set or not.

Macbeth98 (commented):
LGTM!

pkg/ast/spl/spl.peg
AndrewHess: That's a good point. In fact I didn't notice initially that you meant this to be an invalid `TimeUnit`. Currently, 0 is mapped to `TMMicrosecond`:
```
type TimeUnit uint8

const (
	TMMicrosecond TimeUnit = iota
	TMMillisecond
	TMCentisecond
	TMDecisecond
	TMSecond
	TMMinute
	TMHour
	TMDay
	TMWeek
	TMMonth
	TMQuarter
	TMYear
)
```
You could add something like `TMInvalid` as the first one. Then in your code:
```
TimeScale: utils.TMInvaild
```
Which would make your intent more clear

AndrewHess (approved):

PR #1218, opened 2024-07-02
Add mad_over_time promql

venkateshamatam:
# Description
Add mad_over_time promql
Fixed the promql parser to parse mad_over_time correctly.

# Testing
- Unit tests pass 
- Added CI CD test case

# Checklist:


AndrewHess (commented):
Looks pretty good so far.

For `evaluateMADOverTime()` I think you can keep what you have, I just have one question about the structure: it looks like you're iterating through each point, doing a search and iteration to get the previous points in the same window, and then computing the MAD on that list; is there a reason you're not instead keeping a list of points in the current time window, clearing that list whenever you enter a new time window, and adding the current point to that list? It seems like that would be less work, and (to me) more intuitive, but you can keep what you have

venkateshamatam:
@AndrewHess  The approach you suggested would reduce code complexity and eliminate the need for repeated searches and slicing operations as in the current method. I will look into refactoring this as part of an enhancement in a separate PR. Currently, I see the evaluated results in the result `ts` right after the `evaluateMADOverTime()` method, but they are missing in the final `ts` when returning the API response only for the `mad_over_time` command. I may be missing some flow, I'll check and add that, along with CICD test cases.

Macbeth98 (commented):
Looks Good!

AndrewHess (approved):

PR #1220, opened 2024-07-02
feat: remove automatic cleanup and added it as a command option

Macbeth98:
# Description
- Removed Auto Cleanup after a LoadTest when succeeded or failed due to an error.
- Added it as a optional command argument
- The test also will not exit when timeout occurs.

# Testing
- Tested it by running the command with 10 Alerts.

# Checklist:

### DIFF
--- a/tools/sigclient/pkg/alerts/alerts_loadtest.go
+++ b/tools/sigclient/pkg/alerts/alerts_loadtest.go
@@ -62,6 +62,7 @@ const (
 	VectorModeDisabled VectorMode = -1
 	VectorModeEnabled  VectorMode = 1
 	VectorModeOptional VectorMode = 0
+	LOG_MSG_DELIM      string     = "___"
 )
 
 func setUpLoggingToFileAndStdOut() string {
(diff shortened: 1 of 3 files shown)


tools/sigclient/pkg/alerts/alerts_loadtest.go
```diff
@@ -265,13 +266,11 @@ func trackNotifications(webhookChan chan alertutils.WebhookBody, numAlerts int,
...
-					summary.MinuteData[minute] = minuteData
+					delete(summary.MinuteData, minute)
 				}
 			}
 		case <-timeout.C:
-			log.Errorf("Timed out waiting for alerts")
-			successChan <- false
-			return
+			log.Errorf("%vCurrentMinute=%v,Timeout=%v,ErrorMessage=Timed out waiting for alerts", LOG_MSG_DELIM, currentMinute, true)
 		case <-exitChan:
```
AndrewHess, line 274: Do you want to also reset the timeout timer here? Or just log the error?

Macbeth98: Yeah that's a good Point. But I have actually executed and checked the scenario where it would throw this timeout and then I let the SigLens Server send out Alerts and then the Alerts started showing and the timeout got reset. So I stopped the SigLens server and then again the timeout error logs printed after two minutes.

So I think this is fine.

image

AndrewHess (approved):

PR #1230, opened 2024-07-03
Provide Add to Alert btn on metrics explorer

sunitakawane:
# Description
Summarize the change.

# Testing

# Checklist:


sonamgupta21 (commented):
For this case, the query is not being created correctly.

image

On Alerts - 
image

static/metrics-explorer.html
```diff
@@ -172,13 +172,17 @@ <h1 class="myOrg-heading">Metrics Explorer</h1>
                     </div>
                 </div>
                 <hr>
-                <div class="mt-4">
+                <div class="mt-4 d-flex">
+                    <div class="graph-view-container btn" id="alert-from-metrics-container">
+                        <button class="btn alert-from-metrics-btn below-btn-img" id="alert-from-metrics-btn" title="Create Alert"></button>
+                        <label class="label" style="cursor: pointer;">Create Alert</label>
```
sonamgupta21, line 178: Should we just add the bell icon to the right of the toggle button as we do on the logs screen?

sonamgupta21 (commented):
image

Get error when I try to save the alert.
I think it is opening in editing mode.

static/js/alert.js
```diff
@@ -470,6 +483,27 @@ async function displayAlert(res){
         fetchLogsPanelData(data,-1).then((res)=>{
             alertChart(res);
         });
+        $('#query').val(res.queryParams.queryText);
+        $(`.ranges .inner-range #${res.queryParams.startTime}`).addClass('active');
+        datePickerHandler(res.queryParams.startTime, res.queryParams.endTime, res.queryParams.startTime)
+    } else if (res.alert_type === 3){
```
sonamgupta21, line 489: Looks good.
nit: Instead of adding a third type (type 3), we should create a separate function to handle this, to avoid confusion later on.

We can keep the type as 2 and then create a separate function to display all the things, similar to what we have for logs, `createAlertFromLogs`.

sonamgupta21 (approved):

PR #1242, opened 2024-07-04
Added label for default tables

Divyanshu11011:
# Description
Added label for default tables

# Snapshot
image


static/js/dashboards-home.js
```diff
@@ -211,27 +211,35 @@ function createSiglensDashboard(inputdbname) {
...
         this.starIcon.style.backgroundImage = favoriteDBsSet.has(params.data.uniqId) ? starFilledURL : starOutlineURL;
 
-        //Disable delete for default dashboards
-		if (defaultDashboardIds.includes(params.data.uniqId)) {
-            this.dButton.disabled = true;
-            this.dButton.title = "Delete disabled";
-			this.dButton.classList.add('default-dashboard-delete'); 
+        //Disable delete for default dashboards and show "Default" label
+        if (defaultDashboardIds.includes(params.data.uniqId)) {
+            const defaultLabel = document.createElement('span');
```
sonamgupta21, line 233: 1. Add a class to the span, write the style in the CSS file.
2. Align the delete and tag elements to the right.
3. Remove the underline from the default text.
4. Set the background color to `#364fef`.

image

static/css/siglens.css
```diff
@@ -2468,6 +2468,22 @@ input[type="submit" i] {
...
+    background-color: #364fef;
+    padding: 2px 6px;
+    border-radius: 3px;
+    font-size: 12px;
+    margin-right: 6px;
+    text-decoration: none !important;
+}
+
+
+a {
```
sonamgupta21, line 2483: Looks good.
nit: Do not directly apply styles to HTML elements. Instead, give them a specific ID or class; otherwise, the styles will apply to all elements present.

sonamgupta21: https://github.com/siglens/siglens/assets/71771131/72ab7666-10e4-4ac9-9554-3248b758b6d7

When we hover over the delete icon it moves forward.

Divyanshu11011: The 'a' change was inconsistent, so I removed it. The issue with the moving thing was due to changes in margins, but I've fixed it now!

sonamgupta21 (approved):

PR #1246, opened 2024-07-07
Implementation of SPL Bin Command

vasukalariya:
# Description
Summarize the change.
Implemented SPL Bin Command

# Testing

```

app_name=Bracecould | stats count as cnt | bin span=5 cnt as bincnt

app_name=Bracecould | stats count as cnt by group | bin span=5 cnt as bincnt

app_name=Bracecould | bin span=5 http_status as bin_http | fields http_status, bin_http

app_name=Bracecould | bin span=5s http_status as bin_http | fields http_status, bin_http

app_name=Bracecould | bin span=3s aligntime=@d timestamp as bin_tstmp | fields timestamp, bin_tstmp

app_name=Bracecould | bin span=3s aligntime=@d timestamp

app_name=Bracecould | bin span=2log3 http_status

city=Boston | sort http_status | bin span=log1.2 http_status as htp | fields http_status, htp

city=Boston | sort http_status | bin span=log1.2 http_status as htp | fields http_status, htp | stats count by htp

city = Boston | bin bins=10 http_status | stats count by http_status

app_name=Bracecould | bin http_status | stats count by http_status

app_name=Bracecould | bin minspan=100 http_status | stats count by http_status

app_name=Bracecould | bin minspan=1000 http_status | stats count by http_status

* | bin http_status | stats count by http_status

* | bin minspan=20 http_status | stats count by http_status

* | bin bins=50000 minspan=2 http_status | stats count by http_status

app_name=Bracecould | bin span=1y timestamp

app_name=Bracecould | bin span=4mon timestamp as tmp | fields timestamp, tmp

app_name=Bracecould | bin span=4qtr timestamp as tmp | fields timestamp, tmp

app_name=Bracecould | bin bins=500 aligntime=1720046379000 timestamp as tmp | fields tmp

app_name=Bracecould | bin bins=500 aligntime=1720046381000 timestamp as tmp | fields tmp

app_name=Bracecould | bin bins=500 minspan=20s aligntime=1720046379000 timestamp as tmp | fields tmp

app_name=Bracecould | bin bins=500 minspan=1m aligntime=1720046381000 timestamp as tmp | fields tmp

app_name=Bracecould | bin bins=500 minspan=1mon aligntime=1720046381000 timestamp as tmp | fields tmp

* | sort http_status | bin http_status

* | tail 300 | bin http_status | stats count by http_status

```

# Checklist:


pkg/ast/spl/spl.peg
```diff
@@ -961,12 +961,19 @@ TcOptionCMD <- option:("usenull" / "useother" / "nullstr" / "otherstr") {
     return optionStr, nil
 }
 
-BinSpanLenOption <- number:(FloatAsString / IntegerAsString) timeScale:(TimeScale)? {
+BinTimeScale <- timeUnit:(Second / Month / Subseconds / Minute / Hour / Day / Week / Quarter / Year) {
```
Macbeth98, line 964: Did you refactored the `TimeScale` to `BinTimeScale`?

pkg/segment/aggregations/segaggs.go
```diff
@@ -2408,6 +2414,665 @@ func performValueColRequest(nodeResult *structs.NodeResult, aggs *structs.QueryA
 	return nil
 }
 
+func getNumericValue(v interface{}) (float64, bool) {
+	switch num := v.(type) {
+	case int:
```
Macbeth98, line 2419: If you just need float value, you can directly use this method:

```
ConvertToFloat(exp interface{}, bytes int)
```

in `dtypeutils.go`. And you can remove this method.

pkg/segment/aggregations/segaggs.go
```diff
@@ -2408,6 +2414,665 @@ func performValueColRequest(nodeResult *structs.NodeResult, aggs *structs.QueryA
...
+		err := performBinRequestOnMeasureResults(nodeResult, letColReq)
+		if err != nil {
+			return fmt.Errorf("performBinRequest: Error while performing bin request on measure results, err: %v", err)
+		}
+	}
+
+	return nil
+}
+
+func performBin(value float64, spanOptions *structs.BinSpanOptions, binReq *structs.BinCmdOptions) (interface{}, error) {
```
Macbeth98, line 2639: May be refactor this to `performBinWithSpanOptions`. Since you are checking specifically for `SpanOptions`.

Macbeth98 (commented):
@AndrewHess 
@vasukalariya in the error Logs, he is logging `recordKey` or `rowIndex`, whenever there is an error. For Example, check these:
1. https://github.com/vasukalariya/siglens/blob/bin_basic/pkg/segment/aggregations/segaggs.go#L2901
2. https://github.com/vasukalariya/siglens/blob/bin_basic/pkg/segment/aggregations/segaggs.go#L2768

But I feel like, these won't give us any proper information and might make the logs more clumsy, especially the `recordKey`. And the error that he is printing has the actual error message, that is informational. 

So do we want to keep them or remove them?

pkg/segment/aggregations/segaggs.go
```diff
@@ -2408,6 +2414,665 @@ func performValueColRequest(nodeResult *structs.NodeResult, aggs *structs.QueryA
...
+		fieldCValue, exist := fieldToValue[field]
+		if !exist {
+			return 0, fmt.Errorf("getFloatValForBin: field %s does not exist in record", field)
+		}
+		fieldValue = fieldCValue.CVal
+	}
+
+	fieldValueFloat, isNumeric := getNumericValue(fieldValue)
+	if !isNumeric {
+		return 0, fmt.Errorf("getFloatValForBin: field %s is not a numeric, err: %v", field, err)
```
Macbeth98, line 2475: Log the `fieldValue` also here.

pkg/segment/aggregations/segaggs.go
```diff
@@ -2408,6 +2414,665 @@ func performValueColRequest(nodeResult *structs.NodeResult, aggs *structs.QueryA
...
+	if num == 12 {
+		finalTime = time.Date(utcTime.Year(), 1, 1, 0, 0, 0, 0, time.UTC)
+	} else {
+		currMonth := int(utcTime.Month())
+		month := ((currMonth-1)/num)*num + 1
+		finalTime = time.Date(utcTime.Year(), time.Month(month), 1, 0, 0, 0, 0, time.UTC)
+	}
+
+	return uint64(finalTime.UnixMilli())
+}
```
AndrewHess, line 2707: This is a bit confusing. You should rename `num` to something more descriptive. I think it's the month, but it's not clear, and I'm not sure why you also need the UTC time. Also for `num` it's not clear what the range of valid inputs is (e.g. 1-12, or maybe something else)

pkg/segment/aggregations/segaggs.go
AndrewHess, line 1: It would also be good to add some unit tests for some of your functions (e.g., `performBinWithSpan` and `findBucketMonthQuarter`). It should be really quick/easy to add these unit tests since you designed them well so they're small and just do one thing

AndrewHess:
> @AndrewHess @vasukalariya in the error Logs, he is logging `recordKey` or `rowIndex`, whenever there is an error. For Example, check these:
> 
> 1. https://github.com/vasukalariya/siglens/blob/bin_basic/pkg/segment/aggregations/segaggs.go#L2901
> 2. https://github.com/vasukalariya/siglens/blob/bin_basic/pkg/segment/aggregations/segaggs.go#L2768
> 
> But I feel like, these won't give us any proper information and might make the logs more clumsy, especially the `recordKey`. And the error that he is printing has the actual error message, that is informational.
> 
> So do we want to keep them or remove them?

I think you're right that it probably won't add valuable debugging info, so it can get removed.

It might be a good idea to include the key-value pairs of the row in the error message though, provided this doesn't cause a flood of errors for some situations; I _think_ the `PostQueryBucketCleaning()` would just log and return the first error though, so it wouldn't be an issue

pkg/ast/spl/spl.peg
vasukalariya: TimeScale does not have Year in it. So had to create a similar one. I also cannot use the one from RelativeTimestamp because it does not have subseconds.

pkg/segment/aggregations/segaggs.go
```diff
@@ -2408,6 +2414,665 @@ func performValueColRequest(nodeResult *structs.NodeResult, aggs *structs.QueryA
...
+	fieldValueFloat, isNumeric := getNumericValue(fieldValue)
+	if !isNumeric {
+		return 0, fmt.Errorf("getFloatValForBin: field %s is not a numeric, err: %v", field, err)
+	}
+
+	return fieldValueFloat, nil
+}
+
+// Function to find the span range length
+func findSpan(minValue float64, maxValue float64, maxBins uint64, minSpan *structs.BinSpanLength, field string) *structs.BinSpanOptions {
```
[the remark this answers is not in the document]
vasukalariya: minSpan is something that the user passes. minValue is the minimum value in the data distribution of the field. minSpan refers to the minimum granularity of span that needs to be present to create bins. The span could be more that minSpan but should never be less than that.

pkg/segment/aggregations/segaggs.go
```diff
@@ -2408,6 +2414,665 @@ func performValueColRequest(nodeResult *structs.NodeResult, aggs *structs.QueryA
...
+			}
+		}
+	}
+
+	return &structs.BinSpanOptions{
+		BinSpanLength: &structs.BinSpanLength{
+			Num:       spanRange,
+			TimeScale: segutils.TMInvalid,
+		},
+	}
```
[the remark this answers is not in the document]
vasukalariya: Yeah it is correct to use Invalid because we are operating on numerical data and not timedata. If the field is timestamp we would get the estimated timespan from findEstimatedTimeSpan method in line 2484.

pkg/ast/spl/spl.peg
Macbeth98: got it. But I think you can still use the same one like this:

```
BinTimeScale <- timeUnit: (TimeScale / Year) 
```

And for RelTimeUnit

```
RelTimeUnit <- timeUnit: (TimeScale) !(Subseconds)
```

Or even better we can refactor this TimeUnit like this:

```
AllTimeScale <- timeUnit:(Second / Month / Subseconds / Minute / Hour / Day / Week / Quarter / Year) {
    return timeUnit, nil
}

TimeScale <- timeUnit:(AllTimeScale) !(Year) {
    return timeUnit, nil
}

BinTimeScale <- timeUnit:(AllTimeScale) {
    return timeUnit, nil
}

RelTimeUnit <- timeUnit:(AllTimeScale) !(Subseconds) {
    return strconv.Itoa(int(timeUnit.(utils.TimeUnit))), nil
}

```

So in this way `AllTimeScale` contains all the time unit and the other will specify what they do not want to match.

pkg/segment/aggregations/segaggs.go
vasukalariya: addressed in the latest commit.

pkg/segment/aggregations/segaggs.go
vasukalariya: Wouldn't this name be confusing, because we already have performBinWithSpan and perfrormBinWithSpanTime?

pkg/segment/aggregations/segaggs.go
```diff
@@ -2408,6 +2414,665 @@ func performValueColRequest(nodeResult *structs.NodeResult, aggs *structs.QueryA
...
+
+		var binValue interface{}
+		binValue, err = performBin(fieldValueFloat, letColReq.BinRequest.BinSpanOptions, letColReq.BinRequest)
+
+		if err != nil {
+			return fmt.Errorf("performBinRequestOnRawRecordWithSpan: Error while performing bin on record %v, err: %v", recordKey, err)
+		}
+
+		record[letColReq.NewColName] = binValue
+		finalCols[letColReq.NewColName] = true
```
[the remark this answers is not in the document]
vasukalariya: addressed in the latest commit.

pkg/segment/aggregations/segaggs.go
vasukalariya: addressed in the latest commit.

pkg/segment/aggregations/segaggs.go
```diff
@@ -2408,6 +2414,665 @@ func performValueColRequest(nodeResult *structs.NodeResult, aggs *structs.QueryA
...
+	case segutils.TMMinute:
+		return minSpan.Num * 60
+	case segutils.TMHour:
+		return minSpan.Num * 3600
+	case segutils.TMDay:
+		return minSpan.Num * 86400
+	default:
+		// default returning num*(seconds in a month)
+		return minSpan.Num * 2592000
+	}
```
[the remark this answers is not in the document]
vasukalariya: No, the default value should be months when timescale is larger than a day like week, month, quarter or year. I am just returning a larger number which would be give us the estimated span in months.

pkg/segment/aggregations/segaggs.go
```diff
@@ -2408,6 +2414,665 @@ func performValueColRequest(nodeResult *structs.NodeResult, aggs *structs.QueryA
...
+}
+
+// These time ranges are estimated based on different queries executed in splunk, no documentation is present
+func findEstimatedTimeSpan(minValue float64, maxValue float64, maxBins uint64, minSpan *structs.BinSpanLength) *structs.BinSpanOptions {
+	minSpanSecs := getSecsFromMinSpan(minSpan)
+	intervalSec := (maxValue/1000 - minValue/1000) / float64(maxBins)
```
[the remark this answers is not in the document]
vasukalariya: addressed in the latest commit.

pkg/segment/aggregations/segaggs.go
```diff
@@ -2408,6 +2414,665 @@ func performValueColRequest(nodeResult *structs.NodeResult, aggs *structs.QueryA
...
+	}
+
+	return nil, fmt.Errorf("performBin: BinSpanOptions is nil")
+}
+
+// This function either returns a float or a string
+func performBinWithSpan(value float64, spanOpt *structs.BinSpanOptions) (interface{}, error) {
+	if spanOpt.BinSpanLength != nil {
+		lowerBound, upperBound := getBinRange(value, spanOpt.BinSpanLength.Num)
+		if spanOpt.BinSpanLength.TimeScale == 0 {
```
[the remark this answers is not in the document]
vasukalariya: addressed in the latest commit.

pkg/segment/aggregations/segaggs.go
vasukalariya: addressed in the latest commit.

pkg/segment/aggregations/segaggs.go
vasukalariya: addressed in the latest commit.

pkg/ast/spl/spl.peg
AndrewHess: I think that's a good idea, but I'm not sure this will do what you want:
```
RelTimeUnit <- timeUnit: (TimeScale) !(Subseconds)
```
I _think_ that rule would match any `TimeScale` (including `Subseconds`) followed by anything that's not `Subseconds`—but I might be wrong

If that doesn't work, another option would be to use `AllTimeScale` everywhere and then inside the Go part of the rule check to ensure the `AllTimeScale` value is valid for the particular command you're implementing

pkg/segment/aggregations/segaggs.go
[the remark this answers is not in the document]
AndrewHess: Ok, then you should handle those cases explicitly like:
```
case segutils.TMWeek, segutils.TMMonth, segutils.TMQuarter, ...:
    // do what you did in the default case
default:
    // log and return an error
```
This will help us detect errors faster (e.g., what happens for `TMMicroseconds`? It probably shouldn't be treated the same as `TMMonth`, right?), and if we ever add a new enum value (e.g., `TMNanosecond`) we'd be able to detect the issue from the log messages

pkg/ast/spl/spl.peg
vasukalariya: addressed in the new commit.

pkg/segment/aggregations/segaggs.go
vasukalariya: addressed in the new commit.

pkg/segment/aggregations/segaggs.go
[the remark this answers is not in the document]
vasukalariya: addressed in the new commit.

pkg/ast/spl/spl.peg
Macbeth98: Did you change `TimeScale` also? I don't see it on the changes?

pkg/segment/aggregations/segaggs.go
```diff
@@ -2408,6 +2414,648 @@ func performValueColRequest(nodeResult *structs.NodeResult, aggs *structs.QueryA
...
+	default:
+		return 0, fmt.Errorf("getSecsFromMinSpan: Invalid time unit: %v", minSpan.TimeScale)
+	}
+}
+
+// These time ranges are estimated based on different queries executed in splunk, no documentation is present
+func findEstimatedTimeSpan(minValueMillis float64, maxValueMillis float64, maxBins uint64, minSpan *structs.BinSpanLength) (*structs.BinSpanOptions, error) {
+	minSpanSecs, err := getSecsFromMinSpan(minSpan)
+	if err != nil {
+		return nil, fmt.Errorf("findEstimatedTimeSpan: Error while getting seconds from minspan, err: %v", err)
```
AndrewHess, line 2534: You should also log these errors on lines 2526 and 2534

pkg/ast/spl/spl.peg
vasukalariya: Yes I have removed the TimeScale as it was used only at one place. I have added the check to throw error if timeUnit is year in the SpanLength rule that was using it. I have also added a parser test which verifies this.

AndrewHess (approved):
You have to fix some merge conflicts, but then I think this will be ready to merge

AndrewHess (approved):

PR #1247, opened 2024-07-08
feat: Send the query link through Alert Message for Logs

Macbeth98:
# Description
- Send the query link for that caused the Alert to Trigger in the Alert Notification Message.

# Testing
- Tested by sending an Alert Notification for Logs.

# Checklist:


pkg/alerts/alertsHandler/cronJobHandler.go
```diff
@@ -218,8 +221,31 @@ func handleAlertCondition(alertToEvaluate *alertutils.AlertDetails, isAlertCondi
...
 }
 
+func getLogsQueryLinkForTheAlert(alertDetails *alertutils.AlertDetails, timeRange *dtypeutils.TimeRange) string {
+	httpScheme := "http://"
+
+	if config.IsTlsEnabled() {
+		httpScheme = "https://"
+	}
+
+	baseURL := httpScheme + config.GetQueryHostname() + ":5122/index.html"
```
Macbeth98, line 231: My assumption of `:5122` might not be true for all the scenario's. Users can host SigLens through a proxy server and their query link can be just: `https://example.com/index.html?searchText=*&....`

How do we want to resolve this?

Macbeth98: Or do you think, the host name will be given as `example.com:5122`? If SigLens is being server through that port?

pkg/alerts/alertsHandler/cronJobHandler.go
```diff
@@ -218,8 +221,36 @@ func handleAlertCondition(alertToEvaluate *alertutils.AlertDetails, isAlertCondi
 	return nil
 }
 
+func getLogsQueryLinkForTheAlert(alertDetails *alertutils.AlertDetails, timeRange *dtypeutils.TimeRange) string {
+	if timeRange == nil {
+		log.Errorf("ALERTSERVICE: getLogsQueryLinkForTheAlert: TimeRange is nil. Alert=%+v", alertDetails.AlertName)
+		return ""
+	}
+
```
AndrewHess, line 229: Also check if `alertDetails == nil`

pkg/alerts/alertsHandler/cronJobHandler.go
```diff
@@ -218,8 +221,36 @@ func handleAlertCondition(alertToEvaluate *alertutils.AlertDetails, isAlertCondi
...
+	httpScheme := "http://"
+
+	if config.IsTlsEnabled() {
+		httpScheme = "https://"
+	}
+
+	baseURL := httpScheme + config.GetQueryHostname() + "/index.html"
```
AndrewHess, line 236: I think this can be just
```
config.GetQueryServerBaseUrl() + "/index.html"
```

pkg/config/config.go
```diff
@@ -715,6 +715,10 @@ func ExtractConfigData(yamlData []byte) (common.Configuration, error) {
 		config.S3IngestBufferSize = 1000
 	}
 
+	if config.QueryHostname == "" {
+		config.QueryHostname = fmt.Sprintf("localhost:%v", config.QueryPort)
+	}
```
AndrewHess, line 720: This is pretty similar to `GetQueryServerBaseUrl()`; can we just have one? Or maybe we should keep this but refactor `GetQueryServerBaseUrl()` a little since that's also checking if the hostname is the empty string. If needed, maybe `GetQueryServerBaseUrl()` should have flags for whether to include certain things (like `http/https` and the port)

pkg/alerts/alertsHandler/cronJobHandler.go
```diff
@@ -242,7 +273,9 @@ func evaluateLogAlert(alertToEvaluate *alertutils.AlertDetails, job gocron.Job)
 
 	isAlertConditionMatched := evaluateConditions(serResVal, &alertToEvaluate.Condition, alertToEvaluate.Value)
 
-	err = handleAlertCondition(alertToEvaluate, isAlertConditionMatched, "")
+	alertDataMessage := fmt.Sprintf("Check the logs for the Alert Query: %v", getLogsQueryLinkForTheAlert(alertToEvaluate, timeRange))
```
AndrewHess, line 276: nit: `Check the logs for the Alert Query:` is a little unclear to me, but you can keep it if you want. I think it's because to me `Check the logs` sounds like you're telling me to check the logs file rather than to run the query on the UI. Maybe something like `Run the Alert Query` or `View the query results` would be better

Macbeth98: I think I will change it to: `View the Query Results`.

pkg/config/config.go
Macbeth98: I kept this if check and assigning default. So that `GetQueryHostname()` will return a valid value. And I refactored the `GetQueryServerBaseUrl` and `GetUIDomain` functions to not check for empty `hostname`.

AndrewHess (approved):

PR #1250, opened 2024-07-09
Bug: SPL Eval subcommand replace

vasukalariya:
# Description
Summarize the change.
This function can take in a string literal or a field value. Here we are just assuming user will pass a field, when a string literal is passed the program crashes due to segfault because self.Val does not have a NumericExpr.

# Testing
app_name = Bracecould | eval masked_email=replace("xyz@abc.com", "^([^@]+)@", "xxxxx@") | fields email, masked_email

app_name = Bracecould | eval email = "abc@xyz.com" | fields email | eval masked_email=replace(email, "^([^@]+)@", "xxxxx@") | fields email, masked_email

# Checklist:


AndrewHess (commented):
Looks good. You should also add some tests to makes sure we never break this again. Adding to ingest.csv and restart.csv should be sufficient, but feel free to add other unit tests if you think it's necessary to properly test this

AndrewHess (approved):

PR #1260, opened 2024-07-11
feat: fixed Alert queries for all types of Logs queries

Macbeth98:
# Description
- Fixed the Alert logs Query so that:
   - It checks for all of the series in a `Group By` request.
   -  And now it will also work for queries with nested Group By.

**NOTE:** As of now I am rejecting the query with Multiple Measure Columns even with `Eval` or `where` condition set. But the queries like this with Multiple Columns will still work, if the they are followed by a stats query at the end like below.
```
* | rex field=app_version "(?<major>\d+)\.(?<minor>\d).*" 
| stats count(*) as count, sum(major) as sum_major by minor 
| where minor < sum_major 
| stats count(*) as count
```

- This PR also includes the building and sending the Metrics Query Link while sending an Alert Notification.

# Testing
- Tested by executing different types of Logs Queries and verified that the Alert Notification is being sent.
- Tested the metrics query link by creating an Alert for Metric Queries and verified that the URL received in the Notification by decoding the url using online URL-decoders.

# Checklist:

### DIFF
--- a/pkg/segment/structs/segstructs.go
+++ b/pkg/segment/structs/segstructs.go
@@ -461,6 +461,14 @@ func (ma *MeasureAggregator) String() string {
 	return ma.StrEnc
 }
 
+func GetMeasureAggregatorStrEncColumns(measureAggs []*MeasureAggregator) []string {
+	var columns []string
+	for _, ma := range measureAggs {
+		columns = append(columns, ma.String())
+	}
+	return columns
+}
+
 func (ss *SegStats) Merge(other *SegStats) {
 	ss.Count += other.Count
 	ss.Records = append(ss.Records, other.Records...)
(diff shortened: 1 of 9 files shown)


pkg/segment/structs/segstructs.go
```diff
@@ -461,6 +461,14 @@ func (ma *MeasureAggregator) String() string {
 	return ma.StrEnc
 }
 
+func GetMeasureAggregatorStrEncColumns(measureAggs []*MeasureAggregator) []string {
+	var columns []string
+	for _, ma := range measureAggs {
+		columns = append(columns, ma.String())
+	}
+	return columns
+}
```
AndrewHess, line 470: No need to change this. I just want to point out that this could be a good type of function to make generic: one that takes a list of some type and a function on that type to output another type. Maybe some other code could also use a similar function

```
func convertSlice[T, R](slice []T, func(T) R) []R {....}
```

AndrewHess (approved):
Looks good

pkg/segment/structs/segstructs.go
Macbeth98: yes got it.

PR #1269, opened 2024-07-12
Parser for SPL Streamstats

vasukalariya:
# Description
Summarize the change.
Parser for streamstats

# Testing
Added parser tests

# Checklist:


pkg/ast/spl/spl.peg
```diff
@@ -671,6 +690,185 @@ GroupbyBlock <- BY fields:FieldNameList {
...
+
+    return aggNode, nil
+}
+/ PIPE CMD_STREAMSTATS aggs:(CommonAggregatorBlock) {
+    aggNode := aggs.(*structs.QueryAggregators)
+
+    aggNode.StreamStatsOptions = initializeStreamStatsOptions()
+
+    return aggNode, nil
+}
```
AndrewHess, line 870: nit: do you need both of these sections, or can you delete 864-870 and change the `ssOptionList` on line 857 to be optional and then you just need that block?

pkg/ast/spl/spl.peg
```diff
@@ -671,6 +690,185 @@ GroupbyBlock <- BY fields:FieldNameList {
...
+            separator := restSlice[i - 1].([]any)
+            ssOptArg = separator[1].(*StreamStatsOptionArgs)
+        }
+
+        argOption := ssOptArg.argOption
+        val, exist := optionWasSpecified[argOption]
+        if exist && val {
+            return nil, fmt.Errorf("Invalid Syntax, option: %v is already specified", argOption)
+        }
+        optionWasSpecified[argOption] = true
```
AndrewHess, line 830: You don't have to change this. I just want to point out that you have very similar code for option lists for other SPL commands, so it could be good at some point to make a general function to do this logic. That would reduce duplicate code and make it a little easier for you to add similar parts for handling other SPL commands since you'd just have to call a function

AndrewHess (commented):
Looks pretty good

pkg/ast/spl/spl.peg
```diff
@@ -671,6 +690,185 @@ GroupbyBlock <- BY fields:FieldNameList {
...
+
+StreamStatsBlock <- PIPE CMD_STREAMSTATS ssOptionList:(StreamStatsOptionList) SPACE aggs:(CommonAggregatorBlock) {
+    aggNode := aggs.(*structs.QueryAggregators)
+
+    aggNode.StreamStatsOptions = ssOptionList.(*structs.StreamStatsOptions)
+
+    return aggNode, nil
+}
+/ PIPE CMD_STREAMSTATS aggs:(CommonAggregatorBlock) {
+    aggNode := aggs.(*structs.QueryAggregators)
```
Macbeth98, line 865: Do you need this as a separate branch? You can write in the above main one like this right? 

```
StreamStatsBlock <- PIPE CMD_STREAMSTATS ssOptionList:(StreamStatsOptionList)? SPACE aggs:(CommonAggregatorBlock) {
    aggNode := aggs.(*structs.QueryAggregators)
   
    if ssOptionList != nil { 
          aggNode.StreamStatsOptions = ssOptionList.(*structs.StreamStatsOptions)
    } else {
        aggNode.StreamStatsOptions = initializeStreamStatsOptions()
     }

    return aggNode, nil
}
```

pkg/ast/spl/spl.peg
```diff
@@ -556,7 +568,7 @@ FieldSelectBlock <- PIPE CMD_FIELDS op:("-" / "+")? EMPTY_OR_SPACE fields:FieldN
     return queryAggregator, nil
 }
 
-AggregatorBlock <- PIPE CMD_STATS aggs:AggregationList byFields:(GroupbyBlock)? options:(StatsOptions)?{
+CommonAggregatorBlock <- aggs:AggregationList byFields:(GroupbyBlock)? {
```
Macbeth98, line 571: I think it is better to keep this as `AggregatorBlock` and then rename the `stats` one as `StatsAggregatorBlock` or just `StatsBlock`.

pkg/ast/spl/spl.peg
vasukalariya: yes if we use this and user does not use streamstats option. They will need to enter 2 spaces.
`... | streamstats  count as cnt by method`. The only reason why I split this into 2 parts is because of space.

pkg/ast/spl/spl.peg
AndrewHess: I think the point of this was to share some common code between splunk commands that do something like `<agg_function>(<field1>) by <field2>`, so I think what Vasu has is good

pkg/ast/spl/spl.peg
AndrewHess: See https://github.com/siglens/siglens/pull/1269#discussion_r1678275562

AndrewHess (approved):

PR #1272, opened 2024-07-15
Add support for SPL mvindex,mvjoin and mvcount, mvjoin

venkateshamatam:
# Description
Add support for SPL mvindex,mvjoin and mvcount

# Testing
For the below ingested multi-value fields 
image

tested various queries - 
mvindex: 
`* | makemv delim="," name | eval n=mvindex(name,2)`
`* | makemv delim="," name | eval n= mvindex(name,0,3)`
`* | makemv delim="," name | eval n= mvindex(name,-1)`
`* | makemv delim="," name | eval n= mvindex(name,-3, -1)`
`* | makemv delim=";" name | eval n= mvindex(name, -1)`
mvjoin: 
`*  | makemv delim=";" name | eval n= mvjoin(name, "-")`
`*  | makemv delim="," name | eval n= mvjoin(name, "/")`
mvcount: 
`*  | makemv delim="," name | eval n= mvcount(name)`
`*  | makemv delim=";" name | eval n= mvcount(name)`

# Checklist:


pkg/segment/structs/evaluationstructs.go
```diff
@@ -1782,20 +1782,149 @@ func (self *TextExpr) EvaluateText(fieldToValue map[string]utils.CValueEnclosure
...
+			return "", fmt.Errorf("TextExpr.EvaluateText: mvField.CVal is nil")
+		}
+
+		mvString, ok := mvField.CVal.(string)
+		if !ok {
+			return "", fmt.Errorf("TextExpr.EvaluateText: expected mvField.CVal to be a string, got %T", mvField.CVal)
+		}
+
+		mvStringTrimmed := strings.Trim(mvString, "[]")
+		mvSlice := strings.Split(mvStringTrimmed, " ")
```
AndrewHess, line 1910: A lot of this is shared between `mvindex`, `mvjoin`, and `mvcount` (and probably others in the future). You should extract this into a help function so there's a lot less duplicate code

pkg/segment/structs/evaluationstructs.go
```diff
@@ -1782,20 +1782,149 @@ func (self *TextExpr) EvaluateText(fieldToValue map[string]utils.CValueEnclosure
...
+
+		count := len(mvSlice)
+
+		// Handle edge cases
+		if count == 1 && mvSlice[0] == "<nil>" {
+			// If the field has no values or explicitly marked as "<nil>", return NULL (or an equivalent)
+			return "", nil // Assuming an empty string represents NULL
+		}
```
AndrewHess, line 1918: fyi: This is another indication we should probably change our representation of a multivalue value

AndrewHess (commented):
You should also add tests to ingest.csv and restart.csv

pkg/segment/structs/evaluationstructs.go
```diff
@@ -1784,20 +1815,81 @@ func (self *TextExpr) EvaluateText(fieldToValue map[string]utils.CValueEnclosure
...
+		regex := self.Regex
+
+		for index, value := range mvSlice {
+			if regex.GetCompiledRegex().MatchString(value) {
+				return strconv.Itoa(index), nil
+			}
+		}
+
+		// If no match is found
+		return "NULL", nil
```
AndrewHess, line 1886: Does Splunk return the string `"NULL"` here, or some other representation of NULL? Will our later processing correctly handle the string `"NULL"` (e.g., if we pipe this into the `isnull()` SPL command)?

Same comment for line 1847

AndrewHess (commented):
Looking good

pkg/segment/structs/evaluationstructs.go
venkateshamatam: I just checked on Splunk. Splunk doesn't return the string `"NULL"`; instead, it returns an empty value in the field. The `isNull` function works fine on Splunk, but it doesn't work as expected in our case because our implementation of `isNull` checks for `DType` `Backfill` and not a string representation of `null`. Since this implementation is in EvaluateText, we cannot return `null` or set the `DType` to `Backfill`, as it will be overridden before it goes to the processing of `isNull` and it will be processed as the string that is returned by the `EvaluateText` method. How do you think we should handle this? Should we return an empty string and update `isNull` to consider empty strings as `null` values? or return the string `"NULL"` and update `isNull` to check for `backfill` or string `"NULL"` to determine if the value is null.

AndrewHess: I think there's two ok options:
1. As you suggested, return an empty string and update our code to treat that the same as null
2. Create a new generic type like `Option[T]` that will either store a value or indicate that there's no value there. (Similar to `Option` in Rust, if you're familiar). Then to return a NULL string we'd just return a `Option[string]` with no string value.

Option 1 (and similar) will be faster to implement, but leaves you with an small bug that you can never really remove until you do Option 2 (or similar).

I think that unless you really want to do Option 2, you can do Option 1 for now

venkateshamatam: Sounds good, Andrew. I looked into the implementation of Option 2, but I believe it requires updating the return type and the flow currently implemented. We can adopt this flow and use it for all other commands that can return `null` so that `isnull` can recognize them as `null`. This approach is not limited to `mvindex`; it applies to other commands as well where we return `null`. We'll need to update those to follow the same approach as Option 2. 

For now, I have pushed a commit for Option 1.

pkg/segment/structs/evaluationstructs.go
```diff
@@ -1784,20 +1819,86 @@ func (self *TextExpr) EvaluateText(fieldToValue map[string]utils.CValueEnclosure
...
+		}
+
+		for index, value := range mvSlice {
+			if compiledRegex.MatchString(value) {
+				return strconv.Itoa(index), nil
+			}
+		}
+
+		// If no match is found
+		return "NULL", nil
```
AndrewHess, line 1895:
```suggestion
		return "", nil
```

AndrewHess (approved):

PR #1279, opened 2024-07-16
fix: Close Webhook Requests After sending an Alert Notification

Macbeth98:
# Description
- Added closing of webhook request, once an Alert Notification is sent.
- Without closing the request causes the number of file descriptors to keep increasing and eventually it crosses the `os` limits and will not send the requests.

# Testing
- Tested before making this fix by running the alerts load-test client and the error of sending request would occur approximately around 54th minute for 1000 Alerts and also verified that the number of open files kept increasing over time. 
- Tested after making the fix by running the alerts load-test client and it has been running without any issues for the past 170 minutes for 2000 Alerts. I have verified that the number of open files also stayed consistent with the value that was at the start of SigLens server and before starting the client.

# Checklist:

### DIFF
--- a/pkg/alerts/alertsHandler/notificationHandler.go
+++ b/pkg/alerts/alertsHandler/notificationHandler.go
@@ -184,10 +184,11 @@ func sendWebhooks(webhookUrl, subject, message string, alertDataMessage string, 
 
 	r.Header.Add("Content-Type", "application/json")
 	client := &http.Client{}
-	_, err1 := client.Do(r)
-	if err1 != nil {
+	resp, err := client.Do(r)
+	if err != nil {
 		log.Errorf("sendWebhooks: Error sending request. WebhookURL=%v, Error=%v", webhookUrl, err)
 	}
+	resp.Body.Close()
 	return err
 }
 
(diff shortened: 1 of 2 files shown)


pkg/alerts/alertsHandler/notificationHandler.go
```diff
@@ -184,11 +184,13 @@ func sendWebhooks(webhookUrl, subject, message string, alertDataMessage string,
...
-	_, err1 := client.Do(r)
-	if err1 != nil {
+	resp, err := client.Do(r)
+	if err != nil {
 		log.Errorf("sendWebhooks: Error sending request. WebhookURL=%v, Error=%v", webhookUrl, err)
+		return err
 	}
-	return err
+	resp.Body.Close()
+	return nil
```
AndrewHess, line 193: Looks good except for one thing. On lines 181-183, when we get an error we log it but don't return it. With your changes, if we got an error in that block and then got no error on line 187, we'd log an error (line 182), but still return a `nil` error (line 193).

I'm not sure there's a reason we avoided returning an error just after line 182; I think you can return the error there

Macbeth98: Yes, I just return the `err` after logging it.

AndrewHess (approved):

PR #1293, opened 2024-07-18
fix: building of PromQL Metrics query from formulas

Macbeth98:
# Description
- Fixes the issue of building queries through formulas, for fomula's like this: `atan(abs(ceil(123 * a)))`

# Testing
- Tested by executing queries through postman.
- Tested by updating the unit test case.

# Checklist:


Macbeth98:
> Looks good.
> 
> One question: when we replace a variable with it's expression, do we also have to wrap that expression in parentheses? I'm not sure if a case where this would matter is valid for promql, but if you have something like:
> 
> ```
> a: 5 + 2
> b: 3 + 7
> ```
> 
> and you want to do `a * b`, we'd want
> 
> ```
> (5 + 2) * (3 + 7)
> ```
> 
> rather than
> 
> ```
> 5 + 2 * 3 + 7
> ```

Yes, I guess, if we allow for Multiple queries insider a single query then this would become an issue. But since we allow `a` to have only one query, I do not think this will be an issue. And when writing the formula the user should add parentheses based on how they want the query to be executed.

Issue #646, opened 2024-03-19
[BUG] Visualization tab doesn't update


AndrewHess:
# Description
When you run a query that produces a visualization and then go to the visualization tab and then run a different query, you see the visualization you were already seeing.

# Steps to Reproduce
1. Ingest some test data
2. Run this query on the UI: `* | timechart avg(latency) by http_method span=1h`
3. Beneath the query, there's two tabs: Table, and Visualization. Select the Visualization tab and you should see a bar graph with 6 bars.
4. Stay on Visualization and run `* | timechart avg(latency) span=1h`

Notice that the bar graph remains unchanged, but is incorrect. If you go to the Table tab and then back to Visualization, you'll see the correct graph, which has just one bar. Also you can run it with a query that doesn't have a visualization, like `* | http_status=20*`. If you're on the Visualization tab already, it doesn't get updated. But once you go to Table and then back to Visualization, you correctly see the message "This query does not contain visualization data"

# Expected Behavior
Clearly describe what you expect to happen.

# Environment
- Commit Hash: 557458f

# (Optional) Additional Context
Add anything else we should know about this problem.

sharma2910:
Hi, Can i work on this ?

nbhavana:
go ahead

sharma2910:
@nbhavana I have fixed the issue and created the following PR - https://github.com/siglens/siglens/pull/679

PR #1298, opened 2024-07-18
Visualization Tab - Display Bar Graph for Stats Query on Logs Screen & Errors related to Log-Screens

### LINKED ISSUE #1045: [STORY] Visualization Tab - Display Bar Graph for Stats Query on Logs Screen
# Description
When running a stats query on the logs screen within the visualization tab, the expected outcome is to display the results in a bar graph format similar to what we show on dashboards when we run a stats query and choose chart type bar chart.

# Steps to Reproduce
1. Navigate to the logs screen.
2. Open the visualization tab.
3. Run a stats query. Ex- `batch="batch-1" | stats avg(latency) BY city` 

# Expected Behavior
The stats query results should be displayed as a bar graph within the visualization tab.

Divyanshu11011:
# Description
1) Added Bar Graph for Stats Query on Logs Screen.
2) Support multilple Stats Query .
3) Supports Multiple Group By Values.
4) Different Timestamps for timestamp / category data
5) Fixed error related to logs screen and Chart Not Displaying for First Query until Refreshed
6) Graph Auto-Adjusts As Per Mangnitude
7) Made the chart responsive

Fixes #1045 #646 

# Snapshots

https://github.com/user-attachments/assets/f091c53c-bf12-46c4-9a01-285028eb5f73

image

Multiple Group-By


AndrewHess:
I think it would be good to sort the x-axis values. I tested with
```
* | stats count by weekday, group
```
and the bars seem to be in a random order.

However, the normal `Table` tab also has them in what looks like a random order, so maybe sorting them belongs in a different PR that would do it for both the `Table` and `Visualization` tabs. What do you think @sonamgupta21 @Divyanshu11011 ?

static/js/search.js
```diff
@@ -941,17 +942,27 @@ function parseInterval(interval) {
 }
 
 function timeChart() {
-    if (isTimechart) {
```
sonamgupta21, line 944: If I run a query that supports a bar graph and then run another query that does not, the chart should go away and show the prompt `This query doesn't support visualization.`

Don't remove this code, we already have a condition check for the time chart query, so you can add the logic to check if it is an aggregation query.

sonamgupta21 (approved):

PR #1301, opened 2024-07-19
Allow Raw Metrics Query For Alerts and Dashboards

sonamgupta21:
# Description
Summarize the change.

# Testing

# Checklist:


sunitakawane:
Once raw query is executed, it does not change even if we change the query builder query. Is that the correct behaviour?

https://github.com/user-attachments/assets/b6f076dc-3072-4f54-b895-0f75abc190c4

sonamgupta21:
> Once raw query is executed, it does not change even if we change the query builder query. Is that the correct behaviour?

Yes, this is how we wanted it. If a user searches a raw query, we want to maintain both states separately: builder and raw.

sunitakawane (approved):

sunitakawane (approved):

PR #1306, opened 2024-07-20
Fix #1030: Linearly interpolate percentile results

### LINKED ISSUE #1030: [STORY] Linearly interpolate percentile results
# Description
We have a function that finds a percentile for a given slice:
```
func FindPercentileData(arr []uint64, percentile int) uint64 {
    ...
}
```
But it always returns an element from the input slice. Make it so if the desired percentile falls between two elements, it linearly interpolates between those to values. For example, if we want the 50-th percentile and it falls a quarter way between the elements with values 10 and 20, this function should return 12.5 (that is, `10 + (20 - 10) * 0.25`)

# (Optional) Testing
There's already unit tests for `FindPercentileData()`. Update them.

# Acceptance Criteria
- The function linearly interpolates
- You've added/updated unit tests and they all pass 
- The function runtime remains `O(n)`

Axiself:
# Description
 - FindPercentileData function no longer returns a uint64, and instead a float64
 - When a non-integer index is found with a given percentile, it now calculates the weight of the index after and before it and returns that, instead of rounding down the index
 - Fixed minor calculation error when calculating index: ``float64(percentile*(len(arr)-1)) / float64(100)`` instead of ``float64(percentile*(len(arr))) / float64(100)``
 - Changed expected test results to match new behavior and added one new test
 - Changed struct RedMetrics' P50, P90, P95 and P99 variable types to match new FindPercentileData return type

Fixes #1030 

# Testing
I compared the results given by the tests with percentile calculation of the arrays with outside tools (such as online calculators).

# Checklist:

### DIFF
--- a/pkg/segment/tracing/structs/tracestruct.go
+++ b/pkg/segment/tracing/structs/tracestruct.go
@@ -62,10 +62,10 @@ type Span struct {
 type RedMetrics struct {
 	Rate      float64 // Number of entry spans divided by 60 seconds
 	ErrorRate float64 // Percentage of entry spans that errored
-	P50       uint64  // p50, p90, p95, p99 latencies are calculated by the list of durations.
-	P90       uint64
-	P95       uint64
-	P99       uint64
+	P50       float64 // p50, p90, p95, p99 latencies are calculated by the list of durations.
+	P90       float64
+	P95       float64
+	P99       float64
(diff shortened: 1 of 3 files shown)


AndrewHess (approved):
Looks good, thanks for contributing!

PR #1307, opened 2024-07-20
Fix #1041: User-created dashboard with default dashboard name

### LINKED ISSUE #1041: [BUG] Duplicate Dashboard Names Allowed Between Default and User-Created Dashboards
# Description
Currently, we check for duplicate names within user-created dashboards when creating or updating a dashboard. However, this check does not extend to the names of default dashboards. This allows for the creation of a user-created dashboard with the same name as a default dashboard, resulting in two dashboards with the same names.

# Steps to Reproduce
1. Navigate to the dashboards page.
2. Attempt to create a new dashboard, using the same name as an existing default dashboard.
3. Observe that the new dashboard is created successfully, despite the name conflict with the default dashboard.

# Expected Behavior
We should prevent the creation of dashboards with duplicate names, regardless of whether the existing dashboard is a default or user-created one. The name check should be extended to include both default and user-created dashboards.

Literalmentnerd:
# Description
 - Added a check in createDashboard that sees if the name of the dashboard we are trying to create is a default dashboard 
 name
 - Added a simple test to see if the fix is working

Fixes #1041 

# Testing
Made a simple test in dashboard_test.go that checks first if it got an error from createDashboard and if it does, checks if its the intended one for this case (Trying to create a dashboard with a Default dashboard name).

# Checklist:


pkg/dashboards/dashboards.go
```diff
@@ -226,12 +226,25 @@ func dashboardNameExists(dname string, orgid uint64) bool {
 	return exists
 }
 
+var defaultDashboardNames = map[string]bool{
+	"Siglens Query DB":     true,
+	"Siglens Ingestion DB": true,
+	"Siglens Data DB":      true,
+	"Sample Dashboard":     true,
+}
```
AndrewHess, line 234: It would be good to remove this duplication. As I'm sure you know, these names are defined in allids.json:
```
{
    "10329b95-47a8-48df-8b1d-0a0a01ec6c42": "Siglens Ingestion DB",
    "a28f485c-4747-4024-bb6b-d230f101f852": "Siglens Query DB",
    "bd74f11e-26c8-4827-bf65-c0b464e1f2a4": "Siglens Data DB",
    "53cb3dde-fd78-4253-808c-18e4077ef0f1": "Sample Dashboard"
}
```
With the changes you have, it's fairly likely that later we would add/rename/delete a default dashboard and just update allids.json but forget to update this map you've defined, and then we'll get basically the bug you're fixing now.

I think a better approach would be that on startup (so we only have to do it once), you parse allids.json to get the names of the default dashboards and store those. That way if we add/rename/delete a default dashboard in a later release, we'll only have to update one thing instead of two things.

Side note: it looks like you're using this `map` as a `set`. Golang doesn't have a `set`, but you can use `map[string]struct{}` like a set. This has the benefit of being a little more clear for your use case—it also uses slightly less memory, but that doesn't matter for this particular case.

AndrewHess (commented):
This looks correct, but you can improve it so we avoid this bug in the future as well.

Also your JS changes only remove a newline at the end of the file. You should revert those changes.

pkg/dashboards/dashboards.go
Literalmentnerd: So I should do the parsing in the startup.go file and not on dashboard.go?

AndrewHess: No, the function to setup this map/set should be in dashboards.go. To get it to run on startup, you could either:
- Make it a public function, and call it inside startup.go
- Call it from `func init() { ... }` in dashboards.go. Golang automatically runs all `init()` functions on startup, as long as the `main` package includes the specified package (in this case `dashboards`). So as long as `startup.go` includes `dashboards` either directly or indirectly, the `init()` inside `dashboards` will get run.

Literalmentnerd: I have been testing my changes and when I test it manually, and by this I mean opening localhost and trying to create a dashboard with a default name by clicking in the New Dashboard Button, it seems to be working as intended as it does brings the error message and does not create the dashboard. But whenever i run make test,  the Test_Dashboard_storage_methods fails due to the InitDashboards function returning (level=error msg="InitDashboard: Failed to read default dashboards file, err=open defaultDBs/allids.json: no such file or directory") and i cant seem to find a fix for it, i dont think its because of the path since both dashboards.go and dashboards_test.go are in the same folder
image_2024-07-22_191426292

AndrewHess: I think the tests get run from within the package. You can test this with something like:
```
pwd := os.Getenv("PWD")
log.Errorf("(testing) InitDashboard: PWD=%v", pwd)
```
One option is that for the unit test you could write the file (in the directory the tests are run in) before calling `InitDashboards()`, and then remove the file at the end of the test.

Alternatively, maybe you can find a good way to read from the real `defaultDBs/allids.json` file

Literalmentnerd:
Ok I think I was able to make the changes requested

Literalmentnerd:
For some reason one of the pipelines seem to be stuck, even though the program runs in my computer and running make passes all tests

pkg/dashboards/dashboard_test.go
```diff
@@ -28,10 +28,38 @@ import (
 
 func Test_dashboard_storage_methods(t *testing.T) {
 
+	//Create the defaultDB directory
+	defaultDBPath := "defaultDBs"
+	e := os.Mkdir(defaultDBPath, 0755)
+	if e != nil {
+		t.Fatalf("Failed to create defaultDB directory: %v", e)
+	}
```
AndrewHess, line 36: nit: use `err` instead of `e`, since we use `err` throughout the codebase as the error type value

pkg/dashboards/dashboard_test.go
```diff
@@ -28,10 +28,38 @@ import (
...
+		"10329b95-47a8-48df-8b1d-0a0a01ec6c42": "Siglens Ingestion DB",
+		"a28f485c-4747-4024-bb6b-d230f101f852": "Siglens Query DB",
+		"bd74f11e-26c8-4827-bf65-c0b464e1f2a4": "Siglens Data DB",
+		"53cb3dde-fd78-4253-808c-18e4077ef0f1": "Sample Dashboard"
+	}`
+	filePath := defaultDBPath + "/allids.json"
+	e = os.WriteFile(filePath, []byte(jsonContent), 0644)
+	if e != nil {
+		t.Fatalf("Failed to write JSON content to file: %v", e)
+	}
```
AndrewHess, line 49: Also use `err` instead of `e` here

pkg/dashboards/dashboard_test.go
```diff
@@ -174,6 +202,25 @@ func Test_parseUpdateDashboardRequest(t *testing.T) {
 	assert.Equal(t, "mydashboard", details["note"])
 }
 
+// TestCreateDashboardWithDefaultName tests that creating a dashboard with a default name is not allowed
+func TestCreateDashboardWithDefaultName(t *testing.T) {
```
AndrewHess, line 206: Just to be consistent with our other test function naming, this should be:
```suggestion
func Test_CreateDashboardWithDefaultName(t *testing.T) {
```

static/js/common.js
```diff
@@ -1063,4 +1063,4 @@ function updateQueryModeUI(queryMode) {
         $('#query-mode-options #mode-option-2').addClass('active');
         $('#query-mode-btn span').html('Code');
     }
-}
+}
```
AndrewHess, line 1066: Revert this change

pkg/dashboards/dashboards.go
```diff
@@ -36,6 +36,7 @@ import (
 )
 
 var defaultDashboardIds map[string]struct{}
+var defaultDashboardNames map[string]bool
```
AndrewHess, line 39: Since you're only ever storing `true` as the value in your new map, you should change `bool` to `struct{}`.

pkg/dashboards/dashboards.go
AndrewHess, line 1: Also use `err` in this file instead of `e` for an `error` variable

static/js/common.js
AndrewHess: Actually, this may be from an automatic formatter that's part of our PR pipeline. If that's the case you can ignore my comment and keep this change. But if this is not from the automatic formatting, you should revert it

AndrewHess:
> For some reason one of the pipelines seem to be stuck, even though the program runs in my computer and running make passes all tests

Yeah it looks like the job never got started. I'm not sure why, and I'm not finding an easy way to manually run it for a PR from a fork. But when you commit and push again it should trigger the job again

Literalmentnerd:
Ok changed what you asked except of the bool to struct as it was breaking the code and i dont know how to fix it hope its not super important or a bother

Literalmentnerd:
Nvm was able to fix it easily my bad

AndrewHess (approved):
Looks good!

PR #1317, opened 2024-07-23
Add parsing and implementation for relative_time command and implemen…

venkateshamatam:
# Description
https://docs.splunk.com/Documentation/SCS/current/SearchReference/DateandTimeFunctions#relative_time.28.26lt.3Btime.26gt.3B.2C.26lt.3Bspecifier.26gt.3B.29

# Testing
- Test cases pass.

# Checklist:


pkg/segment/structs/evaluationstructs.go
```diff
@@ -167,10 +167,17 @@ type NumericExpr struct {
...
+	Right        *NumericExpr
+	Val          *StringExpr
+	RelativeTime RelativeTimeExpr
+}
+
+type RelativeTimeExpr struct {
+	Snap     string
+	Offset   int64
+	TimeUnit utils.TimeUnit
 }
```
AndrewHess, line 181: I think this has exactly the same data as `RelativeTimeModifier`. Can we use that instead, or does it make sense to define this new type you have?

pkg/segment/structs/evaluationstructs.go
```diff
@@ -1607,13 +1616,56 @@ func (self *NumericExpr) Evaluate(fieldToValue map[string]utils.CValueEnclosure)
...
+			var err error
+			// 1. Direct epoch input (e.g., relative_time(1234567890, "-1d")) is stored in self.Value.
+			// 2. Epoch from another command (e.g., relative_time(now(), "-1d")) after evaluation is stored in left, leaving self.Value as an empty string.
+			if self.Value != "" {
+				epochTime, err = strconv.ParseInt(self.Value, 10, 64)
+				if err != nil {
+					return 0, fmt.Errorf("NumericExpr.Evaluate: failed to parse epoch time from Value %v, err: %v", self.Value, err)
+				}
+			} else if left >= 0 {
+				epochTime = int64(left)
```
AndrewHess, line 1634: Would it make sense to always store it in the left so we don't have to have this logic. Or will that complicate things so doing it this way is better?

pkg/segment/structs/evaluationstructs.go
```diff
@@ -1607,13 +1616,56 @@ func (self *NumericExpr) Evaluate(fieldToValue map[string]utils.CValueEnclosure)
...
 
+			return float64(relTime), nil
 		default:
 			return 0, fmt.Errorf("NumericExpr.Evaluate: unexpected operation: %v", self.Op)
 		}
 	}
 }
 
+func CalculateAdjustedTimeForRelativeTimeCommand(timeModifier RelativeTimeExpr, currTime time.Time) (int64, error) {
+	var err error
```
AndrewHess, line 1652: Maybe this function should be moved into a util package. What do you think?

AndrewHess:
You should also add end-to-end tests, but probably use a hardcoded unix epoch for it because using `now()` could cause problems

pkg/ast/spl/tests/splParser_test.go
```diff
@@ -4180,7 +4180,9 @@ func Test_evalFunctionsRelativeTime(t *testing.T) {
 	assert.Equal(t, aggregator.Next.Next.OutputTransforms.LetColumns.ValueColRequest.NumericExpr.Op, "relative_time")
 	assert.Equal(t, aggregator.Next.Next.OutputTransforms.LetColumns.ValueColRequest.NumericExpr.Left.Op, "now")
 	assert.True(t, aggregator.Next.Next.OutputTransforms.LetColumns.ValueColRequest.NumericExpr.Left.IsTerminal)
-	assert.Equal(t, aggregator.Next.Next.OutputTransforms.LetColumns.ValueColRequest.NumericExpr.Val.RawString, "-1d@d")
+	assert.Equal(t, aggregator.Next.Next.OutputTransforms.LetColumns.ValueColRequest.NumericExpr.RelativeTime.Offset, int64(-1))
+	assert.Equal(t, aggregator.Next.Next.OutputTransforms.LetColumns.ValueColRequest.NumericExpr.RelativeTime.TimeUnit, utils.TMDay)
+	assert.Equal(t, aggregator.Next.Next.OutputTransforms.LetColumns.ValueColRequest.NumericExpr.RelativeTime.Snap, "8")
```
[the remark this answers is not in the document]
venkateshamatam: Snap Value `"8"` here refers to the the constants defined in the [TimeUnit type](https://github.com/venkateshamatam/siglens/blob/2c52ce0ccbb5f46d90fa8ee2c74be83a80002a3a/pkg/segment/utils/timeconsts.go#L28) and these are then used in [ApplySnap](https://github.com/venkateshamatam/siglens/blob/2c52ce0ccbb5f46d90fa8ee2c74be83a80002a3a/pkg/segment/utils/timeconsts.go#L133). Snap value `"8"` corresponds to `TMDay` in the above test case.

pkg/segment/structs/evaluationstructs.go
venkateshamatam: The value stored in `left` comes from nested commands like `now()` or `time()`, and the current parsing of the `relative_time` command does not handle this. Do you recommend that we also store the parsed time from the `relative_time` command in `left` in this case? I thought it might be better if we store the evaluated value from a command like `now()` or `time()` separately from the passed value in the query, such as epoch `1234567890`. If we update it to be stored in `left`, we would also have to test that an evaluated command used in the query in a different way does not override the parsed value from `relative_time` command.

pkg/segment/structs/evaluationstructs.go
venkateshamatam: Yes, `RelativeTimeExpr` contains the same data as `RelativeTimeModifier` from the `ast` package. However, the time modifiers are used as part of the initial search and may have more options when we support chained time modifiers in the future. The new `RelativeTimeExpr` type is specifically for the `relative_time` command and is not related to Time Modifiers.

I thought having separate types for each would make it easier to manage if we add more options to either type in the future. `RelativeTimeModifier` is used in the initial pipe search, while `RelativeTimeExpr` is used for evaluation.

pkg/segment/structs/evaluationstructs.go
venkateshamatam: Yes Andrew, I moved it to `utils` package.

pkg/ast/spl/tests/splParser_test.go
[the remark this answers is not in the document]
AndrewHess: Ok. Then you should use `string(TMDay)` instead, since it's much more clear that way, and won't break if we change the actual value of `TMDay` later

pkg/segment/structs/evaluationstructs.go
AndrewHess: I'm not sure why we'd want to store an exact value like `1234567890` in a different place from something like `now()` that gets evaluated to `1234567890`. Why do you think it would be good to store those separately? To me it seems like we'd want to store them both similarly so we could do something like `self.Left.Evaluate()` and get `1234567890`, whether that was a direct value specified by the user or the result of running `now()`

Also I don't understand what you're saying here:
>  If we update it to be stored in left, we would also have to test that an evaluated command used in the query in a different way does not override the parsed value from relative_time command.

venkateshamatam: Andrew, it looks like both exact value and evaluated values were stored in `left`. `self.Value` would have the exact value only, so we can ignore the `self.Value` and just use the values from `left`, since `left` would have the value irrespective of whether it is an exact value or an evaluated value. I have updated the logic to use only the value from `left` in all cases and added test case using `now()` and `time()` as well.

pkg/ast/spl/tests/splParser_test.go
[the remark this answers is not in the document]
venkateshamatam: Done, I just pushed it.

AndrewHess (approved):

PR #1310, opened 2024-07-22
feat: Add Parser and Implementation for SPL Command: fillnull

Macbeth98:
# Description
- Add Parser and Implementation for SPL command: `fillnull`
- There is an issue with discrepancy in number of columns for async queries. I have mentioned the issue clearly in the slack.

# Testing
- Added unit test cases for parser.
- And verified that the queries work.

# Checklist:

- [x] I have self-reviewed this PR.
- [x] I have removed all print-debugging and commented-out code that should not be merged.
- [x] I have added sufficient comments in my code, particularly in hard-to-understand areas.
- [x] I have formatted the code, if applicable. For Go, I have run `goimports -w .`.

PR #1318, opened 2024-07-23
feat: Implemented Fix for SPL: fillNull command for discrepancy in Columns

Macbeth98:
# Description
- This PR fixes the issue of discrepancy in number of columns for async queries that is mentioned in the PR: #1310 
- Implemented this by fetching all the columns (rotated and unrotated) by the given time range and applied any column transforms that are applicable, so that new columns won't be added.

# Testing
- Tested by running various queries through the UI.

# Checklist:


pkg/segment/structs/segstructs.go
```diff
@@ -378,33 +379,34 @@ type QueryCount struct {
...
+	RecsAggsBlockResults        interface{}              // Evaluates to *blockresults.BlockResults
+	RecsAggsColumnKeysMap       map[string][]interface{} // map of column name to column keys for GroupBy Recs
+	RecsAggsProcessedSegments   uint64
+	RecsRunningSegStats         []*SegStats
+	TransactionEventRecords     map[string]map[string]interface{}
+	TransactionsProcessed       map[string]map[string]interface{}
+	ColumnsOrder                map[string]int
+	RawSearchFinished           bool
+	CurrentSearchResultCount    int
+	AllSearchColumnsByTimeRange map[string]bool
```
AndrewHess, line 409: nit: this should be a `map[string]struct{}` because the `bool` is never `false`, right?

pkg/segment/structs/segstructs.go
```diff
@@ -835,6 +837,44 @@ func (qa *QueryAggregators) HasRexBlockInChainWithStats() bool {
...
+
+func (qa *QueryAggregators) HasFillNullExpr() bool {
+	if qa != nil && qa.OutputTransforms != nil && qa.OutputTransforms.LetColumns != nil && qa.OutputTransforms.LetColumns.FillNullRequest != nil {
+		return true
+	}
+	return false
+}
+
+func (qa *QueryAggregators) HasFillNullExprInChain() bool {
+	if qa == nil {
```
AndrewHess, line 854: You don't have to change anything here for this PR. But we have a lot of these `Has_blank_InChain()` functions with the same exact structure. At some point we could make a higher-level of this and then for the implementation of this we could do just
```
func (qa *QueryAggregators) HasFillNullExprInChain() bool {
    return qa.hasInChain(qa.HasFillNullExpr)
}

File: pkg/segment/tracing/utils/lineartimefinding_test.go [hunk 3]
```diff
@@ -54,8 +55,12 @@ func TestFindPercentileData(t *testing.T) {
...
 	}
 
+	precision := 8
+	scale := math.Pow10(precision)
+
 	for _, tc := range testCases {
 		result := FindPercentileData(tc.arr, tc.percentile)
+		result = math.Round(result*scale) / scale
 		if result != tc.expected {
 			t.Errorf("Expected %d percentile to be %f, but got %f", tc.percentile, tc.expected, result)
```
AndrewHess, line 65: Just curious: was this test flaky for you?

Macbeth98: Yes it was, so the `result` is `29.040000000000003` but the `tc.expected` is `29.04`. So I had to set the precision.

pkg/segment/structs/segstructs.go
Macbeth98: Probably we should create this as a internal JIRA issue or can we create this as a Github issue?

pkg/segment/structs/segstructs.go
Macbeth98: Yes, that might be case but I want to keep the type same as `finalCols`, so that I can use it in functions like: `applyColNameTransform` in `rrcreader.go` and probably in future with some other functions where I want to send this value instead of `finalCols`

AndrewHess: Ok you can keep this is if you want, or change it if you want

AndrewHess (approved):

pkg/segment/structs/segstructs.go
AndrewHess: I can make a GitHub Issue for it

PR #1324, opened 2024-07-24
Fix: Corrected the SPL Eval + Stats

vasukalariya:
# Description
Summarize the change.
Corrected the eval implementation for stats. Added support for multiple fields and value expressions for eval in stats.

# Testing
Added tests in ingest.csv

# Checklist:


pkg/segment/aggregations/evalaggs.go
```diff
@@ -23,297 +23,543 @@ import (
...
+			if err != nil {
+				return currResult, fmt.Errorf("PerformEvalAggForCount: there are some errors in the eval function that is inside the count function: %v", err)
+			}
+			if boolResult {
+				finalValue++
+			}
+		} else {
+			// return count as 1 if expression is not boolean
+			finalValue = 1
 		}
```
AndrewHess, line 419: I don't understand this. Why is it not `finalValue++` on both 414 and 418?

pkg/segment/aggregations/evalaggs.go
```diff
@@ -323,93 +569,114 @@ func ComputeAggEvalForAvg(measureAgg *structs.MeasureAggregator, sstMap map[stri
 	return nil
 }
 
-func ComputeAggEvalForCardinality(measureAgg *structs.MeasureAggregator, sstMap map[string]*structs.SegStats, measureResults map[string]utils.CValueEnclosure, runningEvalStats map[string]interface{}) error {
+// Always pass valid strSet when using this function
+func PerformAggEvalForCardinality(measureAgg *structs.MeasureAggregator, strSet map[string]struct{}, fieldToValue map[string]utils.CValueEnclosure) (float64, error) {
```
AndrewHess, line 573: > // Always pass valid strSet when using this function

What do you mean by `valid`? Do you mean anything that's not nil, or something more specific?

pkg/segment/aggregations/evalaggs.go
```diff
@@ -452,38 +719,26 @@ func AddMeasureAggInRunningStatsForCount(m *structs.MeasureAggregator, allConver
 	return idx, nil
 }
 
-func AddMeasureAggInRunningStatsForAvg(m *structs.MeasureAggregator, allConvertedMeasureOps *[]*structs.MeasureAggregator, allReverseIndex *[]int, colToIdx map[string][]int, idx int) (int, error) {
-
+func SetupMeasureAgg(m *structs.MeasureAggregator, allConvertedMeasureOps *[]*structs.MeasureAggregator, measureFunc utils.AggregateFunctions, allReverseIndex *[]int, colToIdx map[string][]int, idx int) (int, error) {
```
AndrewHess, line 722: nit: rename `m` to something more clear

pkg/segment/aggregations/evalaggs.go
```diff
@@ -23,297 +23,543 @@ import (
...
-		measureResults[measureAgg.String()] = enclosure
+func PopulateFieldToValueFromSegStats(fields []string, measureAgg *structs.MeasureAggregator, sstMap map[string]*structs.SegStats, fieldToValue map[string]utils.CValueEnclosure, i int) error {
+	for _, field := range fields {
+		sst, ok := sstMap[field]
+		if !ok {
+			return fmt.Errorf("PopulateFieldToValueFromSegStats: applyAggOpOnSegments sstMap was nil for aggCol %v", measureAgg.MeasureCol)
+		}
+
+		if i >= len(sst.Records) {
+			return fmt.Errorf("PopulateFieldToValueFromSegStats: Incorrect length of field: %v for aggCol: %v", field, measureAgg.String())
```
AndrewHess, line 296: For the `aggCol`, why do you specify it as `measureAgg.MeasureCol` in one log but as `measureAgg.String()` in another?

AndrewHess (commented):
I still need to review the files in `pkg/segment/results`

pkg/segment/aggregations/evalaggs.go
```diff
@@ -23,297 +23,543 @@ import (
...
+		if measureAgg.ValueColRequest.BooleanExpr != nil {
+			boolResult, err := measureAgg.ValueColRequest.BooleanExpr.Evaluate(fieldToValue)
+			if err != nil {
+				return currRangeStat, fmt.Errorf("PerformEvalAggForRange: there are some errors in the eval function that is inside the range function: %v", err)
+			}
+			if boolResult {
+				finalRangeStat.Max = 1
+				finalRangeStat.Min = 1
+			}
+		} else {
```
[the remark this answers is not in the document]
vasukalariya: Yes, if booleanExpr evaluates to false the record is dropped and won't be processed and thus, you will never see a 0 as a result.

pkg/segment/aggregations/evalaggs.go
```diff
@@ -23,297 +23,543 @@ import (
...
+				return currResult, fmt.Errorf("PerformEvalAggForMinOrMax: there are some errors in the eval function that is inside the %v function: %v", measureAgg.MeasureFunc, err)
+			}
+			if boolResult {
+				finalResult.Dtype = utils.SS_DT_FLOAT
+				finalResult.CVal = float64(1)
+				return finalResult, nil
+			} else {
+				// return current result when no value needs to be updated
+				return currResult, nil
+			}
```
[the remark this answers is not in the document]
vasukalariya: Yes, if booleanExpr evaluates to false the record is dropped and won't be processed and thus, you will never see a 0 as a result.

pkg/segment/aggregations/evalaggs.go
```diff
@@ -23,297 +23,543 @@ import (
...
-	if isMin {
-		edgeValue = math.MaxFloat64
+			if !exists {
+				if isNumeric {
+					finalResult.Dtype = utils.SS_DT_FLOAT
+					finalResult.CVal = floatValue
+				} else {
+					finalResult.Dtype = utils.SS_DT_STRING
+					finalResult.CVal = strValue
+				}
```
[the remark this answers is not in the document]
vasukalariya: exists is just a parameter to denote whether a value exists or not. Basically, it says whether currResult should be considered or not.

pkg/segment/aggregations/evalaggs.go
```diff
@@ -23,297 +23,543 @@ import (
...
-	if (isMin && eValFloat > edgeValue) || (!isMin && eValFloat < edgeValue) {
-		enclosure.CVal = edgeValue
-		measureResults[measureAgg.String()] = enclosure
+	if finalRangeStat.Min > currRangeStat.Min {
+		finalRangeStat.Min = currRangeStat.Min
 	}
-	return nil
+	if finalRangeStat.Max < currRangeStat.Max {
+		finalRangeStat.Max = currRangeStat.Max
+	}
```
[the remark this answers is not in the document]
vasukalariya: Ideally, we would only like to extend the range by comparing these relative values, because say if currRangeStat is not set then we could overwrite the finalRangeStat.Max with maxFloat. So, I think this is safe and is more readable.

pkg/segment/aggregations/evalaggs.go
vasukalariya: Yes, I mean that is not nil. Will update it.

pkg/segment/aggregations/evalaggs.go
vasukalariya: I used the existing comment. They technically refer to the same thing i.e. the field's name used in the measure function. I will update it to make it consistent.

pkg/segment/aggregations/evalaggs.go
vasukalariya: You can use either as this function would set the final value to 1 in two cases when a bool expression evaluates to true or just defaults to 1 if it is not a boolean expression. finalValue++ is equivalent to finalValue = 1 in this scenario.

pkg/segment/aggregations/evalaggs.go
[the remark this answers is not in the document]
vasukalariya: addressed in the latest commit.

pkg/segment/aggregations/evalaggs.go
vasukalariya: addressed in the latest commit.

pkg/segment/aggregations/evalaggs.go
vasukalariya: addressed in the latest commit.

pkg/segment/aggregations/evalaggs.go
vasukalariya: addressed in the latest commit.

pkg/segment/results/segresults/segresults.go
```diff
@@ -238,10 +238,6 @@ func (sr *SearchResults) UpdateSegmentStats(sstMap map[string]*structs.SegStats,
 	sr.updateLock.Lock()
 	defer sr.updateLock.Unlock()
 	for idx, measureAgg := range measureOps {
-		if len(sstMap) == 0 {
-			continue
-		}
```
AndrewHess, line 243: Why don't we want to continue here?

pkg/segment/aggregations/evalaggs.go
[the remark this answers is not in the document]
AndrewHess: Ok I think renaming it would help clarify. Maybe something like `currResultExists`

pkg/segment/results/segresults/segresults.go
vasukalariya: I forgot to revert this change. Sorry for the oversight. I have addressed this in the latest commit.

pkg/segment/aggregations/evalaggs.go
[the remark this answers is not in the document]
vasukalariya: addressed in the latest commit.

AndrewHess (approved):

AndrewHess (approved):

PR #1328, opened 2024-07-25
metrics explorer - handle constant value formula using backend

sunitakawane:
# Description
Summarize the change.

# Testing

# Checklist:


pkg/integrations/prometheus/promql/parser.go
```diff
@@ -607,6 +608,29 @@ func handleVectorSelector(mQueryReqs []*structs.MetricsQueryRequest, intervalSec
...
+	} else if mQueryReqs[0].MetricsQuery.HashedMName == uint64(0) {
+		// This means there is a common group by on multiple metrics separated by operators.
+		// So, we need to append the aggregations to each of the Request in LHS and RHS.
+		// And we can discard the current aggregation in the mQueryReqs[0]
+		mQueryReqs = appendMetricAggsToTheMQuery(lhsRequest, &mQueryReqs[0].MetricsQuery)
+	} else {
+		mQueryReqs = append(mQueryReqs, lhsRequest...)
+	}
+	return mQueryReqs, queryArithmetic, nil
+}
```
Macbeth98, line 633: I do not understand what is the purpose of this function. And based on the way the parser is written. We do not want to do this. As this will cause issues with `BinaryExpr` and `CallExpr`.

Macbeth98 (commented):
From the PR title: handle constant value: 

Is it for cases where formula is: `1 + 2 + a`, or `1 * 2 *a + b`, or just `1 + 2`. If it is for any of the above, there is no need for `handleNumberLiteral`. The parser already handles those cases. 

Then you would only need the changes you have in the `FetchScalarMetricsForUi` function for Interval calculation. 

But if it is for a case where it just a single number value as formula, such as this: `1` or `23` or `987`, then the current parser will not handle it and it would go to the `*parser.NumberLiteral` case, but we do not want to handle it here.

The other way to do this is: In the `ConvertPromQLToMetricsQuery` which is called to parse the PromQL query, check if the `query` string can be converted to a Float value and if it is, then convert it the `query` to be `0 + floatVal` or `floatVal + 0`. In this way, the `BinaryExpr` will handle this case without any issues. 

If you need I can submit this change in a separate PR and you can remove the parser changes.

sunitakawane:
> From the PR title: handle constant value:
> 
> Is it for cases where formula is: `1 + 2 + a`, or `1 * 2 *a + b`, or just `1 + 2`. If it is for any of the above, there is no need for `handleNumberLiteral`. The parser already handles those cases.
> 
> Then you would only need the changes you have in the `FetchScalarMetricsForUi` function for Interval calculation.
> 
> But if it is for a case where it just a single number value as formula, such as this: `1` or `23` or `987`, then the current parser will not handle it and it would go to the `*parser.NumberLiteral` case, but we do not want to handle it here.
> 
> The other way to do this is: In the `ConvertPromQLToMetricsQuery` which is called to parse the PromQL query, check if the `query` string can be converted to a Float value and if it is, then convert it the `query` to be `0 + floatVal` or `floatVal + 0`. In this way, the `BinaryExpr` will handle this case without any issues.
> 
> If you need I can submit this change in a separate PR and you can remove the parser changes.

Yes, we want to handle the case of single number value as formula. It would be great if you could create a separate PR to handle this. thanks @Macbeth98

PR #1330, opened 2024-07-25
SPL command GenTimes

vasukalariya:
# Description
Summarize the change.
Gentimes command splunk

# Testing
Tests in ingest.csv

# Checklist:

### DIFF
--- a/pkg/ast/spl/spl.peg
+++ b/pkg/ast/spl/spl.peg
@@ -313,6 +313,19 @@ func getDefaultFormatResultsExpr() *structs.FormatResultsRequest {
     }
 }
 
+func chainAggregators(curQueryAgg *structs.QueryAggregators, queryAggs []any) {
+    for ; curQueryAgg.Next != nil; curQueryAgg = curQueryAgg.Next {}
+
+    // Link the remaining chains.
+    for i := range queryAggs {
+        queryAgg := queryAggs[i].(*structs.QueryAggregators)
+        curQueryAgg.Next = queryAgg
+
+        // Go to the end of this chain.
+        for ; curQueryAgg.Next != nil; curQueryAgg = curQueryAgg.Next {}
+    }
+}
+
 type aggregator struct {
     measureAgg          *structs.MeasureAggregator
     renameOutputField   bool
@@ -427,16 +440,8 @@ Start <- SPACE? initialSearch:(InitialSearchBlock) filterBlocks:(FilterBlock)* q
 
             // Go to the end of the first chain.
             curQueryAgg := q.PipeCommands
-            for ; curQueryAgg.Next != nil; curQueryAgg = curQueryAgg.Next {}
-
-            // Link the remaining chains.
-            for i := range queryAggSlice[1:] {
-                queryAgg := queryAggSlice[i + 1].(*structs.QueryAggregators)
-                curQueryAgg.Next = queryAgg
-
-                // Go to the end of this chain.
-                for ; curQueryAgg.Next != nil; curQueryAgg = curQueryAgg.Next {}
-            }
+            
+            chainAggregators(curQueryAgg, queryAggSlice[1:])
         }
     }
 
@@ -453,18 +458,9 @@ Start <- SPACE? initialSearch:(InitialSearchBlock) filterBlocks:(FilterBlock)* q
 
     if queryAggBlocks != nil {
         queryAggSlice := queryAggBlocks.([]any)
+        curQueryAgg := q.PipeCommands
 
-        if len(queryAggSlice) > 0 {
-            curQueryAgg := q.PipeCommands
-            // Link the remaining chains.
-            for i := range queryAggSlice {
-                queryAgg := queryAggSlice[i].(*structs.QueryAggregators)
-                curQueryAgg.Next = queryAgg
-
-                // Go to the end of this chain.
-                for ; curQueryAgg.Next != nil; curQueryAgg = curQueryAgg.Next {}
-            }
-        }
+        chainAggregators(curQueryAgg, queryAggSlice)
     }
 
     return q, nil
@@ -477,7 +473,7 @@ PartialTimestamp <- [0-9][0-9] '/' [0-9][0-9] '/' [0-9][0-9][0-9][0-9] {
     return completeTimestamp, nil
 }
 
-IntegerAsTime <- intStr:IntegerAsString {
+IntegerAsTimeToUnixEpochMs <- intStr:IntegerAsString {
     timeOffset, err := strconv.ParseInt(intStr.(string), 10, 64)
     if err != nil {
         return "", fmt.Errorf("Spl peg: Error while converting the integer: %v", err)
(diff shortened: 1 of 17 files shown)


pkg/ast/spl/spl.peg
```diff
@@ -505,6 +442,170 @@ Start <- SPACE? initialSearch:(InitialSearchBlock) filterBlocks:(FilterBlock)* q
 
     return q, nil
 }
+/ SPACE? PIPE CMD_GENTIMES SPACE genTimesOption:(GenTimesOptionList) queryAggBlocks:(QueryAggergatorBlock)* SPACE? EOF {
```
Macbeth98, line 445: So the search command will not be there at all when `gentimes` is used?

pkg/ast/spl/spl.peg
```diff
@@ -505,6 +442,170 @@ Start <- SPACE? initialSearch:(InitialSearchBlock) filterBlocks:(FilterBlock)* q
...
+        if len(queryAggSlice) > 0 {
+            curQueryAgg := q.PipeCommands
+            // Link the remaining chains.
+            for i := range queryAggSlice {
+                queryAgg := queryAggSlice[i].(*structs.QueryAggregators)
+                curQueryAgg.Next = queryAgg
+
+                // Go to the end of this chain.
+                for ; curQueryAgg.Next != nil; curQueryAgg = curQueryAgg.Next {}
+            }
```
Macbeth98, line 466: May be we can extract this as a separate function? So that we don't do duplication?

pkg/ast/spl/spl.peg
```diff
@@ -505,6 +442,170 @@ Start <- SPACE? initialSearch:(InitialSearchBlock) filterBlocks:(FilterBlock)* q
...
+    daySnap := fmt.Sprintf("%v", utils.TMDay)
+    finalTime, err = utils.ApplySnap(daySnap, finalTime)
+    if err != nil {
+        return "", fmt.Errorf("Spl peg: Error while applying the snap: %v", err)
+    }
+
+    return finalTime.UnixMilli(), nil
+}
+
+ConvertedTime <- timeStamp:(FullTimeStamp / PartialTimestamp) {
```
Macbeth98, line 498: nit: Rename it something more descriptive. May be `DateTimeToUnixEpochMs` or `ConvertDateTimeToUnixEpochMs`. Something like this.

pkg/ast/spl/spl.peg
```diff
@@ -505,6 +442,170 @@ Start <- SPACE? initialSearch:(InitialSearchBlock) filterBlocks:(FilterBlock)* q
...
+    return &GenTimesOptionArgs{
+        argOption: "start",
+        genTimesOption: &structs.GenTimes{
+            StartTime: uint64(timeStamp.(int64)),
+        },
+    }, nil
+}
+
+GenTimesOptionIncrement <- "increment" EQUAL intStr:IntegerAsString unitStr:("s" / "m" / "d" / "h")? {
+    spanNum, err := strconv.ParseInt(intStr.(string), 10, 64)
```
Macbeth98, line 530: Is the `unitStr` Optional? when Increment is given? No right?

pkg/ast/spl/spl.peg
```diff
@@ -505,6 +442,170 @@ Start <- SPACE? initialSearch:(InitialSearchBlock) filterBlocks:(FilterBlock)* q
...
+                genTimeOption.Interval = genTimeOptArg.genTimesOption.Interval
+            default:
+                return nil, fmt.Errorf("Not a Valid GenTimes Command Option: %v", argOption)
+        }
+    }
+
+    startSpecified, exist := optionWasSpecified["start"]
+    if !exist || !startSpecified {
+        return nil, fmt.Errorf("Invalid Syntax, start time is not specified")
+    }
```
Macbeth98, line 605: Just checking for `exist` should be fine, right? As seems like the values if present in `optionWasSpecified` will always be `true`.  Also, if that is the case you should change this `optionWasSpecified` type to `map[string]struct{}` instead of `map[string]bool`.

pkg/segment/aggregations/generateevents.go
```diff
@@ -0,0 +1,91 @@
...
+	interval := aggs.GenerateEvent.GenTimes.Interval.Num
+	if interval < 0 {
+		start, end = end, start
+		interval = -interval
+	}
+
+	// No need to generate events
+	if start >= end {
+		return nil
+	}
```
Macbeth98, line 55: Shouldn't this be thrown as an error? What does Splunk do in this scenario?

pkg/segment/aggregations/generateevents.go
```diff
@@ -0,0 +1,91 @@
...
+	"time"
+
+	"github.com/siglens/siglens/pkg/segment/structs"
+	"github.com/siglens/siglens/pkg/segment/utils"
+)
+
+func getFormattedTime(t time.Time) string {
+	return t.Format("Mon Jan 2 15:04:05 2006 -0700")
+}
+
```
Macbeth98, line 14: nit: Move this to `timutils.go` and rename this function to something more descriptive. I guess taking inspiration from Splunk: `formatToHumanReadableTime`. 
And probably also add a comment to this function with the format.

pkg/segment/reader/record/rrcreader.go
```diff
@@ -396,12 +401,17 @@ func GetJsonFromAllRrc(allrrc []*utils.RecordResultContainer, esResponse bool, q
...
+		processSingleSegment("generate_event", "generate_event", nil, true)
+	}
+
 	if !(tableColumnsExist || (aggs != nil && aggs.OutputTransforms == nil) || hasQueryAggergatorBlock || transactionArgsExist) {
 		allRecords, finalCols = applyHardcodedColumns(hardcodedArray, renameHardcodedColumns, allRecords, finalCols)
 		if len(hardcodedArray) > 0 {
 			numProcessedRecords = 1
 		}
 	} else {
+
```
Macbeth98, line 414: you should probably keep this in the below `if-else` block. 

I believe this PR might have some conflicts with my recent PR that got merged, there even with no `segmap`, in the else the function will call `processSegment`. So once you pull the latest changes you might understand.

Macbeth98 (commented):
I think you should also probably set `numTotalSegments`, `finishedSegments`, `resultCount`, `rawSearchFinished` to true in the `queryStatus.go` file. 

Probably whatever variables are there here: https://github.com/siglens/siglens/blob/develop/pkg/segment/reader/record/rrcreader.go#L231-L242

This is so that, if you have `stats` or some other aggregations which require these variables, may not work properly.

I think the current logic handles the setting or updating of `finishedSegments` through this function: `IncrementNumFinishedSegments`. But `totalSegments` value will be zero. And since it is zero, the stats and other aggs which needs all the segments processed might work. But it is better to set it as `1` and also the other variables, I have mentioned above.

Some of these were added in the latest merged PR.

pkg/segment/aggregations/generateevents.go
vasukalariya: It just returns 0 results.

pkg/segment/aggregations/generateevents.go
vasukalariya: addressed in the latest commit.

pkg/ast/spl/spl.peg

pkg/ast/spl/spl.peg
vasukalariya: It seems mandatory as per the Splunk documentation but upon testing on Splunk, it seems that if only a number is present second is used as the time unit. So, we are also making it optional similar to it.

pkg/ast/spl/spl.peg

pkg/ast/spl/spl.peg

pkg/ast/spl/spl.peg
vasukalariya: No, gentimes has to be the first command in the search pipeline.

pkg/segment/reader/record/rrcreader.go
vasukalariya: addressed in the latest commit.

Macbeth98 (approved):

PR #1340, opened 2024-07-29
fix all_slos hook

sunitakawane:
# Description
Summarize the change.
Fix this to show/hide all_slos page 

# Testing

# Checklist:

### DIFF
--- a/static/all-slos.html
+++ b/static/all-slos.html
@@ -71,7 +71,11 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
         <div id="all-slos">
             <h1 class="myOrg-heading">SLOs</h1>
-            {{ .OrgSLOs | EntMsg }}
+            {{ if .OrgSLOs}}
+            {{ .OrgSLOs |  safeHTML }}
+            {{else}}
+            {{ safeHTML "ent"}}
+            {{end }}
         </div>
 
         <div id="app-footer">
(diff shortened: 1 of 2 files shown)


static/all-slos.html
```diff
@@ -71,7 +71,11 @@
 
         <div id="all-slos">
             <h1 class="myOrg-heading">SLOs</h1>
-            {{ .OrgSLOs | EntMsg }}
+            {{ if .OrgSLOs}}
+            {{ .OrgSLOs |  safeHTML }}
+            {{else}}
+            {{ safeHTML "ent"}}
+            {{end }}
```
AndrewHess, line 78: Personally I find using `safeHTML "ent"` a little odd. What if you added an `EnterpriseEnabled` flag and then did something like:
```
{{ if .EnterpriseEnabled }}
{{ .OrgSLOs | safeHTML }}
{{ else }}
{{ EntMsg }}
{{ end }}
```

Or if you like your version better, I can approve what you have

sunitakawane: By adding EnterpriseEnabled flag you mean adding a flag in server.yaml and adding it to the readConfig and related test code? I think I like {{ safeHTML "ent"}} better than adding one more config variable.

AndrewHess: I meant adding it here: https://github.com/siglens/siglens/blob/develop/pkg/hooks/hooks.go#L123-L128. And then set it where we're setting those variables, not in server.yaml.

But I'll approve this if you want to keep what you have

AndrewHess (approved):

static/all-slos.html
sunitakawane: ok, got it, I will change that

AndrewHess (approved):

PR #1346, opened 2024-07-29
Add implementation for mvexpand

venkateshamatam:
# Description
https://docs.splunk.com/Documentation/Splunk/9.2.2/SearchReference/Mvexpand

# Testing
- `city=Boston 
| eval names="Frank,Grace,Heidi,Ivan"
| makemv delim="," names 
| mvexpand names
`
- `city=Boston 
| eval names="Frank,Grace,Heidi,Ivan"
| makemv delim="," names 
| mvexpand names limit=2
`
- `city=Boston 
| eval names="Frank,Grace,Heidi,Ivan"
| eval cities="Boston, Chicago, Miami"
| makemv delim="," cities | mvexpand cities
`
- `city=Boston 
| eval names="Frank,Grace,Heidi,Ivan"
| eval cities="Boston, Chicago, Miami"
| makemv delim="," cities | mvexpand cities limit=2`
# Checklist:

### DIFF
--- a/pkg/segment/aggregations/segaggs.go
+++ b/pkg/segment/aggregations/segaggs.go
@@ -1970,15 +1970,13 @@ func performMultiValueColRequestWithoutGroupby(letColReq *structs.LetColumnsRequ
 func performMVExpand(fieldValue interface{}) []interface{} {
 	var values []interface{}
 
-	switch v := fieldValue.(type) {
-	case []string:
-		for _, val := range v {
-			values = append(values, val)
-		}
-	case []interface{}:
-		values = v
-	default:
+	isArrayOrSlice, v, _ := utils.IsArrayOrSlice(fieldValue)
+	if !isArrayOrSlice {
 		return nil
+	}
+
+	for i := 0; i < v.Len(); i++ {
+		values = append(values, v.Index(i).Interface())
 	}
 
 	return values
--- a/pkg/segment/aggregations/segaggs.go
+++ b/pkg/segment/aggregations/segaggs.go
@@ -1981,11 +1981,6 @@ func performMVExpand(fieldValue interface{}) []interface{} {
 		}
 	case []interface{}:
 		values = v
-	case string:
-		splitValues := strings.Split(v, ",")
-		for _, val := range splitValues {
-			values = append(values, val)
-		}
 	default:
 		return nil
 	}
--- a/pkg/segment/aggregations/segaggs.go
+++ b/pkg/segment/aggregations/segaggs.go
@@ -1948,15 +1948,11 @@ func performMultiValueColRequestWithoutGroupby(letColReq *structs.LetColumnsRequ
 			}
 			delete(recs, key)
 			for i, expandedValue := range expandedRecs {
-				expandedValueStr, ok := expandedValue.(string)
-				if !ok {
-					expandedValueStr = fmt.Sprintf("%v", expandedValue)
-				}
 				newRec := make(map[string]interface{})
 				for k, v := range rec {
 					newRec[k] = v
 				}
-				newRec[mvColReq.ColName] = expandedValueStr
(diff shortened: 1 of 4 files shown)


pkg/segment/aggregations/segaggs.go
```diff
@@ -1931,14 +1933,60 @@ func performMultiValueColRequestWithoutGroupby(letColReq *structs.LetColumnsRequ
...
+			// Apply limit if mvColReq.Limit is greater than 0
+			if mvColReq.Limit > 0 && len(expandedRecs) > int(mvColReq.Limit) {
+				expandedRecs = expandedRecs[:mvColReq.Limit]
+			}
+			delete(recs, key)
+			for i, expandedValue := range expandedRecs {
+				expandedValueStr, ok := expandedValue.(string)
+				if !ok {
+					expandedValueStr = fmt.Sprintf("%v", expandedValue)
+				}
```
Macbeth98, line 1954: Why are you converting this to string? You can just assign the value you are getting right? We might not want to convert Number types to strings.

pkg/segment/aggregations/segaggs.go
```diff
@@ -1937,14 +1939,60 @@ func performMultiValueColRequestWithoutGroupby(letColReq *structs.LetColumnsRequ
...
+		for _, val := range v {
+			values = append(values, val)
+		}
+	case []interface{}:
+		values = v
+	case string:
+		splitValues := strings.Split(v, ",")
+		for _, val := range splitValues {
+			values = append(values, val)
+		}
```
Macbeth98, line 1988: I am missing something here? If it is a string, then it is not a Multi Value field right? 
Or is this how splunk does? if you give `mvexpand` on a string field?

pkg/segment/aggregations/segaggs.go
venkateshamatam: our representation of multi-value is always []string right? so even if we have numbers it always comes as string. But I can change it back to use `expandedValue` as it is so if we update the multi-value representation, then numbers or other data types don't get converted to strings.

pkg/segment/aggregations/segaggs.go
venkateshamatam: Handling just multi-value types should be enough here since we don't interpret comma separated strings as multi-value fields without using `makemv`

pkg/segment/aggregations/segaggs.go
```diff
@@ -1937,14 +1939,55 @@ func performMultiValueColRequestWithoutGroupby(letColReq *structs.LetColumnsRequ
...
+func performMVExpand(fieldValue interface{}) []interface{} {
+	var values []interface{}
+
+	switch v := fieldValue.(type) {
+	case []string:
+		for _, val := range v {
+			values = append(values, val)
+		}
+	case []interface{}:
+		values = v
```
Macbeth98, line 1979: This will not catch any other cases like: `[]int`, `[]float`, etc. This will only catch the case if its type is set as `[]interface{}`. 

Instead of this use the function: `IsArrayOrSlice` that is here: 
https://github.com/siglens/siglens/blob/develop/pkg/utils/sliceutils.go#L70

Macbeth98: you can directly remove the switch case altogether. And if it is true, then just return the values  and if false, return `nil` like how you do here.

Macbeth98 (commented):
Add the queries in `ingest.csv`, and `restart.csv`. 

And also probably add some unit tests for `mvexpand` case.

pkg/segment/aggregations/segaggs.go
venkateshamatam: Done, I just pushed the change to use the `IsArrayOrSlice` from utils. Also added CI/CD tests and unit tests

Macbeth98 (approved):

PR #1347, opened 2024-07-30
fix: database is locked issue in Alerting

Macbeth98:
# Description
- Fixes the database is locked issue when updating the Alert state and last sent notification time.
- This fix includes: 
    - Enabling `WAL` for SQLite and setting a busy timeout wait of `5000` ms. The busy timeout is for the scenarios where the database is locked for a relatively small time. The DB operation will wait for the set time before throwing a database is locked error.
    - Setting retry for 5 times with increasing delays before a retry.
- This fix seems to work perfectly fine after testing on my local device with 2000 Alerts.
- In future, we may need to do the updates in batches if this not work for more work loads. At that time we can use `chan` to collect the Alerts data and call update every set interval ms.

# Testing
- Tested by running alerts load-test up to 2000 Alerts per Minute.

# Checklist:

### DIFF
--- a/pkg/alerts/alertsqlite/alerts_sqlite.go
+++ b/pkg/alerts/alertsqlite/alerts_sqlite.go
@@ -40,6 +40,7 @@ type Sqlite struct {
 }
 
 const maxRetries = 5
+const baseRetryDelay = 50 * time.Millisecond
 
 func (p *Sqlite) SetDB(dbConnection *gorm.DB) {
 	p.db = dbConnection
@@ -125,7 +126,7 @@ func retry(operation func(attemptCount int) error) error {
 		}
 		if strings.Contains(err.Error(), "database is locked") {
 			log.Warnf("retry: database is locked, retrying attempt %d. Error=%v", i+1, err)
-			time.Sleep(time.Duration(math.Pow(2, float64(i))) * time.Millisecond * 50)
+			time.Sleep(time.Duration(math.Pow(2, float64(i))) * baseRetryDelay)
 			continue
 		}
 		return err
@@ -658,11 +659,13 @@ func (p Sqlite) DeleteContactPoint(contact_id string) error {
 func updateLastSentTimeAndAlertState(tx *gorm.DB, alert_id string, alertState alertutils.AlertState) error {
 	currentTime := time.Now().UTC()
 
-	if err := tx.Model(&alertutils.Notification{}).Where("alert_id = ?", alert_id).
+	err := tx.Model(&alertutils.Notification{}).Where("alert_id = ?", alert_id).
 		Updates(map[string]interface{}{
 			"last_sent_time":   currentTime,
 			"last_alert_state": alertState,
-		}).Error; err != nil {
+		}).Error
+
+	if err != nil {
 		err = fmt.Errorf("UpdateLastSentTimeAndAlertState: unable to update, AlertId=%v, Error=%+v", alert_id, err)
 		return err
 	}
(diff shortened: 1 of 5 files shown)


pkg/alerts/alertsqlite/alerts_sqlite.go
```diff
@@ -103,6 +116,23 @@ func isValid(str string) bool {
...
+func retry(operation func(attemptCount int) error) error {
+	var err error
+	for i := 0; i < maxRetries; i++ {
+		err = operation(i + 1)
+		if err == nil {
+			return nil
+		}
+		if strings.Contains(err.Error(), "database is locked") {
+			log.Warnf("retry: database is locked, retrying attempt %d. Error=%v", i+1, err)
+			time.Sleep(time.Duration(math.Pow(2, float64(i))) * time.Millisecond * 50)
```
AndrewHess, line 128: nit: we might want `50 ms` to be a constant defined next to `maxRetries` at the top of the file. What do you think?

pkg/alerts/alertsqlite/alerts_sqlite.go
```diff
@@ -625,49 +655,73 @@ func (p Sqlite) DeleteContactPoint(contact_id string) error {
 }
 
 // update last_sent_time and last_alert_state in notification_details table
-func (p Sqlite) UpdateLastSentTimeAndAlertState(alert_id string, alertState alertutils.AlertState) error {
+func updateLastSentTimeAndAlertState(tx *gorm.DB, alert_id string, alertState alertutils.AlertState) error {
```
AndrewHess, line 658: nit: I'm not quite sure what `tx` means (transaction?). Is this a really common name for a gorm DB? If not, a more descriptive name would be better, or perhaps just `db`

pkg/alerts/alertsqlite/alerts_sqlite.go
```diff
@@ -625,49 +655,73 @@ func (p Sqlite) DeleteContactPoint(contact_id string) error {
...
-func (p Sqlite) UpdateAlertStateAndIncrementNumEvaluations(alert_id string, alertState alertutils.AlertState) error {
+func updateAlertStateAndIncrementNumEvaluations(tx *gorm.DB, alert_id string, alertState alertutils.AlertState) error {
+	if err := tx.Model(&alertutils.AlertDetails{}).Where("alert_id = ?", alert_id).
+		Updates(map[string]interface{}{
+			"state":                 alertState,
+			"num_evaluations_count": gorm.Expr("num_evaluations_count + ?", 1),
+		}).Error; err != nil {
+		err = fmt.Errorf("UpdateAlertStateAndIncrementNumEvaluations: unable to update alert state and increment evaluations count with AlertId=%v, Error=%+v", alert_id, err)
+		return err
+	}
```
AndrewHess, line 681: nit: I think this syntax you're using:
```
if err := foo(); err != nil {
    // handle error
}
```
is fine when the whole condition is on one line, but since your condition spans several lines it gets a bit harder to read. Can you do it this way instead?
```
err := foo
if err != nil {
    // handle error
}
```

AndrewHess (commented):
Looks pretty good

pkg/alerts/alertsqlite/alerts_sqlite.go
Macbeth98: Hmm, I am not sure what to actually name here, that is definitely a `db` instance from a Transaction. I think we need a name that represents that this is the `gorm.DB` instance from a Transaction. 
The only one I can think of `transactionDB`. Please let me know if you have any other suggestions.

AndrewHess: I'd say there's two options, depending on what information is important to the reader.
1. If for this particular function it doesn't matter that the DB is a DB from a transaction, then calling it `transactionDB` gives the reader information that isn't relevant, and—this probably doesn't matter for this specific case, but is helpful to think about in general—it could deter from other people reusing this function for a similar use case but with a non-transaction DB, because the name `transactionDB` makes them think the DB has to be from a transaction. So if this is the case, just call it `db` (or similar)
2. If it is important (just within the scope of this function) that this is a DB from a transaction, then I think `transactionDB` is a good name

I _think_ Option 1 is correct here, but I'm not certain

Macbeth98: Yes that make sense. As you have mentioned Option 1 is correct. If considered individually then that is just a `gorm.DB` instance. So I will just rename it to `db`.

PR #1359, opened 2024-07-31
Parsing and Implementation of SPL Command Inputlookup

vasukalariya:
# Description
Summarize the change.
SPL command inputlookup

# Testing
Added test cases in ingest.csv

# Checklist:

### DIFF
--- a/pkg/ast/pipesearch/searchQueryParser.go
+++ b/pkg/ast/pipesearch/searchQueryParser.go
@@ -138,6 +138,18 @@ func createMatchAll(qid uint64) *ASTNode {
 	return rootNode
 }
 
+func updatePositionForGenEvents(aggs *QueryAggregators) {
+	node := aggs
+	position := 1
+	for node != nil {
+		if node.PipeCommandType == GenerateEventType {
+			node.GenerateEvent.EventPosition = position
+			position++
+		}
+		node = node.Next
+	}
+}
+
 func parsePipeSearch(searchText string, queryLanguage string, qid uint64) (*ASTNode, *QueryAggregators, error) {
 	var leafNode *ASTNode
 	var res interface{}
@@ -190,6 +202,9 @@ func parsePipeSearch(searchText string, queryLanguage string, qid uint64) (*ASTN
 		log.Errorf("qid=%d, parsePipeSearch: searchPipeCommandsToASTnode error: %v", qid, err)
 		return nil, nil, err
 	}
+
+	updatePositionForGenEvents(pipeCommands)
+
 	return boolNode, pipeCommands, nil
 }
 
--- a/pkg/ast/pipesearch/searchQueryParser.go
+++ b/pkg/ast/pipesearch/searchQueryParser.go
@@ -142,7 +142,7 @@ func updatePositionForGenEvents(aggs *QueryAggregators) {
 	node := aggs
 	position := 1
 	for node != nil {
-		if node.PipeCommandType == GenerateEventType {
+		if node.GenerateEvent != nil {
 			node.GenerateEvent.EventPosition = position
 			position++
 		}
--- a/pkg/ast/pipesearch/searchQueryParser.go
+++ b/pkg/ast/pipesearch/searchQueryParser.go
@@ -382,9 +382,6 @@ func parseGenerateCmd(node *structs.GenerateEvent, qid uint64) (*QueryAggregator
 	}
 	if node.InputLookup != nil {
 		aggNode.GenerateEvent.InputLookup = node.InputLookup
-		if aggNode.GenerateEvent.InputLookup.Max == 0 {
-			aggNode.GenerateEvent.InputLookup.Max = 1000000000
-		}
 	}
 
 	return aggNode, nil
--- a/pkg/ast/pipesearch/searchQueryParser.go
+++ b/pkg/ast/pipesearch/searchQueryParser.go
@@ -365,6 +365,12 @@ func parseGenerateCmd(node *structs.GenerateEvent, qid uint64) (*QueryAggregator
 	if node.GenTimes != nil {
 		aggNode.GenerateEvent.GenTimes = node.GenTimes
 	}
+	if node.InputLookup != nil {
+		aggNode.GenerateEvent.InputLookup = node.InputLookup
+		if aggNode.GenerateEvent.InputLookup.Max == 0 {
+			aggNode.GenerateEvent.InputLookup.Max = 1000000000
+		}
+	}
 
 	return aggNode, nil
 }
--- a/pkg/ast/spl/spl.peg
+++ b/pkg/ast/spl/spl.peg
@@ -4535,6 +4535,14 @@ PositiveIntegerAsString <- [0-9]+ {
     return string(c.text), nil
 }
 
+PositiveInteger <- intStr:PositiveIntegerAsString {
+    num, err := strconv.ParseUint(intStr.(string), 10, 64)
+    if err != nil {
+        return nil, fmt.Errorf("Spl peg: Invalid Positive Integer value: (%v): , err: %v", intStr.(string), err)
+    }
+    return num, nil
+}
+
 EqualityOperator <- EMPTY_OR_SPACE op:("=" / "!=") EMPTY_OR_SPACE {
     opStr, err := transferUint8ToString(op)
     if err != nil {
@@ -5221,7 +5229,7 @@ MvexpandLimit <- SPACE "limit" EQUAL intValue:IntegerAsString {
     return intValue, nil
 }
 
-OptionalWhereClause <- SPACE CMD_WHERE condition:BoolExpr {
+WhereClause <- SPACE CMD_WHERE condition:BoolExpr {
     return condition.(*structs.BoolExpr), nil
 }
 
@@ -5243,15 +5251,7 @@ InputLookupOptionStrict <- "strict" EQUAL boolVal:Boolean {
     }, nil
 }
 
-GetPositiveInteger <- intStr:PositiveIntegerAsString {
-    num, err := strconv.ParseUint(intStr.(string), 10, 64)
-    if err != nil {
-        return nil, fmt.Errorf("Spl peg: Invalid Positive Integer value: (%v): , err: %v", intStr.(string), err)
-    }
-    return num, nil
-}
-
-InputLookupOptionStart <- "start" EQUAL posInt:GetPositiveInteger {
+InputLookupOptionStart <- "start" EQUAL posInt:PositiveInteger {
     return &InputLookupOptionArgs{
         argOption: "start",
         inputLookupOption: &structs.InputLookup{
@@ -5260,7 +5260,7 @@ InputLookupOptionStart <- "start" EQUAL posInt:GetPositiveInteger {
     }, nil
 }
 
-InputLookupOptionMax <- "max" EQUAL posInt:GetPositiveInteger {
+InputLookupOptionMax <- "max" EQUAL posInt:PositiveInteger {
     return &InputLookupOptionArgs{
         argOption: "max",
         inputLookupOption: &structs.InputLookup{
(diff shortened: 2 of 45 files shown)


pkg/segment/structs/evaluationstructs.go
```diff
@@ -712,6 +712,95 @@ func (self *BoolExpr) Evaluate(fieldToValue map[string]utils.CValueEnclosure) (b
...
+				return leftFloat <= rightFloat, nil
+			case ">=":
+				return leftFloat >= rightFloat, nil
+			default:
+				return false, fmt.Errorf("BoolExpr.EvaluateForInputLookup: invalid ValueOp %v for floats", self.ValueOp)
+			}
+		} else if errLeftStr == nil && errRightStr == nil {
+			leftStr = strings.ToLower(leftStr)
+			rightStr = strings.ToLower(rightStr)
+			match, err := filepath.Match(rightStr, leftStr)
```
AndrewHess, line 747: `filepath.Match` does more than just allowing `*`. Do you really want `filepath.Match`, or do you just want to allow `*` as "match anything, any number of characters"? If you just want `*`, you might want to use https://github.com/siglens/siglens/blob/develop/pkg/common/dtypeutils/dtypeutils.go#L599-L616

pkg/segment/structs/evaluationstructs.go
AndrewHess, line 1: You should add some unit tests for this in evaluationstructs_test.go

pkg/segment/reader/record/recordreader.go
```diff
@@ -42,11 +42,6 @@ import (
 func GetRecordsFromSegment(segKey string, vTable string, blkRecIndexes map[uint16]map[uint16]uint64,
 	tsKey string, esQuery bool, qid uint64,
 	aggs *structs.QueryAggregators, colsIndexMap map[string]int) (map[string]map[string]interface{}, map[string]bool, error) {
-
-	if aggs.HasGenerateEvent() {
-		return aggs.GenerateEvent.GeneratedRecords, aggs.GenerateEvent.GeneratedCols, nil
-	}
```
AndrewHess, line 48: Why don't you need this anymore?

pkg/segment/aggregations/generateevents.go
```diff
@@ -86,3 +99,112 @@ func PerformGenTimes(aggs *structs.QueryAggregators) error {
...
+}
+
+// Ensure record and column lengths are valid before calling this function.
+func createRecord(columnNames []string, record []string) map[string]interface{} {
+	recordMap := make(map[string]interface{})
+	for i, col := range columnNames {
+		recordMap[col] = record[i]
+	}
+	return recordMap
+}
```
AndrewHess, line 114: nit: this might be better as a generic function in maputils.go. But you can keep it here if you prefer

pkg/segment/aggregations/generateevents.go
```diff
@@ -86,3 +99,112 @@ func PerformGenTimes(aggs *structs.QueryAggregators) error {
...
+
+func PerformInputLookup(aggs *structs.QueryAggregators) error {
+	var reader *csv.Reader
+	if aggs.GenerateEvent.InputLookup == nil {
+		return fmt.Errorf("PerformInputLookup: InputLookup is nil")
+	}
+	filename := aggs.GenerateEvent.InputLookup.Filename
+
+	if !checkCSVFormat(filename) {
+		return fmt.Errorf("PerformInputLookup: Only CSV format is currently supported")
```
AndrewHess, line 124: I think this will be a bit more informative:
```suggestion
		return fmt.Errorf("PerformInputLookup: Only .csv and .csv.gz formats are currently supported")
```

pkg/segment/aggregations/generateevents.go
```diff
@@ -86,3 +99,112 @@ func PerformGenTimes(aggs *structs.QueryAggregators) error {
...
+	}
+
+	workingDir, err := os.Getwd()
+	if err != nil {
+		return fmt.Errorf("PerformInputLookup: Error while getting current working directory, err: %v", err)
+	}
+	filepath := workingDir + "/" + config.GetLookupPath() + filename
+
+	file, err := os.Open(filepath)
```
AndrewHess, line 133: Why do you have to get the working directory? Without the working directory you should get a relative path, and you can `os.Open()` that.

Also when you do `config.GetLookupPath() + filename`, you're assuming the lookup path ends with a `/`. You should use https://pkg.go.dev/path/filepath#Join instead, as that handles both cases

pkg/segment/aggregations/generateevents.go
```diff
@@ -86,3 +99,112 @@ func PerformGenTimes(aggs *structs.QueryAggregators) error {
...
+func createRecord(columnNames []string, record []string) map[string]interface{} {
+	recordMap := make(map[string]interface{})
+	for i, col := range columnNames {
+		recordMap[col] = record[i]
+	}
+	return recordMap
+}
+
+func PerformInputLookup(aggs *structs.QueryAggregators) error {
+	var reader *csv.Reader
```
AndrewHess, line 117: nit: `reader` isn't used for a while, so you might want to declare it just before it's actually set up (around line 139 I think).

However, you can keep what you have if you prefer this way

pkg/segment/aggregations/generateevents.go
```diff
@@ -86,3 +99,112 @@ func PerformGenTimes(aggs *structs.QueryAggregators) error {
...
+			}
+			return fmt.Errorf("PerformInputLookup: Error reading record, err: %v", err)
+		}
+
+		if len(csvRecord) != len(columnNames) {
+			return fmt.Errorf("PerformInputLookup: Error reading record, column count mismatch")
+		}
+
+		count++
+		record := createRecord(columnNames, csvRecord)
```
AndrewHess, line 190: nit: I personally find it really odd to increment `count` (the loop counter) in the middle of the loop instead of at the beginning or the end of the loop. I'd suggest moving it to the beginning since you have a `continue` lower

pkg/ast/pipesearch/searchQueryParser.go
```diff
@@ -138,6 +138,18 @@ func createMatchAll(qid uint64) *ASTNode {
...
+	node := aggs
+	position := 1
+	for node != nil {
+		if node.GenerateEvent != nil {
+			node.GenerateEvent.EventPosition = position
+			position++
+		}
+		node = node.Next
+	}
+}
```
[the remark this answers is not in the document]
vasukalariya: This function is required to assign the generate events an order number i.e. the position of generate event in the aggs list. This is required for creating keys. I think using a specified position/order from start to end of the list is better than using a random number.

pkg/ast/spl/spl.peg
```diff
@@ -5203,8 +5221,148 @@ MvexpandLimit <- SPACE "limit" EQUAL intValue:IntegerAsString {
     return intValue, nil
 }
 
+OptionalWhereClause <- SPACE CMD_WHERE condition:BoolExpr {
+    return condition.(*structs.BoolExpr), nil
+}
```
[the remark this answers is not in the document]
vasukalariya: Marked this rule as optional because it could be used at other places too in the future where we need an optional where clause just like how it is used in the input lookup.

pkg/ast/spl/spl.peg
```diff
@@ -5203,8 +5221,148 @@ MvexpandLimit <- SPACE "limit" EQUAL intValue:IntegerAsString {
...
+
+InputLookupAggBlock <- inputLookupBlock:InputLookupBlock {
+    inputLookup := inputLookupBlock.(*structs.QueryAggregators)
+    if !inputLookup.GenerateEvent.InputLookup.Append {
+        return nil, fmt.Errorf("Inputlookup after search must have append as true")
+    }
+    inputLookup.GenerateEvent.InputLookup.HasPrevResults = true
+    
+    return inputLookup, nil
+}
```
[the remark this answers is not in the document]
vasukalariya: Technically, it could be after any generating command like gentimes, inputlookup, etc. and not specifically after a search so named it as an agg block.

pkg/segment/reader/record/recordreader.go
vasukalariya: I have made the code to use a similar framework as inputlookup, when hasGenerateEventWithoutSearch is present the flow would go inside the postQueryBucketcleaning and there we would populate the recs, recordIndexInFinal and finalCols via performAggOnResult method.

pkg/ast/spl/spl.peg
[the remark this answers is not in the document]
AndrewHess: Ok but in other rules you use it like:
```
whereClause:(OptionalWhereClause)?
```
In those rules, the `?` is what makes it optional. I think naming this `OptionalWhereClause` is misleading because there's nothing inherently optional about the rule; it's just how you're using it in other rules that makes it optional

pkg/ast/pipesearch/searchQueryParser.go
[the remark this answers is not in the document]
AndrewHess: >  I think using a specified position/order from start to end of the list is better than using a random number.

I agree. Are you saying that if we set the `EventPosition` while building the `QueryAggregators` with `searchPipeCommandsToASTnode()` then we'd have to set the value to a random number on each element instead of use consecutive numbers? Don't we build each `QueryAggregator` in order?

pkg/ast/spl/spl.peg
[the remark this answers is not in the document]
AndrewHess: Ok, then you should probably update your error message to clarify

pkg/ast/spl/spl.peg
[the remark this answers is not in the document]

pkg/ast/spl/spl.peg
```diff
@@ -5203,8 +5221,148 @@ MvexpandLimit <- SPACE "limit" EQUAL intValue:IntegerAsString {
...
+    }, nil
+}
+
+GetPositiveInteger <- intStr:PositiveIntegerAsString {
+    num, err := strconv.ParseUint(intStr.(string), 10, 64)
+    if err != nil {
+        return nil, fmt.Errorf("Spl peg: Invalid Positive Integer value: (%v): , err: %v", intStr.(string), err)
+    }
+    return num, nil
+}
```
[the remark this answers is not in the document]

pkg/ast/spl/spl.peg
```diff
@@ -5203,8 +5221,148 @@ MvexpandLimit <- SPACE "limit" EQUAL intValue:IntegerAsString {
...
+    }
+
+    return &structs.QueryAggregators{
+        PipeCommandType: structs.GenerateEventType,
+        GenerateEvent: &structs.GenerateEvent{
+            InputLookup: inputLookup,
+        },
+    }, nil
+}
+/ PIPE CMD_INPUTLOOKUP filename:(String) whereClause:(OptionalWhereClause)? {
```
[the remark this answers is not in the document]

pkg/ast/spl/spl.peg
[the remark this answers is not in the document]

pkg/segment/structs/evaluationstructs.go
vasukalariya: addressed in the latest commit.

pkg/segment/structs/evaluationstructs.go
vasukalariya: addressed in the latest commit.

pkg/segment/aggregations/generateevents.go
vasukalariya: addressed in the latest commit.

pkg/segment/aggregations/generateevents.go
vasukalariya: addressed in the latest commit.

pkg/segment/aggregations/generateevents.go
vasukalariya: addressed in the latest commit.

pkg/segment/aggregations/generateevents.go
vasukalariya: addressed in the latest commit.

pkg/segment/aggregations/generateevents.go
vasukalariya: addressed in the latest commit.

AndrewHess (approved):

PR #1363, opened 2024-08-01
Add support for SPL chained relative time offsets

venkateshamatam:
# Description
Add support for chained relative time offsets
https://docs.splunk.com/Documentation/Splunk/latest/Search/Specifytimemodifiersinyoursearch#Examples_of_chained_relative_time_offsets

# Testing
- Existing test cases for other types of offsets pass.
- Added test cases for chained relative time offsets

# Checklist:


pkg/ast/spl/spl.peg
```diff
@@ -1360,7 +1360,7 @@ BinOptionAlignTime <- "aligntime" EQUAL utcEpoch:(PositiveIntegerAsString) {
 
     return binOptionArgs, nil
 }
-/ "aligntime" EQUAL timestamp:(AbsoluteTimestamp / RelativeTimestamp) {
+/ "aligntime" EQUAL timestamp:(AbsoluteTimestamp / RelativeTimestamp / ChainedRelativeTimestamps) {
```
Macbeth98, line 1363: I don't see `ChainedRelativeTimestamps` defined. I see only `ChainedRelativeTimestamp`. Is this just a spelling mistake? 
Also pigeon did not throw this as an error? Since it is not defined?

Macbeth98: Same comment for all other palces.

pkg/ast/spl/spl.peg
```diff
@@ -4276,12 +4276,45 @@ EarliestAndLatest <- CMD_EARLIEST EQUAL earliestTime:(AbsoluteTimestamp / Relati
...
+
+        if i < len(earliestTimeModifier.ChainedSnaps) {
+            relativeTimeModifier.Snap = earliestTimeModifier.ChainedSnaps[i]
+        }
+
+        startEpoch, err = CalculateRelativeTime(ast.TimeModifier{RelativeTime: relativeTimeModifier}, time.UnixMilli(startEpoch))
+        if err != nil {
+            return nil, err
+        }
     }
```
Macbeth98, line 4296: Extract this `for` loop into a function. Since it seems you have used the same one at other places also.

pkg/ast/spl/spl.peg
```diff
@@ -4365,14 +4415,56 @@ Offset <- off:(RelIntegerAsString) tuOff:(RelTimeUnit) {
...
+
+    // Append zero values to ensure corresponding offsets and snaps
+    // This ensures that each snap and offset correspond to the same index
+    appendZeroValues := func() {
+        if len(offsets) > len(snaps) {
+            snaps = append(snaps, "")
+        } else if len(snaps) > len(offsets) {
+            offsets = append(offsets, ast.RelativeTimeOffset{})
+        }
+    }
```
Macbeth98, line 4430: It seems that you are assigning default values to the `offsets` and `snaps` slices. So I think it is better to rename it to `appendDefaultValues` or something like that, unless there is a specific reason why it is name as `appendZeroValues`.

pkg/ast/spl/spl.peg
```diff
@@ -4365,14 +4415,56 @@ Offset <- off:(RelIntegerAsString) tuOff:(RelTimeUnit) {
...
+    // Process the first element
+    if first != nil {
+        switch v := first.(type) {
+        case ast.RelativeTimeOffset:
+            offsets = append(offsets, v)
+            snaps = append(snaps, "") // Append zero value for Snap
+        case string:
+            snaps = append(snaps, v)
+            offsets = append(offsets, ast.RelativeTimeOffset{}) // Append zero value for Offset
+        }
```
Macbeth98, line 4441: Also change this comment, if you change the variable name. As assigning empty `string` or `ast.RelativeTimeOffset` and the comment indicating `Zero` values does not make sense.

pkg/ast/spl/spl.peg
```diff
@@ -4365,14 +4415,56 @@ Offset <- off:(RelIntegerAsString) tuOff:(RelTimeUnit) {
...
-// Reference Doc: https://docs.splunk.com/Documentation/Splunk/9.1.2/Search/Specifytimemodifiersinyoursearch 
-// TODO: Chained relative modifiers
-RelativeTimestamp <- defaultTime:("now" / "1")? offset:(Offset)? snapParam:(Snap)? {
+RelativeTimestamp <- defaultTime:("now" / "1")? chained:(ChainedRelativeTimestamp)? {
 
     if defaultTime != nil {
-        if offset != nil || snapParam != nil {
-            return nil, fmt.Errorf("Invalid Relative timestamp cannot combine default with offset or snap")
+        if chained != nil {
+            return nil, fmt.Errorf("Invalid Relative timestamp cannot combine default with chained offsets/snaps")
```
Macbeth98, line 4467: nit: Add a `,` or `:` after `Invalid Relative timestamp...`

pkg/ast/spl/spl.peg
Macbeth98: Also probably a good idea to assign `len(*)` to variables and use them in the `for` loop. 

Also based on your `ChainedRelativeTimestamp` definition, the `ChainedOffsets` and `ChainedSnaps` should have the same length right? So may be you can remove the double check and simplify the code? 

Unless there can be a scenario where the length will not be same.

pkg/ast/structs.go
```diff
@@ -81,8 +81,10 @@ type ParseError struct {
 }
 
 type TimeModifier struct {
-	RelativeTime RelativeTimeModifier
-	AbsoluteTime string
+	RelativeTime   RelativeTimeModifier
+	AbsoluteTime   string
+	ChainedOffsets []RelativeTimeOffset
+	ChainedSnaps   []string
```
Macbeth98, line 87: Seems like this is only used for Calculating the `startEpoch` and `endEpoch` and there is a `ChainedRelativeTimestamp` struct below. 

I would suggest to refactor the code so that you replace these two with `chainedRelativeTimestamp: *ChainedRelativeTimestamp ` or you remove the `ChainedRelativeTimestamp` completely and keep these two `fields` and return `ast.TimeModifier` for `ChainedRelativeTimestamp` definition instead of `ast.ChainedRelativeTimestamp`.

Macbeth98 (commented):
Also add some queries to `ingest.csv`.

venkateshamatam:
> Also add some queries to `ingest.csv`.

I'm not sure if we can add test cases for this in `ingest.csv` since its dependent on current time, we do the entire calculations for epochs in spl parser itself and i added tests for that to ensure the parsed value matches the query, relative to the current time.

pkg/ast/spl/spl.peg
venkateshamatam: Addressed it in the latest commit.

pkg/ast/spl/spl.peg
venkateshamatam: Addressed it in the latest commit.

pkg/ast/spl/spl.peg
venkateshamatam: Addressed it in the latest commit.

pkg/ast/spl/spl.peg
venkateshamatam: Addressed it in the latest commit.

Macbeth98 (approved):

PR #1365, opened 2024-08-01
Add utils for debugging

AndrewHess:
# Description
Add some utils for debugging the flow of functions, and allows you to log extra relevant info. These functions are intended to be used just while debugging.

For example, at the top of the function you can add:
```
utils.TraceEnter(fmt.Sprintf("len recs: %v", len(recs)))
defer utils.TraceExit(func() string { return fmt.Sprintf("len recs: %v", len(recs)) })
```
Then when that function is called, you'll get something like:
```
Entering performAggOnResult (segaggs.go:183) from PostQueryBucketCleaning at segaggs.go:146; len recs: 1000
```
And when the function exits you'll see:
```
Exiting PostQueryBucketCleaning at segaggs.go:162; len recs: 2000
```
Also the the line number it mentions here (segaggs.go:162), is the exact line number of the `return` that got triggered, so if the function has multiple returns, you can see exactly which one got triggered.

# Testing

# Checklist:

### DIFF
--- a/pkg/utils/debugutils.go
+++ b/pkg/utils/debugutils.go
@@ -27,14 +27,14 @@ import (
 )
 
 // Use this at the start of a function like this:
-// TraceEnter(fmt.Sprintf("parameter foo=%v", foo))
+// SigDebugEnter(fmt.Sprintf("parameter foo=%v", foo))
 // OR
-// TraceEnter("")
-func TraceEnter(extraMessage string) {
+// SigDebugEnter("")
+func SigDebugEnter(extraMessage string) {
 	// Get information about the function we're entering.
 	programCounter, file, line, ok := runtime.Caller(1)
 	if !ok {
-		log.Warnf("TraceEnter: Unable to get caller information")
+		log.Warnf("SigDebugEnter: Unable to get caller information")
 		return
 	}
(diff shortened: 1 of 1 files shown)


pkg/utils/debugutils.go
```diff
@@ -0,0 +1,90 @@
...
+	}
+
+	log.Infof(message)
+}
+
+// Use this at the start of a function like this:
+// defer TraceExit(func() string { return fmt.Sprintf("final foo=%v", foo) })
+// OR
+// defer TraceExit(nil)
+func TraceExit(computeExtraMessage func() string) {
```
nkunal, line 66: when we introduce tracing, we may have wrappers named like TraceEnter/Exit, should we name the functions in this file to be slightly different ?

AndrewHess: Good point. Maybe `LogEnter/LogExit` then? Or `DebugEnter/DebugExit`, but that could be confusing because the functions use `log.Infof` since you might not want to see other `log.Debugf` messages.

Or do you have other suggestions?

nkunal: maybe `SigDebugEnter/Exit`

Macbeth98 (approved):

PR #1369, opened 2024-08-02
Bug Fix fields command after generating commands.

vasukalariya:
# Description
Summarize the change.
Earlier for non-search events like gentimes, etc. fields command was erroring out due to nil map for nodeResult.ColumnsOrder at this [line](https://github.com/siglens/siglens/blob/a5bda9c262fe5e30552d62355eee5fb0f0c2ebf2/pkg/segment/aggregations/segaggs.go#L834).

nodeResult.ColumnsOrder must be initialized like over [here](https://github.com/siglens/siglens/blob/a5bda9c262fe5e30552d62355eee5fb0f0c2ebf2/pkg/segment/reader/record/rrcreader.go#L216).

# Testing

```
| gentimes start=-2 increment=5m | fields starttime
```

# Checklist:

### DIFF
--- a/pkg/segment/reader/record/rrcreader.go
+++ b/pkg/segment/reader/record/rrcreader.go
@@ -224,6 +224,7 @@ func GetJsonFromAllRrc(allrrc []*utils.RecordResultContainer, esResponse bool, q
 		} else {
 			recs = make(map[string]map[string]interface{})
 			finalCols = nodeRes.FinalColumns
+			nodeRes.ColumnsOrder = colsIndexMap
 		}
 
 		if hasQueryAggergatorBlock || transactionArgsExist {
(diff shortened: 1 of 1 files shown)


pkg/segment/reader/record/rrcreader.go
```diff
@@ -224,6 +224,7 @@ func GetJsonFromAllRrc(allrrc []*utils.RecordResultContainer, esResponse bool, q
 		} else {
 			recs = make(map[string]map[string]interface{})
 			finalCols = nodeRes.FinalColumns
+			nodeRes.ColumnsOrder = colsIndexMap
```
AndrewHess, line 227: nit: now you're doing `nodeRes.ColumnsOrder = colsIndexMap` in both the `if` (line 216) and the `else`, so you should move it out to above/below the `if/else`

vasukalariya: addressed in the latest commit.

AndrewHess (approved):

PR #1378, opened 2024-08-06
refac: Refactored smrLock to use `RWMutex`

Macbeth98:
# Description
- Refactored `smrLock` to use `RWMutex` instead of `Mutex`. 
- Changed all the read functions to use `RLock()` instead of `Lock()`.
- This also made the `clusterStats` endpoint to load the data faster.
- Fixed an issue with `nil pointer` error in `segmentmicroindex` while getting the `CMIsForBlock`.  

# Testing
- All the tests passed.
- Verified that the cluster stats is showing correctly and the queries are working.

# Checklist:

### DIFF
--- a/pkg/segment/query/metadata/segmentmicroindex.go
+++ b/pkg/segment/query/metadata/segmentmicroindex.go
@@ -104,7 +104,7 @@ func (smi *SegmentMicroIndices) GetCMIsForBlock(blkNum uint16) (map[string]*stru
 		return nil, fmt.Errorf("no cmis are loaded")
 	}
 
-	if int(blkNum) >= len(smi.blockCmis)-1 {
+	if int(blkNum) >= len(smi.blockCmis) {
 		return nil, fmt.Errorf("blkNum %+v does not exist", blkNum)
 	}
 	cmis := smi.blockCmis[blkNum]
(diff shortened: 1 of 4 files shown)


pkg/segment/query/metadata/segmentmicroindex.go
```diff
@@ -99,7 +99,12 @@ func (smi *SegmentMicroIndices) clearMicroIndices() {
 
 // Returns all columnar cmis for a given block or any errors encountered
 func (smi *SegmentMicroIndices) GetCMIsForBlock(blkNum uint16) (map[string]*structs.CmiContainer, error) {
-	if int(blkNum) >= len(smi.blockCmis) {
+	if len(smi.blockCmis) == 0 {
+		log.Errorf("GetCMIsForBlock: No block cmis are loaded. loadedMicroIndices: %+v", smi.loadedMicroIndices)
+		return nil, fmt.Errorf("no cmis are loaded")
+	}
+
+	if int(blkNum) >= len(smi.blockCmis)-1 {
```
nkunal, line 107: our `blkNum` start with 0, so when len(smi.blockCmis) is 1, so even though 1 block has been loaded, we will incorrectly return error here ?

Macbeth98: Yes, I will remove this.

nkunal (approved):

PR #1379, opened 2024-08-06
Add support for SPL stats list

venkateshamatam:
# Description
https://docs.splunk.com/Documentation/SCS/current/SearchReference/Multivaluefunctions#list.28.26lt.3Bvalue.26gt.3B.29

# Testing
- Tested queries manually.
- Will add unit tests and push again.

# Checklist:


pkg/segment/results/segresults/segresults.go
```diff
@@ -368,6 +368,26 @@ func (sr *SearchResults) UpdateSegmentStats(sstMap map[string]*structs.SegStats,
 			}
 			sort.Strings(uniqueStrings)
 			strVal := strings.Join(uniqueStrings, "&nbsp")
+			sr.segStatsResults.measureResults[measureAgg.String()] = utils.CValueEnclosure{
+				Dtype: utils.SS_DT_STRING,
+				CVal:  strVal,
+			}
+			continue
+		case utils.List:
+			strList := make([]string, 0, 100)
```
AndrewHess, line 377: nit: It looks like you only return at most 100 because that's what the splunk documentation says:
> If more than 100 values are in the field, only the first 100 are returned.

It would be good to add a comment here, saying something like that, just to clarify to future readers of this code where the number 100 came from

pkg/segment/results/segresults/segresults.go
```diff
@@ -368,6 +368,26 @@ func (sr *SearchResults) UpdateSegmentStats(sstMap map[string]*structs.SegStats,
...
+				strList = utils.AppendWithLimit(strList, sr.runningSegStat[idx].StringStats.StrList, 100)
+			}
+
+			sort.Strings(strList)
+
+			strVal := strings.Join(strList, " ")
+
 			sr.segStatsResults.measureResults[measureAgg.String()] = utils.CValueEnclosure{
 				Dtype: utils.SS_DT_STRING,
 				CVal:  strVal,
```
AndrewHess, line 393: Is this how we store a multivalue value? I thought we changed it so we store a list of strings

pkg/segment/results/segresults/segresults.go
```diff
@@ -368,6 +368,26 @@ func (sr *SearchResults) UpdateSegmentStats(sstMap map[string]*structs.SegStats,
...
+			if currSst != nil && currSst.StringStats != nil && currSst.StringStats.StrList != nil {
+				strList = utils.AppendWithLimit(strList, currSst.StringStats.StrList, 100)
+			}
+
+			if sr.runningSegStat[idx] != nil && sr.runningSegStat[idx].StringStats != nil {
+				strList = utils.AppendWithLimit(strList, sr.runningSegStat[idx].StringStats.StrList, 100)
+			}
+
+			sort.Strings(strList)
+
```
AndrewHess, line 388: Why do we sort this? The splunk documentation says:
> The order of the values reflects the order of the events.

Am I missing something?

AndrewHess (commented):
Looks pretty good. Can you add some end-to-end tests in ingest.csv and restart.csv?

Macbeth98 (approved):

pkg/segment/results/segresults/segresults.go
venkateshamatam: You're right, we don't need to sort it for the `list` command. Addressed this in the latest commit.

pkg/segment/results/segresults/segresults.go
venkateshamatam: Addressed this in the latest commit.

pkg/segment/results/segresults/segresults.go
venkateshamatam: Added the comment in the latest commit.

PR #1386, opened 2024-08-06
Bug: Panicking on certain queries

vasukalariya:
# Description
Summarize the change.
When specific query is where the GetSortLimit method fails. After the qa.HasSortBlock fails there is no nil check for query aggregator. Added the check to prevent seg fault.
nodeResult.NextQueryAgg will not be set and will come up as [nil](https://github.com/siglens/siglens/blob/239c57ac284c608db219d02a533173c7f412794a/pkg/segment/aggregations/segaggs.go#L144) because len of recs is zero for the case where condition returns 0 records before stats.

Updated the head queries that were giving inconsistent results due to different ingestion order.

# Testing
Tested the queries manually as well by running the ingest.csv

# Checklist:

### DIFF
--- a/cicd/ingest.csv
+++ b/cicd/ingest.csv
@@ -171,8 +171,8 @@ city=Boston | stats count AS Count BY http_method | eval newField=lower(http_met
 "* | head 2 | stats count",now-1d,now,*,countRecord:count(*):*,eq,2,Splunk QL
 "* | head | stats count",now-1d,now,*,countRecord:count(*):*,eq,10,Splunk QL
 "city=Boston | head limit=3 | stats count",now-1d,now,*,countRecord:count(*):*,eq,3,Splunk QL
-"app_name = Bracecould | head gender=""female"" | stats count",now-1d,now,*,countRecord:count(*):*,eq,3,Splunk QL
-"app_name = Bracecould | head gender=""female"" keeplast=true | stats count",now-1d,now,*,countRecord:count(*):*,eq,4,Splunk QL
+"app_name = Bracecould | sort gender | head gender=""female"" | stats count",now-1d,now,*,countRecord:count(*):*,eq,3,Splunk QL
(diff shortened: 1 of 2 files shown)


cicd/ingest.csv
```diff
@@ -171,8 +171,8 @@ city=Boston | stats count AS Count BY http_method | eval newField=lower(http_met
 "* | head 2 | stats count",now-1d,now,*,countRecord:count(*):*,eq,2,Splunk QL
 "* | head | stats count",now-1d,now,*,countRecord:count(*):*,eq,10,Splunk QL
 "city=Boston | head limit=3 | stats count",now-1d,now,*,countRecord:count(*):*,eq,3,Splunk QL
-"app_name = Bracecould | head gender=""female"" | stats count",now-1d,now,*,countRecord:count(*):*,eq,3,Splunk QL
-"app_name = Bracecould | head gender=""female"" keeplast=true | stats count",now-1d,now,*,countRecord:count(*):*,eq,4,Splunk QL
+"app_name = Bracecould | sort gender | head gender=""female"" | stats count",now-1d,now,*,countRecord:count(*):*,eq,3,Splunk QL
+"app_name = Bracecould | sort gender | head gender=""female"" keeplast=true | stats count",now-1d,now,*,countRecord:count(*):*,eq,4,Splunk QL
```
AndrewHess, line 175: Why did you add sorting? Since this PR is about fixing an issue, you probably shouldn't change any of the tests unless there's a good reason to. But it looks like the query is supposed to give the same results either way, right?

AndrewHess (commented):
The segstructs.go change looks good. Were you able to verify that the change not only fixes the panic, but also that now the panicking queries give the correct results?

cicd/ingest.csv
vasukalariya: So, when multiple segments are used, the order of records is not certain. This test picks the records if gender is female. In the multiple segment case, the gender male often comes up as first record and this query would return zero. To have a consistent query for checking the results we would need a sort. In the default case when a single segment ingest is done. records where female is the gender are ingested first.

AndrewHess (approved):

PR #1394, opened 2024-08-07
Display query latency since install and query latency since last restart on cluster stats page

venkateshamatam:
# Description
Display query latency since install and query latency since last restart on the cluster stats page. 

- Query latency since install is calculated as: `(totalRespTime since install) / (queries since install)`
- Query latency since last restart is calculated as: `(totalRespTime since restart) / (queries since restart)`

# Testing
- Tested with existing records in `usageStats`.
- Tested without existing records in `usageStats`.
- When `queries since install` is 0, verified that `queries since install` equals the sum of `queries since restart` values across multiple restarts.

image

# Checklist:


pkg/health/clusterStatsHandler.go
```diff
@@ -94,10 +94,16 @@ func ProcessClusterStatsHandler(ctx *fasthttp.RequestCtx, myid uint64) {
...
+		httpResp.QueryStats["Average Query Latency (since install)"] = fmt.Sprintf("%v", utils.ToFixed(totalResponseTimeSinceInstall, 3)) + " ms"
+	}
+
 	if queryCount > 1 {
-		httpResp.QueryStats["Average Query Latency (since install)"] = fmt.Sprintf("%v", utils.ToFixed(totalResponseTime/float64(queryCount), 3)) + " ms"
+		httpResp.QueryStats["Average Query Latency (since restart)"] = fmt.Sprintf("%v", utils.ToFixed(totalResponseTimeSinceRestart/float64(queryCount), 3)) + " ms"
 	} else {
-		httpResp.QueryStats["Average Query Latency (since install)"] = fmt.Sprintf("%v", utils.ToFixed(totalResponseTime, 3)) + " ms"
+		httpResp.QueryStats["Average Query Latency (since restart)"] = fmt.Sprintf("%v", utils.ToFixed(totalResponseTimeSinceRestart, 3)) + " ms"
 	}
```
AndrewHess, line 107: On lines 97 and 103, why is the condition `> 1` instead of `> 0`?

pkg/ssa/ssa.go
```diff
@@ -376,17 +376,22 @@ func populateIngestSsa(m map[string]interface{}, myid uint64) {
...
 	m["region"] = IPAddressInfo.Region
 	m["country"] = IPAddressInfo.Country
 	if queryCount > 1 {
-		m["avg_query_latency_ms"] = fmt.Sprintf("%v", utils.ToFixed(totalResponseTime/float64(queryCount), 3)) + " ms"
+		m["avg_query_latency_ms_since_last_restart"] = fmt.Sprintf("%v", utils.ToFixed(totalResponseTimeSinceRestart/float64(queryCount), 3)) + " ms"
 	} else {
-		m["avg_query_latency_ms"] = fmt.Sprintf("%v", utils.ToFixed(totalResponseTime, 3)) + " ms"
+		m["avg_query_latency_ms_since_last_restart"] = fmt.Sprintf("%v", utils.ToFixed(totalResponseTimeSinceRestart, 3)) + " ms"
+	}
+	if querieSinceInstall > 1 {
```
AndrewHess, line 391: Same question here. Why not `> 0`?

pkg/usageStats/stats.go
```diff
@@ -123,12 +125,28 @@ func ReadQueryStats(orgid uint64) error {
 		return err
 	}
 	if len(val) > 0 {
-		flushedQueriesSinceInstall, err := strconv.ParseUint(val[len(val)-1][0], 10, 64)
+		lastRecord := val[len(val)-1]
+		if len(lastRecord) < 1 {
+			return fmt.Errorf("readQueryStats: last record has insufficient fields %v", lastRecord)
+		}
```
AndrewHess, line 131: You should also log this error, just in case the caller of this function doesn't log the error. You can do it like this:
```
return utils.TeeErrorf("readQueryStats: last record has insufficient fields %v", lastRecord)
```

Same comment some other places in this function

pkg/usageStats/stats.go
```diff
@@ -408,27 +427,30 @@ func UpdateMetricsStats(metricsBytesCount uint64, incomingMetrics uint64, orgid
...
 		}
 	}
-	atomic.AddUint64(&QueryStatsMap[orgid].QueryCount, queryCount)
-	atomic.AddUint64(&QueryStatsMap[orgid].QueriesSinceInstall, queryCount)
-	QueryStatsMap[orgid].TotalRespTime += totalRespTime
+	qs := QueryStatsMap[orgid]
+	atomic.AddUint64(&qs.QueryCount, queryCount)
+	atomic.AddUint64(&qs.QueriesSinceInstall, queryCount)
+	qs.TotalRespTimeSinceRestart += respTime
+	qs.TotalRespTimeSinceInstall += respTime
```
AndrewHess, line 453: Why do you have to atomically update `qs.QueryCount` but you don't have to atomically update `qs.TotalRespTimeSinceRestart`?

AndrewHess (commented):
Looks pretty good. My main question is about thread safety; the other comments are minor

pkg/health/clusterStatsHandler.go
venkateshamatam: We are using `>1` instead of `>0` to ensure that the average latency calculation is only performed when there are multiple queries and dividing by 1 is redundant because the result would be the same as the original value.

pkg/usageStats/stats.go
venkateshamatam: We atomically update `qs.QueryCount` because it is a `uint64` and Go provides atomic operations for integer types to ensure thread safety. We were previously doing the same too; we weren't using atomic updates for `qs.TotalRespTime` because Go does not provide atomic operations for floating-point types.

We can probably use a `sync.Mutex` on `QueryStatsMap` to ensure thread safety for updating `qs.TotalRespTimeSinceRestart` and `qs.TotalRespTimeSinceInstall`? What do you think?

AndrewHess: Yeah I'm not sure if this code will get called by multiple concurrent threads, but the atomic updates make me think it is. If it is, then yes we should use a lock for the floating point numbers. Or another option would be to change the internal structure so they we use small enough units for the floats that we can use ints instead and then convert back to the units they currently are (seconds?) whenever we show them (e.g., send them to frontend)

venkateshamatam: I have updated the code to ensure thread safety for the floating-point numbers by using a `sync.Mutex`

AndrewHess (approved):

PR #1398, opened 2024-08-07
Fix: Some queries were not working for multi segment/index data.

vasukalariya:
# Description
Summarize the change.
Setup
Upon ingesting the same amount of data across multiple indexes as given below

```
go run main.go ingest esbulk -n 10 -g benchmark -d http://localhost:8081/elastic -t 100_000
go run main.go query esbulk -d http://localhost:5122/ -f ../../cicd/ingest.csv
```

These 8 queries were failing. The rest of the queries were tested for 50+ runs and worked fine.

```
"app_name=Bracecould | bin span=5s http_status as bin_http | stats count by bin_http | stats count",now-1d,now,*,countRecord:count(*):*,eq,3,Splunk QL
"app_name=Bracecould | bin span=2log3 http_status | stats count by http_status | stats count",now-1d,now,*,countRecord:count(*):*,eq,2,Splunk QL
"app_name=Bracecould | stats values(eval(http_status > 400)) as values",now-1d,now,*,group:values:*,eq,1,Splunk QL
"app_name=Bracecould | stats values(eval(http_status+10)) as values",now-1d,now,*,group:values:*,eq,"311&nbsp312&nbsp410&nbsp510",Splunk QL
"app_name=Bracecould | stats values(eval(if(http_status > 400, 100, ""abc""))) as values",now-1d,now,*,group:values:*,eq,"100&nbspabc",Splunk QL
"app_name=Bracecould | stats values(eval(if(http_method = ""POST"", ""HEAD"", http_method))) as values",now-1d,now,*,group:values:*,eq,"HEAD&nbspPATCH&nbspPUT",Splunk QL
"city=Boston | eval nullField1=null() | eval nullField2=null() | eval nullStatus=nullif(http_status, 200) | fillnull value=FILLING nullField1 nullField2 | fields city, http_status, nullField1, nullField2, nullStatus | where http_status=200 | stats count(*)",now-1d,now,*,countRecord:count(*):*,eq,14,Splunk QL
"city=Boston | eval nullField1=null() | eval nullField2=null() | eval nullStatus=nullif(http_status, 200) | fillnull value=FILLING nullField1 nullField2 | fields city, http_status, nullField1, nullField2, nullStatus | where nullStatus=null() | stats count(*)",now-1d,now,*,countRecord:count(*):*,eq,14,Splunk QL
```

For bin queries, the issue was with the aggregations that is being done at a later stage i.e. from GetJsonFromAllRrc. For some segments the len of records was 0, so aggregators were being skipped from group by to measure. Ideally, we process the first groupby/measure aggregator for all segments and then rest of the ones are processed.
We cannot remove the check all-together to return the nodeResult if the performAggsOnRecs is true, because for default segmentStats or GroupBy queries with say a statistic expression, we would want to ignore this as computation is already done during the search phase. We only have to return the nodeResult directly when we are coming from GetJsonFromAllRrc and this is the only place from where we will send a non-nil recs map.

For values queries, the runningEvalStats was initialized to a different map rather than strSet. When runningEvalStats does not have a valid strSet we should allocate it with the strSet initialized before so that the same is being passed to ComputeAggEvalForValues.

For the nullif queries we need to returns all the records if there is a stats in the chain, earlier only the first 100 records were being fetched and results were computed. In multi index ingest the order of ingestion differed so results differed. Now, the count shows the correct value that is the number of records with http_status = 200. Corrected the test queries too.

For input lookup one edge case was there that if recordIndexInFinal does not have any entries i.e. the segment does not have any entries, the offset has to be 0 not 1.

# Testing
Before fix ran for 50+ iterations to identify the issues.
Ran the tests with the same kind of data ingest for 10+ iterations after fix. 

# Checklist:

### DIFF
--- a/pkg/segment/aggregations/segaggs.go
+++ b/pkg/segment/aggregations/segaggs.go
@@ -141,7 +141,7 @@ func PostQueryBucketCleaning(nodeResult *structs.NodeResult, post *structs.Query
 		if len(nodeResult.TransactionEventRecords) > 0 {
 			nodeResult.NextQueryAgg = agg
 			return nodeResult
-		} else if nodeResult.PerformAggsOnRecs && len(recs) > 0 {
+		} else if nodeResult.PerformAggsOnRecs && recs != nil {
 			nodeResult.NextQueryAgg = agg
 			return nodeResult
 		}
@@ -330,7 +330,7 @@ func performInputLookup(nodeResult *structs.NodeResult, agg *structs.QueryAggreg
 
 	// When the first block of the last segment arrives update the records and record index once
 	if !agg.GenerateEvent.InputLookup.UpdatedRecordIndex && agg.GenerateEvent.InputLookup.NumProcessedSegments == numTotalSegments-1 {
-		offset := 0
+		offset := -1
 		for _, recIndex := range recordIndexInFinal {
 			if recIndex > offset {
 				offset = recIndex
--- a/cicd/ingest.csv
+++ b/cicd/ingest.csv
@@ -171,8 +171,8 @@ city=Boston | stats count AS Count BY http_method | eval newField=lower(http_met
 "* | head 2 | stats count",now-1d,now,*,countRecord:count(*):*,eq,2,Splunk QL
 "* | head | stats count",now-1d,now,*,countRecord:count(*):*,eq,10,Splunk QL
 "city=Boston | head limit=3 | stats count",now-1d,now,*,countRecord:count(*):*,eq,3,Splunk QL
-"app_name = Bracecould | head gender=""female"" | stats count",now-1d,now,*,countRecord:count(*):*,eq,3,Splunk QL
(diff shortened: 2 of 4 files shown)


pkg/segment/aggregations/segaggs.go
```diff
@@ -141,7 +141,7 @@ func PostQueryBucketCleaning(nodeResult *structs.NodeResult, post *structs.Query
 		if len(nodeResult.TransactionEventRecords) > 0 {
 			nodeResult.NextQueryAgg = agg
 			return nodeResult
-		} else if nodeResult.PerformAggsOnRecs && len(recs) > 0 {
+		} else if nodeResult.PerformAggsOnRecs && recs != nil {
 			nodeResult.NextQueryAgg = agg
 			return nodeResult
 		}
```
AndrewHess, line 147: Just curious: does this change actually change any behavior? I thought Go implements `len()` so that `len(nil) == 0`, so the only potential difference is when `recs` is non-nil but has zero items, but you're not using `recs` inside the block. So it looks like nothing changes; or am I missing something?

AndrewHess (approved):
Looks good

PR #1406, opened 2024-08-08
PQS Delete API

vasukalariya:
# Description
Summarize the change.
Added the Delete API to remove the pqmr, pqmeta files.
Also calling pqsClear to update the pqinfo.bin and other persistent info files.

# Testing
Tested it by ingesting a bunch of data across different segments. 
Run few queries to check the pqids, meta and pqmr files being populated.
Call the delete API and verify that segmeta.json is updated and all the pqids, meta and pqmr files are deleted across all segments.

# Checklist:


pkg/segment/writer/segwriter.go
```diff
@@ -1042,29 +1044,76 @@ func BackFillPQSSegmetaEntry(segsetkey string, newpqid string) {
...
 		}
 	}
-	wfd, err := os.OpenFile(localSegmetaFname, os.O_WRONLY|os.O_CREATE|os.O_TRUNC, 0644)
+	err = WriteSegMeta(localSegmetaFname, preservedSmEntries)
 	if err != nil {
-		log.Errorf("BackFillPQSSegmetaEntry: Failed to open SegMetaFile name=%v, err:%v", localSegmetaFname, err)
-		return
+		log.Errorf("BackFillPQSSegmetaEntry: failed to write Segmeta: err=%v", err)
+	}
+}
```
AndrewHess, line 1051: nit: I'd recommend actually keeping the `return` in this error block, even though it is the last part of this function so it doesn't make any difference currently. My reasoning is that later if we add more logic to this function, we will want the `return` here if we get an error

pkg/segment/writer/segwriter.go
```diff
@@ -1042,29 +1044,76 @@ func BackFillPQSSegmetaEntry(segsetkey string, newpqid string) {
...
+
+func WriteSegMeta(segmetaFname string, segMetaEntries []*structs.SegMeta) error {
+	wfd, err := os.OpenFile(segmetaFname, os.O_WRONLY|os.O_CREATE|os.O_TRUNC, 0644)
```
AndrewHess, line 1054: nit: maybe rename this to something like `WriteOverSegMeta` or `ClearAndWriteSegMeta` or similar, just to clarify that the old segmeta file gets removed rather than appended to

pkg/segment/writer/segwriter.go
```diff
@@ -1042,29 +1044,76 @@ func BackFillPQSSegmetaEntry(segsetkey string, newpqid string) {
...
 
-	for _, smentry := range preservedSmEntries {
+	for _, smentry := range segMetaEntries {
 
 		segmetajson, err := json.Marshal(*smentry)
 		if err != nil {
-			log.Errorf("BackFillPQSSegmetaEntry: failed to Marshal: err=%v", err)
-			return
+			return fmt.Errorf("WriteSegMeta: failed to Marshal: segmeta filename=%v: err=%v", segmetaFname, err)
 		}
```
AndrewHess, line 1065: I'm not sure this `err` will give us enough info to debug. I think you should include in this log message the `smentry` that you tried to marshal

Also I'd recommend both logging and returning the error, just to make sure that we do log this error even if the caller of this function doesn't log the error. You can do this with:
```
return utils.TeeErrorf("WriteSegMeta: failed to Marshal: segmeta filename=%v: err=%v", segmetaFname, err)
```

pkg/server/query/entryHandlers.go
```diff
@@ -385,6 +386,17 @@ func postPqsClearHandler() func(ctx *fasthttp.RequestCtx) {
 	}
 }
 
+func postPqsDeleteHandler() func(ctx *fasthttp.RequestCtx) {
+	return func(ctx *fasthttp.RequestCtx) {
+		err := writer.DeletePQSData()
+		if err != nil {
+			utils.SetInternalError(ctx, "Error while deleting PQS data")
+			return
+		}
```
AndrewHess, line 395: you should also log this `err`. One option is to use `utils.SendInternalError()`:
```
		err := writer.DeletePQSData()
		if err != nil {
			utils.SendInternalError(ctx, "Error while deleting PQS data", "", err)
			return
		}
```
See https://github.com/siglens/siglens/blob/develop/pkg/utils/httpserverutils.go#L902-L941

AndrewHess (commented):
A few small comments, but this is looking good

pkg/segment/writer/segwriter.go
vasukalariya: Actually, the callers of this method are logging the error so that we know what was the source and prevent duplicate logs. What do you think?

pkg/server/query/entryHandlers.go
vasukalariya: The error are already logged in DeletePQSData, do we want duplicate logs ?

pkg/segment/writer/segwriter.go

pkg/segment/writer/segwriter.go

pkg/segment/writer/segwriter.go

pkg/server/query/entryHandlers.go
vasukalariya: addressed in the latest commit.

AndrewHess (approved):
Looks good

PR #1407, opened 2024-08-08
Enable PQS by default

Macbeth98:
# Description
- Changed the `server.yaml` to enable `pqs` by default

# Testing
- Tested by running the server and checked that the PQS is enabled.

# Checklist:


nkunal (dismissed):

AndrewHess:
Maybe also default it to `true` when it's not set in server.yaml.
https://github.com/siglens/siglens/blob/develop/pkg/config/config.go#L601-L603

nkunal (dismissed):

nkunal (approved):

PR #1408, opened 2024-08-08
Global Logger to reduce Log Counts

vasukalariya:
# Description
Summarize the change.
Added a global logger to log common error and prevent log fill up.
For now, the ErrorInfo struct has Count, later we can add more meta data like segmentkey etc.

# Testing
This particular query produces multiple "Did not find index for record identifier" errors for many records.
Verified the count and errors reported.
```
country=United* | fields app_name, app_version, country, city, user_email, http_status, http_method | transaction app_name startswith=eval(http_status=301 OR http_status=200) endswith=eval(http_status>=400)
```

# Checklist:

### DIFF
--- a/pkg/ast/pipesearch/wsSearchHandler.go
+++ b/pkg/ast/pipesearch/wsSearchHandler.go
@@ -283,6 +283,20 @@ func processQueryUpdate(conn *websocket.Conn, qid uint64, sizeLimit uint64, scro
 	return numRrcsAdded
 }
 
+func LogGlobalErrors(qid uint64) {
+	nodeRes, err := query.GetOrCreateQuerySearchNodeResult(qid)
+	if err != nil {
+		log.Errorf("LogGlobalErrors: Error getting query search node result for qid: %v", qid)
+		return
+	}
+	for errMsg, errInfo := range nodeRes.GlobalErrors {
+		if errInfo == nil {
+			continue
+		}
+		log.Errorf("qid=%v, %v, Count: %v", qid, errMsg, errInfo.Count)
+	}
+}
+
 func processCompleteUpdate(conn *websocket.Conn, sizeLimit, qid uint64, aggs *structs.QueryAggregators) {
 	queryC := query.GetQueryCountInfoForQid(qid)
 	totalEventsSearched, err := query.GetTotalsRecsSearchedForQid(qid)
@@ -333,6 +347,7 @@ func processCompleteUpdate(conn *websocket.Conn, sizeLimit, qid uint64, aggs *st
 			log.Errorf("qid=%d, processCompleteUpdate: failed to write error response to websocket! err: %+v", qid, wErr)
 		}
 	}
+	LogGlobalErrors(qid)
 }
 
 func processMaxScrollComplete(conn *websocket.Conn, qid uint64) {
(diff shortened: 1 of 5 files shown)


pkg/segment/structs/segstructs.go
```diff
@@ -439,6 +439,7 @@ type NodeResult struct {
 	CurrentSearchResultCount    int
 	AllSearchColumnsByTimeRange map[string]bool
 	FinalColumns                map[string]bool
+	GlobalErrors                map[string]*ErrorInfo // maps global error from error message -> error stats
```
AndrewHess, line 442: Should this be merged with the `ErrList` field, or should they stay separate? If they stay separate, I think it would be good to move one of these fields so they're both adjacent

pkg/ast/pipesearch/wsSearchHandler.go
```diff
@@ -333,6 +347,7 @@ func processCompleteUpdate(conn *websocket.Conn, sizeLimit, qid uint64, aggs *st
 			log.Errorf("qid=%d, processCompleteUpdate: failed to write error response to websocket! err: %+v", qid, wErr)
 		}
 	}
+	LogGlobalErrors(qid)
 }
```
AndrewHess, line 351: I'm worried about cases that don't get here, and so we don't ever print the error messages (e.g., the query times out or is canceled). Maybe a better way to do this would be to `defer` it in `ProcessPipeSearchWebsocket()` once we have the `qid`. On that note, you'll also want to ensure this gets called when we handle queries with `ProcessPipeSearchRequest()` (that's the handler for non-UI queries, like if you run a query via curl)

pkg/segment/reader/record/rrcreader.go
```diff
@@ -346,7 +346,7 @@ func GetJsonFromAllRrc(allrrc []*utils.RecordResultContainer, esResponse bool, q
 				// it's an async query we're running this function with
 				// len(segmap)=1 because we try to process the data as the
 				// searched complete.
-				log.Infof("qid=%d, GetJsonFromAllRrc: Did not find index for record indentifier %s.", qid, recInden)
+				nodeRes.StoreGlobalError("GetJsonFromAllRrc: Did not find index for record identifier")
```
AndrewHess, line 349: From the comment in the code, I think getting here is expected in some cases. Are you sure we want to upgrade this from an `Info` log to an `Error` log?

pkg/segment/structs/segstructs.go
nkunal: hmm.. ErrList should eventually be deprecated and we should move to `GlobalErrors`, primarily because ErrList is an array and we append too many of the same errors to it, it will get large. 

And maybe we should call this as `GlobalSearchErrors`
Agree with moving these two adjacent and adding a note about eventually replacing ErrList to this new var.

pkg/segment/structs/segstructs.go
```diff
@@ -461,6 +462,10 @@ type StringStats struct {
 	StrSet map[string]struct{}
 }
 
+type ErrorInfo struct {
```
nkunal, line 465: nit: maybe rename this to `SearchErrorInfo`

pkg/ast/pipesearch/wsSearchHandler.go
vasukalariya: I have moved the logging in DeleteQuery and CancelQuery. We cannot directly defer this function as query info would be deleted after DeleteQuery. This should handle the ProcessPipeSearchRequest case also because after Executing the query it calls DeleteQuery.

AndrewHess (approved):

PR #1414, opened 2024-08-09
SPL: Update multi value expressions from &nbsp strings to []strings

vasukalariya:
# Description
Summarize the change.
Updated the parser and implementation for multi-value expressions.
Updated the values to return an array of strings, not a &nbsp joined strings.
Updated the sigclient query checker to be able to verify a slice of string.

# Testing
Updated the parser tests, added extra tests in ingest.csv, restart.csv and evaluationstructs_test.go

# Checklist:


pkg/segment/structs/evaluationstructs.go
```diff
@@ -2364,6 +2442,10 @@ func (self *TextExpr) GetFields() []string {
 		if self.LengthExpr != nil {
 			fields = append(fields, self.LengthExpr.GetFields()...)
 		}
+		if self.MultiValueExpr != nil {
+			fields = append(fields, self.MultiValueExpr.GetFields()...)
+		}
```
AndrewHess, line 2447: This looks good. I think you also need this logic inside `ValueExpr.GetFields()`, right? Or maybe I missed that

tools/sigclient/pkg/query/query.go
```diff
@@ -541,6 +541,50 @@ func runContinuousQueries(client *http.Client, requestStr string, bearerToken st
...
+func createString(value []interface{}) string {
+	var str string
+	for i, v := range value {
+		str += fmt.Sprintf("%v", v)
+		if i != len(value)-1 {
+			str += ","
+		}
+	}
+	return str
+}
```
AndrewHess, line 553: This is alright, but I think this would be better:
```
valueStrings := make([]string, 0, len(value))
for _, item := value {
    valueStrings = append(valueStrings, fmt.Sprintf("%v", v))
}

return strings.Join(valueStrings, ",")
```

Also, since you only call this function in one place, you might want to inline this code at that spot instead of making this function. You can have this separate function if you want though; in that case, I'd recommend renaming `createString` and `value` to something more clear

AndrewHess (commented):
Nice work, this looks almost ready to merge

pkg/segment/structs/evaluationstructs.go
vasukalariya: ValueExpr.GetFields() already has this [logic](https://github.com/siglens/siglens/blob/5b4465a6e8298316723cf9c3543bbafa03446fe2/pkg/segment/structs/evaluationstructs.go#L1109) to call the GetFields of MultiValueExpr which will handle this.

Although, I missed this [check](https://github.com/siglens/siglens/blob/5b4465a6e8298316723cf9c3543bbafa03446fe2/pkg/segment/structs/evaluationstructs.go#L1124) and have added it.

Addressed in the latest commit.

tools/sigclient/pkg/query/query.go
vasukalariya: Addressed in the latest commit.

AndrewHess (approved):
Looks good

PR #1424, opened 2024-08-09
Faster stats eval

AndrewHess:
# Description
This improves the performance of SPL queries that have `eval` inside `stats`, like this:
```
* | stats count(eval(http_status=500)) as "Internal Server Errors"
```
It does this essentially by converting it to this query instead:
```
http_status=500 | stats count(eval(http_status=500)) as "Internal Server Errors"
```
The actual query doesn't get modified though; we parse it to `ast.Node` and `QueryAggregators` like normal, but then modify those so that they're what would have been generated from the optimized query.

# Testing
## Correctness
New unit tests, and a some new e2e tests to ensure I didn't break something.

## Performance
I ran this query on my local machine with 110 million rows:
```
* | stats count(eval(http_status=500)) as "Internal Server Errors"
```
With this PR, it took 12.6 seconds; without this PR, it took 75.9 seconds. The result was the same.

# Checklist:


cicd/ingest.csv
```diff
@@ -262,6 +262,10 @@ city=Boston | stats count AS Count BY http_method | eval newField=lower(http_met
 "app_name=Bracecould | stats count(eval(http_status > 300 AND http_method=""PATCH"")) as count BY http_method",now-1d,now,*,group:count:PATCH,eq,1,Splunk QL
 "app_name=Bracecould | stats values(eval(if(http_status > 302, ""Error"", http_status))) as values BY http_method",now-1d,now,*,group:values:POST,eq,"301&nbspError",Splunk QL
 "app_name=Companywill | stats dc(eval(if(http_status > 310, ""Error"", http_status))) as dc by http_method",now-1d,now,*,group:dc:DELETE,eq,2,Splunk QL
+"* | stats count(eval(http_status=500)) as count",now-1d,now,*,group:count:*,eq,"16,711",Splunk QL
+"* | stats count(eval(http_status=500)) as badCount, count(eval(http_status=200)) as goodCount",now-1d,now,*,group:goodCount:*,eq,"16,763",Splunk QL
+"http_status > 200 | stats count(eval(http_status<400)) as count",now-1d,now,*,group:count:*,eq,"33,361",Splunk QL
+"http_status > 200 | eval http_status=200 | stats count(eval(http_status=200))",now-1d,now,*,group:count:*,eq,"83,237",Splunk QL
```
nkunal, line 268: we should also add the query that would not go through this new logic: `* | stats count(eval(http_status=500+100)) as "Internal Server Errors"` OR `* | eval http_status=http_status/2 | stats count(eval(http_status=500))`

nkunal (commented):
overall looks good. Is there a way where we can log a logline, if we do end up optimizing the query. It will help in debugging, if the query AST nodes were modified

nkunal (approved):
LGTM, approving it. 
the couple of feedbacks about, few more queries in cicd and the log line to indicate query ast was modified can come in next PR

PR #1425, opened 2024-08-09
Migrate Update PQIDs from wipBlock to segStore

vasukalariya:
# Description
Summarize the change.
Migrated code to update PQIDs from resetWipBlock to resetSegStore

# Testing
Checked the promotion of PQIDs when new segments are ingested.

# Checklist:


pkg/utils/maputils.go
```diff
@@ -65,3 +65,25 @@ func CreateRecord(columnNames []string, record []string) (map[string]interface{}
 	}
 	return recordMap, nil
 }
+
+// SetDifference returns the added and removed elements between two sets.
+func SetDifference[T1, T2 any](newSet map[string]T2, oldSet map[string]T1) ([]string, []string, bool) {
```
nkunal, line 70: add UTs for this function

pkg/segment/writer/segstore.go
```diff
@@ -280,7 +280,31 @@ func (segstore *SegStore) resetSegStore(streamid string, virtualTableName string
...
+				mrSize = uint(numPrevRec)
+			}
+			segstore.pqMatches[pqid] = pqmr.CreatePQMatchResults(mrSize)
+		}
+		segstore.pqTracker.addSearchNode(pqid, pNode)
+	}
+
+	promoted, demoted, change := toputils.SetDifference(segstore.pqMatches, segstore.LastSegPqids)
+	if change {
+		log.Infof("resetSegStore: PQIDs Promoted: %v, Demoted: %v", promoted, demoted)
```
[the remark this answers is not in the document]

pkg/utils/maputils.go
vasukalariya: addressed in the latest commit.

pkg/utils/maputils.go
```diff
@@ -69,21 +69,20 @@ func CreateRecord(columnNames []string, record []string) (map[string]interface{}
...
 	for key := range newSet {
 		added = append(added, key)
 		if _, exists := oldSet[key]; !exists {
-			change = true
+			newEntries = true
 		}
 	}
 
-	return added, removed, change
+	return added, removed, newEntries
```
AndrewHess, line 87: nit: for a boolean, I think `newEntries` is a little unclear because (without type information) it could either mean "a list of keys that were added" or "the sets are different". Maybe `hasChange` or `setsDiffer` would be slightly better.

Also, is this `newEntries` variable actually useful? It seems like the caller could get the same info from `len(added) > 0`, right?

pkg/segment/writer/segwriter.go
```diff
@@ -445,7 +443,7 @@ func getSegStore(streamid string, ts_millis uint64, table string, orgId uint64)
 			return nil, fmt.Errorf("getSegStore: max allowed segstores reached (%d)", maxAllowedSegStores)
 		}
 
-		segstore = &SegStore{Lock: sync.Mutex{}, OrgId: orgId, firstTime: true}
+		segstore = &SegStore{Lock: sync.Mutex{}, OrgId: orgId, firstTime: true, pqMatches: make(map[string]*pqmr.PQMatchResults), LastSegPqids: make(map[string]struct{})}
```
AndrewHess, line 446: nit: now that this has gotten a bit longer, it's probably worth putting each field on a separate line

pkg/utils/maputils.go
AndrewHess: nit: this could be a little more generic like this:
```suggestion
func SetDifference[K comparable, T1, T2 any](newSet map[K]T2, oldSet map[K]T1) ([]K, []K, bool) {
```
but it's probably fine as it is since we can always make it more generic later if we actually need it to be

pkg/utils/maputils.go
vasukalariya: No len(added) will always be greater than 0, if PQIDs are there. Added records will actually be old records that are present in the newSet + newly added records in the newSet.

AndrewHess: Oh, is that actually what you want to return (I'm talking about the `added` variable)? If it is, then you need to change some documentation to clarify that:
- `added` should be renamed to something indicating that it's all of the keys in the new set, not just the ones added to the old set to make the new set
- In you function comment: `// SetDifference returns the added and removed elements between two sets.` you should also make that clarification

pkg/utils/maputils.go
vasukalariya: addressed in the latest commit.

pkg/utils/maputils.go
vasukalariya: Upon clarification, we will only log changes that means we will only log newly added entries in promoted and removed entries in demoted. Have removed this newEntries/change flag and updated the function and tests for the same.

pkg/segment/writer/segwriter.go

AndrewHess (commented):
Looks good. @nkunal do you want to do a test of this before it gets merged?

nkunal (dismissed):

nkunal:
@AndrewHess this is good to be merged . @vasukalariya please resolve the conflicts and merge

AndrewHess (approved):

PR #1428, opened 2024-08-11
avoid revereKeyIndex lookups per record and do it once every block

nkunal:
# Description

During multicol file reading, record by record, pull out the fileReader ptr that is inside MultiColReader for a column by passing in the keyIndex, instead of doing a lookup in a colNameToKeyIndex every record, this way we only find all the keyIndices once every block instead of once every record

With this change, if there are `5 Million` records and `270` blocks in one segfile then we do the 
`sfr := mcsr.allColsReverseIndex[cName]` 270 times instead of 5 Million times per segfile

# Testing

# Checklist:

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -508,7 +508,7 @@ func GetTestConfig(dataPath string) common.Configuration {
 
 func InitializeTestingConfig(dataPath string) {
 	InitializeDefaultConfig(dataPath)
-	SetDebugMode(true)
+	SetDebugMode(false)
 }
 
 func ReadRunModConfig(fileName string) (common.RunModConfig, error) {
--- a/pkg/segment/reader/record/recordreader.go
+++ b/pkg/segment/reader/record/recordreader.go
@@ -254,8 +254,6 @@ func readAllRawRecords(orderedRecNums []uint16, blockIdx uint16, segReader *segr
 
 			if cname == config.GetTimeStampKey() {
 				isTsCol = true
-			} else {
-				isTsCol = false
 			}
 
 			cValEnc, err := segReader.ExtractValueFromColumnFile(colKeyIdx, blockIdx, recNum,
--- a/pkg/segment/reader/record/recordreader.go
+++ b/pkg/segment/reader/record/recordreader.go
@@ -205,6 +205,7 @@ func readAllRawRecords(orderedRecNums []uint16, blockIdx uint16, segReader *segr
 	results := make(map[uint16]map[string]interface{})
 
 	dictEncCols := make(map[string]bool)
+	allColKeyIndices := make(map[int]string)
 	for _, colInfo := range segReader.AllColums {
 		col := colInfo.ColumnName
 		if !esQuery && (col == "_type" || col == "_id") {
@@ -220,6 +221,10 @@ func readAllRawRecords(orderedRecNums []uint16, blockIdx uint16, segReader *segr
 			dictEncCols[col] = true
 			allMatchedColumns[col] = true
 		}
+		cKeyidx, ok := segReader.GetColKeyIndex(col)
+		if ok {
+			allColKeyIndices[cKeyidx] = col
+		}
 	}
 
 	var mathColMap map[string]int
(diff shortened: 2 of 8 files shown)


pkg/segment/search/conditioncheck.go
```diff
@@ -141,24 +128,17 @@ func ApplyColumnarSearchQuery(query *SearchQuery, multiColReader *segread.MultiC
...
-			// we skip rawsearching for columns that are dict encoded,
-			// since we already search for them in the prior call to applyColumnarSearchUsingDictEnc
-			_, ok := dictEncColNames[colInfo.ColumnName]
-			if ok {
-				continue
-			}
+		for colKeyIndex := range allColKeyIndices {
 
-			rawColVal, err := multiColReader.ReadRawRecordFromColumnFile(colInfo.ColumnName, blockNum, recordNum, qid)
+			rawColVal, err := multiColReader.ReadRawRecordFromColumnFile(colKeyIndex, blockNum, recordNum, qid, false)
```
Macbeth98, line 141: Why are we searching `allColkeyIndices` when we are supposed to skip raw searching for `dictEncColNames`.

In `filtersearch.go` where we are building the `allColKeyIndices`, we used `multiColReader.AllColums` which will contain all the columns.  Check the code [here](https://github.com/siglens/siglens/pull/1428/files#diff-1c06c829ef6cb4902da6c66dee810690b4f50a3cd971683a29bba74f52cdff6eR200-R210).

But we just need to loop through over `cmiPassedNonDictColKeyIndices`, since we are supposed to skip `dictEncColNames`?  

If my reasoning is true we can avoid creating the `allColKeyIndices` altogether, I think.

pkg/segment/reader/record/recordreader.go
```diff
@@ -233,42 +238,49 @@ func readAllRawRecords(orderedRecNums []uint16, blockIdx uint16, segReader *segr
...
 			if ok {
 				continue
 			}
 
-			cValEnc, err := segReader.ExtractValueFromColumnFile(col, blockIdx, recNum, qid)
+			if cname == config.GetTimeStampKey() {
+				isTsCol = true
+			} else {
+				isTsCol = false
+			}
```
Macbeth98, line 259: nit: We don't need to the `else` block here. As by default `isTsCol` value is `false`.

Macbeth98: Same comment here also: https://github.com/siglens/siglens/pull/1428/files#diff-9ecb737019841f5110cc60e7192afc17333e9b7d86756fb829d607417d80683dR195-R197

pkg/segment/search/searchaggs.go
```diff
@@ -177,10 +177,41 @@ func addRecordToAggregations(grpReq *structs.GroupByRequest, timeHistogram *stru
...
+	} else {
+		for _, col := range grpReq.GroupByColumns {
+			cKeyidx, ok := multiColReader.GetColKeyIndex(col)
+			if ok {
+				groupbyColKeyIndices[cKeyidx] = struct{}{}
+			}
+		}
+	}
+
+	measColKeyIdxAndIndices := make(map[int][]int)
```
Macbeth98, line 207: nit: It might be better to rename this as `measureColKeyIdxAndIndices`

pkg/segment/search/conditioncheck.go
nkunal: previously for `MatchDictArrayAllColumns` queryType we were doing `for _, colInfo := range multiColReader.AllColums {`

so I kept the same case here. For dictArray searching we do not skip dictEncColNames,

Macbeth98: yes, we are looping through `multiColReader.AllColums` but inside that we are checking if the column exists in `dictEncColNames` and if `true` we are skipping the raw search. See the below screenshot for `MatchDictArrayAllColumns`. See the below screenshot or check the respective code here: https://github.com/siglens/siglens/blob/develop/pkg/segment/search/conditioncheck.go#L152-L160


image

nkunal: yup, you are right, the previous PR removed that dict_enc check and moved it one stack higher but that PR did not adjust this part of the code. Will fix it

pkg/segment/reader/record/recordreader.go
nkunal: addressed in recent commit

pkg/segment/search/searchaggs.go
nkunal: done

Macbeth98 (approved):

nkunal:
@AndrewHess I am merging this PR, but you should still review and add your feedback here. I will address them in next PR.

pkg/config/config.go
```diff
@@ -508,7 +508,7 @@ func GetTestConfig(dataPath string) common.Configuration {
 
 func InitializeTestingConfig(dataPath string) {
 	InitializeDefaultConfig(dataPath)
-	SetDebugMode(true)
+	SetDebugMode(false)
 }
```
AndrewHess, line 512: Why do you we want debug off for testing?

pkg/segment/reader/record/recordreader.go
AndrewHess: I'm pretty sure removing this `else` adds a bug. `isTsCol` is defined outside the loops, so once it gets set to `true` it's never reset to `false`. Am I missing something? You could also do
```
isTsCol = (cname == config.GetTimeStampKey())
```
unless you think that's less readable 

I'm just referring to the code here; I think the other code Mani linked is fine since there's not a loop

pkg/segment/reader/segread/multicolreader.go
```diff
@@ -89,6 +90,7 @@ func initNewMultiColumnReader(segKey string, colFDs map[string]*os.File, blockMe
 		allColsReverseIndex: colRevserseIndex,
 		timeStampKey:        tsKey,
 		segKey:              segKey,
+		maxColIdx:           -1,
```
AndrewHess, line 93: nit: it looks like `-1` is a special value to indicate it's invalid. This would be a great place to use `utils.Option[int]` :) But I don't think it's necessary to change the code since you already have it this way

PR #1437, opened 2024-08-12
Use pre-alloc byte slice instead of buf.Bytes during AddMeasureToRes

nkunal:
# Description
For each bucket key we create during historgram queries, we were using bytes.Buffer and growing it by adding all the groupby cols to it. This bytes buffer was reset on every record. Now pre-allocate a byte slice per block worker and reuse that bytes buffer during each record processing. This helps in runtime memory allocations.

# Testing

# Checklist:

### DIFF
--- a/pkg/segment/search/searchaggs.go
+++ b/pkg/segment/search/searchaggs.go
@@ -273,9 +273,9 @@ func addRecordToAggregations(grpReq *structs.GroupByRequest, timeHistogram *stru
 
 			// resize the working buf if we cannot accomodate the max value of any
 			// column's record
-			if len(aggsKeyWorkingBuf) < len(groupbyColKeyIndices) * 65_000 {
+			if len(aggsKeyWorkingBuf) < len(groupbyColKeyIndices)*65_000 {
 				aggsKeyWorkingBuf = toputils.ResizeSlice(aggsKeyWorkingBuf,
-					len(groupbyColKeyIndices) * 65_000)
+					len(groupbyColKeyIndices)*65_000)
 			}
 			for _, colKeyIndex := range groupbyColKeyIndices {
 				rawVal, err := multiColReader.ReadRawRecordFromColumnFile(colKeyIndex, blockNum, recNum, qid, false)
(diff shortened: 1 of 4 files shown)


nkunal:
@AndrewHess this PR is ready for review

pkg/segment/search/searchaggs.go
```diff
@@ -264,13 +272,22 @@ func addRecordToAggregations(grpReq *structs.GroupByRequest, timeHistogram *stru
 				}
 			}
 		} else {
+
+			// resize the working buf if we cannot accomodate the max value of any
+			// column's record
+			if len(aggsKeyWorkingBuf) < len(groupbyColKeyIndices)*65_000 {
+				aggsKeyWorkingBuf = toputils.ResizeSlice(aggsKeyWorkingBuf,
+					len(groupbyColKeyIndices)*65_000)
+			}
```
AndrewHess, line 281: Where is 65,000 coming from? Maybe this should be a constant

AndrewHess (dismissed):
Looks good; one question

pkg/segment/search/searchaggs.go
```diff
@@ -264,13 +272,22 @@ func addRecordToAggregations(grpReq *structs.GroupByRequest, timeHistogram *stru
 				}
 			}
 		} else {
+
+			// resize the working buf if we cannot accomodate the max value of any
+			// column's record
+			if len(aggsKeyWorkingBuf) < len(groupbyColKeyIndices)*65_000 {
+				aggsKeyWorkingBuf = toputils.ResizeSlice(aggsKeyWorkingBuf,
+					len(groupbyColKeyIndices)*65_000)
```
Macbeth98, line 280: nit: It would be better if we use variable or a constant such as `MAX_COL_VALUE_SIZE` or something instead of directly using `65_000`.

pkg/segment/search/searchaggs.go
nkunal: it is a max value of any column value, switching it to a constant in the next commit

pkg/segment/search/searchaggs.go
nkunal: yup fixed it in latest commit

Macbeth98 (approved):

PR #1439, opened 2024-08-13
reuse cval per segment and avoid using it per rec

nkunal:
# Description

Reuse Cval in each blockworker/segment worker while extracting values from csg files. By doing this we save 10 GB of runtime mem alloc on a 456 million data size for a query like `"* | stats avg(http_status) by hobby, http_method`

# Testing

# Checklist:


Macbeth98 (approved):
Looks Good! 

Do you think we should add a `nil` checks for `CVal` in `ExtractValueFromColumnFile` and `GetCvalFromRec`?

As of now, with the current code, I think we are sure, there will be no `nil` issue. So may be we don't need it for now.

PR #1450, opened 2024-08-13
Use pre-compiled regex for applying match filter on records

vasukalariya:
# Description
Summarize the change.
Pass the compiled regex with a fast flag for applying match filter on records. Compile the regex at block level.

CPU Time
                flat flat%  sum%    cum  cum%
Before: 19.52s 3.92% 58.88%  135.90s 27.32%
After: 15.23s 3.13% 62.73%  128.84s 26.46%
Before Total time: 161.73s
After Total time: 158.17s

Memory Usage:
No change, used 512.06kB memory in regexp.Match for both the cases.

Summary:
Total time reduced by ~3s.
Time reduced in this function ~4s
Memory consumption: No change

# Testing

# Checklist:

### DIFF
--- a/pkg/segment/reader/segread/dechecker.go
+++ b/pkg/segment/reader/segread/dechecker.go
@@ -40,15 +40,15 @@ returns:
 */
 func (sfr *SegmentFileReader) ApplySearchToMatchFilterDictCsg(match *structs.MatchFilter,
 	bsh *structs.BlockSearchHelper) (bool, error) {
-	var regexp *regexp.Regexp
+	var compiledRegex *regexp.Regexp
 	var err error
 
 	if len(match.MatchWords) == 0 {
 		return false, nil
 	}
 
-	if match.MatchType == structs.MATCH_PHRASE && match.MatchOperator == utils.And {
-		regexp, err = match.GetRegexp()
+	if match.MatchType == structs.MATCH_PHRASE {
+		compiledRegex, err = match.GetRegexp()
 		if err != nil {
 			log.Errorf("ApplySearchToMatchFilterDictCsg: error getting match regex: %v", err)
 			return false, err
@@ -56,7 +56,7 @@ func (sfr *SegmentFileReader) ApplySearchToMatchFilterDictCsg(match *structs.Mat
 	}
 
 	for dwordIdx, dWord := range sfr.deTlv {
-		matched, err := writer.ApplySearchToMatchFilterRawCsg(match, dWord, regexp, true)
+		matched, err := writer.ApplySearchToMatchFilterRawCsg(match, dWord, compiledRegex)
 		if err != nil {
 			return false, err
 		}
--- a/pkg/segment/reader/segread/dechecker.go
+++ b/pkg/segment/reader/segread/dechecker.go
@@ -19,10 +19,12 @@ package segread
 
 import (
 	"errors"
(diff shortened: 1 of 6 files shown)


pkg/segment/search/filtersearch.go
```diff
@@ -201,11 +205,19 @@ func filterRecordsFromSearchQuery(query *structs.SearchQuery, segmentSearch *Seg
 			queryInfoColKeyIndex = cKeyidx
 		}
 
+		if query.MatchFilter != nil && query.MatchFilter.MatchType == structs.MATCH_PHRASE && query.MatchFilter.MatchOperator == utils.And {
+			regexp, err = query.MatchFilter.GetRegexp()
+			if err != nil {
+				log.Errorf("ApplySearchToMatchFilterRawCsg: error getting match regex: %v", err)
```
nkunal, line 211: log msg funcname is not correct

pkg/segment/reader/segread/dechecker.go
```diff
@@ -38,13 +40,23 @@ returns:
...
 func (sfr *SegmentFileReader) ApplySearchToMatchFilterDictCsg(match *structs.MatchFilter,
 	bsh *structs.BlockSearchHelper) (bool, error) {
+	var regexp *regexp.Regexp
+	var err error
 
 	if len(match.MatchWords) == 0 {
 		return false, nil
 	}
 
+	if match.MatchType == structs.MATCH_PHRASE && match.MatchOperator == utils.And {
```
nkunal, line 50: we can remove the `AND` operator check here (even though we had that check in the main method). So that in future we don't have to worry about changing this path. Basically we should pre-compile the regexp if the matchTYpe is match_phrase

pkg/segment/writer/rawchecker.go
```diff
@@ -61,10 +65,12 @@ func ApplySearchToMatchFilterRawCsg(match *MatchFilter, col []byte) (bool, error
 	if match.MatchOperator == And {
 		var foundQword bool = true
 		if match.MatchType == MATCH_PHRASE {
-			regexp, err := match.GetRegexp()
-			if err != nil {
-				log.Errorf("ApplySearchToMatchFilterRawCsg: error getting match regex: %v", err)
-				return false, err
+			if !fast {
```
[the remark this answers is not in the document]
vasukalariya: It can try to fire the GetRegexp calculation for queries where regexp is nil/does not exist. Although GetRegexp will check for it. I thought this would be more clear.

pkg/segment/search/filtersearch.go
vasukalariya: addressed in the latest commit.

pkg/segment/writer/rawchecker.go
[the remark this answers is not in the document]
vasukalariya: addressed in the latest commit.

pkg/segment/reader/segread/dechecker.go
vasukalariya: addressed in the latest commit.

pkg/segment/writer/rawchecker.go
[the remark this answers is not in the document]
nkunal: Good point

nkunal (approved):

PR #1464, opened 2024-08-14
Added Lookups UI Support

sonamgupta21:
# Description
Added UI Support to Upload `.csv` and `.csv.gz` lookup files and to see all the uploaded lookup file names.

image
image

# Testing

# Checklist:


pkg/lookups/lookups.go
```diff
@@ -0,0 +1,153 @@
...
+	if !strings.HasSuffix(strings.ToLower(fileName), allowedExtCSV) &&
+		!strings.HasSuffix(strings.ToLower(fileName), allowedExtCSVGZ) {
+		if isCSVGZ {
+			fileName += allowedExtCSVGZ
+		} else {
+			fileName += allowedExtCSV
+		}
+	}
+
+	fullLookupsDir := filepath.Join(config.GetDataPath(), lookupsDir)
```
AndrewHess, line 74: There's already a function for this:
```suggestion
	fullLookupsDir := config.GetLookupPath()
```

And then I think you can delete the `lookupsDir` global constant

pkg/lookups/lookups.go
```diff
@@ -0,0 +1,153 @@
...
+		ctx.Error("Error creating lookups directory", fasthttp.StatusInternalServerError)
+		return
+	}
+
+	dstPath := filepath.Join(fullLookupsDir, fileName)
+	if _, err := os.Stat(dstPath); !os.IsNotExist(err) {
+		log.Errorf("UploadLookupFile: File already exists: %s", fileName)
+		ctx.Error("A file with the same name already exists", fasthttp.StatusConflict)
+		return
+	}
```
AndrewHess, line 86: Do we want to throw an error here always, or provide a way to write over the existing file? If we can't write over a file, they'd have to manually delete the file and then upload the new version which seems a little annoying. But if we do want to add a way to overwrite a file, I think we could do that in a separate small PR if you want

AndrewHess (commented):
It would be good to add the new APIs to the API_README.md file

I didn't check the frontend code

sunitakawane (approved):

pkg/lookups/lookups.go
sonamgupta21: I'll address this in a new PR.

PR #1467, opened 2024-08-14
Optimize Reduce function by reusing CVal

vasukalariya:
# Description
Summarize the change.
Reuse the CVal enclosures instead of creating new ones.
Query = `* | stats sum(http_status) by hobby, http_method`

Performance improvement is a bit tricky in this case. Here the major time taken is during the operation, in the above example it is taken by the addition operation in the Reduce function and this is quite random.

Ran multiple tests for 10 iterations on around 100M events.
The minimum execution time difference observed between before and after change for Reduce function is around 2.5s (75.68s to 72.97s). And same is the case for for the place where we have optimized by resuing the CVal, the difference is about 2s.
In very rare executions, I have observed the original code to perform better than this one. But, it could majorly be due to operation being executed faster (which is a bit random).

For memory
Have seen reduction in memory usage from 7.49GB to 7.44GB.

# Testing

# Checklist:


AndrewHess (approved):
Looks good

PR #1468, opened 2024-08-14
Accumulate and log the errors at segment level during ingestion for PQS queries

vasukalariya:
# Description
Summarize the change.
Earlier applySearchSingleQuery function was causing log flood due to errors.
We removed the error logs in this [PR](https://github.com/siglens/siglens/pull/1418). Fixed the logging from record level to segment level.
Added a global logger for segStore which will accumulate the errors and log it when it gets reset. 

# Testing
Tested it manually by throwing the error for a certain type of query.
Segment logger seems to count and log the errors appropriately before rotation.
image

# Checklist:


pkg/segment/structs/segstructs.go
```diff
@@ -465,6 +465,7 @@ type StringStats struct {
 type SearchErrorInfo struct {
 	Count    int
 	LogLevel log.Level
+	Error    error // extra information about the error
```
AndrewHess, line 468: nit: this comment doesn't seem to add info, so you don't really need it

pkg/segment/structs/segstructs.go
```diff
@@ -1196,14 +1197,20 @@ func (qa *QueryAggregators) IsStatsAggPresentInChain() bool {
 	return qa.HasInChain(statsAggPresentInCur)
 }
 
-func (nodeRes *NodeResult) StoreGlobalSearchError(errMsg string, logLevel log.Level) {
-	if nodeRes.GlobalSearchErrors == nil {
-		nodeRes.GlobalSearchErrors = make(map[string]*SearchErrorInfo)
+func (nodeRes *NodeResult) StoreGlobalSearchError(errMsg string, logLevel log.Level, err error) {
+	nodeRes.GlobalSearchErrors = StoreError(nodeRes.GlobalSearchErrors, errMsg, logLevel, err)
+}
```
AndrewHess, line 1202: `errMsg` and `err` seem redundant. Can we just keep `err`, or is there a reason we need `errMsg` as well?

pkg/segment/writer/segstore.go
```diff
@@ -282,6 +296,11 @@ func (segstore *SegStore) resetSegStore(streamid string, virtualTableName string
 	segstore.wipBlock.colWips = make(map[string]*ColWip)
 	segstore.wipBlock.clearPQMatchInfo()
 
+	if segstore.SegmentErrors != nil {
+		segstore.LogSegmentErrors()
+		segstore.SegmentErrors = nil
+	}
```
AndrewHess, line 302: nit: If we don't have to log errors multiple times, it might make sense to rename `LogSegmentErrors()` to `FlushSegmentErrors()` or `LogAndFlushSegmentErrors()`—actually, you probably don't need "Segment" because it's a method on a segstore, which is already only for one segment—and then just call that function here.

This might just be my personal preference, but I also think you don't need this check `segstore.SegmentErrors != nil` because it's easier for the reader if they don't need to know that `LogSegmentErrors()` uses `SegmentErrors`; `LogSegmentErrors()` should just handle that case (e.g., do nothing and don't panic)

pkg/segment/structs/segstructs.go
vasukalariya: `errMsg` is a common message, whereas `err` would have extra details about the error. We cannot always directly use `err` because it could be the case that somewhere we are printing record-specific details. If we map using this error, the keys could be unique and can result in a log flood. The reason for keeping err is that we can get extra details about this error if present and will be stored [once](https://github.com/siglens/siglens/blob/dd55833aba79270dbf78ad634d90dcd42683148e/pkg/segment/structs/segstructs.go#L1210).

pkg/segment/structs/segstructs.go
vasukalariya: addressed in the latest commit.

pkg/segment/writer/segstore.go

AndrewHess (approved):

PR #1469, opened 2024-08-14
Use cache for storing groupby values

AndrewHess:
# Description
In this query:
```
* | timechart avg(latency) by http_method span=1h
```
A significant amount of time is spent in `ConvertGroupByKey()` extracting the group-by value for a certain record. This PR adds a cache to reduce redundant computation.

# Testing
I profiled the above query on my local machine with 200 million records. Without this PR it took about 10.9 seconds and used 10.6 GB of memory total. With this PR it took about 9.0 seconds and used 6 GB of memory total.

# Alternatives Considered
I tried several other ways of optimizing this:
- Use a global cache: with this there's some increased complexity because you have to deal with concurrency and cache invalidation. I didn't attempt cache invalidation in my proof of concept work. I did find that a normal `map` with a `sync.RWMutex` was actually quite a bit slower than the current `develop` code, due to a lot of locking and releasing (with read locks). Using `sync.Map` was a bit better than `develop`, but worse than this PR. And a procedure where readers don't have to get a lock to check the cache was a bit faster than `sync.Map` but added more complexity (I think it was also slower than this PR); this procedure used a read-only cache and a write-only cache, and every so often (10 ms), all the entries in the read-only cache would be written to the write-only cache and then the caches would get swapped.
- I also tried using a sync.Pool inside `ConvertGroupByKey()` to store `strings.Builder` structs, but that didn't improve performance. In fact, `Builder.Reset()` actually discards the underlying buffer, so it gets garbaged collected anyways. I tried with my own implementation of a similar struct where `Reset()` would just reset the length of the underlying buffer (but keep the capacity, so it doesn't get garbage collected), but that also did not improve performance.

# Checklist:


pkg/segment/search/searchaggs.go
```diff
@@ -254,9 +256,20 @@ func addRecordToAggregations(grpReq *structs.GroupByRequest, timeHistogram *stru
...
+						if err != nil {
+							log.Errorf("addRecordToAggregations: failed to extract raw key: %v", err)
+						}
+
+						// I'm pretty sure we need to actually copy the string
+						// here to insert it into the map, since we made the
+						// string previously with an unsafe conversion from a
+						// []byte, and that []byte will change later.
+						rawValStr = string(rawVal)
+						groupByCache[rawValStr] = strs
```
nkunal, line 272: if we have failed to convertGroupBY.. then do we still want to insert it into the cache ?

nkunal (dismissed):
one minor comment , all else looks good

pkg/segment/search/searchaggs.go
AndrewHess: Good question. I'm actually not sure what we should do. Currently `ConvertGroupByKey` is deterministic and doesn't rely on any external state, so if it fails for a certain input, it will always fail on that input. So caching here is safe for now, but may change if `ConvertGroupByKey` changes

In general though maybe we wouldn't want to cache results that return an error because it could have errored due to some external state. What do you think we should do here?

nkunal: for now we should not store it in cache

nkunal (approved):

nkunal (dismissed):

nkunal (approved):

PR #1470, opened 2024-08-15
My Org Usage Stats takes too much time to load

vasukalariya:
# Description
Summarize the change.
The main bottleneck here is reading the segmeta files, which is happening 4 times: twice for ingestion stats and the rest for trace stats. Moreover, we are reading the localSegMeta file along with allSegMeta which is ideally is covered in allSegMetas. This is another redundant read. 
If we just read all the segmetas once and pass that around, it greatly improves the performance (reducing 4 reads to just 1).

I have also refactored a bit of the code. Earlier we were using a `map[string]utils.AllIndexesStats` which is weird because `AllIndexesStats` maps index -> stats. This [PR](https://github.com/siglens/siglens/pull/1400) mentions it in detail.
I have refactored it to just use AllIndexesStats. This could even be simplified by just using `map[string]IndexStats` which is a member of [AllIndexesStats](https://github.com/siglens/siglens/blob/d90f19f92a33c899a171f0d3fa9b44fb917a538d/pkg/utils/httpserverutils.go#L373).
But, this `AllIndexesStats` struct might have more fields in the future, so I have kept it as is.
Also, there is no point in sorting the indices over [here](https://github.com/siglens/siglens/blob/d90f19f92a33c899a171f0d3fa9b44fb917a538d/pkg/health/clusterStatsHandler.go#L298) because the iteration order of the map is not deterministic. Sort is now applied while creating the final data slice.
I am also not sure if we should filter out the segmentCount based on OrgId over [here](https://github.com/siglens/siglens/blob/d90f19f92a33c899a171f0d3fa9b44fb917a538d/pkg/health/clusterStatsHandler.go#L317). As it wasn't done over here, I kept it as is. Need clarification on this.

# Testing
Downloaded usage stats from 20TB machine. Earlier it was taking around 3s to load the entire stats. Now, it has reduced to 1s at max.

# Checklist:

### DIFF
--- a/pkg/health/clusterStatsHandler.go
+++ b/pkg/health/clusterStatsHandler.go
@@ -54,11 +54,14 @@ func ProcessClusterStatsHandler(ctx *fasthttp.RequestCtx, myid uint64) {
 			return
 		}
 	}
-	indexData, logsEventCount, logsIncomingBytes, logsOnDiskBytes := getIngestionStats(myid)
+
+	allSegMetas := writer.ReadAllSegmetas()
+
+	indexData, logsEventCount, logsIncomingBytes, logsOnDiskBytes := getIngestionStats(myid, allSegMetas)
 	queryCount, totalResponseTimeSinceRestart, totalResponseTimeSinceInstall, queriesSinceInstall := usageStats.GetQueryStats(myid)
 
 	metricsIncomingBytes, metricsDatapointsCount, metricsOnDiskBytes := GetMetricsStats(myid)
-	traceIndexData, traceSpanCount, totalTraceBytes, totalTraceOnDiskBytes := GetTracesStats(myid)
+	traceIndexData, traceSpanCount, totalTraceBytes, totalTraceOnDiskBytes := GetTracesStats(myid, allSegMetas)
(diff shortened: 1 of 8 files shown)


pkg/health/clusterStatsHandler.go
```diff
@@ -127,34 +130,40 @@ func calculateStorageSavedPercentage(incomingBytes, onDiskBytes float64) float64
...
 
-func convertDataToSlice(indexToAllIndexStats map[string]utils.AllIndexesStats, volumeField, countField,
+func convertDataToSlice(allIndexStats utils.AllIndexesStats, volumeField, countField,
 	segmentCountField string) []map[string]map[string]interface{} {
 
+	indices := make([]string, 0)
+	for index := range allIndexStats.IndexToStats {
+		indices = append(indices, index)
+	}
+	sort.Strings(indices)
```
nkunal, line 140: do we need to sort ?

nkunal (approved):
Looks great. As you said, we may not need the sort on indexname

pkg/health/clusterStatsHandler.go
vasukalariya: it would make the results look more consistent.

PR #1472, opened 2024-08-15
memory and misc fixes for agile trees

nkunal:
# Description
1. Fix a bug with duplicate groupby columns being passed to agiletree
2. Limit number of concurrent agile trees
3. Reuse memory for string array for agile tree reverseDictEnc
4. Only pick columns that have cardinality less than 1k for agile tree groupby
5. Increase PQS query removal time to 3 days
6. Set limit on concurrent trees
7. Use pre-allocated responses for es bulk responses
8. Add bench suite for agiletree ingest
9. Switch to ptr array for node pool
10. Drop Agile Tree if cardinality of any column goes above a limit

# Testing

# Checklist:


pkg/segment/writer/segstore.go
```diff
@@ -823,8 +843,37 @@ func (segstore *SegStore) computeStarTree() {
...
 		}
+
+		hasTreeSpace := false
+		atreeCounterLock.Lock()
+		if currentAgileTreeCount < MaxConcurrentAgileTrees {
+			// for now the first 5 segstores will get to AgileTree but
+			// we should add some smart logic on how we can rotate this
+			// amongst other indices
+			currentAgileTreeCount++
+			hasTreeSpace = true
```
Macbeth98, line 854: This function is being called from `AppendWipToSegfile`. This function is called when we want to flush the Block. So basically the function `computeStarTree` is being called per block and so the `computeStarTree` will only be executed for `5` blocks. 

Is that what we want? 

I did not understand the comment saying first 5 `segstores`. Are you referring to first 5 segments or first 5 indices?

nkunal: good point. I was missing the case where the segstore (index) has to give back this counter during rotation, so that some other segstore (index) can pick it up 
This function gets called on every wip flush but the creation/initialization of startree happens only on block num 0

Macbeth98: Ahh, okay I see, the counter is getting incremented only when the `numBlocks` is `zero`. 
But the same segstore (index) can also pick it up again for the next segment.

Also, since `MaxConcurrentAgileTrees` is set to `3` in the latest commit, I think we should update this comment. Instead of stating a specific number like `3` or `5`, refactor the comment to say: `For now, the first MaxConcurrentAgileTrees number of segstores will get to ...`

Macbeth98 (approved):

pkg/segment/writer/segstore.go
```diff
@@ -57,6 +57,11 @@ import (
 const MaxAgileTreeNodeCount = 8_000_000
 const colWipsSizeLimit = 2000 // We shouldn't exceed this during normal usage.
 
+const MaxConcurrentAgileTrees = 3
+
+var currentAgileTreeCount int
+var atreeCounterLock sync.Mutex = sync.Mutex{}
```
AndrewHess, line 63: Should you use a semaphore here instead?

pkg/segment/writer/segstore.go
```diff
@@ -655,6 +662,18 @@ func (segstore *SegStore) checkAndRotateColFiles(streamid string, forceRotate bo
 	}
 
 	if segstore.OnDiskBytes > maxSegFileSize || forceRotate || onTimeRotate || onTreeRotate {
+
+		if config.IsAggregationsEnabled() && segstore.usingSegTree {
+			nc := segstore.sbuilder.GetNodeCount()
+			cnc := segstore.sbuilder.GetEachColNodeCount()
+			log.Infof("checkAndRotateColFiles: stree nc: %v , Each Col NodeCount: %v", nc, cnc)
```
AndrewHess, line 669: nit:
```suggestion
			log.Infof("checkAndRotateColFiles: stree node count: %v , Each Col NodeCount: %v", nc, cnc)
```
or 
```suggestion
			log.Infof("checkAndRotateColFiles: stree total node count: %v , Each Col NodeCount: %v", nc, cnc)
```

pkg/segment/writer/segstore.go
```diff
@@ -792,12 +811,18 @@ func (segstore *SegStore) initStarTreeCols() ([]string, []string) {
 		}
 
 		cest := uint32(segstore.AllSst[cname].Hll.Estimate())
-		gcMap[cname] = cest
+
+		// skip columns if card is above threshold
+		if cest > uint32(wipCardLimit*2) {
+			continue
+		}
```
AndrewHess, line 818: nit: this comment isn't really useful, especially if you rename `cest` to something more clear

pkg/segment/writer/segstore.go
```diff
@@ -823,8 +848,37 @@ func (segstore *SegStore) computeStarTree() {
...
+		if sizeToAdd > 0 {
+			newArr := make([][]string, sizeToAdd)
+			segstore.stbDictEncWorkBuf = append(segstore.stbDictEncWorkBuf, newArr...)
+		}
+		for colNum := 0; colNum < len(sortedGrpCols); colNum++ {
+			if len(segstore.stbDictEncWorkBuf[colNum]) < MaxAgileTreeNodeCount {
+				// we know each stree col won't have more encodings than max stree node limit
+				segstore.stbDictEncWorkBuf[colNum] = make([]string, MaxAgileTreeNodeCount)
+			}
+		}
```
AndrewHess, line 878: nit: I'm a little confused by this comment, because you only know that since the `if` condition was true. So this comment isn't useful? It looks like you might have moved this logic along with the comment, and now that it's here the comment isn't needed

pkg/segment/writer/segstore.go
nkunal: My goal is to not have any limits here, hence this temp code. If I do end up having limits then will switch to semaphore

benchmark_test.go
```diff
@@ -137,9 +138,16 @@ func Benchmark_EndToEnd(b *testing.B) {
...
+
+		"* | timechart avg(latency) by http_method span=1h",
+
+		/*
+				"* | timechart avg(latency) by http_method span=1h",
+				"* | stats avg(http_status) by hobby, http_method",
+				"* | stats count(*) by http_status",
+				"* | stats sum(http_status) by hobby, http_method",
+			"whatever*",
+		*/
```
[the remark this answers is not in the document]
nkunal: yup, addressed in latest commit

benchmark_test.go
```diff
@@ -399,25 +441,30 @@ func Benchmark_agileTreeIngest(b *testing.B) {
...
-		if err != nil {
-			log.Errorf("ERROR: err=%v", err)
-			break
-		}
-		if processedCount == 0 {
-			log.Errorf("ERROR: processedCount was 0, resp=%v", response)
-			break
+	for i := 0; i < numSegs; i++ {
+		for bcnt := 0; bcnt < 5; bcnt++ {
+			processedCount, response, err := eswriter.HandleBulkBody(bulkData, nil, 0, 0, false)
```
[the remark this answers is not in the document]
nkunal: switched to `bulkCnt`

pkg/es/writer/esBulkHandler.go
```diff
@@ -112,11 +121,20 @@ func HandleBulkBody(postBody []byte, ctx *fasthttp.RequestCtx, rid uint64, myid
 
 	var bytesReceived int
 	// store all request index
-	var items = make([]interface{}, 0)
+	items_len := 1000
```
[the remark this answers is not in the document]
nkunal: addressed in latest commit

pkg/otlp/otlp.go
```diff
@@ -112,7 +114,7 @@ func ProcessTraceIngest(ctx *fasthttp.RequestCtx) {
 				}
 
 				lenJsonData := uint64(len(jsonData))
-				err = writer.ProcessIndexRequest(jsonData, now, indexName, lenJsonData, shouldFlush, localIndexMap, orgId, 0 /* TODO */)
+				err = writer.ProcessIndexRequest(jsonData, now, indexName, lenJsonData, shouldFlush, localIndexMap, orgId, 0, idxToStreamIdCache /* TODO */)
```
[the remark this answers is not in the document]
nkunal: fixed in latest commit

pkg/segment/writer/packer.go
```diff
@@ -651,7 +650,7 @@ func encJsonNumber(key string, numType SS_IntUintFloatTypes, intVal int64, uintV
 	switch numType {
 	case SS_INT64:
 		copy(wipbuf[idx:], VALTYPE_ENC_INT64[:])
-		copy(wipbuf[idx+1:], utils.Int64ToBytesLittleEndian(int64(intVal)))
+		utils.Int64ToBytesLittleEndianInplace(int64(intVal), wipbuf[idx+1:])
 		valSize = 1 + 8
```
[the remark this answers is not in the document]
nkunal: good point. Let me see how to fix this

sunitakawane (approved):

AndrewHess (approved):

PR #1473, opened 2024-08-16
feat: Read only the Columns Present in the Aggs

Macbeth98:
# Description
- When there is a `stats` command in the query, this PR would go through each Aggregator and get all the fields until the `stats` Block.
- When reading the Column files or initiating the `InitSharedMultiColumnReaders` only the cols that are present in Aggs will be read.
- This makes it so that we don't read unnecessary columns that are not in the final output, but reads all the intermediate columns.

# Testing
- Tested by running `circle` config script on my local and cicd passed.

# Profiling Stats:

Query used: `* | rex field=app_version "(?<major>\d+)\.(?<minor>\d).*" | stats count by major`
Records count: 2.1 Million. Records
On Tip of `develop`:
  - CPU: 56.39s
  - MEM: 693
  - Overall AVG ~= 25 seconds
  - Overall MEM: 30 GB

With this PR Changes:
  - CPU: 17.55 seconds
  - MEM: 602 MB
  - Overall AVG ~= 7 seconds
  - Overall Mem = 8.60GB

**NOTE**: The Query still slow for very huge number of records and will still timeout even with this PR.

# Checklist:


pkg/segment/query/querystatus.go
```diff
@@ -106,6 +106,7 @@ type RunningQueryState struct {
 	currentSearchResultCount int
 	nodeResult               *structs.NodeResult
 	totalRecsToBeSearched    uint64
+	AllColsInAggs            map[string]struct{}
```
Macbeth98, line 109: I am not exactly sure, where else we could set this value. This probably will not work for synchronous queries. As the query seems to be deleted by the time it reaches the `read raw records` block. Please suggest me if there is a better place where I could store and retrieve this value.

nkunal: you could probably store it in segquery or segsearch

nkunal: Or you could do a call to extract out these columns and just pass it around right from segquery in all the search/apply functions, so that you don't have to depend on this other module which sets/gets it via a lock

pkg/segment/structs/evaluationstructs.go
```diff
@@ -1380,7 +1396,7 @@ func (self *StatisticExpr) SortBucketResult(results *[]*BucketResult) error {
 
 // Only display fields which in StatisticExpr
 func (self *StatisticExpr) RemoveFieldsNotInExprForBucketRes(bucketResult *BucketResult) error {
-	groupByCols := self.GetGroupByCols()
+	groupByCols := self.GetFields()
```
nkunal, line 1399: hmm... could we come up with a better name for this func ? It is getting the column names but then assign it to a "group by column name field"

Macbeth98: It is getting `byClause` and `FieldList`. This is the definition of this function.

```
func (self *StatisticExpr) GetGroupByCols() []string {
	return append(self.FieldList, self.ByClause...)
}
```

pkg/segment/query/querystatus.go
Macbeth98: yeah okay I guess I can do that. Let me check.

pkg/segment/structs/segstructs.go
```diff
@@ -145,6 +146,8 @@ type TransactionGroupState struct {
 	Timestamp uint64
 }
 
+// Update this function: GetAllColsInAggsIfStatsPresent() to return all columns in the query aggregators if stats are present.
+// This function should return all columns in the query aggregators if stats are present.
 type QueryAggregators struct {
```
AndrewHess, line 151: I'm a little confused by this comment. Are you saying if we add fields to `QueryAggregators` we might need to update `GetAllColsInAggsIfStatsPresent()`?

pkg/segment/structs/segstructs.go
```diff
@@ -961,6 +973,201 @@ func (qtype QueryType) String() string {
...
+	if qa.HasStatsBlock() {
+		return true
+	}
+
+	if qa.Next != nil {
+		return qa.Next.HasStatsBlockInChain()
+	}
+
+	return false
+}
```
AndrewHess, line 994: nit: there's a helper function `HasInChain()` that will reduce this boilerplate code (see `HasQueryAggergatorBlockInChain()`)

pkg/segment/structs/segstructs.go
```diff
@@ -961,6 +973,201 @@ func (qtype QueryType) String() string {
...
+
+// returns all columns in the query aggregators if stats are present.
+// Update this function whenever a new struct or an is added to the query aggregators.
```
AndrewHess, line 997: nit: you're missing a word
> ... a new struct or an \<blank> is added ...

pkg/segment/structs/segstructs.go
```diff
@@ -961,6 +973,201 @@ func (qtype QueryType) String() string {
...
+	AddAllColumnsInOutputTransforms(cols, qa.OutputTransforms)
+	AddAllColumnsInMeasureAggs(cols, qa.MeasureOperations)
+	AddAllColumnsInGroupByRequest(cols, qa.GroupByRequest)
+	AddAllColumnsInTransactionArguments(cols, qa.TransactionArguments)
+	AddAllColumnsInStreamStatsOptions(cols, qa.StreamStatsOptions)
+
+	if qa.HasStatsBlock() {
+		// We want to stop at the first stats block
+		return
+	}
```
AndrewHess, line 1028: This looks like it gets all the columns through the first stats block rather than stopping just before the first stats block. Is that right? If so, I think you should change the name to indicate that

pkg/segment/structs/segstructs.go
```diff
@@ -961,6 +973,201 @@ func (qtype QueryType) String() string {
...
+	if outputTransforms == nil {
+		return
+	}
+
+	if outputTransforms.HarcodedCol != nil {
+		sutils.ConvertToSetFromSlice(cols, outputTransforms.HarcodedCol)
+	}
+
+	if outputTransforms.RenameHardcodedColumns != nil {
+		sutils.ConvertToSetFromMap(cols, outputTransforms.RenameHardcodedColumns)
```
AndrewHess, line 1045: nit: changing the names to `ConvertSliceToSet` and `ConvertMapToSet` would be slightly more concise. But you can keep what you have if you want

pkg/utils/maputils.go
```diff
@@ -84,3 +86,37 @@ func SetDifference[K comparable, T1, T2 any](newSet map[K]T2, oldSet map[K]T1) (
...
+		set[k] = struct{}{}
+	}
+}
+
+// ConvertToSetFromSlice converts a slice to a set. The set is represented as a map with the key as the element of the set
+func ConvertToSetFromSlice[K comparable](set map[K]struct{}, source []K) {
+	for _, k := range source {
+		set[k] = struct{}{}
+	}
+}
```
AndrewHess, line 102: nit:
1. `ConvertMapToSet` and `ConvertSliceToSet` would be slightly more concise, but you can keep the `ConvertToXFromY` format if you want
2. Using `Add` instead of `Convert` in the function name seems more precise
3. For `ConvertToSetFromMap()` it would be good to specify in the name that you're adding the map keys rather than the values

pkg/utils/maputils.go
```diff
@@ -84,3 +86,37 @@ func SetDifference[K comparable, T1, T2 any](newSet map[K]T2, oldSet map[K]T1) (
...
+
+func AddToSet[K comparable](set map[K]struct{}, key K) {
+	if _, ok := set[key]; !ok {
+		set[key] = struct{}{}
+	}
+}
```
AndrewHess, line 108: nit: you don't need this check `if _, ok := set[key]; !ok {` because there's only one possible value in the map for a given key. So this function would be just `set[key] = struct{}{}`, so I don't really see the need for this function (at least, not unless it was a method on a new `type Set struct {...}`)

pkg/segment/structs/segstructs.go
Macbeth98: Yes we should get all the Columns through first stats block. I have changed the function name and comment to reflect that.

pkg/utils/maputils.go
```diff
@@ -87,24 +87,22 @@ func SetDifference[K comparable, T1, T2 any](newSet map[K]T2, oldSet map[K]T1) (
 	return added, removed
 }
 
-// ConvertToSetFromMap converts a map to a set. The set is represented as a map with the key as the element of the set
-func ConvertToSetFromMap[K comparable, V any](set map[K]struct{}, source map[K]V) {
+// AddMapKeysToSet converts a map to a set. The set is represented as a map with the key as the element of the set
+func AddMapKeysToSet[K comparable, V any](set map[K]struct{}, source map[K]V) {
```
AndrewHess, line 91: nit: with this name change, I don't think you even need the comment any more

Same comment for `AddSliceToSet()`

pkg/segment/reader/record/recordreader.go
```diff
@@ -57,6 +57,14 @@ func GetRecordsFromSegment(segKey string, vTable string, blkRecIndexes map[uint1
 			return nil, allCols, errors.New("failed to get column names for segkey in rotated and unrotated files")
 		}
 	}
+
+	// if len(allColsInAggs) > 0, then we need to intersect the allCols with allColsInAggs
+	// this is because we only need to read the columns that are present in the aggregators.
+	// if len(allColsInAggs) == 0, then we ignore this step, as this would mean that the
+	// query does not have a stats block, and we need to read all columns.
+	if len(allColsInAggs) > 0 {
```
AndrewHess, line 65: nit: this could be a good spot to use `utils.Option[T]` instead. Then instead of `if len(allColsInAggs) > 0 {` you'd do:
```
if allCols, ok := allColsInAggs.Get(); ok {
```
(where `allColsInAggs` is now a `Option[map[string]struct{}]` instead of a `map[string]struct{}`). And in this case, you'd still do the intersection if it was a non-nil map of zero elements.

However, you can ignore this and keep what you have if you want

AndrewHess (approved):
Looks good. I had a few nits but they're very minor, so you don't have to address them if you don't want to

AndrewHess:
@nkunal I think this PR is ready for final performance testing

pkg/segment/reader/record/recordreader.go
Macbeth98: Got it. But that would mean then I need to change the type of `allColsInAggs` from where I have first initialized this variable right? 

For now I would like to keep this as it is. But for future PRs I will use `utils.Option[T]`.

AndrewHess (approved):

Issue #1103, opened 2024-06-12
[STORY] Refactor Spl Stats Functions


Jiale-Fang:
# Description
In `spl.peg` file, stats functions share a lot of similarities. And we have 20-30 plus functions parsing by this way. You should create a general method to remove redundant code. 
https://github.com/siglens/siglens/blob/20f25abf1899d96e45a99bbbc92e4db6a7664e6b/pkg/ast/spl/spl.peg#L2133-L2203

# (Optional) Testing
doc: https://pkg.go.dev/github.com/mna/pigeon#section-readme
```
go get -u github.com/mna/pigeon	//Download library
pigeon -o pkg/ast/spl/spl.go pkg/ast/spl/spl.peg 
```
# Acceptance Criteria
List the concrete items that should be completed for this task to be considered completed. Be as clear and thorough as you can.

saniya16:
I would like to work on this

AndrewHess:
Great, go ahead

deva-rushi:
I will be working on it

AndrewHess:
Sorry, there's actually an open PR already for this

deva-rushi:
It’s ok, are there any other issues that I can work on?

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

AtharvaKeshre:
I will be working on it

Dhruv1681:
I will be working on it

Macbeth98:
Sorry @Dhruv1681 @AtharvaKeshre , there is already a PR out for this, please choose another Issue. Thanks!

PR #1478, opened 2024-08-16
SPL: Refactored aggregate functions

vasukalariya:
# Description
Summarize the change.
Refactored a lot of duplicate code in the parser.
Corrected parser for functions that are yet to be implemented
Refactored parser tests. Added stats + eval tests for yet to be implemented functions

https://github.com/siglens/siglens/issues/1103

# Testing

# Checklist:

### DIFF
--- a/pkg/ast/spl/tests/splParser_test.go
+++ b/pkg/ast/spl/tests/splParser_test.go
@@ -10009,7 +10009,7 @@ func Test_FillNull_ValueArg_FieldList(t *testing.T) {
 	assert.Equal(t, []string{"field1", "field2"}, aggregator.OutputTransforms.LetColumns.FillNullRequest.FieldList)
 }
 
-func GetMeasureFuncStr(measureFunc utils.AggregateFunctions) (string, string) {
+func getMeasureFuncStr(measureFunc utils.AggregateFunctions) (string, string) {
 	switch measureFunc {
 	case utils.Cardinality:
 		return "dc", ""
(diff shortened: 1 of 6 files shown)


pkg/ast/spl/tests/splParser_test.go
```diff
@@ -10132,8 +10009,62 @@ func Test_FillNull_ValueArg_FieldList(t *testing.T) {
 	assert.Equal(t, []string{"field1", "field2"}, aggregator.OutputTransforms.LetColumns.FillNullRequest.FieldList)
 }
 
-func performCommon_aggEval_BoolExpr(t *testing.T, query []byte, measureFunc utils.AggregateFunctions, strEnc string) {
+func GetMeasureFuncStr(measureFunc utils.AggregateFunctions) (string, string) {
```
AndrewHess, line 10012: nit: since this is just for testing, it can be private
```suggestion
func getMeasureFuncStr(measureFunc utils.AggregateFunctions) (string, string) {
```

pkg/segment/utils/segconsts.go
```diff
@@ -300,6 +301,7 @@ const (
 	Estdc
 	EstdcError
 	ExactPerc
+	Mean
```
AndrewHess, line 304: We already have `Avg`; why do we need `Mean`?

AndrewHess (commented):
I think this is good, just a few small comments

P.S. this would have been a bit easier to review if you separated the PEG changes and the test changes a little more. You already have them in different commits which was helpful, but your PEG changes broke one of the tests, and it was a bit difficult for me to find exactly what changes to testing you made for that. So a better split could have been:
1. Refactor PEG
2. Make small changes to fix the tests
3. Do the rest of the tests refactoring

Either all of these as separate commits would be good, or having 1 and 2 in one PR and step 3 in a separate PR would work too. Also with it separated like that, it's easier to check that your PEG refactoring didn't break anything

pkg/ast/spl/tests/splParser_test.go
vasukalariya: addressed in the latest commit.

pkg/segment/utils/segconsts.go
vasukalariya: addressed in the latest commit.

pkg/utils/testutils.go
```diff
@@ -22,11 +22,25 @@ import (
...
 
-func RandomStringWithCharset(length int) string {
+func GetRandomString(length int, stringType string) string {
+	var charset string
+	switch stringType {
+	case "alpha":
+		charset = alphabets
+	case "numeric":
+		charset = digits
+	case "alphanumeric":
```
AndrewHess, line 38: nit: use an enum instead of strings so that we get a compile-time error if there's a typo

PR #1480, opened 2024-08-16
feat: Record Column value Length at Segment level if the length is constant for all the Records

Macbeth98:
# Description
- Adds a new Field called `cValSize` that records the length of a constant Column Value for records at segment level.
- If the length of the column value is constant for all the records in the segment, then that value will be record, otherwise `-1` is set.
- `getCurrentRecordLength` uses this value if `+ve` otherwise, it proceeds through normal process.

# Testing
- Added a test case that tests the sizes.

# Checklist:


pkg/segment/structs/segmeta.go
```diff
@@ -24,8 +24,9 @@ import (
 const MAX_SEGMETA_FSIZE = 10_000_000 // 10 MB
 
 type ColSizeInfo struct {
-	CmiSize uint64 `json:"cmiSize"`
-	CsgSize uint64 `json:"csgSize"`
+	CmiSize  uint64 `json:"cmiSize"`
+	CsgSize  uint64 `json:"csgSize"`
+	CValSize int8   `json:"cValSize"`
```
nkunal, line 29: Since this is just meta and will get store once in segmeta per segment, then we can just go with a type of `int` , that should handle the cases where the string length is greater than 127 and we don't have to worry about overflows.
Plus it is not of big hit of 1 byte v/s 4 bytes for this var type

pkg/segment/writer/packer.go
```diff
@@ -550,7 +558,7 @@ func (ss *SegStore) initAndBackFillColumn(key string, value interface{}, matched
 	if !ok {
 		colWip = InitColWip(ss.SegmentKey, key)
 		allColWip[key] = colWip
-		ss.AllSeenColumns[key] = true
+		// ss.AllSeenColumns[key] = true
```
nkunal, line 561: remove commented code

pkg/segment/writer/packer.go
```diff
@@ -672,6 +680,38 @@ func encJsonNumber(key string, numType SS_IntUintFloatTypes, intVal int64, uintV
 	return valSize
 }
 
+func (ss *SegStore) updateColValueSizeInAllSeenColumns(colName string, size uint32) {
+	currentSize, ok := ss.AllSeenColumns[colName]
+	if !ok {
+		if ss.RecordCount > 0 {
+			// Not the first record in the segment, so the col value size is not consistent.
```
nkunal, line 687: update the comment to something of ...
// column appearing first time in the middle of the wip, so past recNums will be filled with BackFill_CVAL_TYpe, so mark this as inconsistent

pkg/segment/writer/packer.go
```diff
@@ -672,6 +680,38 @@ func encJsonNumber(key string, numType SS_IntUintFloatTypes, intVal int64, uintV
...
+			ss.AllSeenColumns[colName] = -1
+			currentSize = -1
+		} else {
+			ss.AllSeenColumns[colName] = 0
```
nkunal, line 691: we can just set `ss.AllSeenColumns[colName] = inSize` and return from here

pkg/segment/writer/packer.go
```diff
@@ -672,6 +680,38 @@ func encJsonNumber(key string, numType SS_IntUintFloatTypes, intVal int64, uintV
 	return valSize
 }
 
+func (ss *SegStore) updateColValueSizeInAllSeenColumns(colName string, size uint32) {
+	currentSize, ok := ss.AllSeenColumns[colName]
+	if !ok {
+		if ss.RecordCount > 0 {
+			// Not the first record in the segment, so the col value size is not consistent.
+			ss.AllSeenColumns[colName] = -1
+			currentSize = -1
```
nkunal, line 689: we don't need to set currentSize and just return from here. (note see below comment, we can just have 1 return statement for this whole if !ok block

pkg/segment/writer/packer.go
```diff
@@ -672,6 +680,38 @@ func encJsonNumber(key string, numType SS_IntUintFloatTypes, intVal int64, uintV
...
+
+	if size > math.MaxInt8 {
+		ss.AllSeenColumns[colName] = -1
+		return
+	}
+
+	if currentSize == 0 {
+		ss.AllSeenColumns[colName] = int8(size)
+		return
+	}
```
nkunal, line 708: we won't need this if check based on above comments

pkg/segment/writer/packer.go
```diff
@@ -672,6 +680,38 @@ func encJsonNumber(key string, numType SS_IntUintFloatTypes, intVal int64, uintV
...
+
+	if currentSize != int8(size) {
```
nkunal, line 710: would be great if we can avoid type conversions, since this is the fast path code, understand the need of -1, but we can pick some other value for noting down inconsistency

pkg/segment/writer/segstore.go
```diff
@@ -71,7 +71,7 @@ type SegStore struct {
 	lastUpdated        time.Time
 	VirtualTableName   string
 	RecordCount        int
-	AllSeenColumns     map[string]bool
+	AllSeenColumns     map[string]int8 // Map of Column to Column Value size. The value is a positive int if the size is consistent across records and -1 if it is not.
```
nkunal, line 74: since we are repurposing this var for multiple things, maybe we should rename this to something like `AllSeenColumnsAndSizes` (you could find a better name than my suggestion)

pkg/segment/structs/segmeta.go
nkunal, line 29: maybe we can call this `ConsistentCvalSize` .. something that denotes not the size of the whole column values but the consistency of the columnVal size across the whole segment

nkunal (commented):
overall looks fantastic, a few naming nits and data type suggestion.
Great Job!!

pkg/segment/query/metadata/segmentmicroindex.go
```diff
@@ -260,8 +260,16 @@ func (sm *SegmentMicroIndex) readCmis(blocksToLoad map[uint16]map[string]bool, a
...
-		retVal[k] = true
+	for colName := range sm.ColumnNames {
+		retVal[colName] = true
+	}
+	return retVal
+}
+
+func (sm *SegmentMicroIndex) GetAllColumnsRecSize() map[string]int8 {
+	retVal := make(map[string]int8, len(sm.ColumnNames))
+	for colName, colValueSize := range sm.ColumnNames {
```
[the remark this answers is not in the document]
Macbeth98: yes, it should be `ColSizeInfo`. I thought I named it `ColSizeInfo`. Correcting it.

pkg/segment/writer/packer.go
Macbeth98: Yes, I did this way so that I can check the size. But now since there is no limit. I can just assign the size here and return.

pkg/segment/writer/packer.go
Macbeth98: Then I think I can make the type as `uint32` and for inconsistency the value will be `0`.  As the size by default comes in `uint32`.

pkg/segment/reader/segread/segreader.go
```diff
@@ -257,6 +259,13 @@ func (sfr *SegmentFileReader) iterateNextRecord() error {
 }
 
 func (sfr *SegmentFileReader) getCurrentRecordLength() (uint32, error) {
+	// if we have the +ve column value rec len, that is the current record length
+	// This value comes from the segment metadata, where we store the column value size
+	// at segment level. If the value is +ve and is != INCONSISTENT_CVAL_SIZE it means all records in the segment for this column
+	// have the same length and if it is equal to INCONSISTENT_CVAL_SIZE, it means the records have different lengths
```
AndrewHess, line 265: What is `+ve`?

pkg/segment/structs/segmeta.go
```diff
@@ -24,8 +24,9 @@ import (
 const MAX_SEGMETA_FSIZE = 10_000_000 // 10 MB
 
 type ColSizeInfo struct {
-	CmiSize uint64 `json:"cmiSize"`
-	CsgSize uint64 `json:"csgSize"`
+	CmiSize            uint64 `json:"cmiSize"`
+	CsgSize            uint64 `json:"csgSize"`
+	ConsistentCvalSize uint32 `json:"cValSize"`
```
AndrewHess, line 29: nit: It might be worth adding a comment here to describe what `ConsistentCvalSize` means

AndrewHess (commented):
This looks good; just some minor comments

@nkunal I think this is about ready to merge, if you want to do some performance testing before it gets merged

pkg/segment/reader/segread/segreader.go
Macbeth98: I mean if the value is `>0`

Macbeth98:
> This looks good; just some minor comments
> 
> @nkunal I think this is about ready to merge, if you want to do some performance testing before it gets merged

Please wait, I am doing an optimization where I am fetching the Column sizes before at `QuerySegmentRequests` rather than fetching it at `InitSharedMultiColumnReaders`. I am not exactly sure how much performance this will improve, but we don't need an extra lock and unlock through this way.

Macbeth98:
@nkunal @AndrewHess Can you review the PR again? I have changed the getting of the Column sizes to segment level instead of calling it per block.  I have ran the benchmarks also with this latest PR.

```
Develop
Total Run: 626s
getCurrentRecordLength: 30s
1st run: 1m42s
2nd run: 1m47s

This PR
Total Run: 591s
getCurrentRecordLength: 20s
1st run: 1m36s
2nd run: 1m43s
```

pkg/segment/structs/segmeta.go
```diff
@@ -24,8 +24,9 @@ import (
 const MAX_SEGMETA_FSIZE = 10_000_000 // 10 MB
 
 type ColSizeInfo struct {
-	CmiSize uint64 `json:"cmiSize"`
-	CsgSize uint64 `json:"csgSize"`
+	CmiSize            uint64 `json:"cmiSize"`
+	CsgSize            uint64 `json:"csgSize"`
+	ConsistentCvalSize uint32 `json:"cValSize"` // The size of the column value, given that the size is consistent across all records. The value is set to `1MB` if the column values length is not consistent.
 }
```
AndrewHess, line 30: Using 1MB seems quite odd to me. Why not use 0 or the max uint32 value to indicate this special case?

Alternatively, maybe it would make sense to use a `utils.Option[uint32]` here instead (although with that I think you'd have to write the JSON encode/decode functions, which may cause problems for non-literal types, but perhaps you could only handle certain types like uint32 for now)

This `ColSizeInfo` is JSON encoded and then written to segmeta.json, so whatever you do, we're going to have to support that for a long time

nkunal:
Maybe use uint32 max

*Best Regards*
*-- Kunal Nawale ,**Founder/CEO, SigScalr*

pkg/segment/structs/segmeta.go
Macbeth98: Before we decided to use `1MB` because as we will not allow the value size more than `65_535` bytes. So the size will never reach `1MB`. So Should I continue using the same or change it to max of `uint32`?

AndrewHess: Let's use the max `uint32`

AndrewHess (approved):

PR #1484, opened 2024-08-19
Backend APIs for Get and Delete Specific Lookup File

vasukalariya:
# Description
Summarize the change.
Get and Delete API for specific lookup files.
Simplified CICD to use a flag for migrating lookup_file from cicd folder to lookups folder for testing.

# Testing
Try multiple scenarios after deleting, creating lookups files and then retrieving data.
image
Ran E2E tests to check the simplified CICD.

# Checklist:


pkg/lookups/lookups.go
```diff
@@ -150,3 +152,69 @@ func GetAllLookupFiles(ctx *fasthttp.RequestCtx) {
...
+				utils.SendInternalError(ctx, "Error while reading the file", "", err)
+				return
+			}
+			break
+		}
+		if _, err := ctx.Write(buf[:n]); err != nil {
+			utils.SendInternalError(ctx, "Error while sending the data", "", err)
+			return
+		}
+	}
```
AndrewHess, line 196: Instead of reading into a buffer and then writing that buffer to `ctx`, I'm pretty sure you can simply use `io.Copy`, which does buffering internally. Also it looks like, `os.File` implements `Reader`, so you don't even need to open a buffered reader for the file:
```
bytesWritten, err := io.Copy(ctx, file)
if err != nil {
    // handle errors
}
```

tools/sigclient/cmd/sigclient.go
```diff
@@ -141,6 +142,15 @@ var esQueryCmd = &cobra.Command{
 		log.Infof("filePath : %+v\n", filepath)
 		log.Infof("randomQueries: %+v\n", randomQueries)
 		log.Infof("bearerToken : %+v\n", bearerToken)
+
+		if cicd {
+			err := query.MigrateLookupsForCicd()
```
AndrewHess, line 147: Naming the flag `cicd` is a bit confusing; it's not immediately obvious what this would mean, and from that name I wouldn't guess what it actually does (move a lookup file).

Maybe a better way to do this would be to have a flag like `--lookups file1.csv file2.csv ...` (for now we could just support one file, if that's easier and we don't currently need to have multiple).

Also, you should add new command line options to the sigclient README

AndrewHess (commented):
Looks pretty good, but you need to add the new command line option to the sigclient README

AndrewHess (approved):

PR #1485, opened 2024-08-19
Add list sub command

### LINKED ISSUE #746: [STORY] SPL Add sub command list
# Description
Implement the sub command `list` under the `stats` SPL command
reference:https://docs.splunk.com/Documentation/SplunkCloud/9.1.2312/SearchReference/Multivaluefunctions

sample query:
`search index=*-wfd-dmq-app host=*** sourcetype=rabbitmq (mnesia_down) OR (*partition*) | stats list(_raw) by host _time`

sathkrith:
# Description
Add list sub-command to stat.

Fixes #746 (link all the GitHub issues this addresses)
https://github.com/siglens/siglens/issues/746

# Testing
Tested the code with 2kevents.json with these queries:
* | stats min(latency), count(hobby), list(gender) 
* | stats list(eval(gender="male"))
city=Boston | stats min(latency), count(hobby), list(gender) As gender
city=Boston | stats list(eval(gender="male")) 

# Checklist:


pkg/segment/structs/segstructs.go
```diff
@@ -578,9 +578,25 @@ func (ss *SegStats) Merge(other *SegStats) {
...
+}
+
+func (ss *StringStats) Merge(other *StringStats) {
+	if ss.StrSet != nil {
+		for key, value := range other.StrSet {
+			ss.StrSet[key] = value
+		}
+	}
+	if ss.StrList != nil {
+		ss.StrList = append(ss.StrList, other.StrList...)
```
AndrewHess, line 598: If `ss.StrSet == nil` or `ss.StrList == nil`, you should probably make the set/list and then insert the `other` values into it

pkg/segment/aggregations/evalaggs.go
```diff
@@ -600,6 +600,35 @@ func PerformAggEvalForCardinality(measureAgg *structs.MeasureAggregator, strSet
...
+			if boolResult {
+				strList = append(strList, "1")
+			}
+		} else {
+			cellValueStr, err := measureAgg.ValueColRequest.EvaluateToString(fieldToValue)
+			if err != nil {
+				return []string{}, fmt.Errorf("PerformAggEvalForList: Error while evaluating value col request, err: %v", err)
+			}
+			strList = append(strList, cellValueStr)
+		}
```
AndrewHess, line 627: Lines 607-611 and 622-626 look the same to me. Am I missing something? If they are the same, you should rearrange this code so we only have that block once.

pkg/segment/aggregations/evalaggs.go
```diff
@@ -600,6 +600,35 @@ func PerformAggEvalForCardinality(measureAgg *structs.MeasureAggregator, strSet
 	return float64(len(strSet)), nil
 }
 
+func PerformAggEvalForList(measureAgg *structs.MeasureAggregator, strList []string, fieldToValue map[string]utils.CValueEnclosure) ([]string, error) {
```
AndrewHess, line 603: nit: it would be good to rename `strList` to something more meaningful, as `strList` just repeats its type

pkg/segment/aggregations/evalaggs.go
```diff
@@ -693,6 +721,62 @@ func ComputeAggEvalForValues(measureAgg *structs.MeasureAggregator, sstMap map[s
...
+		}
+		finalList = list
+	} else {
+		sst, ok := sstMap[fields[0]]
+		if !ok {
+			return fmt.Errorf("ComputeAggEvalForList: sstMap did not have segstats for field %v, measureAgg: %v", fields[0], measureAgg.String())
+		}
+
+		length := len(sst.Records)
+		for i := 0; i < length; i++ {
```
AndrewHess, line 754: nit: this is more idiomatic for Go
```suggestion
		for i := range sst.Records {
```

pkg/segment/aggregations/evalaggs.go
```diff
@@ -693,6 +721,62 @@ func ComputeAggEvalForValues(measureAgg *structs.MeasureAggregator, sstMap map[s
...
+				return fmt.Errorf("ComputeAggEvalForList: Error while performing eval agg for list, err: %v", err)
+			}
+			finalList = append(finalList, list...)
+		}
+	}
+	sort.Strings(finalList)
+	// splunk list is limited to 100 results, take first 100 results match splunk.
+	if len(finalList) > 100 {
+		finalList = finalList[:100]
+	}
```
AndrewHess, line 772: Hmm, now you're repeating information that we already have here: https://github.com/siglens/siglens/blob/develop/pkg/segment/results/segresults/segresults.go#L377-L386

Could you please make a global constant that both of these code sections can access and use that instead of hardcoding `100`?

I think you also use this in other places (at least in blockresult.go)

pkg/segment/results/segresults/segresults.go
```diff
@@ -374,17 +374,27 @@ func (sr *SearchResults) UpdateSegmentStats(sstMap map[string]*structs.SegStats,
...
+				sort.Strings(currSst.StringStats.StrList)
 				strList = utils.AppendWithLimit(strList, currSst.StringStats.StrList, 100)
 			}
 
 			if sr.runningSegStat[idx] != nil && sr.runningSegStat[idx].StringStats != nil {
+				sort.Strings(sr.runningSegStat[idx].StringStats.StrList)
 				strList = utils.AppendWithLimit(strList, sr.runningSegStat[idx].StringStats.StrList, 100)
 			}
-
+			sort.Strings(strList)
```
AndrewHess, line 397: Why do you sort here? I had the same question on the original PR that added `list`: https://github.com/siglens/siglens/pull/1379#discussion_r1714424294

Am I missing something? I think you're sorting somewhere else as well

pkg/segment/structs/segstructs.go
sathkrith: Made the changes, It will pick the other.StrList and other.StrSet when they are not nil.

pkg/segment/aggregations/evalaggs.go
sathkrith: Yes, they are the same. 
I have optimized the code.

pkg/segment/aggregations/evalaggs.go
sathkrith: Changed to currentList

pkg/segment/aggregations/evalaggs.go
sathkrith: Switched to range loop

pkg/segment/aggregations/evalaggs.go
sathkrith: Switched to MAX_LIST_LEN constant defined in segconsts

pkg/segment/results/segresults/segresults.go
sathkrith: There is no need to sort.
I got confused by:
> Add the stats command with the list function to the search. The numbers are returned in ascending order in a single, multivalue result. 

Removed sort in latest commit.

sathkrith:
> This is looking pretty good. I had some nitpicks (comments prefixed with `nit`), but the overall structure looks correct.
> 
> Also, you should add some end-to-end tests in ingest.csv and restart.csv (you can add the same tests to both). These get run by our CICD pipeline when PRs are updated. You can look at `.circleci/config.yml` to see exactly what commands get executed so you can run them on your local machine

Testing the **stats list** is challenging with the current setup. The list command orders results by event timestamp, and since most logs ingested during the CI/CD tests have the same timestamp, the results from the list are inconsistent.

To work around this, I've designed the tests to produce only a single result for each query.

I attempted to use sort, but there appears to be a bug when using sort followed by values or list. Any query that involves **sort | list/values** causes SigLens to crash. Is this a known issue?

pkg/segment/utils/segconsts.go
```diff
@@ -111,6 +111,8 @@ const MS_IN_MIN = 60_000     // 60 * 1000
 const MS_IN_HOUR = 3_600_000 // 60 * 60 * 1000
 const MS_IN_DAY = 86_400_000 // 24 * 60 * 60 * 1000
 
+const MAX_LIST_SIZE = 100
```
AndrewHess, line 114: nit: just from this it's not obvious that this is for the splunk `list` command (as opposed to a normal list in Go), and it's not obvious why it's `100`.

You should rename it to something like `MAX_SPL_LIST_SIZE` (for the Search Processing Language that splunk has), and add a comment about why it's 100 (perhaps linking to splunk documentation)

AndrewHess (commented):
This looks good, just one more thing to change

AndrewHess:
> I attempted to use sort, but there appears to be a bug when using sort followed by values or list. Any query that involves **sort | list/values** causes SigLens to crash. Is this a known issue?

Good catch. I just made an Issue for this: https://github.com/siglens/siglens/issues/1502

The end-to-end tests you have for this PR with just a single value in the list should be fine for this PR, and we can do more testing after the new bug you found is fixed

pkg/segment/utils/segconsts.go
sathkrith: switched to MAX_SPL_LIST_SIZE , and added some description to where the constant comes from.

AndrewHess (approved):
Looks good!

PR #1489, opened 2024-08-20
Cleanup incorrect entries in measureColInfoUsage

vasukalariya:
# Description
Summarize the change.
This [method](https://github.com/siglens/siglens/blob/efd321d637320ac577daf25a05cbdcdd6b4c062f/pkg/querytracker/querytracker.go#L226) is responsible for returning the most used group by and measure columns.
We were directly using `MeasureCol` value to populate this usage map. This is not ideal for cases where stats has an eval inside of it. In these cases the value of `MeasureCol` is the value of `strEnc` which is the whole expression like `avg(eval(http_status+10))`.
Now, it checks if the measureAgg has a valueColRequest then it will get the fields and use that to populate the column usage map.
Moreover, we do not need to store entries like `timestamp` and `*`. So, removed these from both the maps at the end.

# Testing
- Insert data
- Run queries `city=Boston | stats avg(eval(if(http_status > 400, "100", "10"))) as avg BY http_method`, `* | stats count by timestamp`.
- Insert new data after a while.
- Earlier it used to show the gcols and measureInfoUsage cols with values like `avg(eval(if(http_status > 400, "100", "10")))`, `timestamp`, `*`
- Now, it fetches the columns properly and cleans up the irrelevant cols.
gcols:  map[http_method:{}]
inMesCols:  map[http_status:true]

Moreover, ran this on the SPL test suite (ingest.csv). From now, on the columns would be fetched properly for stats with eval queries.

# Checklist:


pkg/querytracker/querytracker.go
```diff
@@ -215,6 +215,27 @@ type colUsage struct {
...
+		}
+	} else {
+		measureInfoUsage[measureAgg.MeasureCol] = true
+	}
+}
+
+func RemoveIrrelevantCols[T any](cols map[string]T, irrelevantCols []string) {
+	for _, col := range irrelevantCols {
+		delete(cols, col)
+	}
```
Macbeth98, line 236: I think you can make this as a generic function in `maputils.go`.

pkg/querytracker/querytracker.go
```diff
@@ -304,6 +325,10 @@ func GetTopPersistentAggs(table string) (map[string]struct{}, map[string]bool) {
 		}
 	}
 
+	irrelevantCols := []string{config.GetTimeStampKey(), "*"}
```
Macbeth98, line 328: nit: Also may be you can create this as a global constant instead of declaring and initializing in the function.

Macbeth98: When you do that also rename this so that it reflects the intent of this variable better.

pkg/querytracker/querytracker.go
```diff
@@ -215,6 +215,27 @@ type colUsage struct {
...
+	}
+
+	if measureAgg.ValueColRequest != nil {
+		fields := measureAgg.ValueColRequest.GetFields()
+		for _, field := range fields {
+			measureInfoUsage[field] = true
+		}
+	} else {
+		measureInfoUsage[measureAgg.MeasureCol] = true
+	}
```
AndrewHess, line 230: Is the section in this `if` actually useful to us?
```
		fields := measureAgg.ValueColRequest.GetFields()
		for _, field := range fields {
			measureInfoUsage[field] = true
		}
```
Or should we just skip this `measureAgg` when `ValueColRequest != nil`?

I think this information is only used for `initStarTreeCols()` (which calls `GetTopPersistentAggs()` to get the info), and I think if we have a stats eval query we won't be able to use this info to answer the query anyways, but I might be wrong. Is there a way we can use this for the query, or some other use case where this would be useful?

cc: @nkunal

pkg/querytracker/querytracker.go
vasukalariya: addressed in the latest commit.

pkg/querytracker/querytracker.go
vasukalariya: addressed in the latest commit. We cannot create constant arrays in go. Also creating it as a global var does not seem to call config.GetTimeStampKey(). I think it might be due to the timing of variable and package initialization.

pkg/querytracker/querytracker.go
vasukalariya: addressed in the latest commit.

AndrewHess (approved):

PR #1496, opened 2024-08-21
Fix: Accumulation of search results

vasukalariya:
# Description
Summarize the change.
Earlier a common SegmentSearchStatus was used to accumulate the results without accounting for the operator type for certain cases.
For example,
`* | search country=Sweden http_method=HEAD | search city=def OR city=Boston`
A search like this will combine the results of `country=Sweden` and `http_method=HEAD` with an `AND` condition.
Results of `city=def` and `city=Boston` will be merged on the previous results with `OR` condition. Which is incorrect, we have to take accountability for the fact that the later search is on top of the existing search (`search country=Sweden http_method=HEAD`). 
Due to this, it was returning 1099 (69 matches for `country=Sweden AND http_method=HEAD`, 1032 matches for `city=Boston`, and 2 matches were common satisfying `country=Sweden AND http_method=HEAD AND city=Boston`. So 1032+69-2 = 1099) results instead of 4 (which satisfies the condition`country=Sweden AND http_method=HEAD` AND (city=Boston OR city=def)`).

Due to the nature of the Search Tree, the results have to be accumulated recursively. Moreover, the block search only searches a part of the query, so cannot use a single global result and apply local operators (OR in the above case).
In a later PR, we can optimize the current implementation that uses local SegmentSearchStatus to pass on results.

# Testing
Have added the test cases in ingest.csv and restart.csv.

# Checklist:


pkg/segment/search/segsearch.go
```diff
@@ -489,42 +490,63 @@ func rawSearchSingleSPQMR(multiReader *segread.MultiColSegmentReader, req *struc
...
+	return searchRes
+}
+
+func mergeSegmentSearchStatus(baseSearch *SegmentSearchStatus, searchToMerge *SegmentSearchStatus, op utils.LogicalOperator) {
+	if searchToMerge == nil {
+		return
+	}
+	for blkNum, blkStatus := range searchToMerge.AllBlockStatus {
+		_ = baseSearch.updateMatchedRecords(blkNum, blkStatus.allRecords, op)
+	}
```
AndrewHess, line 549: Why are you ignoring the `error` type that's returned? We probably want to log it

AndrewHess (commented):
Looks pretty good

pkg/segment/search/segsearch.go
vasukalariya: sorry missed that. Addressed in the latest commit.

AndrewHess (approved):

PR #1503, opened 2024-08-22
Dropping of column in agileTree

vasukalariya:
# Description
Summarize the change.
Added Drop column functionality to agileTrees.
This also fixes a bug for aggregations (this got created when we moved from utils.CValueEnclosure to *utils.Number).
Aggregate was copying the pointers instead of values when it sees the first children. So, when child gets modified so does the parent. Now, it properly copies the child.

Merge approach:
- To remove a level say x from the tree, the parent of x has to be linked with the child of x.
- These new sets of children can have duplicates, this duplicacy can carry forward in further children nodes.
- To tackle this, we will collect the common key nodes at each step and add them as commonChildren.
- The first child in each set of commonChildren will be picked as the final child and will have the correct hierarchy.
- Upon recursively returning we will properly link the parent to the children and clean up commonChildren.

This function can only drop a single column in one call. To drop multiple columns we will need to call this multiple times.

Upon verification, we can decide whether to use the currently proposed merging approach or recreate the tree by reading the data again.

# Testing
Added unit tests to further validate this.
- Set number of records per WIP to be around 500
- Ingest data
- Run PQS queries
```
"city=F* | stats min(latitude) BY city",
"city=F* | stats max(longitude) BY city",
"city=F* | stats sum(http_status) BY city",
"city=F* | stats count(http_status) BY city",
"app_name=Basketd* | stats min(latitude) BY app_name",
"app_name=Basketd* | stats max(longitude) BY app_name",
"app_name=Basketd* | stats sum(http_status) BY app_name",
"app_name=Basketd* | stats count(http_status) BY app_name",
"country=I* | stats min(latitude) BY country",
"country=I* | stats max(longitude) BY country",
"country=I* | stats sum(http_status) BY country",
"country=I* | stats count(http_status) BY country",
"* | stats min(latitude) BY http_method",
"* | stats max(longitude) BY http_method",
"* | stats sum(http_status) BY http_method",
"* | stats count(http_status) BY http_method",
```
- Ingest some more data (agileTrees creation should start, notice dropping of column app_name during blockNum 2)
- Verify results by querying (remember to set the config.Aggs to true so that agileTrees can be used)

```
* | stats count(longitude) BY city | sort city
Verification: city=<some city> | stats count(longitude)
* | stats sum(http_status) BY country | sort country
Verification: country=<some country> | stats sum(http_status)
* | stats sum(http_status) BY http_method | sort http_method
Verification: http_method=<some http_method> | stats sum(http_status) // Checked this for all of them

// For multiple group BY
* | stats sum(http_status) by city, http_method | sort city
Verification: city=<some city> | stats sum(http_status) BY http_method
```
During ingestion/agileTree creation verify the number of nodes
`
INFO[2024-08-29 07:39:12] checkAndRotateColFiles: Release STB, segkey: data/sigsingle.FqXHsNRJFR8tfvhSd6sdJQ/active/ind-0/0-0-3544697602014606120/2/2, stree node count: 11266 , Each Col NodeCount: map[app_name:0 city:98 country:249 http_method:6] 
`
For a normal ingest scenario nodes seems to be similar
`
INFO[2024-08-29 07:42:25] checkAndRotateColFiles: Release STB, segkey: data/sigsingle.4KGn6NwYSP2yMpxVxcBnC6/active/ind-0/0-0-3544697602014606120/2/2, stree node count: 10266 , Each Col NodeCount: map[city:98 country:249 http_method:6]
`

# Checklist:


pkg/segment/writer/agiletree.go
```diff
@@ -279,6 +280,149 @@ func (stb *StarTreeBuilder) insertIntoTree(node *Node, colVals []uint32, recNum
...
+		if err != nil {
+			return err
+		}
+		delete(node.children, child.myKey)
+	}
+
+	return nil
+}
+
+func (stb *StarTreeBuilder) removeLevelFromTree(node *Node, currIdx uint, idxToRemove uint, lastIdx uint) error {
```
AndrewHess, line 375: nit: at some point it would be good to make a wrapper for this like `removeColumn(colName string)` that just calls this method with the correct parameters. That would also make your unit tests a little more clear because you can test on `removeColumn()` instead of testing `removeLevelFromTree()` directly

pkg/segment/writer/agiletree.go
```diff
@@ -279,6 +280,149 @@ func (stb *StarTreeBuilder) insertIntoTree(node *Node, colVals []uint32, recNum
...
+		if err != nil {
+			log.Errorf("updateAggVals: error in count err:%v", err)
+			return err
+		}
+	}
+
+	return nil
+}
+
+func (stb *StarTreeBuilder) cleanupCommonChildren(currNode *Node) error {
```
AndrewHess, line 326: So this recursively merges nodes lower in the tree where the nodes are the same now that a level somewhere above was deleted, right?

AndrewHess (commented):
I think as a draft this is looking good. I think this approach should work, but even if we go with a different approach we should be able to keep the unit tests you've added here

pkg/segment/writer/startree_test.go
```diff
@@ -594,3 +594,377 @@ func TestStarTreeMediumEncodingDecoding(t *testing.T) {
...
+
+	gcWorkBuf := make([][]string, len(groupByCols))
+	for colNum := 0; colNum < len(groupByCols); colNum++ {
+		gcWorkBuf[colNum] = make([]string, MaxAgileTreeNodeCount)
+	}
+
+	// basic test
+	var builder StarTreeBuilder
+	for trial := 0; trial < 10; trial += 1 {
+		builder.ResetSegTree(groupByCols, mColNames, gcWorkBuf)
```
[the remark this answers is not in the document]
vasukalariya: I just took the reference from the above-mentioned tests. I think it's just to ensure there is never a case for non-determinism.

pkg/segment/writer/agiletree.go
vasukalariya: it properly links the children of the common nodes to the first node along with the aggs. Then, it recursively calls this for their own children and so on. While returning it links the node with the first child and cleans up accumulated nodes i.e. `commonChildren`.

pkg/segment/writer/startree_test.go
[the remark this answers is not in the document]
AndrewHess: Ok if there's not a good reason to do it multiple times, you should do it just once. (it's fine for now, but when the PR is marked Ready for Review it should be changed unless there's a good reason to keep it)

pkg/segment/writer/agiletree.go
AndrewHess: Ok nice. When the PR is marked Ready for Review it would probably be good to rename this function to make it more obvious. I think there were a few other things that could be good to rename, but I didn't note them during my review since the PR is Draft

pkg/segment/writer/startree_test.go
```diff
@@ -769,6 +770,61 @@ func getTotalLevels(node *Node) int {
...
+	for recNum := uint16(0); recNum < numRecs; recNum += 1 {
+		for i, grpValues := range recGroupByValues[recNum] {
+			mp[builder.groupByKeys[i]][grpValues] = builder.wipRecNumToColEnc[i][recNum]
+		}
+	}
+
+	return mp, nil
+}
+
+func getTree(t *testing.T, root *Node, ans map[string][]utils.CValueEnclosure, key string) {
```
AndrewHess, line 789: nit: rename `ans`; I'm not sure what it means

pkg/segment/writer/startree_test.go
```diff
@@ -769,6 +770,61 @@ func getTotalLevels(node *Node) int {
...
+	mp := make(map[string]map[string]uint32)
+	for colNum := range builder.groupByKeys {
+		mp[builder.groupByKeys[colNum]] = make(map[string]uint32)
+	}
+
+	for recNum := uint16(0); recNum < numRecs; recNum += 1 {
+		for i, grpValues := range recGroupByValues[recNum] {
+			mp[builder.groupByKeys[i]][grpValues] = builder.wipRecNumToColEnc[i][recNum]
+		}
+	}
```
AndrewHess, line 784: nit: this code is hard for me to understand, so it takes me a while to review. Here's the questions I had reading this:
1. What is `DeriveEnc` supposed to do? Perhaps a different name would make it more clear.
2. What does `mp` store? It looks like you wanted to name it `map` but that's a keyword so you went with `mp`. Just from `mp := make(map[string]map[string]uint32)`, I have very little idea what this is going to store

pkg/segment/writer/startree_test.go
```diff
@@ -769,6 +770,61 @@ func getTotalLevels(node *Node) int {
...
+		keys = append(keys, fmt.Sprintf("%v", encMap[col][grpVals[i]]))
+	}
+	return strings.Join(keys, "_")
+}
+
+func testAggs(t *testing.T, expected []int, aggVals []utils.CValueEnclosure) {
+	assert.Equal(t, len(expected), len(aggVals))
+
+	for i := 0; i < len(expected); i++ {
+		if i%TotalMeasFns == MeasFnCountIdx {
```
AndrewHess, line 818: nit: it looks like this function implicitly assumes `expected` (and `aggVals`) are for [these](https://github.com/siglens/siglens/blob/develop/pkg/segment/writer/agiletree.go#L42-L45) measure functions, in exactly that order. It might be good to make that assumption more explicit

pkg/segment/writer/startree_test.go
```diff
@@ -817,25 +892,52 @@ func TestStarTree2(t *testing.T) {
...
-
-	gcWorkBuf := make([][]string, len(groupByCols))
-	for colNum := 0; colNum < len(groupByCols); colNum++ {
-		gcWorkBuf[colNum] = make([]string, MaxAgileTreeNodeCount)
-	}
-
 	// basic test
 	var builder StarTreeBuilder
-	for trial := 0; trial < 10; trial += 1 {
+	for trial := 0; trial < 1; trial += 1 {
```
AndrewHess, line 897: nit: Now you don't need the loop

pkg/segment/writer/startree_test.go
```diff
@@ -817,25 +892,52 @@ func TestStarTree2(t *testing.T) {
...
 		_, err = builder.EncodeStarTree(ss.SegmentKey)
 		assert.NoError(t, err)
 
+		// check the tree structure
+		assert.Equal(t, 13, len(tree))
+		key := createKey(root, encMap, groupByCols, []string{"toyota", "green", "sedan"})
+		testAggs(t, []int{10, 10, 10, 1, 9, 9, 9, 1, 5, 5, 5, 1}, tree[key])
+		key = createKey(root, encMap, groupByCols, []string{"toyota", "yellow", "sedan"})
+		testAggs(t, []int{8, 8, 8, 1, 7, 7, 7, 1, 3, 3, 3, 1}, tree[key])
+		key = createKey(root, encMap, groupByCols, []string{"toyota", "red", "suv"})
```
AndrewHess, line 918: This is hard for me to verify.

It might be good to have this for one of your tests, to make sure the tree structure is exactly what we expect. But for most of your tests, it might be better to do something like this:
1. Ingest some data
2. Run various queries on it (not necessarily a full SPL query; just get the relevant data out of the tree and aggregated and/or bucketed correctly) and ensure they all give the correct results (it should be fairly easy for someone to read the code and verify they're the expected results)
3. Drop a column (or multiple columns)
4. Run the same queries as before that you still have all the columns for, and check that they give the same results

pkg/segment/writer/startree_test.go
AndrewHess: Actually I think this is the `answer`—the result you want from running `getTree`, right? In that case, maybe renaming `getTree` would be good to clarify what exactly we are getting

pkg/segment/writer/agiletree.go
vasukalariya: addressed in the latest commit.

pkg/segment/writer/agiletree.go
vasukalariya: addressed in the latest commit.

pkg/segment/writer/startree_test.go
vasukalariya: addressed in the latest commit.

pkg/segment/writer/startree_test.go
vasukalariya: Have renamed it and added a comment explaining what it does. TLDR: There is no specific way to get the reverse mapping of colValues to enc. So iterate over the records and manually create this reverse encoding map.

pkg/segment/writer/startree_test.go
vasukalariya: addressed in the latest commit.

pkg/segment/writer/startree_test.go
vasukalariya: addressed in the latest commit.

pkg/segment/writer/startree_test.go
vasukalariya: I have verified this using an end-to-end test procedure mentioned above. Creating an end-to-end unit test will take time because a lot of setup is necessary.

pkg/segment/writer/agiletree.go
```diff
@@ -142,7 +143,11 @@ func (stb *StarTreeBuilder) GetNodeCount() int {
 func (stb *StarTreeBuilder) GetEachColNodeCount() map[string]uint32 {
 	res := make(map[string]uint32)
 	for colIdx, lastNum := range stb.segDictLastNum {
-		res[stb.groupByKeys[colIdx]] = lastNum
+		if stb.droppedCols[colIdx] {
+			res[stb.groupByKeys[colIdx]] = 0
+		} else {
+			res[stb.groupByKeys[colIdx]] = lastNum
+		}
```
AndrewHess, line 150: This seems odd to me that we need to know which columns were dropped when we get the node count for each column. Why doesn't dropping a column put the tree in the same state as it would be in if the column was never added?

pkg/segment/writer/agiletree.go
```diff
@@ -216,6 +222,34 @@ func (stb *StarTreeBuilder) DropSegTree(stbDictEncWorkBuf [][]string) {
...
+	if dropLevel == -1 {
+		return fmt.Errorf("DropColumn: column to drop %v not found", colToDrop)
+	}
+	if stb.droppedCols[dropLevel] {
+		return fmt.Errorf("DropColumn: column %v already dropped", colToDrop)
+	}
+	nextLevel := stb.getNextLevel(-1)
+	if nextLevel == -1 {
+		return fmt.Errorf("DropColumn: cannot drop last column")
+	}
```
AndrewHess, line 242: Why are there restrictions on which columns we can drop?

pkg/segment/writer/segstore.go
```diff
@@ -831,21 +831,31 @@ func CleanupUnrotatedSegment(segstore *SegStore, streamId string, resetSegstore
 	return nil
 }
 
-func (segstore *SegStore) isAnyAtreeColAboveCardLimit() (string, bool, uint64) {
+func (segstore *SegStore) getColsAboveCardLimit() ([]string, []uint64) {
 
-	for _, cname := range segstore.stbHolder.stbPtr.GetGroupByKeys() {
+	highCardCols := []string{}
+	highCardVals := []uint64{}
```
AndrewHess, line 837: nit: These are tightly coupled—they should always have the same length. Would it be better to combine them into a `map[string]uint64`? If not, what about combining them into a `[]someNewType` where `someNewType` has fields for the column name and cardinality?

pkg/segment/writer/segstore.go
```diff
@@ -938,18 +948,28 @@ func (segstore *SegStore) computeStarTree() {
...
+			log.Warnf("computeStarTree: Dropping SegTree All remaining cols %v found with high card: %v, blockNum: %v",
+				highCardCols, highCardVals, segstore.numBlocks)
 			segstore.stbHolder.stbPtr.DropSegTree(segstore.stbDictEncWorkBuf)
 			segstore.stbHolder.ReleaseSTB()
 			segstore.stbHolder = nil
 			return
 		}
+		for idx, cname := range highCardCols {
+			log.Warnf("computeStarTree: Dropping column, found cname: %v with high card: %v, blockNum: %v",
+				cname, highCardVals[idx], segstore.numBlocks)
```
AndrewHess, line 963: nit: for log messages, it would be good to be a little more clear
```suggestion
			log.Warnf("computeStarTree: Dropping column, found cname: %v with high cardinality: %v, blockNum: %v",
				cname, highCardVals[idx], segstore.numBlocks)
```

Same for lines 954-955

pkg/segment/writer/startree_test.go
AndrewHess, line 1: It would be good to add a simple test that tries to drop invalid columns

Also thanks for renaming `DeriveEnc -> GetColValueEncodings` and `getTree -> populateAggsFromTree`. The new names are much easier for me to understand

AndrewHess (commented):
This is looking pretty good. A few notes:
1. I think you're adding some unnecessary complexity with keeping track of which columns have been dropped. Then in several places when you iterate over the tree looking for certain things, you have an extra check to see if that column's been dropped. I think it would be cleaner to make dropping a column put the tree in the same state as if the tree was never built with that column. What do you think?
2. The change from `CValueEnclosure` to `Number` probably could have been a separate small PR after this one, since it seems independent of dropping columns. But it's not a problem that it's in this PR

pkg/segment/writer/agiletree.go
vasukalariya: It's just that I don't want to move all of these fields around. Even Kunal agreed that doing this will cause memory and performance issues.

pkg/segment/writer/agiletree.go
vasukalariya: Sorry the comment is incorrect. It means that we don't have any columns to drop.

pkg/segment/writer/segstore.go
vasukalariya: Have modified the code to take in columns to drop as an array of strings. If we create someNewType we would need to pass it to the agileTree and it won't be modular. I can however create a map, just let me know if this needs change as per latest code.

pkg/segment/writer/agiletree.go
```diff
@@ -222,30 +216,64 @@ func (stb *StarTreeBuilder) DropSegTree(stbDictEncWorkBuf [][]string) {
...
 	if err != nil {
 		return err
 	}
+	stb.numGroupByCols--
+	stb.groupByKeys = newGrpByKeys
 
-	stb.droppedCols[dropLevel] = true
+	// Rest of the cleanup is done in DropColumns for efficiency
 
 	return nil
```
AndrewHess, line 278: You're inside `DropColumn()` which is a public function, so you can't rely on `DropColumns()` being the the caller. Either make `DropColumn()` private, or do the cleanup here

pkg/segment/writer/segstore.go
AndrewHess: Ok, I think it would be good to combine these into one map, but I haven't thought too much about this. So use your judgement about whether that will make the code cleaner.

pkg/segment/writer/agiletree.go
```diff
@@ -222,30 +216,64 @@ func (stb *StarTreeBuilder) DropSegTree(stbDictEncWorkBuf [][]string) {
...
+		}
+		dropIndexes[colIdx] = struct{}{}
+	}
+
+	stb.segDictMap = toputils.RemoveElements(stb.segDictMap, dropIndexes)
+	stb.segDictEncRev = toputils.RemoveElements(stb.segDictEncRev, dropIndexes)
+	stb.segDictLastNum = toputils.RemoveElements(stb.segDictLastNum, dropIndexes)
+
+	// No need to update wipRecNumToColEnc, since it will be reset in each ComputeStarTree based on groupByKeys
+	stb.wipRecNumToColEnc = stb.wipRecNumToColEnc[:stb.numGroupByCols]
```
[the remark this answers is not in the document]
vasukalariya: Just resizing it directly. I am not finding the correct column indexes and filter those out properly just like I did for the previous 3 members.
```
stb.segDictMap = toputils.RemoveElements(stb.segDictMap, dropIndexes)
stb.segDictEncRev = toputils.RemoveElements(stb.segDictEncRev, dropIndexes)
stb.segDictLastNum = toputils.RemoveElements(stb.segDictLastNum, dropIndexes)
```
So, technically you are not processing it explicitly, we are just resizing it, directly cause in the ComputeStarTree call we will create the encodings and update this member.

pkg/segment/writer/agiletree.go
vasukalariya: addressed in the latest commit.

pkg/segment/writer/agiletree.go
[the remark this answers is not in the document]
vasukalariya: Made the comment more clear.

pkg/segment/writer/agiletree.go
```diff
@@ -222,30 +216,64 @@ func (stb *StarTreeBuilder) DropSegTree(stbDictEncWorkBuf [][]string) {
...
+		colIdx, exists := mapColNameToIdx[colToDrop]
+		if !exists {
+			return fmt.Errorf("DropColumns: column to drop %v not found", colToDrop)
+		}
+		err := stb.DropColumn(colToDrop)
+		if err != nil {
+			return fmt.Errorf("DropColumns: Error while dropping column %v, err: %v", colToDrop, err)
+		}
+		dropIndexes[colIdx] = struct{}{}
+	}
```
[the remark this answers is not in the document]
vasukalariya: addressed in the latest commit.

pkg/segment/writer/segstore.go
vasukalariya: addressed in the latest commit. Used the `map[string]uint64`, have not used generics on the DropColumns method, because if we want to optimize further in future regarding some estimation of the order of dropping columns based on cardinality, it would be good to have it here and currently this function is explicitly used for dropping high cardinality columns.

pkg/segment/writer/agiletree.go
```diff
@@ -230,12 +230,16 @@ func (stb *StarTreeBuilder) DropColumns(colsToDrop []string) error {
 	}
 
 	// check if all columns to drop are present
-	for _, colToDrop := range colsToDrop {
-		colIdx, exists := mapColNameToIdx[colToDrop]
-		if !exists {
-			return fmt.Errorf("DropColumns: column to drop %v not found", colToDrop)
+	for colcolsToDrop := range colsToDrop {
+		if _, exists := mapColNameToIdx[colcolsToDrop]; !exists {
+			return fmt.Errorf("DropColumns: column to drop %v not found", colcolsToDrop)
```
AndrewHess, line 235: nit: rename `colcolsToDrop` to `colToDrop`, or `col`, or similar

pkg/segment/writer/agiletree.go
```diff
@@ -216,7 +216,7 @@ func (stb *StarTreeBuilder) DropSegTree(stbDictEncWorkBuf [][]string) {
 	stb.ResetSegTree(stb.groupByKeys, stb.mColNames, stbDictEncWorkBuf)
 }
 
-func (stb *StarTreeBuilder) DropColumns(colsToDrop []string) error {
+func (stb *StarTreeBuilder) DropColumns(colsToDrop map[string]uint64) error {
```
AndrewHess, line 219: What is the `uint64` for in this map? It doesn't look like you use the `uint64` in this function, so you probably shouldn't have it.

It looks like this is actually the map of column names and their cardinality that you're getting like this:
```
colsToDrop := segstore.getColsAboveCardLimit()
```
Is that right?

Here's what I would suggest:
1. Find a one-liner way to get the keys from a map. Getting them either as a `[]string` or a `map[string]struct{}` should be fine. Write a util in `maputils.go` if you need to
2. Change `colsToDrop` either back to `[]string` or to `map[string]struct{}`
3. Keep `colsToDrop := segstore.getColsAboveCardLimit()`, but then call your one-liner to extract the keys as a slice/set, and pass that to `DropColumns()`

pkg/segment/writer/agiletree.go
```diff
@@ -246,13 +250,14 @@ func (stb *StarTreeBuilder) DropColumns(colsToDrop []string) error {
 	stb.segDictEncRev = toputils.RemoveElements(stb.segDictEncRev, dropIndexes)
 	stb.segDictLastNum = toputils.RemoveElements(stb.segDictLastNum, dropIndexes)
 
-	// No need to update wipRecNumToColEnc, since it will be reset in each ComputeStarTree based on groupByKeys
+	// No need to update wipRecNumToColEnc based on index, since it will be repopulated based on update groupByKeys in creatEnc
 	stb.wipRecNumToColEnc = stb.wipRecNumToColEnc[:stb.numGroupByCols]
```
AndrewHess, line 254: I'm still confused
1. What does "based on index" mean? I don't see a variable `index`
2. You're not calling `creatEnc()` in this (public) function though, so you don't know that the caller will invoke `creatEnc()`. If the caller doesn't, will this tree be in an invalid state?

vasukalariya: Above members are cleaned up based on the index of dropped columns. This particular member doesn't need to be cleaned up since it is repopulated in ComputeStarTree. Have added comment to clarify this.

pkg/segment/writer/agiletree.go
vasukalariya: addressed in the latest commit.

pkg/segment/writer/agiletree.go
vasukalariya: addressed in the latest commit.

AndrewHess (approved):
Looks good

AndrewHess (approved):

PR #1509, opened 2024-08-23
Numeric dict encoded values

vasukalariya:
# Description
Summarize the change.
For queries like `app_name=Bracecould | stats min(http_status)`, `http_status` which is a numeric dict encoded value was being treated as nonDeCol. Over [here](https://github.com/siglens/siglens/blob/ed6891239e65e973446a5f0783b75e32d911242b/pkg/segment/search/searchaggs.go#L988) there are no checks for int64 and float64 type values which also can be dict [encoded](https://github.com/siglens/siglens/blob/ed6891239e65e973446a5f0783b75e32d911242b/pkg/segment/reader/segread/segreader.go#L427). Added numeric stats accumulation similar to how its done over [here](https://github.com/siglens/siglens/blob/ed6891239e65e973446a5f0783b75e32d911242b/pkg/segment/search/searchaggs.go#L906).

# Testing

# Checklist:

### DIFF
--- a/pkg/segment/search/searchaggs.go
+++ b/pkg/segment/search/searchaggs.go
@@ -987,6 +987,10 @@ func applySegmentStatsUsingDictEncoding(mcr *segread.MultiColSegmentReader, filt
 				switch val := rawVal.(type) {
 				case string:
 					stats.AddSegStatsStr(lStats, colName, val, bb, aggColUsage, hasValuesFunc, hasListFunc)
+				case int64:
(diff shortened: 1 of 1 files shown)


AndrewHess (approved):
I think this is good, but I'm curious what actual issue this solves. E.g., when you ran the query before this PR, was it slower? Or did it give incorrect results, or log a lot of errors?

vasukalariya:
> I think this is good, but I'm curious what actual issue this solves. E.g., when you ran the query before this PR, was it slower? Or did it give incorrect results, or log a lot of errors?

It logged a lot of errors saying the CName is not DictEncoded. Ideally, for dict encoded shouldn't we use this?

nkunal (approved):

PR #1510, opened 2024-08-23
hll benchmark test

Macbeth98:
- Did both axiom and segmentio.

### DIFF
--- a/benchmark_test.go
+++ b/benchmark_test.go
@@ -23,6 +23,8 @@ import (
 	"testing"
 	"time"
 
+	"github.com/axiomhq/hyperloglog"
+	"github.com/cespare/xxhash"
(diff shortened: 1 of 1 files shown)


benchmark_test.go
```diff
@@ -609,3 +612,96 @@ func Benchmark_S3_segupload(b *testing.B) {
...
+
+func createUniqueStrings(n int) []string {
+	uniqueStrings := make([]string, n)
+	for i := 0; i < n; i++ {
+		uniqueStrings[i] = fmt.Sprintf("string-%d", i)
+	}
+	return uniqueStrings
+}
+
+func insertIntoAxiomHll(axiomHll *hyperloglog.Sketch, str string) {
```
nkunal, line 624: Pass the string array here and then call tgis function once

PR #1511, opened 2024-08-23
Optimize Flushing of Bloom

vasukalariya:
# Description
Summarize the change.
Optimize writing of bytesWritten by first writing dummy data and then overwritting it with the correct size.
This prevents writing the bloom twice.

Tested on a sample 20M records 1 index 10 segment test for [Benchmark_agileTreeIngest](https://github.com/siglens/siglens/blob/ed6891239e65e973446a5f0783b75e32d911242b/benchmark_test.go#L379).

Before
Total Memory used: 166.44MB
Memory used by allocation of bufio writer: 25 MB
Memory used by `bi.Bf.WriteTo(bufWriter)`: 96 MB

After
Total Memory used: 46.43MB
Memory used by additional os.Open: 1 MB

We noticed a sudden drop from 166 MB to 46 MB of memory usage. It would be even more efficient for large data sizes.

# Testing

# Checklist:

### DIFF
--- a/pkg/segment/writer/segstore.go
+++ b/pkg/segment/writer/segstore.go
@@ -1092,12 +1092,6 @@ func (ss *SegStore) flushBloomIndex(cname string, bi *BloomIndex) uint64 {
 		return 0
 	}
 
-	err = bffd.Sync()
-	if err != nil {
-		log.Errorf("flushBloomIndex: failed to sync correct bloom size fname=%v, err=%v", fname, err)
-		return 0
-	}
-
 	if len(bi.HistoricalCount) == 0 {
 		bi.HistoricalCount = make([]uint32, 0)
 	}
--- a/pkg/segment/writer/segstore.go
+++ b/pkg/segment/writer/segstore.go
@@ -1023,52 +1023,6 @@ func (segstore *SegStore) WritePackedRecord(rawJson []byte, ts_millis uint64,
 	return nil
 }
 
-// This function will write bloom to file and return the number of bytes written
-// Ensure that the caller has to again write the file with the correct bloom size later.
-func (ss *SegStore) writeBloomToFile(fname string, bi *BloomIndex) (uint32, error) {
-	bffd, err := os.OpenFile(fname, os.O_APPEND|os.O_WRONLY|os.O_CREATE, 0644)
-	if err != nil {
-		log.Errorf("writeBloomToFile: open failed fname=%v, err=%v", fname, err)
-		return 0, err
-	}
-	defer bffd.Close()
-
-	// There is no accurate way to find the size of bloom before writing it to the file.
-	// So, we will first write a dummy 4 bytes of size and then write the actual bloom size later.
-	bytesWritten := uint32(0)
-
-	_, err = bffd.Write([]byte{0, 0, 0, 0})
-	if err != nil {
-		log.Errorf("writeBloomToFile: failed to skip bytes for bloom size fname=%v, err=%v", fname, err)
-		return 0, err
-	}
-	bytesWritten += 4
-
-	// copy the blockNum
-	if _, err = bffd.Write(toputils.Uint16ToBytesLittleEndian(ss.numBlocks)); err != nil {
-		log.Errorf("writeBloomToFile: block num write failed fname=%v, err=%v", fname, err)
-		return 0, err
-	}
-	bytesWritten += utils.LEN_BLKNUM_CMI_SIZE
-
-	// write CMI type
-	if _, err = bffd.Write(utils.CMI_BLOOM_INDEX); err != nil {
-		log.Errorf("writeBloomToFile: CMI Type write failed fname=%v, err=%v", fname, err)
-		return 0, err
-	}
-	bytesWritten += 1
-
-	// write the blockBloom
-	bloomSize, err := bi.Bf.WriteTo(bffd)
-	if err != nil {
-		log.Errorf("writeBloomToFile: write blockbloom failed fname=%v, err=%v", fname, err)
-		return 0, err
-	}
-	bytesWritten += uint32(bloomSize)
-
-	return bytesWritten, nil
-}
-
 // flushes bloom index and returns number of bytes written
 func (ss *SegStore) flushBloomIndex(cname string, bi *BloomIndex) uint64 {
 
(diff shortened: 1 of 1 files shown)


pkg/segment/writer/segstore.go
```diff
@@ -1037,50 +1036,72 @@ func (ss *SegStore) flushBloomIndex(cname string, bi *BloomIndex) uint64 {
...
-	if bferr != nil {
-		log.Errorf("flushBloomIndex: write blockbloom failed fname=%v, err=%v", fname, bferr)
+	bloomSize, err := bi.Bf.WriteTo(bffd)
+	if err != nil {
+		log.Errorf("flushBloomIndex: write blockbloom failed fname=%v, err=%v", fname, err)
+		return 0
+	}
+	bytesWritten += uint32(bloomSize)
+
+	err = bffd.Close()
```
Macbeth98, line 1074: I noticed that the file is opened at the beginning, but there are multiple return points before it's closed. This could lead to the file not being closed properly if an error occurs,

I would suggest refactoring the code that opens and writes to the file into a separate function. In this way, you can do `defer` immediately like you have done when you have opened the file for the second time.

pkg/segment/writer/segstore.go
```diff
@@ -1037,50 +1036,72 @@ func (ss *SegStore) flushBloomIndex(cname string, bi *BloomIndex) uint64 {
...
+	if err != nil {
+		log.Errorf("flushBloomIndex: close failed fname=%v, err=%v", fname, err)
+		return 0
+	}
+
+	bffd, err = os.OpenFile(fname, os.O_WRONLY, 0644)
```
Macbeth98, line 1080: Also, I am curious as to why there is a need to `close` and `open` the file again to write the bytes written. Can you not do `seek` or use `WriteAt` function after writing the bloom data to write the bytes written value and then do `Sync()` and `close`?  

Is there a reason why you have done this way?

pkg/segment/writer/segstore.go
```diff
@@ -1037,50 +1036,72 @@ func (ss *SegStore) flushBloomIndex(cname string, bi *BloomIndex) uint64 {
...
 
-	defer bffd.Close()
-
-	var buf bytes.Buffer
-	bufWriter := bufio.NewWriter(&buf)
-
 	// there is no accurate way to find exactly how many bytes the write.to is going to write
 	// and we need that number , so that we write it first and then the actual bloom data
 	// hence this messiness to write it to some buffer, get the bytesWritten count and then do
 	// the actual write
```
Macbeth98, line 1042: You should also probably change this comment, as we are not writing to a buffer and we are writing directly to a file.

pkg/segment/writer/segstore.go
vasukalariya: Seek is not defined and writeAt is not supported for files opened with O_APPEND flag. That's why I am first append the data and then reopening the file with Write only mode and writing the number of bytes written at its correct position.

pkg/segment/writer/segstore.go
vasukalariya: yeah sorry, I missed it, I have addressed it in the latest commit.

pkg/segment/writer/segstore.go

pkg/segment/writer/segstore.go
```diff
@@ -1024,6 +1023,52 @@ func (segstore *SegStore) WritePackedRecord(rawJson []byte, ts_millis uint64,
...
+
+	// write the blockBloom
+	bloomSize, err := bi.Bf.WriteTo(bffd)
+	if err != nil {
+		log.Errorf("writeBloomToFile: write blockbloom failed fname=%v, err=%v", fname, err)
+		return 0, err
+	}
+	bytesWritten += uint32(bloomSize)
+
+	return bytesWritten, nil
```
nkunal, line 1069: I am pretty sure you need to do a fd sync here, so that the filesystem cache is flushed out. refer to Andrew's PR

pkg/segment/writer/segstore.go
```diff
@@ -1034,56 +1079,37 @@ func (ss *SegStore) flushBloomIndex(cname string, bi *BloomIndex) uint64 {
...
+	if err != nil {
+		log.Errorf("flushBloomIndex: open failed fname=%v for writing the bloom size, err=%v", fname, err)
 		return 0
 	}
+	defer bffd.Close()
 
-	// copy the blockNum
-	if _, err = bffd.Write(toputils.Uint16ToBytesLittleEndian(ss.numBlocks)); err != nil {
-		log.Errorf("flushBloomIndex: bloomsize write failed fname=%v, err=%v", fname, err)
+	_, err = bffd.Seek(-int64(bytesWritten), io.SeekEnd)
```
AndrewHess, line 1095: I think this is fine, but there might be a concurrency issue. Is `flushBloomIndex()` ever called concurrently? If so, then seeking from the end like this could put us at the wrong location. If that's an issue, I believe seeking from the start should fix the issue (and in that case, you could use `WriteAt()` instead of seeking and then writing

AndrewHess (commented):
> Memory used by additional os.Open: 1 MB

Instead of opening the file twice, would it make sense to do this:
1. Open the file in non-append mode
2. Seek to the end of the file
3. Write the 0 uint32 and then write the bloom
4. Seek again to where you started
5. Write the actual size of the bloom

pkg/segment/writer/segstore.go
```diff
@@ -1024,6 +1023,52 @@ func (segstore *SegStore) WritePackedRecord(rawJson []byte, ts_millis uint64,
 	return nil
 }
 
+// This function will write bloom to file and return the number of bytes written
+// Ensure that the caller has to again write the file with the correct bloom size later.
+func (ss *SegStore) writeBloomToFile(fname string, bi *BloomIndex) (uint32, error) {
```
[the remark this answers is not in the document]
vasukalariya: The only use of this function was to prevent duplicacy for fd.Close(). Now we are directly opening the file in WRITE mode so no point in making this function. Addressed in the latest commit.

pkg/segment/writer/segstore.go
```diff
@@ -1034,56 +1079,37 @@ func (ss *SegStore) flushBloomIndex(cname string, bi *BloomIndex) uint64 {
...
-
-	// there is no accurate way to find exactly how many bytes the write.to is going to write
-	// and we need that number , so that we write it first and then the actual bloom data
-	// hence this messiness to write it to some buffer, get the bytesWritten count and then do
-	// the actual write
-	bytesWritten, bferr := bi.Bf.WriteTo(bufWriter)
-	if bferr != nil {
-		log.Errorf("flushBloomIndex: write buf failed fname=%v, err=%v", fname, bferr)
 		return 0
 	}
```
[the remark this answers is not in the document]
vasukalariya: Earlier the function writeBloomToFile was logging the error so it was not needed here.

pkg/segment/writer/segstore.go
vasukalariya: Addressed in the latest commit.

nkunal (commented):
lgtm.

AndrewHess (approved):

PR #1513, opened 2024-08-24
use bitset for storing recnums for dictEnc columns

nkunal:
# Description
1. Switch `recNums` for dictEnc columns from array to bitset
2. This saves about `3.1 GB` of mem  (about `6%`) on a `1 index, 10 segment, 3.2 million recs/seg` test 

# Testing

# Checklist:

### DIFF
--- a/pkg/segment/writer/segstore.go
+++ b/pkg/segment/writer/segstore.go
@@ -63,6 +63,8 @@ const colWipsSizeLimit = 2000 // We shouldn't exceed this during normal usage.
 
 const MaxConcurrentAgileTrees = 5
 
+const BS_INITIAL_SIZE = uint32(1000)
+
 var currentAgileTreeCount int
 var atreeCounterLock sync.Mutex = sync.Mutex{}
 
@@ -130,7 +132,7 @@ func NewSegStore(orgId uint64) *SegStore {
 
 func (ss *SegStore) GetNewBitset(bsSize uint) *bitset.BitSet {
 	if ss.bsPoolCurrIdx >= ss.bsPoolCount {
-		newCount := uint32(1000)
+		newCount := BS_INITIAL_SIZE
 		ss.bsPool = toputils.ResizeSlice(ss.bsPool, int(newCount+ss.bsPoolCount))
 		for i := uint32(0); i < newCount; i++ {
 			newBs := bitset.New(bsSize)
--- a/pkg/segment/writer/segstore.go
+++ b/pkg/segment/writer/segstore.go
@@ -101,7 +101,6 @@ type SegStore struct {
 	SegmentErrors      map[string]*structs.SearchErrorInfo
 	bsPool             []*bitset.BitSet
 	bsPoolCurrIdx      uint32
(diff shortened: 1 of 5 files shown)


pkg/segment/writer/segstore.go
```diff
@@ -124,6 +128,22 @@ func NewSegStore(orgId uint64) *SegStore {
 	return segstore
 }
 
+func (ss *SegStore) GetNewBitset(bsSize uint) *bitset.BitSet {
+	if ss.bsPoolCurrIdx >= ss.bsPoolCount {
+		newCount := uint32(1000)
```
Macbeth98, line 133: nit: Can we make this value as a constant?

nkunal: done

pkg/segment/writer/segstore.go
```diff
@@ -124,6 +130,22 @@ func NewSegStore(orgId uint64) *SegStore {
...
 }
 
+func (ss *SegStore) GetNewBitset(bsSize uint) *bitset.BitSet {
+	if ss.bsPoolCurrIdx >= ss.bsPoolCount {
+		newCount := BS_INITIAL_SIZE
+		ss.bsPool = toputils.ResizeSlice(ss.bsPool, int(newCount+ss.bsPoolCount))
+		for i := uint32(0); i < newCount; i++ {
+			newBs := bitset.New(bsSize)
+			ss.bsPool[ss.bsPoolCount+i] = newBs
+		}
```
AndrewHess, line 140: nit: I've seen this same pattern recently (I think on of Mani's PRs), so we might want to make a new util like:
```
func ResizeSliceWithDefault[T any](slice []T, size int, constructor func() T)
```
(You don't need that change in this PR if you want)

pkg/segment/writer/segstore.go
```diff
@@ -96,6 +99,9 @@ type SegStore struct {
 	stbDictEncWorkBuf  [][]string
 	segStatsWorkBuf    []byte
 	SegmentErrors      map[string]*structs.SearchErrorInfo
+	bsPool             []*bitset.BitSet
+	bsPoolCurrIdx      uint32
+	bsPoolCount        uint32
```
AndrewHess, line 104: I'm not sure what the purpose of `bsPoolCount` is. Why not use `len(bsPool)` instead?

AndrewHess: nit: we might want to move these two a separate struct type that handles all the pooling logic, and just have one field here with that struct. What do you think?

AndrewHess (commented):
Looking good

pkg/segment/writer/segstore.go
nkunal: created a issue for this: https://github.com/siglens/siglens/issues/1525

pkg/segment/writer/segstore.go
nkunal: good point I remove bsPoolCount and reusing the len(bsPool)

AndrewHess (approved):

PR #1515, opened 2024-08-25
avoid string conversion during jp.ParseString and use bytes all along

nkunal:
# Description

The call to `jp.ParseString()` was using up `19 GB` during a `1 index, 10 segments` test. The call inside this function was doing `string(retVal)` where retVal was `[]byte`. All this function was doing was unescaping the incoming bytes and converting the result into string. We don't necessarily need the string and work around by just using the bytes. So switched this processing logic to only do the unescaping and returning the resultant byte array as is. 

This resulted in savings of about `13 GB` or so

# Testing

# Checklist:


AndrewHess (commented):
Looks good; mostly nits

AndrewHess (approved):
Looks good

PR #1520, opened 2024-08-26
Do not run tracing related queries if trace index count is 0

venkateshamatam:
# Description
Do not run tracing related queries if trace index count is 0.
This code uses 0 as `myID` since there is no way to retrieve the orgID from the caller in `startup.go` ([link](https://github.com/venkateshamatam/siglens/blob/29e9253627d4a58870d4df31f19f5a24cede6c39/cmd/startup/startup.go#L291)) and the current implementation uses 0 as myID ([link](https://github.com/venkateshamatam/siglens/blob/29e9253627d4a58870d4df31f19f5a24cede6c39/pkg/segment/tracing/handler/tracehandler.go#L356))
# Testing

# Checklist:


AndrewHess (approved):
Looks good

PR #1524, opened 2024-08-26
StarTreeBuilderPool for managing STBs

vasukalariya:
# Description
Summarize the change.
Created an STBHolderPool to get and release StarTreeBuilder for memory management.
In future PR, we will release memory based on specific timeouts.
Also, `usingSegTree` variable is not needed as this could be checked directly via stbHolder.

# Testing
Manually ran multiple segment ingest and monitored logs to check the allocation of stbHolders to segStore.
Also, tested the same using the [Benchmark_agileTreeIngest](https://github.com/siglens/siglens/blob/ed6891239e65e973446a5f0783b75e32d911242b/benchmark_test.go#L379).

# Checklist:

### DIFF
--- a/pkg/segment/writer/agiletree.go
+++ b/pkg/segment/writer/agiletree.go
@@ -20,12 +20,18 @@ package writer
 import (
 	"fmt"
 	"math"
+	"sync"
 	"time"
 
 	"github.com/siglens/siglens/pkg/segment/utils"
 	toputils "github.com/siglens/siglens/pkg/utils"
 	log "github.com/sirupsen/logrus"
 )
+
+const MaxConcurrentAgileTrees = 5
+
+var currentAgileTreeCount int
+var atreeCounterLock sync.Mutex = sync.Mutex{}
 
 type StarTree struct {
 	Root *Node
--- a/pkg/segment/writer/agiletree.go
+++ b/pkg/segment/writer/agiletree.go
@@ -46,13 +46,12 @@ var STBHolderPool [MaxConcurrentAgileTrees]*STBHolder
 
 type STBHolder struct {
 	stbPtr            *StarTreeBuilder
-	segStore          *SegStore
 	currentlyInUse    bool
 	lastUsedTimestamp time.Time
 	myIdx             uint16
 }
 
-func (ss *SegStore) GetSTB() *STBHolder {
+func GetSTB() *STBHolder {
 	atreeCounterLock.Lock()
 	defer atreeCounterLock.Unlock()
 	if currentAgileTreeCount == MaxConcurrentAgileTrees {
@@ -65,7 +64,6 @@ func (ss *SegStore) GetSTB() *STBHolder {
 			STBHolderPool[i].stbPtr = &StarTreeBuilder{}
 			STBHolderPool[i].currentlyInUse = true
 			STBHolderPool[i].lastUsedTimestamp = time.Now()
-			STBHolderPool[i].segStore = ss
 			STBHolderPool[i].myIdx = uint16(i)
 			currentAgileTreeCount++
 
@@ -90,7 +88,6 @@ func (stbHolder *STBHolder) ReleaseSTB() {
 
 	currentAgileTreeCount--
 	stbHolder.currentlyInUse = false
-	stbHolder.segStore = nil
 }
 
 var IdxToAgFn []utils.AggregateFunctions = []utils.AggregateFunctions{
--- a/pkg/segment/writer/agiletree.go
+++ b/pkg/segment/writer/agiletree.go
@@ -48,7 +48,6 @@ type STBHolder struct {
 	stbPtr            *StarTreeBuilder
 	currentlyInUse    bool
 	lastUsedTimestamp time.Time
-	myIdx             uint16
 }
 
 func GetSTB() *STBHolder {
@@ -60,14 +59,9 @@ func GetSTB() *STBHolder {
 
 	for i := 0; i < MaxConcurrentAgileTrees; i++ {
 		if STBHolderPool[i] == nil {
-			STBHolderPool[i] = &STBHolder{}
-			STBHolderPool[i].stbPtr = &StarTreeBuilder{}
-			STBHolderPool[i].currentlyInUse = true
-			STBHolderPool[i].lastUsedTimestamp = time.Now()
-			STBHolderPool[i].myIdx = uint16(i)
-			currentAgileTreeCount++
-
-			return STBHolderPool[i]
+			STBHolderPool[i] = &STBHolder{
+				stbPtr: &StarTreeBuilder{},
+			}
 		}
 
 		if !STBHolderPool[i].currentlyInUse {
--- a/pkg/segment/writer/agiletree.go
+++ b/pkg/segment/writer/agiletree.go
@@ -59,7 +59,7 @@ type STBHolder struct {
 func GetSTB() *STBHolder {
 	atreeCounterLock.Lock()
 	defer atreeCounterLock.Unlock()
-	if currentAgileTreeCount == MaxConcurrentAgileTrees {
+	if currentAgileTreeCount >= MaxConcurrentAgileTrees {
 		return nil
 	}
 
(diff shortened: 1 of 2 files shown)


pkg/segment/writer/agiletree.go
```diff
@@ -41,6 +42,54 @@ const (
...
+			return STBHolderPool[i]
+		}
+
+		if !STBHolderPool[i].currentlyInUse {
+			STBHolderPool[i].currentlyInUse = true
+			STBHolderPool[i].lastUsedTimestamp = time.Now()
+			currentAgileTreeCount++
+
+			return STBHolderPool[i]
+		}
```
AndrewHess, line 79: nit: you have quite a bit of duplicate code. I'd suggest one of these options:
1. In the case `if STBHolderPool[i] == nil`, just do the basic setup, and then fall through to `if !STBHolderPool[i].currentlyInUse` to do the code that both parts have
2. Change `var STBHolderPool [MaxConcurrentAgileTrees]*STBHolder` to `var STBHolderPool [MaxConcurrentAgileTrees]STBHolder`, so that you don't need the case `if STBHolderPool[i] == nil` at all. `STBHolder` is already pretty small, so this will only use a little memory (and it'll be on the stack)

AndrewHess (commented):
Looks pretty good; just some nits

pkg/segment/writer/agiletree.go
```diff
@@ -41,6 +42,54 @@ const (
...
 )
 
+var STBHolderPool [MaxConcurrentAgileTrees]*STBHolder
+
+type STBHolder struct {
+	stbPtr            *StarTreeBuilder
+	currentlyInUse    bool
+	lastUsedTimestamp time.Time
+	myIdx             uint16
+}
```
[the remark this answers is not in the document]
vasukalariya: Yes it was only used for logging, have removed it.

pkg/segment/writer/segstore.go
```diff
@@ -259,7 +257,11 @@ func (segstore *SegStore) resetSegStore(streamid string, virtualTableName string
 	segstore.LastSegPqids = make(map[string]struct{})
 	segstore.numBlocks = 0
 	segstore.timeCreated = time.Now()
-	segstore.usingSegTree = false
+	if segstore.stbHolder != nil {
+		log.Infof("resetSegStore: Release STB due to reset: %v\n", segstore.stbHolder.myIdx)
```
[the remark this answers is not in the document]
vasukalariya: Addressed in the latest commit.

pkg/segment/writer/agiletree.go
vasukalariya: Went with the first option since we are also allocating StarTreeBuilder when we assign STBHolder and STB has a lot of fields. Addressed in the latest commit.

pkg/segment/writer/agiletree.go
```diff
@@ -41,6 +42,48 @@ const (
...
+var STBHolderPool [MaxConcurrentAgileTrees]*STBHolder
+
+type STBHolder struct {
+	stbPtr            *StarTreeBuilder
+	currentlyInUse    bool
+	lastUsedTimestamp time.Time
+}
+
+func GetSTB() *STBHolder {
+	atreeCounterLock.Lock()
```
nkunal, line 54: since the atree locks are just used in this file, then lets move the definition in to this file. same comment for the MaxConcurrentAgileTrees

pkg/segment/writer/segstore.go
```diff
@@ -698,6 +686,14 @@ func (segstore *SegStore) checkAndRotateColFiles(streamid string, forceRotate bo
 		bytesWritten := segstore.flushStarTree()
 		segstore.OnDiskBytes += uint64(bytesWritten)
 
+		if config.IsAggregationsEnabled() && segstore.stbHolder != nil {
+			nc := segstore.stbHolder.stbPtr.GetNodeCount()
+			cnc := segstore.stbHolder.stbPtr.GetEachColNodeCount()
+			log.Infof("checkAndRotateColFiles:  Release STB, stree node count: %v , Each Col NodeCount: %v", nc, cnc)
```
nkunal, line 692: log the segkey here

pkg/segment/writer/segstore.go
```diff
@@ -875,26 +871,15 @@ func (segstore *SegStore) computeStarTree() {
...
-			// amongst other indices/segstores
-			currentAgileTreeCount++
-			hasTreeSpace = true
-		}
-		atreeCounterLock.Unlock()
-		if !hasTreeSpace {
-			segstore.usingSegTree = false
+		segstore.stbHolder = GetSTB()
+		if segstore.stbHolder == nil {
+			log.Infof("computeStarTree: Failed to get STB")
```
nkunal, line 879: we should not declare this as a "failed to get" , becayse we may have already used up the 5 concurrent trees, we shuold just treat this as `hasTreeSpace` was false like it was previously

pkg/segment/writer/segstore.go
```diff
@@ -908,11 +893,10 @@ func (segstore *SegStore) computeStarTree() {
 			}
 		}
 
-		segstore.sbuilder.ResetSegTree(sortedGrpCols, mCols,
-			segstore.stbDictEncWorkBuf)
+		segstore.stbHolder.stbPtr.ResetSegTree(sortedGrpCols, mCols, segstore.stbDictEncWorkBuf)
 	}
 
-	if !segstore.usingSegTree { // if tree creation had failed on first block, then skip it
+	if segstore.stbHolder == nil { // if tree creation had failed on first block, then skip it
```
nkunal, line 899: the comment needs to be updated based on my above feedback

pkg/segment/writer/agiletree.go
vasukalariya: Addressed in the latest commit.

pkg/segment/writer/segstore.go
vasukalariya: Addressed in the latest commit.

pkg/segment/writer/segstore.go
vasukalariya: I was just logging it for debugging purposes. Addressed in the latest commit.

pkg/segment/writer/segstore.go
vasukalariya: Addressed in the latest commit.

pkg/segment/writer/segstore.go
```diff
@@ -280,7 +273,11 @@ func (segstore *SegStore) resetSegStore(streamid string, virtualTableName string
 	segstore.LastSegPqids = make(map[string]struct{})
 	segstore.numBlocks = 0
 	segstore.timeCreated = time.Now()
-	segstore.usingSegTree = false
+	if segstore.stbHolder != nil {
+		log.Infof("resetSegStore: Release STB due to reset")
```
nkunal, line 277: we can remove this log line, not that helpful

nkunal (commented):
looks great. One minot comment. 
I will also run some tests on this PR tomorrow morning to make sure, everything still works correctly.

pkg/segment/writer/segstore.go
nkunal: @vasukalariya you missed this comment

pkg/segment/writer/agiletree.go
```diff
@@ -41,6 +48,48 @@ const (
...
+	defer atreeCounterLock.Unlock()
+	if currentAgileTreeCount == MaxConcurrentAgileTrees {
```
nkunal, line 62: nit: I think what you have here is fine. But being too strict on the equality check is bit nervous. Just in case if there are bugs in rest of the code base and the`currentAgileTreeCount` goes past `MaxConcurrentAgileTrees` then we will keep adding trees. In such scenarios I am usually follow 
`currentAgileTreeCount >= MaxConcurrentAgileTrees` pattern to be doubly sure about this.
But again up to you if you want to keep it this way or switch to `>=`

pkg/segment/writer/segstore.go
vasukalariya: addressed in the latest commit.

pkg/segment/writer/agiletree.go
vasukalariya: addressed in the latest commit.

PR #1528, opened 2024-08-27
Use number union for node.AggValues

nkunal:
# Description
Switch from using `CvalEnclosure (interface)` to a number union type.
This saves about `1.5 GB` per agile tree

# Testing

# Checklist:

### DIFF
--- a/pkg/segment/utils/number.go
+++ b/pkg/segment/utils/number.go
@@ -1,0 +1,208 @@
+// Copyright (c) 2021-2024 SigScalr, Inc.
+//
+// This file is part of SigLens Observability Solution
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Affero General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+package utils
+
+import (
+	"fmt"
+	"math"
+
+	"github.com/siglens/siglens/pkg/utils"
+)
+
+type numberType = byte
+
+const (
+	invalidType numberType = iota
+	int64Type
+	float64Type
+	backfillType
+)
+
+type Number struct {
+	bytes [9]byte // 8 bytes for the number; last byte for the type
+}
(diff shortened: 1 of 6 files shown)


pkg/segment/utils/number.go
```diff
@@ -0,0 +1,237 @@
...
+	}
+
+	if n.bytes[8] != float64Type {
+		return 0, fmt.Errorf("Not a float64, t: %v", n.bytes[8])
+	}
+
+	return utils.BytesToFloat64LittleEndian(n.bytes[:8]), nil
+}
+
+func ConvertBytes(buf []byte) (int64, float64, SS_DTYPE) {
```
AndrewHess, line 84: Should this be a method on `Number`?

AndrewHess: nit: rename `ConvertBytes`. I don't really know what that means without reading the whole function

pkg/segment/utils/number.go
```diff
@@ -0,0 +1,237 @@
...
+	if n.bytes[8] == int64Type {
+		ni, err := n.Int64()
+		if err != nil {
+			return err
+		}
+		n.SetFloat64(float64(ni))
+		return nil
+	}
+
+	if n.bytes[8] != float64Type {
```
AndrewHess, line 122: nit: maybe use `switch n.bytes[8]` instead of three `if`s. But feel free to keep what you have if you prefer

Same for `ConvertToInt64()`

pkg/segment/utils/number.go
```diff
@@ -0,0 +1,237 @@
...
+	return nil
+}
+
+func (n *Number) Reset() {
+	n.bytes[8] = invalidType
+}
+
+func (n *Number) Type() numberType {
+	return n.bytes[8]
+}
```
AndrewHess, line 155: Does this work? `numberType` isn't exported. Should `Type()` be `type()` instead?

AndrewHess (commented):
Looks pretty good. Can you also add some unit tests for number.go?

pkg/segment/utils/number.go
nkunal: will switch to `ConvertBytesToNumber` , the caller of this function needs to get back the int64, float64 and the type, since the rest of the code where it uses these return variables need that. And there is lot of changes if we switch this to be a method on `Number`.
eventually we would get rid of the `Number` class and use the `NumTypeEnclosure` which was actually used for this purpose.

pkg/segment/utils/number.go
nkunal: it was only used internally, so switched to unexporting it

pkg/segment/utils/number.go
nkunal: changed to `switch`

AndrewHess (approved):

PR #1532, opened 2024-08-27
Bugfix 1502

### LINKED ISSUE #1502: [BUG] stats list after a sort causes a panic
# Description
Running a query like `... | sort colA | stats list(colA)` causes a panic. For example, 
```
country=Pitcairn AND weekday=Monday | sort app_version | stats list(app_version)
```
Without the `sort`, it doesn't panic.

Here's part of the panic output:
```
panic: interface conversion: interface {} is []string, not int64

goroutine 81 [running]:
github.com/siglens/siglens/pkg/segment/search.PerformMeasureAggsOnRecs.func1()
        /home/andrew/sigscalr/hyperion/siglens/pkg/segment/search/searchaggs.go:638 +0x355
github.com/siglens/siglens/pkg/segment/search.PerformMeasureAggsOnRecs(0xc024e2a000, 0xc00939bec0, 0xc024cf7018, 0xc024cf7018?, 0x1, 0x0)
        /home/andrew/sigscalr/hyperion/siglens/pkg/segment/search/searchaggs.go:653 +0x667
github.com/siglens/siglens/pkg/segment/search.PerformAggsOnRecs(0xc024e2a000?, 0xc024a28300?, 0xc00939bec0?, 0x6?, 0xc00939a390?, 0x0?, 0x2?)
        /home/andrew/sigscalr/hyperion/siglens/pkg/segment/search/searchaggs.go:362 +0x54
```

# Steps to Reproduce
1. ...

# Expected Behavior
Should not panic; should give the same results as without the `sort` but with the multivalue list elements sorted

# Environment
- Commit Hash: 9ea9a4a

sathkrith:
# Description
Fixes panic when list, value is run after sort.
Fixes string stats not being initialized in PerformMeasureAggsOnRecs of pkg/segment/search/searchaggs.go
Fixes running string stats not being updated in UpdateSegmentStats of pkg/segment/results/segresults/segresults.go

This does not fix the sort command not affecting stats list. PerformMeasureAgg accepts records as a map, invalidating the changes made by sort. Supporting sort before list would require refactoring PerformMeasureAgg and other functions that come after it.

Fixes #1502  (link all the GitHub issues this addresses)

# Testing
Run query:
app_name=Bracecould | sort http_status | stats list(http_status) as list
app_name=Bracecould| sort http_status | stats values(http_status) as values
or any other query with sort followed by stats list or values

# Checklist:

- [x ] I have self-reviewed this PR.
- [x ] I have removed all print-debugging and commented-out code that should not be merged.
- [x ] I have added sufficient comments in my code, particularly in hard-to-understand areas.
- [x ] I have formatted the code, if applicable. For Go, I have run `goimports -w .`.


pkg/segment/search/searchaggs_test.go
AndrewHess, line 1: Is the only difference between your first and second new tests here that the first has 200 records, so `list` truncates it to 100, and your second test has 40 records so there's no truncation? Just wanted to check if I missed anything else about the differences

AndrewHess (commented):
This looks good! Just a couple minor comments

pkg/segment/search/searchaggs_test.go
sathkrith: Yes, the first test is to check if it is truncating the list and the second is if the lists get merged when we are not exceeding the size limit. I'll add comments to clarify test cases.

pkg/segment/results/blockresults/runningstats.go
```diff
@@ -583,6 +583,9 @@ func (rr *RunningBucketResults) AddEvalResultsForList(runningStats *[]runningSta
 	if err != nil {
 		return 0, fmt.Errorf("RunningBucketResults.AddEvalResultsForList: failed to evaluate ValueColRequest to string, err: %v", err)
 	}
+	if len(result) > utils.MAX_SPL_LIST_SIZE {
+		(*runningStats)[i].rawVal.CVal = result[:utils.MAX_SPL_LIST_SIZE]
+	}
 	(*runningStats)[i].rawVal.CVal = result
```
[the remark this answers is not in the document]
sathkrith: You are correct, fixed this in the latest commit.

pkg/segment/search/searchaggs.go
```diff
@@ -633,6 +658,12 @@ func PerformMeasureAggsOnRecs(nodeResult *structs.NodeResult, recs map[string]ma
...
 			if value.Dtype == utils.SS_DT_FLOAT {
 				value.CVal = humanize.CommafWithDigits(value.CVal.(float64), 3)
+			} else if value.Dtype == utils.SS_DT_STRING_SLICE {
+				strVal, err := value.GetString()
+				if err != nil {
+					log.Errorf("PerformMeasureAggsOnRecs: failed to obtain string representation of slice %v: %v", value, err)
+				}
+				value.CVal = strVal
 			} else {
 				value.CVal = humanize.Comma(value.CVal.(int64))
```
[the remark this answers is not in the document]
sathkrith: Added logs for error cases and moved to switch.

pkg/segment/search/searchaggs.go
```diff
@@ -656,20 +656,25 @@ func PerformMeasureAggsOnRecs(nodeResult *structs.NodeResult, recs map[string]ma
...
+					value.CVal = strVal
 				}
-				value.CVal = strVal
-			} else {
+			case utils.SS_DT_SIGNED_NUM:
 				value.CVal = humanize.Comma(value.CVal.(int64))
+			default:
+				log.Errorf("PerformMeasureAggsOnRecs: Unexpected type %v ", value.Dtype)
+				value.CVal = ""
 			}
```
AndrewHess, line 675: Thanks for adding this extra failsafe. However, setting the `CVal` and not setting the `Dtype` could potentially cause problems (e.g., if we got to this `default` because the `Dtype` is `SS_DT_BOOL`, but now `CVal` has a string, so later we might try to extract the value as a boolean and panic when it's actually a string). I think a few good options are:
1. Return an error here
2. Set the `Dtype` to `SS_INVALID` (then the `CVal` value doesn't matter)

AndrewHess (commented):
Looks good; one more issue

pkg/segment/search/searchaggs.go
sathkrith: Updated the code. Dtype is set to SS_INVALID  and is ignored in case of errors.

AndrewHess (approved):

PR #1536, opened 2024-08-27
Restoring of case-insensitive search PR plus  Bytes Lower case Conversion memory Optimization 

Macbeth98:
# Description
- Reverted the PR #1533, restoring the changes in the PR: #1497 
- Implemented an in place lower case bytes conversion.
- The benchmark now did not show any memory consumption for this lower case conversion.

# Checklist:

### DIFF
--- a/pkg/segment/writer/packer.go
+++ b/pkg/segment/writer/packer.go
@@ -18,6 +18,7 @@
 package writer
 
 import (
+	"bytes"
 	"encoding/json"
 	"errors"
 	"fmt"
@@ -25,6 +26,7 @@ import (
 	"math/rand"
 	"os"
 	"strconv"
+	"strings"
 	"testing"
 	"time"
 
@@ -382,6 +384,8 @@ func (ss *SegStore) encodeSingleDictArray(arraykey string, data []byte, maxIdx u
 			if bi != nil {
 				bi.uniqueWordCount += addToBlockBloom(bi.Bf, []byte(keyName))
 				bi.uniqueWordCount += addToBlockBloom(bi.Bf, []byte(keyVal))
+				bi.uniqueWordCount += addToBlockBloom(bi.Bf, []byte(strings.ToLower(keyName)))
+				bi.uniqueWordCount += addToBlockBloom(bi.Bf, []byte(strings.ToLower(keyVal)))
 			}
 			addSegStatsStrIngestion(ss.AllSst, keyName, []byte(keyVal))
 			if colWip.cbufidx > maxIdx {
@@ -490,6 +494,7 @@ func (ss *SegStore) encodeSingleString(key string, maxIdx uint32,
 
 	if bi != nil {
 		bi.uniqueWordCount += addToBlockBloom(bi.Bf, valBytes)
+		bi.uniqueWordCount += addToBlockBloom(bi.Bf, bytes.ToLower(valBytes))
 	}
 	if !ss.skipDe {
 		ss.checkAddDictEnc(colWip, colWip.cbuf[s:colWip.cbufidx], recNum)
--- a/pkg/segment/writer/packer.go
+++ b/pkg/segment/writer/packer.go
@@ -330,7 +330,7 @@ func (ss *SegStore) encodeSingleDictArray(arraykey string, data []byte, maxIdx u
 				log.Errorf("getNestedDictEntries error %+v", err)
 				return
 			}
-			if keyName == "" || keyType == "" || keyVal == "" {
+			if len(keyName) == 0 || keyType == "" || len(keyVal) == 0 {
 				err = fmt.Errorf("encodeSingleDictArray: Jaeger tags array should have key/value/type values")
 				log.Error(err)
 				return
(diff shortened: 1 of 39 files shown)


pkg/segment/writer/packer.go
```diff
@@ -380,8 +380,12 @@ func (ss *SegStore) encodeSingleDictArray(arraykey string, data []byte, maxIdx u
...
 			}
 			if bi != nil {
-				bi.uniqueWordCount += addToBlockBloom(bi.Bf, []byte(keyName))
-				bi.uniqueWordCount += addToBlockBloom(bi.Bf, []byte(keyVal))
+				bytesKeyName := []byte(keyName)
+				bytesKeyVal := []byte(keyVal)
+				bi.uniqueWordCount += addToBlockBloom(bi.Bf, bytesKeyName)
+				bi.uniqueWordCount += addToBlockBloom(bi.Bf, bytesKeyVal)
+				bi.uniqueWordCount += addToBlockBloom(bi.Bf, utils.BytesToLowerInPlace(bytesKeyName))
+				bi.uniqueWordCount += addToBlockBloom(bi.Bf, utils.BytesToLowerInPlace(bytesKeyVal))
```
Macbeth98, line 388: The new changes that I have done on this PR.

pkg/segment/writer/packer.go
```diff
@@ -490,11 +494,13 @@ func (ss *SegStore) encodeSingleString(key string, maxIdx uint32,
...
 	if bi != nil {
 		bi.uniqueWordCount += addToBlockBloom(bi.Bf, valBytes)
+		bi.uniqueWordCount += addToBlockBloom(bi.Bf, utils.BytesToLowerInPlace(valBytes))
 	}
 	if !ss.skipDe {
 		ss.checkAddDictEnc(colWip, colWip.cbuf[s:colWip.cbufidx], recNum)
 	}
-	addSegStatsStrIngestion(ss.AllSst, key, valBytes)
+	valueLen := uint32(len(valBytes))
+	addSegStatsStrIngestion(ss.AllSst, key, colWip.cbuf[colWip.cbufidx-valueLen:colWip.cbufidx])
```
Macbeth98, line 503: New Changes.

pkg/utils/bitutils.go
```diff
@@ -207,3 +207,39 @@ func SearchStr(needle string, haystack []string) bool {
...
+
+// This function converts the bytes to lower case in place
+func BytesToLowerInPlace(b []byte) []byte {
+	for i, c := range b {
+		if c >= 'A' && c <= 'Z' {
+			b[i] = c + 32
+		}
+	}
+	return b
+}
```
Macbeth98, line 245: New Changes

pkg/segment/writer/packer.go
AndrewHess: This looks good, but there's one more thing I just found. Here you're doing a copy to convert a string to `[]byte`:
```
bytesKeyVal := []byte(keyVal)
```
But `keyVal` is from a call to this function: https://github.com/siglens/siglens/blob/develop/pkg/segment/writer/packer.go#L403-L427 and that function gets a `[]byte` for the `keyVal` and copies it to a `string`. So I think you can make that function return a `[]byte` instead to avoid a copy. And same for `keyName`

But also be mindful you're using these a little lower: https://github.com/siglens/siglens/pull/1536/commits/53066ae4331dd47d3be9462f6dbf473dc77dab3a#diff-f71656920dd3349225bbb82c8a0daf499efcd2971346afb1c050faaa83a874d3R390 so you don't want to accidentally change the case where they're used there

AndrewHess: BTW, you don't have to point out the new changes for this PR; github let's you select a single commit or a range of commits to review, so I just selected your second commit since the initial commit was reverting something

Macbeth98: oh Yes, I forgot about that.

Macbeth98: > But also be mindful you're using these a little lower ... so you don't want to accidentally change the case where they're used there

Should I tag the function name as `UnSafe`? Refactor the function name to `UnsafeBytesToLowerInPlace` or `UnsafeBytesToLower`?

AndrewHess: No, I think what you have currently `BytesToLowerInPlace` is clear already

AndrewHess (approved):
Looks good! Feel free to merge it; I'm just not merging it in case you wanted to run more performance tests first

pkg/utils/bitutils.go
```diff
@@ -207,3 +207,39 @@ func SearchStr(needle string, haystack []string) bool {
...
+
+func PerformBytesEqualityCheck(isCaseInsensitive bool, a, b []byte) bool {
+	if isCaseInsensitive {
+		return BytesCaseInsensitiveEqual(a, b)
+	}
+	return bytes.Equal(a, b)
+}
+
+// This function converts the bytes to lower case in place
+func BytesToLowerInPlace(b []byte) []byte {
```
nkunal, line 238: we should handle unicode conversion in a separate PR

nkunal (approved):

PR #1537, opened 2024-08-27
memory optimizations: create getnumfromrec and move the compWokrBuf from wip to segstore

nkunal:
# Description
1. Create a new method that converts from rec bytes directly to `Number` type instead of the intermediary Cval
2. Move the `workBufForComp` from ColWip to segstore so that it does not get cleared out on every segment rotation.

memory Savings:
1. Saves about `80 MB` per segment
2. Saves about `60 MB` per segment

# Testing

# Checklist:


AndrewHess (dismissed):
Looks good. Minor comments, but nothing blocking

AndrewHess (approved):

PR #1540, opened 2024-08-28
Tests for new data type Number

### LINKED ISSUE #1534: [STORY] Add unit tests for Number
# Description
https://github.com/siglens/siglens/pull/1528 added a `Number` type. Add unit tests for it.

# Acceptance Criteria
- All your unit tests pass
- You've tested setting and getting values
- You've tested some error cases
- You've tested `ReduceFast`
- You've added any other tests that you think are necessary

dongrep:
# Description
Added multiple tests to test different functions available on the type Number

Fixes #1534 

# Testing
added number_test.go file with tests to test the new type

# Checklist:


pkg/segment/utils/number_test.go
```diff
@@ -0,0 +1,402 @@
...
+	assert.True(t, n.ntype() == invalidType)
+}
+
+// Test_invalidTypeReturnsErrorForInt64AndFloat64 tests converting invalidType to Int64 and Float64.
+func Test_invalidTypeReturnsErrorForInt64AndFloat64(t *testing.T) {
+	n := Number{}
+	n.SetInvalidType()
+	if val, err := n.Int64(); err == nil || val != 0 {
+		assert.Fail(t, "Expected 0, got %v with error %v", val, err)
+	}
```
Macbeth98, line 23: nit: the way you are doing the `err` and `nil` verification is fine. But you can refactor that code like this:

```
assert.Nil(t, err, "some Message")  // or assert.NoError(t, err, "some message")
assert.Equal(t, 0, val)
```

Same comment elsewhere.

pkg/segment/utils/number_test.go
```diff
@@ -0,0 +1,402 @@
...
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+// Test_isInvalid tests if the number is invalid. It should return invalidType.
+func Test_isInvalid(t *testing.T) {
+	n := Number{}
+	n.SetInvalidType()
+	assert.True(t, n.ntype() == invalidType)
```
Macbeth98, line 14: This you can change to:
`assert.Equal(t, invalidType, n.ntype())`

Macbeth98: Same comment elsewhere

pkg/segment/utils/number_test.go
```diff
@@ -0,0 +1,402 @@
...
+	n = Number{[9]byte{0, 0, 0, 0, 0, 0, 0, 0, backfillType}}
+	assert.True(t, n.ntype() == backfillType)
+
+	n = Number{[9]byte{0, 0, 0, 0, 0, 0, 0, 0, int64Type}}
+	assert.True(t, n.ntype() == int64Type)
+
+	n = Number{[9]byte{0, 0, 0, 0, 0, 0, 0, 0, float64Type}}
+	assert.True(t, n.ntype() == float64Type)
+
+}
```
Macbeth98, line 103: I am not sure what this test is trying to do? Based on set of functions and given that the `bytes` field is private. I do not think we would initialize the `Number` type in this way.

But also the type checking I think you are already doing this through the `Set**Type` functions. And this initialization to similar to those Set functions.

pkg/segment/utils/number_test.go
```diff
@@ -0,0 +1,402 @@
...
+
+// Test_numberConversion tests converting the number. It should return the correct value.
+func TestNumberConversion(t *testing.T) {
+	n := Number{}
+
+	// Test Int64 to Float64 conversion
+	n.SetInt64(42)
+	if err := n.ConvertToFloat64(); err != nil {
+		assert.Fail(t, "Conversion failed: %v", err)
+	}
```
Macbeth98, line 148: `assert` the type here: `ntype()` and then proceed to verify the `float` value. Sam comment for `float` to `int` conversion.

Macbeth98 (commented):
I have not yet fully reviewed the PR. But overall looks good. Just some few nits.

pkg/segment/utils/number_test.go
dongrep: Right makes sense. Removing the constructor initialization

pkg/segment/utils/number_test.go
dongrep: changed assert.True to Equal everywhere where applicable

pkg/segment/utils/number_test.go
dongrep: changed it to get the variables and then assert.Equal

pkg/segment/utils/number_test.go
```diff
@@ -0,0 +1,396 @@
...
+	assert.Equal(t, n.ntype(), invalidType)
+}
+
+// Test_invalidTypeReturnsErrorForInt64AndFloat64 tests converting invalidType to Int64 and Float64.
+func Test_invalidTypeReturnsErrorForInt64AndFloat64(t *testing.T) {
+	n := Number{}
+	n.SetInvalidType()
+
+	_, err := n.Int64()
+	assert.NotEqual(t, nil, err)
```
Macbeth98, line 23: nit: you can do: `assert.NotNil(t, err, "custom informational message")`. 

Same comment elsewhere

pkg/segment/utils/number_test.go
```diff
@@ -0,0 +1,396 @@
...
+
+	// Test Int64
+	n.SetInt64(42)
+	if val, err := n.Int64(); err != nil || val != 42 {
+		assert.Fail(t, "Expected 42, got %v with error %v", val, err)
+	}
+
+	// Test Float64
+	n.SetFloat64(3.14)
+	if val, err := n.Float64(); err != nil || math.Abs(val-3.14) > 1e-6 {
```
Macbeth98, line 109: Can you not directly do equal? Since `3.14` is also float?

Also use: `assert.Nil()`, `assert.Equal()`, `assert.NotNil()`

pkg/segment/utils/number_test.go
Macbeth98: Ahh I see you did the assertion before the conversion. That is fine. But I actually meant after the conversion. Do an assert on the type.

pkg/segment/utils/number_test.go
```diff
@@ -0,0 +1,396 @@
...
+	}
+}
+
+// Test_numberCopyToBuffer tests copying the number to a buffer. It should return the correct value.
+func Test_numberCopyToBuffer(t *testing.T) {
+	n := Number{}
+	n.SetFloat64(3.14)
+	buf := make([]byte, 9)
+	n.CopyToBuffer(buf)
+	if buf[8] != float64Type {
```
Macbeth98, line 163: User `assert.Equal`. Please change any other places where you have code like this.

pkg/segment/utils/number_test.go
dongrep: Ohh yes right misunderstood it. Comparing the type after conversion.

pkg/segment/utils/number_test.go
dongrep: The NotNil gives more information. Makes sense. Changed it

pkg/segment/utils/number_test.go
dongrep: Changed all places where condition was used to appropriate asserts

pkg/segment/utils/number_test.go
dongrep: Changed

pkg/segment/utils/number_test.go
dongrep: using abs value because of the float precision error. It works without checking precision, but this makes sure it handles the possible issues as well.

.gitignore
```diff
@@ -87,4 +87,11 @@ siglens.db
...
-vector_log_lib/
\ No newline at end of file
+vector_log_lib/
+
+
+# MacOS
+.DS_Store
+
+# VS Code
+.vscode/
```
AndrewHess, line 97: We already have these (lines 20 and 50). Were the existing rules not working for you?

AndrewHess (commented):
The tests look good!

.gitignore
dongrep: I could see that the .vscode folder was being added to the commit. I see that the .gitignore has .vscode/* which ignores files inside it.

Will remove it.

dongrep:
@AndrewHess @Macbeth98 Do we anything else to close this PR?

AndrewHess:
Nope, this looks good. Sorry I forgot to get back to this earlier.

PR #1541, opened 2024-08-28
fix: fixes issue with arithmetic operation on metrics without filters

Macbeth98:
# Description
- This fixes the issue where there are multiple series on both of the metric queries in a formula

# Testing
- Added a new unit test case

# Checklist:

### DIFF
--- a/pkg/segment/segexecution.go
+++ b/pkg/segment/segexecution.go
@@ -91,7 +91,7 @@ func ProcessQueryArithmeticAndLogical(queryOps []structs.QueryArithmetic, resMap
 	IsScalar := false
 	var scalarValue float64
 
-	finalResult, scalarValuePtr, err := processQueryArithmeticNodeOp(&queryOps[0], resMap, &operationCounter, opLabelsDoNotNeedToMatch)
+	finalResult, scalarValuePtr, err := processQueryArithmeticNodeOp(&queryOps[0], resMap, &operationCounter, &opLabelsDoNotNeedToMatch)
 	if err != nil {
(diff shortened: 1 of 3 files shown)


pkg/segment/segexecution.go
```diff
@@ -335,6 +340,9 @@ func HelperQueryArithmeticAndLogical(queryOp *structs.QueryArithmetic, resMap ma
 				for rGroupId := range resultRHS.Results {
 					matchingLabelValTorightGroupID[matchingLabelValStr] = rGroupId
 				}
+			} else {
+				// ManyToMany, Set this to false so that only the matching label sets are processed.
+				*opLabelsDoNotNeedToMatch = false
 			}
```
AndrewHess, line 346: This might be fine; however, I am a little concerned that this is being set so deeply in the call stack (`ProcessQueryArithmeticAndLogical` gets called with a boolean, which passes a pointer to that bool to `processQueryArithmeticNodeOp`, which calls `HelperQueryArithmeticAndLogical`, which sets the boolean here). I think this will make the code harder to reason about if we have to make changes later

What do you think about this: before calling `ProcessQueryArithmeticAndLogical`, we do some extra work to find the correct value of `opLabelsDoNotNeedToMatch`, then pass it as a boolean (not a pointer) through the whole call stack.

Is that feasible? Do you think it would help make the code easier to understand?

Macbeth98: Yes, I have updated the code to reflect your comment.

AndrewHess (approved):

PR #1546, opened 2024-08-28
Added SigLens-Hll and refactored to use existing byte slice

Macbeth98:
# Description
- Added SigLens forked `go-hll` and refactored the code to use in place `ToBytes` for `Hll`.

# Testing
- Added a new unit test case to verify the segstats written to the buffer.
- All test cases pass.

# Checklist:

### DIFF
--- a/pkg/segment/writer/segstore.go
+++ b/pkg/segment/writer/segstore.go
@@ -104,8 +104,8 @@ type PQTracker struct {
 	PQNodes     map[string]*structs.SearchNode // maps pqid to search node
 }
 
-func InitSegStore(segmentKey string, segbaseDir string, suffix uint64,
-	virtualTableName string, skipDe bool, orgId uint64) *SegStore {
+func InitSegStore(segmentKey string, segbaseDir string, suffix uint64, virtualTableName string,
+	skipDe bool, orgId uint64, highTs uint64, lowTs uint64) *SegStore {
 
 	segStore := NewSegStore(orgId)
 	segStore.SegmentKey = segmentKey
@@ -114,6 +114,10 @@ func InitSegStore(segmentKey string, segbaseDir string, suffix uint64,
 	segStore.VirtualTableName = virtualTableName
 	segStore.skipDe = skipDe
 	segStore.OrgId = orgId
+
+	segStore.initWipBlock()
+	segStore.wipBlock.blockSummary.HighTs = highTs
+	segStore.wipBlock.blockSummary.LowTs = lowTs
 
 	return segStore
 }
--- a/pkg/segment/writer/segstore.go
+++ b/pkg/segment/writer/segstore.go
@@ -1490,15 +1490,15 @@ func writeSstToBuf(sst *structs.SegStats, buf []byte) (uint16, error) {
 	copy(buf[idx:], toputils.Uint64ToBytesLittleEndian(sst.Count))
 	idx += 8
 
-	hllData := sst.GetHllBytes()
+	hllDataSize := sst.GetHllDataSize()
 
 	// HLL_Size
-	copy(buf[idx:], toputils.Uint16ToBytesLittleEndian(uint16(len(hllData))))
+	copy(buf[idx:], toputils.Uint16ToBytesLittleEndian(uint16(hllDataSize)))
 	idx += 2
 
 	// HLL_Data
-	copy(buf[idx:], hllData)
-	idx += uint16(len(hllData))
+	hllByteSlice := sst.GetHllBytesInPlace(buf[idx : idx+uint16(hllDataSize)])
+	idx += uint16(len(hllByteSlice))
 
 	if !sst.IsNumeric {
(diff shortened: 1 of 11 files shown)


pkg/segment/writer/segstore_test.go
AndrewHess, line 1: (no need to change) This looks good. I was going to suggest an alternative approach: instead of directly testing `writeSstToBuf()`, call that function to encode it, then call a different function to decode it, and check that you got the original back. One benefit of that approach is that it's not as tightly coupled to our exact encoding process, so if we decided to change our encoding and correctly updated both the reader and writer, we wouldn't have to change the unit test. However, it looks like the reader and writer are in different packages, so it would be a hassle to do that.

AndrewHess (commented):
Looks good; just one question

pkg/segment/writer/segstore.go
```diff
@@ -1490,15 +1490,21 @@ func writeSstToBuf(sst *structs.SegStats, buf []byte) (uint16, error) {
...
 
 	// HLL_Data
-	copy(buf[idx:], hllData)
-	idx += uint16(len(hllData))
+	hllByteSlice := sst.GetHllBytesInPlace(buf[idx : idx+uint16(hllDataSize)])
+	hllByteSliceLen := len(hllByteSlice)
+	if hllByteSliceLen != hllDataSize {
+		// This case should not happen, but if it does, we need to adjust the size
+		log.Errorf("writeSstToBuf: hllByteSlice size mismatch, expected: %v, got: %v", hllDataSize, hllByteSliceLen)
+		copy(buf[idx-2:idx], toputils.Uint16ToBytesLittleEndian(uint16(hllByteSliceLen)))
```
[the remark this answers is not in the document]
Macbeth98: Yes, later I have created a new PR, that includes the `3 header bytes` also in the size. So this case should not happen.

pkg/segment/writer/segstore.go
```diff
@@ -1513,13 +1513,16 @@ func writeSstToBuf(sst *structs.SegStats, buf []byte) (uint16, error) {
 	idx += 2
 
 	// HLL_Data
-	hllByteSlice := sst.GetHllBytesInPlace(buf[idx : idx+uint16(hllDataSize)])
+	hllDataSliceFullCap := buf[idx : idx+uint16(hllDataSize)]
+	hllByteSlice := hllDataSliceFullCap[:len(hllDataSliceFullCap):len(hllDataSliceFullCap)] // len = cap
+	hllByteSlice = sst.GetHllBytesInPlace(hllByteSlice)
```
AndrewHess, line 1518: This is interesting. I'd recommend adding a bit longer of a comment explaining why it's necessary that we go through these extra steps to create a buffer where the length and capacity are the same, since we generally don't need to do that

AndrewHess (commented):
Looks good; minor comment

pkg/segment/writer/segstore.go
```diff
@@ -1506,15 +1506,32 @@ func writeSstToBuf(sst *structs.SegStats, buf []byte) (uint16, error) {
 	copy(buf[idx:], toputils.Uint64ToBytesLittleEndian(sst.Count))
 	idx += 8
 
-	hllData := sst.GetHllBytes()
+	hllDataSize := sst.GetHllDataSize()
```
nkunal, line 1509: this is actually pre-existing issue. `idx` is `uint16` so max val it can hold is 65535. However when we encode hll and other segstats, and if we change some of the hllDefaultSettings , it will cross 65K and we will have an overflow in the buf. (buf can hold upto 2 MB). 
But then the other issue is this version of this sst file allocates only 2 bytes for the hll size, so we will have to change that as well.

pkg/segment/writer/segstore.go
```diff
@@ -1506,15 +1506,32 @@ func writeSstToBuf(sst *structs.SegStats, buf []byte) (uint16, error) {
...
-	idx += uint16(len(hllData))
+	hllDataSliceFullCap := buf[idx : idx+uint16(hllDataSize)]
+
+	// Ensures that the slice has a full capacity where len(slice) == cap(slice).
+	// This is necessary because we're using the slice to get the HLL bytes in place,
+	// and the HLL package relies on cap(slice) to determine where to write the bytes.
+	// It expects the slice to be exactly the size of hllDataSize. But the slice we're
+	// using is larger than that, so we need to ensure that the slice is exactly the size
+	// of hllDataSize.
+	hllByteSlice := hllDataSliceFullCap[:len(hllDataSliceFullCap):len(hllDataSliceFullCap)]
```
nkunal, line 1524: do we need the `hllDataSliceFullCap`, can't we do this:

`hllByteSlice :=buf[idx: idx + hllDataSize : hllDataSize]`

nkunal (commented):
a few minor nits, but looks great overall

nkunal (approved):

PR #1548, opened 2024-08-28
Playground query suite CSV file

Macbeth98:
# Description
- Added the Playground csv file to SigClient.

# Checklist:


tools/sigclient/datafiles/SPL_Playground.csv
```diff
@@ -0,0 +1,119 @@
+* | search hobby=Walking
+search app_name=Codhave
+search http_status>400
+http_status=20*
+* | hobby=Act*
+"""* | hobby=""""Slot car*"""""""
+"""search app_name=Martencan address=""""157 Overpass berg, Buffalo, Alaska 51094"""""""
```
AndrewHess, line 7: Why do you need so many quotes here? I thought `""` would escape the `"` so you could just do:
```
"* | hobby=""Slot car*"""
```
which—after escaping—would effectively be:
```
* | hobby="Slot car*"
```

AndrewHess (commented):
Looks good. A couple questions:
1. How should we run these queries? Is there a script, or is this just meant as a reference that we'll copy-paste?
2. Are these just meant for benchmarking?

Macbeth98:
> Looks good. A couple questions:
> 
> 1. How should we run these queries? Is there a script, or is this just meant as a reference that we'll copy-paste?
> 2. Are these just meant for benchmarking?

Yes there is a function in the SigClient that will let you run these queries. And yes it is for benchmarking specifically checking response time of each query. This is the query suite that we ran on the GCP instances (S1, S2, S3).

These are the options you need to send to enable this in SigClient along with this file path: https://github.com/siglens/siglens/blob/develop/tools/sigclient/README.md?plain=1#L27-L28

AndrewHess (approved):

PR #1560, opened 2024-08-29
Bump version to 0.2.32

Macbeth98:


pkg/config/version.go
```diff
@@ -21,4 +21,4 @@
 
 package config
 
-const SigLensVersion = "0.2.31d"
+const SigLensVersion = "0.2.32"
```
venkateshamatam, line 24: This should be 0.2.31, right?

PR #1564, opened 2024-08-29
feat: update the size limit and encoding of Seg-Stats 

Macbeth98:
# Description
- Now the SegStats can write HLL Data of size more `uint16` and up to max of `uint32`.
- This change also required to change the `SegstatsLen` from `uint16` to `uint32`.
- Maintained version compatibility.

# Testing
- All uint test cases passed

# Checklist:


pkg/segment/utils/segconsts.go
```diff
@@ -148,6 +148,11 @@ var VERSION_TSGFILE = []byte{0x01}
 var VERSION_MBLOCKSUMMARY = []byte{0x01}
 
 var VERSION_SEGSTATS = []byte{2}
+var VERSION_SEGSTATS_LEGACY = []byte{1}
+
```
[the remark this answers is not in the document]
AndrewHess: I partially agree with this. It might be good to mention what `VERSION_SEGSTATS` vs `VERSION_SEGSTATS_BUF` means, but I don't think it's necessary to comment on `VERSION_SEGSTATS` vs `VERSION_SEGSTATS_LEGACY`

pkg/segment/reader/segread/segstatsreader.go
```diff
@@ -102,22 +115,30 @@ func readSingleSst(fdata []byte, qid uint64) (*structs.SegStats, error) {
...
 
 	switch version {
-	case utils.VERSION_SEGSTATS[0]:
+	case utils.VERSION_SEGSTATS_BUF[0]:
+		hllSize = toputils.BytesToUint32LittleEndian(fdata[idx : idx+4])
+		idx += 4
+	case utils.VERSION_SEGSTATS_BUF_LEGACY_2[0], utils.VERSION_SEGSTATS_BUF_LEGACY_1[0]:
+		hllSize = uint32(toputils.BytesToUint16LittleEndian(fdata[idx : idx+2]))
+		idx += 2
+	default:
```
AndrewHess, line 127: Why don't you handle `VERSION_SEGSTATS_BUF_LEGACY_1`? Even if you handle it just by throwing an error because that version is no longer supported, I think that will be more useful/clear to someone debugging than it will be if you let it fall to the `default` case

Macbeth98: I am handling that error case below here: https://github.com/siglens/siglens/pull/1564/files#diff-93111a0cc266401b8aa0d7c5db0fa5cf6ef706e35b8b4ed2d54c0e54459a9c98R132-R134

This is just after the switch case and in the switch case I am reading the `hllSize` for both legacy 1 and legacy 2 versions in the same case. As they are both have size to `2 bytes`. 
And we don't want to error out even for `legacy 1` cause we want to read other segstats.
The corresponding code:

```
if version == utils.VERSION_SEGSTATS_BUF_LEGACY_1[0] {
		log.Infof("qid=%d, readSingleSst: ignoring Hll (old version)", qid)
	}
```

AndrewHess: Oh, my bad. This looks good

AndrewHess (approved):

PR #1565, opened 2024-08-29
Bug: Fix for incorrect results due to scroll

vasukalariya:
# Description
Summarize the change.
Due to the updation of scroll from [here](https://github.com/siglens/siglens/blob/a0fb2f06d888594cd72d209fce3bfccc9112e334/pkg/ast/pipesearch/wsSearchHandler.go#L162), we are modiying the number of [records](https://github.com/siglens/siglens/blob/a0fb2f06d888594cd72d209fce3bfccc9112e334/pkg/segment/query/querystatus.go#L619) to pass for this instance of query execution. This results in loss of records that are present in the later segments but  are present before the scrollFrom records due to sort.

For e.g. Consider that records are sorted according to timestamp in ascending order. By default they are sorted in descending order, so same case as below can be made for it.
time segnum recordnum
0         1         1
1         1         2
2         1         3
3         2         1
4         1         4
5         2         2
6         2         3
7         1         5

Now supposed results are returned and seg 1 finishes first. So it will return 5 records and
`scrollFrom` now becomes 5. After processing of search by segment 2, we would only send 2 records for processing because from time 5 only 2 records of segment 2 are present.  Record 3 is left unprocessed.

# Testing

Ingest data across multiple segments
- `go run main.go ingest esbulk -n 10-g benchmark -d http://localhost:8081/elastic -t 100_000`
- Run query `app_name=Bracecould`, only partial results will show up due to this.

Same is the case for
- `go run main.go ingest esbulk -n 2-g benchmark -d http://localhost:8081/elastic -t 10_000`
- `country=Togo weekday=Friday`
- Here only 7 records come up due to this but stats gives correct results cause its flow does not touch this portion.

# Checklist:


Macbeth98 (commented):
Then this would mean, that we are sending all the records every time the scroll happens on the UI, right? Because `scrollFrom` value will always be `zero`?

AndrewHess (approved):
Looks good!

PR #1566, opened 2024-08-30
fix bug with resizing of slices

nkunal:
# Description

`AppendWIP..` would crash if the num of columns changed between each wip, because we were incorrectly resizing slice.

# Testing

# Checklist:

### DIFF
--- a/pkg/segment/writer/segstore.go
+++ b/pkg/segment/writer/segstore.go
@@ -1434,8 +1434,26 @@ Encoding Scheme for all columns single file
 func (ss *SegStore) FlushSegStats() error {
 
 	if len(ss.AllSst) <= 0 {
-		log.Errorf("FlushSegStats: no segstats to flush")
-		return errors.New("FlushSegStats: no segstats to flush")
+		found := 0
+		tsKey := config.GetTimeStampKey()
+		for cname, cwip := range ss.wipBlock.colWips {
+			if cwip.cbufidx > 0 {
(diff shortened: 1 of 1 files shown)


sonamgupta21 (approved):

nkunal:
@AndrewHess you should still review this , but merging for now

pkg/segment/writer/segstore.go
```diff
@@ -1440,8 +1434,26 @@ Encoding Scheme for all columns single file
...
+				log.Infof("FlushSegStats: sst nil but cname: %v, cwip.cbufidx: %v", cname,
+					cwip.cbufidx)
+				if cname != tsKey {
+					found += 1
+				}
+			}
+		}
+		// Flush is called only one of the cbufidx is >0, but if we find no columns
+		// with cbufidx > 0 other than timestamp column, only then declare this as an error
+		// else we won;t create a sst file
```
AndrewHess, line 1450: typos:
```suggestion
		// Flush is called once one of the cbufidx is >0, but if we find no columns
		// with cbufidx > 0 other than timestamp column, only then declare this as an error
		// else we won't create a sst file
```

AndrewHess: nit: it might be better to move this comment above the `for` loop, since this comment explains why we're doing the loop to find non-timestamp columns

AndrewHess (commented):
Code changes look good; just some nits about a comment

pkg/segment/writer/segstore.go
nkunal: addressed in the next PR

PR #1567, opened 2024-08-30
fix missing lock and de pack buf panics

nkunal:
# Description

1. concurrent map read/write was causing a panic during ingestion. Added missing segstore.lock
2. Don't create dependency graph if traces count is 0
3. Use cbufpool for de packing instead of a const size of 200k (fixes a panic in some scenarios)

# Testing

# Checklist:

### DIFF
--- a/pkg/segment/writer/segwriter.go
+++ b/pkg/segment/writer/segwriter.go
@@ -181,8 +181,10 @@ func GetInMemorySize() uint64 {
 
 	totalSize := uint64(0)
 	for _, s := range allSegStores {
+		s.Lock.Lock()
 		totalSize += s.wipBlock.getSize()
 		totalSize += s.GetSegStorePQMatchSize()
+		s.Lock.Unlock()
 	}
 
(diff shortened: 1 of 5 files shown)


nkunal:
@AndrewHess this is ready for review

pkg/segment/writer/segwriter.go
```diff
@@ -463,12 +465,14 @@ func InitColWip(segKey string, colName string) *ColWip {
 	}
 
 	cbuf := *wipCbufPool.Get().(*[]byte)
+	dePack := *wipCbufPool.Get().(*[]byte)
```
AndrewHess, line 468: You're using the same pool because `cbuf` and `dePack` should require roughly the same amount of memory, right? If they use significantly different amounts of memory, we might want to have separate pools

(this could also be changed in a different PR if we find it helps performance)

AndrewHess (approved):
Looks good! Just one question, but I think this can get merged as is

PR #1568, opened 2024-08-30
Implemented efficient GetString function [STORY] 

### LINKED ISSUE #1538: [STORY] Implement an efficient version of GetString
# Description

Currently in this part of the [code](https://github.com/siglens/siglens/blob/develop/pkg/es/writer/esBulkHandler.go#L233), there are multiple places where a call is made to `jp.GetString`, this method is implemented in this code base here: https://github.com/buger/jsonparser/blob/master/parser.go#L1192. It makes a copy by doing this call `string(v)`, which results in extra memory usage.

Implement a function similar to the `jp.GetString` function in our code base (refer to the code link above) and create a new method in our `pkg/utils/jsonutils.go` which will be like `utils.GetString(data []byte,  workBuf []byte, keys ...string) (val []byte, err error)`
where the user can supply a `workBuf`, if the workBuf cannot accomodate the space then do resizing and then use this workbuf. however do not do string conversion, and copy over the results from input data to this workBuf and return this workBuf. The caller will take care of whether to keep the returned value as byte itself or convert to unsafe.String.  

A similar thing was implemented in this PR: https://github.com/siglens/siglens/pull/1539, but just for the indexname parsing. 

# (Optional) Testing
Describe how this should be tested and/or which files unit tests should be added to.

For documentation changes or other similar scenarios, this section can be deleted.

# Acceptance Criteria
List the concrete items that should be completed for this task to be considered completed. Be as clear and thorough as you can.

smitneu3009:
Optimize JSON string extraction with dynamic buffer management and reduce memory usage

# Description
Summarize the change.

- Implemented a new utils.GetString function to handle JSON string extraction efficiently.
- Introduced a work buffer (workBuf) that can be resized dynamically to accommodate the string value without unnecessary memory allocations.
- Avoided conversions and direct copying to minimize extra memory usage and improve performance.

Fixes #1538

# Checklist:


smitneu3009 (commented):
Enhance GetString function with comments, improved error messages, and buffer resizing test

pkg/utils/jsonutils.go
```diff
@@ -58,3 +60,29 @@ func FlattenSingleValue(key string, m map[string]interface{}, child interface{})
 		m[key] = child
 	}
 }
+
+// GetString extracts a string value from JSON data based on the provided keys.
+// It uses the supplied workBuf to avoid unnecessary allocations and resizing.
+// If the value is not a string or if an error occurs, it returns an appropriate error.
+func GetString(data []byte, workBuf []byte, keys ...string) ([]byte, error) {
```
nkunal, line 67: nit: rename this to be `GetStringFromJson`

pkg/utils/jsonutils.go
```diff
@@ -58,3 +60,28 @@ func FlattenSingleValue(key string, m map[string]interface{}, child interface{})
...
+        return nil, err
+    }
+
+    if dataType != jsonparser.String {
+        if dataType == jsonparser.Null {
+            return nil, errors.New("Null value")
+        }
+        return nil, fmt.Errorf("value is not a string: %s", string(v))
+    }
+
```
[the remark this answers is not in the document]
nkunal: incorporate this feedback

pkg/utils/jsonutils.go
```diff
@@ -58,3 +60,29 @@ func FlattenSingleValue(key string, m map[string]interface{}, child interface{})
...
+    if dataType != jsonparser.String {
+        if dataType == jsonparser.Null {
+            return nil, fmt.Errorf("key %s has a null value", strings.Join(keys, ", "))
+        }
+        return nil, fmt.Errorf("value is not a string: %s", string(v))
+    }
+
+    if cap(workBuf) < len(v) {
+        workBuf = make([]byte, len(v))
```
nkunal, line 81: we should not erase out the old workbuf, but rather __extend__ it. Checkout examples of we have been using the `ResizeSlice` function in such cases

pkg/utils/jsonutils_test.go
```diff
@@ -279,3 +279,18 @@ func TestFlatten(t *testing.T) {
...
 	}
 }
+
+func TestGetString_BufferResizing(t *testing.T) {
+    data := []byte(`{"key": "value"}`)
+    var workBuf []byte // Initialize an empty buffer
+
+    result, err := GetString(data, workBuf, "key")
+    if err != nil {
+        t.Fatalf("expected no error, got %v", err)
```
nkunal, line 289: don't use Fatal but rather asserts, checkout the rest of the test functions how they assert test outputs

pkg/utils/jsonutils.go
smitneu3009: updated

pkg/utils/jsonutils.go
smitneu3009: updated

pkg/utils/jsonutils_test.go
smitneu3009: updated

pkg/utils/jsonutils.go
```diff
@@ -58,3 +61,26 @@ func FlattenSingleValue(key string, m map[string]interface{}, child interface{})
...
+            return nil, fmt.Errorf("key %s has a null value", strings.Join(keys, ", "))
+		}
+        return nil, fmt.Errorf("expected string value for key(s) %s but got %s", strings.Join(keys, ", "), dataType)
+    }
+
+    // Use ResizeSlice to extend workBuf
+    workBuf = ResizeSlice(workBuf, len(v))
+
+    copy(workBuf, v)
+    return workBuf, nil
```
AndrewHess, line 85: Your comment at the top of this function says
> // It uses the supplied workBuf to avoid unnecessary allocations and resizing.

But here you've extracted the value into `v` without using the `workBuf`, and then you copy it to the `workBuf`, so it looks like there's no performance benefit of having a `workBuf`. Can you change how you extract the value so that `workBuf` is actually useful and improves performance?

pkg/utils/jsonutils.go
```diff
@@ -58,3 +61,44 @@ func FlattenSingleValue(key string, m map[string]interface{}, child interface{})
...
+		return nil, fmt.Errorf("expected string value for key(s) %s but got %s", strings.Join(keys, ", "), dataType)
+	}
+
+	// Resize workBuf again to fit the actual length of the extracted value
+	workBuf = ResizeSlice(workBuf, len(value))
+
+	// Copy the extracted value directly into workBuf
+	copy(workBuf, value)
+
+	return workBuf, nil
```
AndrewHess, line 103: You're still only using `workBuf` after you've done all the real work, which defeats the purpose of `workBuf`. I'm pretty sure you'd need to pass it here
```
value, dataType, _, err = jsonparser.Get(data, keys...)
```
for it to be useful (i.e., to prevent allocations). And for that, I think you would have to write an equivalent of `Get()` in our codebase that allows passing a work buffer. Or let me know if I'm missing something

pkg/utils/jsonutils.go
```diff
@@ -58,3 +63,61 @@ func FlattenSingleValue(key string, m map[string]interface{}, child interface{})
...
+		return nil, err
+	}
+	if dataType != jsonparser.String {
+		if dataType == jsonparser.Null {
+			return nil, fmt.Errorf("key %s has a null value", strings.Join(keys, ", "))
+		}
+		return nil, fmt.Errorf("expected string value for key(s) %s but got %s", strings.Join(keys, ", "), dataType)
+	}
+
+	return []byte(strings.Trim(string(workBuf), "\"")), nil
```
AndrewHess, line 122: There seems to be a misunderstanding. The entire point of this PR (to fix https://github.com/siglens/siglens/issues/1538) is to reduce memory usage. Here you're doing something like `[]byte(string(workBuf))`, which is going to use more memory

pkg/utils/jsonutils.go
```diff
@@ -58,3 +63,61 @@ func FlattenSingleValue(key string, m map[string]interface{}, child interface{})
...
+		default:
+			return nil, jsonparser.NotExist, fmt.Errorf("expected object for key %s but found %T", key, obj)
+		}
+	}
+	valueBytes, err := json.Marshal(current)
+	if err != nil {
+		return nil, jsonparser.NotExist, fmt.Errorf("error converting value to bytes: %v", err)
+	}
+	workBuf = ResizeSlice(workBuf, len(valueBytes))
+	copy(workBuf, valueBytes)
```
AndrewHess, line 94: You're still not using the workbuf properly (i.e., even if the user passes a workbuf large enough, you'll still allocate memory and then copy it to the workbuf)

AndrewHess:
Closing due to lack of updates (feel free to reopen if you want to continue working on it)

PR #1571, opened 2024-08-30
Bump version to 0.2.32

venkateshamatam:
Bump version to 0.2.32

### DIFF
--- a/cmd/startup/startup.go
+++ b/cmd/startup/startup.go
@@ -365,7 +365,7 @@ func startQueryServer(serverAddr string) {
 					return emptyHtmlContent
 				},
 				"CSSVersion": func() string {
-					return "0.0.1"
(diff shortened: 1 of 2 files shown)


cmd/startup/startup.go
```diff
@@ -365,7 +365,7 @@ func startQueryServer(serverAddr string) {
 					return emptyHtmlContent
 				},
 				"CSSVersion": func() string {
-					return "0.0.1"
+					return "0.2.32"
```
Macbeth98, line 368: Should we change this to `0.0.2` or the same version as SigLens: `0.2.32`?

nkunal: keep it the same as the release version

Macbeth98 (approved):

PR #1577, opened 2024-09-01
perf improvements: bloom writing optimizations

nkunal:
# Description
1. This avoids looping through the colValue twice and does it only if there is an upper case letter
2. Avoids testAdd call to bloom if there is no upper case letter
3. Add to bloom during the time of wipPacking to avoid bloom adds for the dict columns for the same dict encodings
4. Switch to using `Add` call for bloom instead of `TestAdd`
5. Fix a buf with itemsLen if batchsize was greater than 4k
6. All of these improvements increase the ingestion performance by `40%`

# Testing

# Checklist:

### DIFF
--- a/pkg/es/writer/esBulkHandler.go
+++ b/pkg/es/writer/esBulkHandler.go
@@ -161,7 +161,7 @@ func HandleBulkBody(postBody []byte, ctx *fasthttp.RequestCtx, rid uint64, myid 
 		if inCount >= itemsLen {
 			newArr := make([]interface{}, 100)
 			items = append(items, newArr...)
-			itemsLen += 1000
+			itemsLen += 100
 		}
 
 		esAction, indexName, idVal := extractIndexAndValidateAction(scanner.Bytes(),
--- a/pkg/es/writer/esBulkHandler.go
+++ b/pkg/es/writer/esBulkHandler.go
@@ -141,10 +141,6 @@ func HandleBulkBody(postBody []byte, ctx *fasthttp.RequestCtx, rid uint64, myid 
 	origItems := items
 	defer respItemsPool.Put(&origItems)
 
-	// kunal todo check , if items gets extended and we put back origitems then does the os
-	// delete the old items array ?
-	itemsLen := len(items)
-
 	atleastOneSuccess := false
 	localIndexMap := make(map[string]string)
 
@@ -158,10 +154,9 @@ func HandleBulkBody(postBody []byte, ctx *fasthttp.RequestCtx, rid uint64, myid 
 
 	for scanner.Scan() {
 		inCount++
-		if inCount >= itemsLen {
+		if inCount >= len(items) {
 			newArr := make([]interface{}, 100)
 			items = append(items, newArr...)
-			itemsLen += 100
 		}
 
 		esAction, indexName, idVal := extractIndexAndValidateAction(scanner.Bytes(),
--- a/pkg/segment/utils/segconsts.go
+++ b/pkg/segment/utils/segconsts.go
@@ -66,7 +66,7 @@ const WIP_SIZE = 2_000_000
 const PQMR_SIZE uint = 4000 // init size of pqs bitset
 const WIP_NUM_RECS = 4000
 const BLOOM_SIZE_HISTORY = 5 // number of entries to analyze to get next block's bloom size
-const BLOCK_BLOOM_SIZE = 200 // the default should be on the smaller side. Let dynamic bloom sizing fix the optimal one
+const BLOCK_BLOOM_SIZE = 10
 const BLOCK_RI_MAP_SIZE = 100
 
 var MAX_BYTES_METRICS_BLOCK uint64 = 1e+8         // 100MB
--- a/pkg/segment/writer/segwriter.go
+++ b/pkg/segment/writer/segwriter.go
@@ -1312,9 +1312,7 @@ func addToBlockBloomBothCasesWithBuf(blockBloom *bloom.BloomFilter, fullWord []b
 	hasUpper := utils.HasUpper(copy)
 
 	// add the original full
-	if !blockBloom.TestAndAdd(copy) {
-		blockWordCount += 1
-	}
+	_ = blockBloom.Add(copy)
 
 	var hasSubWords bool
 	for {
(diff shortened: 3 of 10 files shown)


pkg/segment/writer/segwriter.go
```diff
@@ -561,37 +544,69 @@ func getFinalBaseSegDirFromActive(activeBaseSegDir string) (string, error) {
...
+			// if subwords, then add the original full's lowercase
+			// but have to use the original var, since the copy is pointing to last subword
+			if !blockBloom.TestAndAdd(utils.BytesToLowerInPlace(fullWord[:])) {
+				blockWordCount += 1
+			}
+		} else {
+			if !blockBloom.TestAndAdd(utils.BytesToLowerInPlace(copy)) {
+				blockWordCount += 1
+			}
+		}
```
AndrewHess, line 607: nit: can't the `hasSubwords` and `!hasSubwords` cases be handled the same way (using `fullWord`)?

pkg/segment/writer/segwriter.go
```diff
@@ -1174,19 +1187,159 @@ func DeletePQSData() error {
...
+	// we add twice to avoid undersizing for above reason.
+	bi.Bf = bloom.NewWithEstimates(uint(cw.deData.deCount), BLOOM_COLL_PROBABILITY)
+	for dwordkey := range cw.deData.deMap {
+		dword := []byte(dwordkey)
+		switch dword[0] {
+		case VALTYPE_ENC_BACKFILL[0]:
+			// we don't add backfill value to bloom since we are not going to search for it
+			continue
+		case VALTYPE_ENC_SMALL_STRING[0]:
+			// the first 3 bytes are TL(type len)
```
AndrewHess, line 1228: nit: 
```suggestion
			// the first 3 bytes are the type and length
```

Also, it would be good to add the same comment for line 1267

pkg/segment/writer/segwriter.go
```diff
@@ -1174,19 +1187,159 @@ func DeletePQSData() error {
...
 
-	dci := &DwordCbufIdxs{
-		sIdx:     sIdx,
-		wlen:     wlen,
-		recBsIdx: recBsIdx,
+func (cw *ColWip) writeDeBloom(buf []byte, bi *BloomIndex) error {
+	// todo a better way to size the bloom might be to count the num of space and
+	// then add to the cw.deData.deCount, that should be the optimal size
+	// we add twice to avoid undersizing for above reason.
+	bi.Bf = bloom.NewWithEstimates(uint(cw.deData.deCount), BLOOM_COLL_PROBABILITY)
```
AndrewHess, line 1220: I'm not seeing where we "add twice." Where is this happening?

pkg/segment/writer/segwriter.go
AndrewHess, line 1: `addToBlockBloomBothCases()` and `addToBlockBloomBothCasesWithBuf()` are very similar, and I think if we ever need to change one of them we'll have to change the other as well. Is it feasible to remove this duplication by having `addToBlockBloomBothCases()` call `addToBlockBloomBothCasesWithBuf()`? If not, it would be good to at least put these two function next to each other.

pkg/segment/writer/segwriter.go
```diff
@@ -561,37 +561,59 @@
...
+		if !blockBloom.TestAndAdd(utils.BytesToLowerInPlace(copy)) {
+			blockWordCount += 1
+		}
 	}
+
+	// if there is no subword, then we have not added the lower case, add it
+	if !hasSubWords && hasUpper {
+		if !blockBloom.TestAndAdd(utils.BytesToLowerInPlace(copy)) {
+			blockWordCount += 1
+		}
```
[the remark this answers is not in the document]
AndrewHess: I agree with Lindy that some simple unit tests for this function would be good

AndrewHess (commented):
I don't fully understand why you reverted a recent change you made (storing the records for a dict encoding as a bitset instead of as a list of record numbers). I believe you made that change to reduce heap allocations, but can you explain why you undid that in this PR and moved back to a list of record numbers?

pkg/segment/writer/segwriter.go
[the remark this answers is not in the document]
nkunal: filed an GH issue for this

pkg/es/writer/esBulkHandler.go
```diff
@@ -161,7 +161,7 @@ func HandleBulkBody(postBody []byte, ctx *fasthttp.RequestCtx, rid uint64, myid
 		if inCount >= itemsLen {
 			newArr := make([]interface{}, 100)
 			items = append(items, newArr...)
-			itemsLen += 1000
+			itemsLen += 100
 		}
```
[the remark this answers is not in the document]
nkunal: done

pkg/segment/utils/segconsts.go
```diff
@@ -66,7 +66,7 @@ const WIP_SIZE = 2_000_000
 const PQMR_SIZE uint = 4000 // init size of pqs bitset
 const WIP_NUM_RECS = 4000
 const BLOOM_SIZE_HISTORY = 5 // number of entries to analyze to get next block's bloom size
-const BLOCK_BLOOM_SIZE = 200 // the default should be on the smaller side. Let dynamic bloom sizing fix the optimal one
+const BLOCK_BLOOM_SIZE = 10
```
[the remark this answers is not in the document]
nkunal: it is there just as placeholder, we do bloom creation at the end of wip now where we now the size to allocate for bloom. This var is here just for some value to initialize the bloom filter with, since there is some code that still relies on bi.BF != nil. Eventually we will remove the "double loading" of the bloom variable name

pkg/segment/writer/segwriter.go
nkunal: yup, good point. Fixed it

pkg/segment/writer/segwriter.go
nkunal: good catch, I was doing it for NonDeBloom, but forgot to do it for DeBloom. Fixed it

pkg/segment/writer/segwriter.go
nkunal: yup good point, fixed it

AndrewHess (approved):
Looks good

PR #1578, opened 2024-09-01
Filter benchmarking

vasukalariya:
# Description
Summarize the change.
Experimenting alternatives to bloom.
For this benchmark test:
Number of random strings: 10M
Length of each random string: 100

Bloom:
- Takes 17 MB for filter + 17 MB for writing to buffer
- Easy to use

[Cuckoo Linvon](https://github.com/linvon/cuckoo-filter):
- Takes 14 MB for filter + 34.5 MB for Add method (implementation) + 14 MB for encode/writing to file
- More config options and is flexible but has inefficient implementation.

[Cuckoo Seiflotfy](https://github.com/seiflotfy/cuckoofilter)
- Takes 16 MB for filter + 16 MB for encoding/writing to file
- Has higher false positive rate and is not configurable at all.

[Cuckoo Panmari](https://github.com/panmari/cuckoofilter)
- It is a variant of the Seiflotfy impl but with better false positive rate (0.0001 vs 0.03).
- Takes 32 MB for filter and 57 MB for encode/writing to file.
- Not configurable at all.

[Xor Filter](https://github.com/FastFilter/xorfilter)
- Although this is supposed to be efficient, the implementation does not seem to be efficient for production.
- This filter is immutable meaning it has to be created once with all the keys which consumes too much memory.
- Takes 540 MB for filter population and 12 MB for encode/writing to file.
- Needs uint64 hashes instead of bytes as input.

[Binary Fuse Filter](https://github.com/FastFilter/xorfilter)
- It is better in memory consumption than xor while populating but still consumes too much memory.
- Takes 236 MB for filter population and 11 MB for encode/writing to file.
- Needs uint64 hashes instead of bytes as input.

```
// Function to deserialize binaryfuse8 filter from a file
func deserializeBinaryFuse8Filter(filename string) (*xorfilter.BinaryFuse8, error) {
    file, err := os.Open(filename)
    if err != nil {
        return nil, err
    }
    defer func() {
		file.Close()
		os.Remove(file.Name())
	}()

    var filter xorfilter.BinaryFuse8
    decoder := gob.NewDecoder(file)
    err = decoder.Decode(&filter)
    return &filter, err
}

// Function to deserialize xor8 filter from a file
func deserializeXor8Filter(filename string) (*xorfilter.Xor8, error) {
    file, err := os.Open(filename)
    if err != nil {
        return nil, err
    }
    defer func() {
		file.Close()
		os.Remove(file.Name())
	}()

    var filter xorfilter.Xor8
    decoder := gob.NewDecoder(file)
    err = decoder.Decode(&filter)
    return &filter, err
}
```

# Testing

# Checklist:


benchmark_test.go
```diff
@@ -609,3 +619,217 @@ func Benchmark_S3_segupload(b *testing.B) {
...
+	if err != nil {
+		panic(err)
+	}
+}
+
+// // Function to deserialize binaryfuse8 filter from a file
+// func deserializeBinaryFuse8Filter(filename string) (*xorfilter.BinaryFuse8, error) {
+//     file, err := os.Open(filename)
+//     if err != nil {
+//         return nil, err
```
AndrewHess, line 764: nit: Don't add commented-out code. Delete it, or uncomment it

benchmark_test.go
```diff
@@ -609,3 +619,217 @@ func Benchmark_S3_segupload(b *testing.B) {
...
+func benchmarkBloom(strs [][]byte) {
+	numItems := uint(len(strs)) // num of items the filter will store
+	// false positive rate (BLOOM_COLL_PROBABILITY): 0.001
+	bloom := bloom.NewWithEstimates(numItems, utils.BLOOM_COLL_PROBABILITY)
+
+	for _, str := range strs {
+		bloom.Add(str)
+	}
+
+	fname := "bloomfilter.bin"
```
AndrewHess, line 632: Do we want to delete this file after benchmarking?

Same question for other files we write to

benchmark_test.go
```diff
@@ -609,3 +619,217 @@ func Benchmark_S3_segupload(b *testing.B) {
...
+	benchmarkXorFilter(hashedKeys)
+
+	/*
+	   go test -run=Bench -bench=Benchmark_Filters  -cpuprofile cpuprofile.out -o rawsearch_cpu
+	   go tool pprof ./rawsearch_cpu
+
+	   (for mem profile)
+	   go test -run=Bench -bench=Benchmark_Filters -benchmem -memprofile memprofile.out -o rawsearch_mem
+	   go tool pprof ./rawsearch_mem memprofile.out
+	*/
```
AndrewHess, line 834: nit: move this comment near the top of the function

AndrewHess (commented):
Looks good; just some nits

benchmark_test.go
vasukalariya: We cannot uncomment it since it's not used and gives lint error so have deleted it. Have copied it in the PR description and referred the PR link for reference.

benchmark_test.go
vasukalariya: addressed in the latest commit.

benchmark_test.go
vasukalariya: addressed in the latest commit.

AndrewHess (approved):
Looks good

PR #1584, opened 2024-09-03
Check and Migrate Siglens DB

vasukalariya:
# Description
Summarize the change.
Migrate the siglens DB from working directory to data folder.

# Testing
After this change start the siglens server noticed the migration of DB from working directory to the data directory.
Checked that there are no error logs specifying Failed to connect to siglens database.

# Checklist:

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -7,7 +7,7 @@ services:
       - "${UI_PORT}:5122"
     volumes:
       - "${WORK_DIR}/data:/siglens/data"
-      - "${WORK_DIR}/siglens.db:/siglens/siglens.db"
+      - "${WORK_DIR}/data/siglens.db:/siglens/data/siglens.db"
       - "${WORK_DIR}/logs:/siglens/logs"
       - "${WORK_DIR}/${CONFIG_FILE}:/siglens/${CONFIG_FILE}"
     command: ["./siglens", "--config", "${CONFIG_FILE}"]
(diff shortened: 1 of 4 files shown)


nkunal (approved):

docker-compose.yml
```diff
@@ -7,7 +7,7 @@ services:
       - "${UI_PORT}:5122"
     volumes:
       - "${WORK_DIR}/data:/siglens/data"
-      - "${WORK_DIR}/siglens.db:/siglens/siglens.db"
+      - "${WORK_DIR}/data/siglens.db:/siglens/data/siglens.db"
```
Macbeth98, line 10: nit: I don't think we need to mount the `siglens.db` volume separately like this now. Since it is already inside `data` folder.

install.sh
```diff
@@ -718,11 +718,11 @@ chmod a+rwx data || {
...
-touch siglens.db || {
+touch data/siglens.db || {
     post_event "install_failed" "Failed to create file 'siglens.db'."
     print_error_and_exit "Failed to create file 'siglens.db'. Please check your permissions."
 }
-chmod a+rwx siglens.db || {
+chmod a+rwx data/siglens.db || {
     post_event "install_failed" "Failed to change permissions for file 'siglens.db'."
     print_error_and_exit "Failed to change permissions for file 'siglens.db'. Please check your file permissions."
 }
```
Macbeth98, line 728: nit: This part of the code is also not required. Since we are doing this for the entire `data` folder.

Macbeth98 (commented):
@AndrewHess @nkunal Once check my comments. May be we should do this update before next release? Or do we keep in this current way only?

PR #1585, opened 2024-09-03
Bug Test for Number in STB

vasukalariya:
# Description
Summarize the change.
We are copying pointers instead of values [here](https://github.com/siglens/siglens/blob/8c00ecd0c87e1d1daf389b61501ba708defba3d8/pkg/segment/writer/agiletree.go#L305).

# Testing

# Checklist:

### DIFF
--- a/pkg/segment/writer/startree_test.go
+++ b/pkg/segment/writer/startree_test.go
@@ -635,3 +635,103 @@ func TestStarTreeMediumEncodingDecoding(t *testing.T) {
 	fName = fmt.Sprintf("%v.strm", ss.SegmentKey)
 	_ = os.RemoveAll(fName)
 }
+
+// func Test_Aggregate(t *testing.T) {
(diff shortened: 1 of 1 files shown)


AndrewHess (commented):
What is this actually testing? It looks like it only tests that the tree can be built and encoded; it seems that the actual data ingested and the measure columns and group-by columns are arbitrary since the actual values from the tree are not read and asserted. Is that intended? Did this unit test fail before https://github.com/siglens/siglens/pull/1503 got merged?

vasukalariya:
> What is this actually testing? It looks like it only tests that the tree can be built and encoded; it seems that the actual data ingested and the measure columns and group-by columns are arbitrary since the actual values from the tree are not read and asserted. Is that intended? Did this unit test fail before #1503 got merged?

Fixed in this [PR](https://github.com/siglens/siglens/pull/1503)

PR #1595, opened 2024-09-04
Use PLEs to improve ingest performance

nkunal:
# Description
This PR rearranges how we do ingestion to get improved ingestion performance. We do this in two main ways:
1. Do more parsing without holding a lock, for more parallelization when ingesting into one index
2. Batch ingestion so when we get the segment lock, we insert more than one record. This helps avoid thrashing.

This adds a ParsedEventLog type (PLE) for parsing data in an ingest request. It mostly does this in a zero-copy procedure by storing pointers to the relevant parts of the ingest request.

This PR only implements the new ingestion method for the ES bulk handler. Other handlers will be updated in a different PR.

# Testing
## Correctness
Existing unit tests and CICD

## Performance
Running this ingestion:
```
go run main.go ingest esbulk -n 1 -g dynamic-user -d http://localhost:8081/elastic -p 23 -t 10_000_000 -i bidx -b 1000
```
On my local machine, it ran at 64661 events per second before this PR. With this PR I get 113900 events per second (about 120 MB/s)—a 76% increase in throughput.

# Checklist:


pkg/es/writer/esBulkHandler.go
```diff
@@ -196,22 +229,30 @@ func HandleBulkBody(postBody []byte, ctx *fasthttp.RequestCtx, rid uint64, myid
 						}
 					}
 				} else {
-					err := ProcessIndexRequest(rawJson, tsNow, indexName, uint64(numBytes),
-						false, localIndexMap, myid, rid, idxToStreamIdCache,
-						cnameCacheByteHashToStr, jsParsingStackbuf[:])
+					ple := plePool.Get().(*writer.ParsedLogEvent)
+					defer plePool.Put(ple)
```
nkunal, line 233: not 100% sure about this (need to read the language) spec: 
if we put all the ple one by one for defer, it will go on the stack and the stack size might grow.
so do we need to put these in some array and then have just one defer call before the for loop, that would walk through the array and do the `Puts` 
(btw this does not need to be addressed in this PR, just a thought)

pkg/es/writer/esBulkHandler.go
```diff
@@ -196,22 +229,30 @@ func HandleBulkBody(postBody []byte, ctx *fasthttp.RequestCtx, rid uint64, myid
...
+					allPLEs = append(allPLEs, ple)
```
nkunal, line 234: ah, we already have all the PLEs here, so it could work with above comment

pkg/es/writer/esBulkHandler.go
```diff
@@ -239,6 +280,21 @@ func HandleBulkBody(postBody []byte, ctx *fasthttp.RequestCtx, rid uint64, myid
 			items[inCount-1] = resp_status_201
 		}
 	}
+
+	pleBatches := utils.ConvertSliceToMap(allPLEs, func(ple *writer.ParsedLogEvent) string {
+		return ple.GetIndexName()
+	})
```
nkunal, line 286: sweet !

pkg/es/writer/esBulkHandler.go
```diff
@@ -330,21 +386,37 @@ func ProcessIndexRequest(rawJson []byte, tsNow uint64, indexNameIn string,
...
 	cnameCacheByteHashToStr map[uint64]string, jsParsingStackbuf []byte) error {
+	return nil
+}
+
+func ProcessIndexRequestPle(tsNow uint64, indexNameIn string,
+	bytesReceived uint64, flush bool, localIndexMap map[string]string, myid uint64,
+	rid uint64, idxToStreamIdCache map[string]string,
+	cnameCacheByteHashToStr map[uint64]string, jsParsingStackbuf []byte, pleArray []*writer.ParsedLogEvent) error {
+
+	for _, ple := range pleArray {
```
nkunal, line 397: technically the `indexNameIn` and the ple.GetIndexName()` should be exactly same , because when we did the batching, it was grouped based on the indexname.
Or is this here just for a safety check ?

pkg/es/writer/esBulkHandler.go
```diff
@@ -330,21 +386,37 @@ func ProcessIndexRequest(rawJson []byte, tsNow uint64, indexNameIn string,
...
-		cfgkey = "startTimeMillis"
+		tsKey = "startTimeMillis"
 	} else {
 		docType = segment.SIGNAL_EVENTS
 	}
 
-	ts_millis := utils.ExtractTimeStamp(rawJson, &cfgkey)
-	if ts_millis == 0 {
-		ts_millis = tsNow
+	for _, ple := range pleArray {
```
nkunal, line 415: nit: do we need to do array walking here ? because later in the flow, we have to walk the array anyways, so why not do it just once .
(this comment does not need to be handled in this PR, just adding here for later reference)

pkg/segment/writer/segwriter.go
```diff
@@ -139,6 +139,62 @@ type WipBlock struct {
...
+	ple.allCnames = ple.allCnames[:0]
+	ple.allCvals = ple.allCvals[:0]
+	ple.allCvalsTypeLen = ple.allCvalsTypeLen[:0]
+	ple.numCols = 0
+}
+
+func (ple *ParsedLogEvent) MakeSpaceForNewColumn() {
+	ple.allCnames = append(ple.allCnames, "")
+	ple.allCvals = append(ple.allCvals, nil)
+	ple.allCvalsTypeLen = append(ple.allCvalsTypeLen, [9]byte{})
```
nkunal, line 171: nit: lets add a comment here why the 9 bytes

pkg/segment/writer/segwriter.go
```diff
@@ -247,63 +303,186 @@ func cleanRecentlyRotatedInfo() {
...
+
+	tsKey := config.GetTimeStampKey()
 
 	segstore.Lock.Lock()
 	defer segstore.Lock.Unlock()
 
-	if segstore.wipBlock.maxIdx+MAX_RECORD_SIZE >= WIP_SIZE ||
-		segstore.wipBlock.blockSummary.RecCount >= MAX_RECS_PER_WIP {
-		err := segstore.AppendWipToSegfile(streamid, false, false, false)
+	// use the same PLE N times to simulate
```
nkunal, line 438: this comment can be removed

pkg/segment/writer/segwriter.go
```diff
@@ -247,63 +303,186 @@ func cleanRecentlyRotatedInfo() {
...
+				log.Errorf("SegStore.AddEntry: failed to append segkey=%v, err=%v", segstore.SegmentKey, err)
+				return err
+			}
+			instrumentation.IncrementInt64Counter(instrumentation.WIP_BUFFER_FLUSH_COUNT, 1)
+		}
+
+		matchedPCols, err := segstore.doLogEventFilling(ple, &tsKey)
 		if err != nil {
-			log.Errorf("SegStore.AddEntry: failed to append segkey=%v, err=%v", segstore.SegmentKey, err)
+			log.Errorf("AddEntry: failed to do parsed even filling, segkey: %v, err: %v", segstore.SegmentKey, err)
```
nkunal, line 453: nit: typo in the log message

nkunal (commented):
This is looking pretty cool now !

pkg/segment/writer/segwriter.go
AndrewHess: Yeah I updated your comment in the type definition:
```
type ParsedLogEvent struct {
	...
	allCvalsTypeLen [][9]byte // array of all column values type and len (3 bytes for strings; 9 for numbers)
	...
}
```

Do you want me to be more clear about why numbers need 9 bytes?

pkg/es/writer/esBulkHandler.go
AndrewHess: It was just an extra safety check in case we make more calls to this function later. Do you think I should remove it?

pkg/es/writer/esBulkHandler.go
AndrewHess: Good point; I think we will want to do that

pkg/es/writer/esBulkHandler.go
AndrewHess: Ok, let's do that in a separate PR

pkg/es/writer/esBulkHandler.go
```diff
@@ -239,6 +288,21 @@ func HandleBulkBody(postBody []byte, ctx *fasthttp.RequestCtx, rid uint64, myid
 			items[inCount-1] = resp_status_201
 		}
 	}
+
+	pleBatches := utils.ConvertSliceToMap(allPLEs, func(ple *writer.ParsedLogEvent) string {
+		return ple.GetIndexName()
+	})
+
+	for indexName, plesInBatch := range pleBatches {
+		err = ProcessIndexRequestPle(tsNow, indexName, uint64(len(line)),
```
nkunal, line 297: we need to remove the `len(line)` param from this func and figure out the `len(line)` inside the function for each individual PLE.

pkg/es/writer/esBulkHandler.go
```diff
@@ -359,10 +439,10 @@ func ProcessIndexRequest(rawJson []byte, tsNow uint64, indexNameIn string,
 	// json-rsponse formation during query-resp. We should either add it in this AddEntryToInMemBuf
 	// OR in json-resp creation we add it in the resp using the vtable name
 
-	err := writer.AddEntryToInMemBuf(streamid, rawJson, ts_millis, indexNameConverted, bytesReceived, flush,
-		docType, myid, rid, cnameCacheByteHashToStr, jsParsingStackbuf)
+	err := writer.AddEntryToInMemBuf(streamid, indexNameConverted, bytesReceived, flush,
```
nkunal, line 442: same comment here about the `bytesReceived`

pkg/segment/writer/segwriter.go
```diff
@@ -250,63 +306,185 @@ func cleanRecentlyRotatedInfo() {
...
+
+	tsKey := config.GetTimeStampKey()
 
 	segstore.Lock.Lock()
 	defer segstore.Lock.Unlock()
 
-	if segstore.wipBlock.maxIdx+MAX_RECORD_SIZE >= WIP_SIZE ||
-		segstore.wipBlock.blockSummary.RecCount >= MAX_RECS_PER_WIP {
-		err := segstore.AppendWipToSegfile(streamid, false, false, false)
+	for _, ple := range pleArray {
```
nkunal, line 441: we should be taking the bytesReceived from each of the ple's len(rawjson) here add that up and then use it in line 473.

AndrewHess (dismissed):

AndrewHess (approved):

PR #1597, opened 2024-09-05
Alert Unsilencing

sonamgupta21:
# Description

1. Added an API endpoint `api/alerts/unsilenceAlert` that resets the silenceEndTime and silenceMinutes to 0.
2. Included `silenceEndTime` in the alert details to display the "muted for minutes" information on the UI, allowing users to know how long the alert will remain muted.
3. Integrated the unsilence API with the frontend and added an unmute icon.

image

# Testing

# Checklist:


pkg/alerts/alertsHandler/alertsHandler.go
```diff
@@ -221,6 +223,50 @@
 	utils.WriteJsonResponse(ctx, responseBody)
 }
 
+func ProcessUnsilenceAlertRequest(ctx *fasthttp.RequestCtx) {
+	responseBody := make(map[string]interface{})
```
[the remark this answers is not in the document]
Macbeth98: nit: I agree with this. If you could refactor the code, that would be great. You could probably use the same struct as SilenceRequest to unmarshal json input request. The `silenceMinutes` and `silenceEndTime` will be `zero` for `UnsilenceRequest`. And you could pass the type  (Silence or UnSilence) to the internal refactored function and based on the type you could check whether the `silenceMinutes` and `silenceEndTime` should be greater than `zero` or not.

Macbeth98 (commented):
Looks good!

pkg/alerts/alertsHandler/alertsHandler.go
```diff
@@ -193,34 +193,59 @@ func ProcessSilenceAlertRequest(ctx *fasthttp.RequestCtx) {
...
+		if request.SilenceMinutes == 0 {
+			utils.SendError(ctx, "SilenceMinutes must be greater than zero", "", nil)
+			return
+		}
+		if request.SilenceEndTime == 0 {
+			utils.SendError(ctx, "SilenceEndTime must be greater than zero", "", nil)
+			return
+		}
+		alertDataObj.SilenceMinutes = request.SilenceMinutes
+		alertDataObj.SilenceEndTime = request.SilenceEndTime
```
AndrewHess, line 220: A few things:
1. Why do both `SilenceMinutes` and `SilenceEndtime` need to be set? Aren't these redundant pieces of information?
2. What happens if `SilenceEndtime != now() + SilenceMinutes`?

pkg/alerts/alertsHandler/alertsHandler.go
```diff
@@ -193,34 +193,59 @@ func ProcessSilenceAlertRequest(ctx *fasthttp.RequestCtx) {
...
+	} else if mode == alertutils.UnsilenceMode {
+		alertDataObj.SilenceMinutes = 0
+		alertDataObj.SilenceEndTime = 0
+	}
+
```
AndrewHess, line 225: nit: have an `else` here to log an error if we get to an unexpected state. Or perhaps change `mode` from a string to a boolean since there's only two options

static/js/all-alerts.js
```diff
@@ -365,10 +450,26 @@ let alertColumnDefs = [
         width: 100,
         cellRenderer: stateCellRenderer,
     },
+    {
+        headerName: 'Muted For',
```
sunitakawane, line 454: should this column name be changed to `Silenced for` since everywhere else in the UI we are using the word  silence instead of mute

sunitakawane:
Are we not adding mute/unmute events as part of alert history

pkg/alerts/alertsHandler/alertsHandler.go
sonamgupta21: 1. Previously, we only stored the silence duration with `SilenceMinutes`, but it wasn’t easy to calculate how much time was left for the alert to become active again. So, I added `SilenceEndTime` to directly calculate the exact time when the silence ends on the frontend and display it in the UI.

2. I moved the calculation of `SilenceEndTime` to the backend to avoid inconsistencies between `SilenceEndTime` and `now() + SilenceMinutes`.

sunitakawane (approved):

PR #1600, opened 2024-09-05
Fix stats count for MatchAllQuery

vasukalariya:
# Description
Summarize the change.
In the case of matchAllQuery like `* | stats count` we were just picking up a random column from the sstMap and then using that count. If records have NULL values, the count would vary based on the column picked (Note: Map iteration order is not deterministic).
Find the total records in the segment from microIndices and then pass that on to the SegmentQueryRequest. Use this value to get the count.

Showing the message explicitly because, often it is confusing for people to apply operators with `*` which is only supported for count.

`request_id=* | stats count` was giving correct results because we are [populating](https://github.com/siglens/siglens/blob/523968fba532486e00054b05e924757b25cdb7db/pkg/segment/search/searchaggs.go#L958) sstMap for `*` because for raw search for `count` with `*`.

After this fix `* | stats count` should also return same result as above.

# Testing
Insert dummy data with NULL values in the column and then query.
Moreover, tested by ingesting esbulk across multiple segments.

# Checklist:


AndrewHess (approved):
Looks good

PR #1601, opened 2024-09-06
Segment WIP Block Flush max Time Interval

Macbeth98:
# Description
- Adds a new config variable called `MaxWaitWipFlushIntervalSecs`, that is used to set the max interval before a WIP block force flushed due to time.
- Segstore has a new field `lastWipFlushTime` that updates the time whenever a WIP block is flushed/Appended to the Seg File. 
- Refactored the `SegFLushIntervalSecs` to `IdleWipFlushIntervalSecs` to reflect the allow Idle time before flushing.

# Testing
- All unit test cases passed.
- And there were no issues found when running the test while writing small writes every 1 to 3 seconds and validating whether the log is present or not.

# Checklist:


pkg/config/common/common.go
```diff
@@ -88,14 +88,15 @@ assignment in the following functions
...
-	SegFlushIntervalSecs       int      `yaml:"segFlushIntervalSecs"` // Time Interval after which to write to segfile
+	IngestListenIP             string   `yaml:"ingestListenIP"`             // Listen IP used for ingestion server
+	QueryListenIP              string   `yaml:"queryListenIP"`              // Listen IP used for query server
+	IngestPort                 uint64   `yaml:"ingestPort"`                 // Port for ingestion server
+	QueryPort                  uint64   `yaml:"queryPort"`                  // Port used for query server
+	EventTypeKeywords          []string `yaml:"eventTypeKeywords"`          //Required event type keyword
+	QueryNode                  string   `yaml:"queryNode"`                  //Node to enable/disable all query endpoints
+	IngestNode                 string   `yaml:"ingestNode"`                 //Node to enable/disable all ingest endpoints
+	SegFlushIntervalSecs       int      `yaml:"segFlushIntervalSecs"`       // Time Interval after which to write to segfile
+	SegWipFlushMaxIntervalSecs int      `yaml:"segWipFlushMaxIntervalSecs"` // Max time interval after which to write to segfile
```
nkunal, line 99: nit: name this as `wipFlush...` and all the get/set for this function accordingly

pkg/config/config.go
```diff
@@ -285,6 +285,14 @@ func GetSegFlushIntervalSecs() int {
 	return runningConfig.SegFlushIntervalSecs
 }
 
+func GetSegWipFlushMaxIntervalSecs() int {
+	if runningConfig.SegWipFlushMaxIntervalSecs > 60 {
+		log.Errorf("GetSegWipFlushMaxIntervalSecs: SegWipFlushMaxIntervalSecs cannot be more than 1 min")
```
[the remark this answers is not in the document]
nkunal: returning this error and doing this value check during runtime is expensive. 
This check and setting of correct value should be done either on the `Set` call or when we read the config

pkg/config/config_test.go
```diff
@@ -132,6 +133,7 @@ func Test_ExtractConfigData(t *testing.T) {
  segreaderNode: true
  metareaderNode: true
  segFlushIntervalSecs: 1200
+ segWipFlushMaxIntervalSecs: 300
```
nkunal, line 136: if we are only allowing a value upto 60 then setting this to 300 does not make sense

pkg/segment/writer/segwriter.go
```diff
@@ -360,8 +360,9 @@ func updateValuesFromConfig() {
 func timeBasedWIPFlushToFile() {
 	for {
 		sleepDuration := time.Duration(config.GetSegFlushIntervalSecs()) * time.Second
```
nkunal, line 362: we should also rename the old segflush in lieu of introducing this new variable.

pkg/segment/writer/segwriter.go
```diff
@@ -360,8 +360,9 @@ func updateValuesFromConfig() {
...
+		maxFlushDuration := time.Duration(config.GetSegWipFlushMaxIntervalSecs()) * time.Second
 		time.Sleep(sleepDuration)
-		FlushWipBufferToFile(&sleepDuration)
+		FlushWipBufferToFile(&sleepDuration, &maxFlushDuration)
```
nkunal, line 365: do we want to overload the existing go routine to do both things ?
what happens if these two durations are different. We may miss out on the old one ?
we have two options:
1. create seperate go routines and each sleeps for it's own interval 
2. Keep one go routine and take the minimum of these two intervals and sleep for that duration
in both cases the `AppendWip..` call should still be called validating the timings

Macbeth98: Kunal, why would we miss out on the old one? I did not understand that.

Also I did not create a separate go routine, because that go routine also does the same exact functionality but instead of `lastUpdatedTime` it checks for `lastWipFlushTime`. Its like the min and max interval for a WIP flush block. So isn't better to do it in the same go routine?

pkg/config/config.go
```diff
@@ -340,13 +340,17 @@ func SetEventTypeKeywords(val []string) {
 	runningConfig.EventTypeKeywords = val
 }
 
-func SetSegFlushIntervalSecs(val int) {
+func SetIdleWipFlushIntervalSecs(val int) {
 	if val < 1 {
-		log.Errorf("SetSegFlushIntervalSecs: SegFlushIntervalSecs should not be less than 1s")
-		log.Infof("SetSegFlushIntervalSecs: Setting SegFlushIntervalSecs to 1 by default")
+		log.Errorf("SetIdleWipFlushIntervalSecs: IdleWipFlushIntervalSecs should not be less than 1s")
+		log.Infof("SetIdleWipFlushIntervalSecs: Setting IdleWipFlushIntervalSecs to 1 by default")
```
nkunal, line 346: I think we should set this default to `5` instead of `1`

pkg/config/config.go
```diff
@@ -340,13 +340,17 @@ func SetEventTypeKeywords(val []string) {
...
 		val = 1
 	}
-	runningConfig.SegFlushIntervalSecs = val
+	if val > 600 {
+		log.Warnf("SetIdleWipFlushIntervalSecs: IdleWipFlushIntervalSecs cannot be more than 10 mins. Defaulting to 10 mins")
```
nkunal, line 350: and the max should not be 600 but 60s

pkg/config/config_test.go
```diff
@@ -131,7 +132,8 @@ func Test_ExtractConfigData(t *testing.T) {
  seedNode: true
  segreaderNode: true
  metareaderNode: true
- segFlushIntervalSecs: 1200
+ idleWipFlushIntervalSecs: 1200
```
nkunal, line 135: I guess this will get rejected and will default to the value that we allow

pkg/config/config_test.go
```diff
@@ -131,7 +132,8 @@ func Test_ExtractConfigData(t *testing.T) {
...
+ maxWaitWipFlushIntervalSecs: 300
```
nkunal, line 136: same here

pkg/segment/writer/segwriter.go
Macbeth98: I have created two separate go routines, as we might miss out on `maxWaitWipFlushTime` if the intervals were close to each other.

pkg/segment/writer/segwriter.go
```diff
@@ -427,16 +438,32 @@ func timeBasedRotateSegment() {
...
+			continue
+		}
+
+		shouldFlush := false
+
+		if idleWipFlushDuration != nil && time.Since(segstore.lastUpdated) > *idleWipFlushDuration {
+			shouldFlush = true
+		}
+
+		if !shouldFlush && maxWaitWipFlushDuration != nil && time.Since(segstore.lastWipFlushTime) > *maxWaitWipFlushDuration {
```
nkunal, line 456: I think this function is handling both cases of the two pointers that are passed, should they do they are own checks and not worry about whether shouldFlush was set. basically :
`if idleWipFlushDuration != nil {
    check the time and set should flush accordingly
}
if  maxWaitWipFlushDuration != nil {
    check the time and set should flush accordingly
}
if shouldFlush {
 // do the flushing, it can come from any of those above two if checks, it does not matter because if 
// we are above the thresholds of each flush timer then we should flush it irrespective of the sleepDuration.
}
`

pkg/config/config_test.go
Macbeth98: Yes

pkg/config/config_test.go
nkunal: yes , a error log will be printed and we get set to default. Since this is just for testing , then we should set correct value (just a nit btw)

pkg/segment/writer/segwriter.go
nkunal: ah I see makes sense. you can keep it then. Sorry I didn't catch that in the first read

pkg/config/config_test.go
Macbeth98: yes we are testing this particular scenario where we will give more than allowed values and expect to see the correct allowed values.

nkunal (approved):

PR #1602, opened 2024-09-06
Index support for SPL

vasukalariya:
# Description
Summarize the change.
Added a index support for SPL, to override the index for searching.

# Testing
- Ingest multi index data
- Change the index on the UI and run a index specific query
- Verify the results

Added parser tests and tests in ingest.csv and restart.csv

# Checklist:

### DIFF
--- a/pkg/ast/pipesearch/searchQueryParser.go
+++ b/pkg/ast/pipesearch/searchQueryParser.go
@@ -39,26 +39,30 @@ import (
 	log "github.com/sirupsen/logrus"
 )
 
-func ParseRequest(searchText string, startEpoch, endEpoch uint64, qid uint64, queryLanguageType string, indexName string) (*ASTNode, *QueryAggregators, error) {
+func ParseRequest(searchText string, startEpoch, endEpoch uint64, qid uint64, queryLanguageType string, indexName string) (*ASTNode, *QueryAggregators, string, error) {
 	var err error
 	var queryAggs *QueryAggregators
 	var boolNode *ASTNode
-	boolNode, queryAggs, err = ParseQuery(searchText, qid, queryLanguageType)
+	var parsedIndexName string
+	boolNode, queryAggs, parsedIndexName, err = ParseQuery(searchText, qid, queryLanguageType)
 	if err != nil {
-		return nil, nil, err
+		return nil, nil, "", err
+	}
+	if parsedIndexName != "" {
+		indexName = parsedIndexName
 	}
 
 	if boolNode == nil && queryAggs == nil {
 		err := fmt.Errorf("qid=%d, ParseRequest: boolNode and queryAggs are nil for searchText: %v", qid, searchText)
 		log.Errorf(err.Error())
-		return nil, nil, err
+		return nil, nil, "", err
 	}
 
 	if boolNode.TimeRange == nil {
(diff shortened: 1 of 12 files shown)


pkg/ast/structs.go
```diff
@@ -28,6 +28,7 @@ import (
 type QueryStruct struct {
 	SearchFilter *Node
 	PipeCommands *structs.QueryAggregators
+	IndexName    string
```
nkunal, line 31: is this field used somewhere ?

nkunal (commented):
overall looks great

pkg/ast/pipesearch/searchQueryParser.go
```diff
@@ -150,13 +155,13 @@ func updatePositionForGenEvents(aggs *QueryAggregators) {
 	}
 }
 
-func parsePipeSearch(searchText string, queryLanguage string, qid uint64) (*ASTNode, *QueryAggregators, error) {
+func parsePipeSearch(searchText string, queryLanguage string, qid uint64) (*ASTNode, *QueryAggregators, string, error) {
```
AndrewHess, line 158: Maybe this `indexName` being returned should be a `[]string` so we can support `index1 OR index2`, but you can keep it as `string` if this is just for the simplest use cases

pkg/ast/spl/spl.peg
```diff
@@ -576,6 +580,33 @@ Start <- SPACE? initialSearch:(InitialSearchBlock) filterBlocks:(FilterBlock)* q
...
+    for _, v := range rest.([]any) {
+        parts := v.([]any)  // This will be [OR, IndexAssign].
+        index := parts[1].(string)
+        if index == "" {
+            continue
+        }
+        finalIndex = fmt.Sprintf("%s,%s", finalIndex, index)
+    }
+
+    return finalIndex, nil
```
AndrewHess, line 603: I think it will be a little easier to debug/understand if you return a list of strings here, instead of joining the strings with a comma

pkg/ast/spl/spl.peg
```diff
@@ -576,6 +580,33 @@ Start <- SPACE? initialSearch:(InitialSearchBlock) filterBlocks:(FilterBlock)* q
     return q, nil
 }
 
+IndexAssign <- index:("_index") EQUAL indexName:String {
+    return removeQuotes(indexName), nil
+}
+
+IndexExpression <- first:(IndexAssign) rest:(OR IndexAssign)* {
```
AndrewHess, line 587: nit: probably add a TODO here, since we want to allow other index selection later

pkg/ast/structs.go
vasukalariya: Yes it will capture all the indexes passed as comma separated (just like how we pass in the json ind-0, ind-1). this was done so that we can reuse the `InitTableInfo` just like how we do it for a curl request.

pkg/ast/pipesearch/searchQueryParser.go
vasukalariya: It could be but it was better to reuse `InitTableInfo` and make it synchronous with what we have in the curl request.

vasukalariya: addressed in the latest commit.

pkg/ast/spl/spl.peg

pkg/ast/spl/spl.peg

AndrewHess (approved):

PR #1606, opened 2024-09-06
Fix Regex setting in case of valueIsRegex

vasukalariya:
# Description
Summarize the change.
Prevent updating the regex if valueIsRegex is true.

# Testing
Added tests in ingest.csv and restart.csv

# Checklist:


nkunal (commented):
lgtm

pkg/ast/common.go
```diff
@@ -104,7 +104,7 @@ func ProcessSingleFilter(colName string, colValue interface{}, originalColValue
 						log.Errorf("qid=%d, ProcessSingleFilter: Failed to compile regex for %s. This may cause search failures. Err: %v", qid, t, err)
 						return nil, fmt.Errorf("invalid regex: %s", t)
 					}
-					criteria := CreateTermFilterCriteria(colName, compiledRegex, opr, qid, caseConversion)
+					criteria := CreateTermFilterCriteria(colName, compiledRegex, opr, qid, caseConversion, valueIsRegex)
```
Macbeth98, line 107: `valueIsRegex` is in `caseConversion` struct. You do not need to pass it as a param here.

Same for other places.

pkg/ast/common.go
```diff
@@ -258,19 +258,19 @@ func createMatchFilterCriteria(colName, colValue interface{}, opr LogicalOperato
...
 
-func CreateTermFilterCriteria(colName string, colValue interface{}, opr FilterOperator, qid uint64, cci *CaseConversionInfo) *FilterCriteria {
+func CreateTermFilterCriteria(colName string, colValue interface{}, opr FilterOperator, qid uint64, cci *CaseConversionInfo, valueIsRegex bool) *FilterCriteria {
 	cVal, err := CreateDtypeEnclosure(colValue, qid)
 	if err != nil {
 		log.Errorf("qid=%d, createTermFilterCriteria: error creating DtypeEnclosure for ColValue=%v. Error=%+v", qid, colValue, err)
 	} else {
-		if cci != nil {
+		if cci != nil && !valueIsRegex {
 			cVal.UpdateRegexp(cci.caseInsensitive)
```
Macbeth98, line 267: Move this inside the below if condition. We only need to do this at that place.

Macbeth98: `ShouldAlsoSearchWithOriginalCase` will ensure that if `valueIsRegex` set to true, it won't be considered as a case-insensitive search.

pkg/ast/common.go
vasukalariya: addressed in the latest commit.

pkg/ast/common.go
vasukalariya: addressed in the latest commit.

Macbeth98 (approved):

PR #1610, opened 2024-09-09
BugFix: fixes eval in stats after sort 

sathkrith:
# Description
This PR fixes an issue where the eval in stats does not work as expected.
The issue can be reproduced by 
app_name=Bracecould | sort latency | stats avg(eval(ssn/2)) as avg

It also refactors PerformMeasureAggsOnRecs to reuse existing functions.

# Testing
app_name=Bracecould | sort latency | stats avg(eval(ssn/2)) as avg 
should give the same result as 
app_name=Bracecould |  stats avg(eval(ssn/2)) as avg
# Checklist:


pkg/segment/structs/segstructs.go
```diff
@@ -1084,6 +1084,17 @@ func (qa *QueryAggregators) HasListFunc() bool {
 	return false
 }
 
+func HasNumTypeAggForColumn(measureOperations []*MeasureAggregator, column string) bool {
```
AndrewHess, line 1087: nit: You've defined this function in the middle of a list of `QueryAggregators` methods. I'd suggest moving this definition somewhere else, or maybe it makes sense to make this a `QueryAggregators` method.

pkg/segment/search/searchaggs.go
```diff
@@ -553,66 +553,61 @@ func PerformMeasureAggsOnRecs(nodeResult *structs.NodeResult, recs map[string]ma
...
 	for recInden, record := range recs {
 		sstMap := make(map[string]*structs.SegStats, 0)
-
-		for _, mOp := range nodeResult.MeasureOperations {
-			dtypeVal, err := utils.CreateDtypeEnclosure(record[mOp.MeasureCol], qid)
+		for column := range aggCols {
+			if column == "*" {
+				stats.AddSegStatsCount(sstMap, column, 1)
+				continue
+			}
```
AndrewHess, line 567: `column == "*"` only makes sense for the `Count` aggregation, so I think it would be good to double check that here and log an error if we have a different aggregation for `column == "*"`. Or is there a reason you're not doing that?

pkg/segment/reader/segread/segstatsreader_test.go
```diff
@@ -444,3 +444,74 @@ func TestGetAverage_InvalidType(t *testing.T) {
...
+	_, err := GetSegValue(runningSegStat, firstUpdate)
+	assert.NoError(t, err)
+
+	result, err := GetSegValue(runningSegStat, secondUpdate)
+	assert.NoError(t, err)
+
+	expected := []string{"initial", "second", "third"}
+	if !reflect.DeepEqual(result.CVal, expected) {
+		t.Errorf("TestGetSegValue_UpdateRunningStat: Expected running stat to be updated to %v, but got %v", expected, result.CVal)
+	}
```
AndrewHess, line 516: nit: I think doing `assert.Equal(t, expected, actual)` would be simpler because you'll get a decent error message by default if the assert fails. But you can keep it this way since you've already done it

pkg/segment/search/searchaggs_test.go
```diff
@@ -445,3 +445,78 @@ func Test_PerformMeasureAggsOnRecs_WithList(t *testing.T) {
...
+	}
+}
+
+func getDummyMeasureAggregator(field string) *structs.MeasureAggregator {
+	return &structs.MeasureAggregator{
+		MeasureCol:      "",
+		MeasureFunc:     utils.List,
+		StrEnc:          "list(eval(" + field + "/2))",
+		ValueColRequest: getDummyNumericValueExpr(field),
+	}
```
AndrewHess, line 492: nit: you're calling these functions `getDummy...()` but it seems like they're coupled with each other and also with `Test_PerformMeasureAggsOnRecs_WithEvalWithField()`, which assumes they do `field / 2`. So it would be good to rename them to indicate they do `field / 2`

AndrewHess (commented):
This is looking pretty good. However, I don't quite understand how the code changes fix this query you mentioned in the PR description:
```
app_name=Bracecould | sort latency | stats avg(eval(ssn/2)) as avg
```
Could you give me a high-level explanation?

sathkrith:
> This is looking pretty good. However, I don't quite understand how the code changes fix this query you mentioned in the PR description:
> 
> ```
> app_name=Bracecould | sort latency | stats avg(eval(ssn/2)) as avg
> ```
> 
> Could you give me a high-level explanation?


The eval aggs group of functions that compute evaluations use` structs.SegStats.Records` to evaluate value.
The issue with the previous implementation of `PerformMeasureAggsOnRecs `, which is used to update stats when output modifiers like sort, was that we were not updating `structs.SegStats.Records` when `ValueColRequest` was not nil.

I could have just checked for ValueColRequest and populated the `structs.SegStat` in `sstMap `with Records. However, we already have functions to update `SegStat `based on columns and aggregation types. So instead of updating in `PerformMeasureAggsOnRecs`, I re-used the existing functions to do the same.

A similar approach is used in `applySegStatsToMatchedRecords/segmentStatsWorker`. 

This does not solve all issues with `eval` in PerformMeasureAggsOnRecs when output modifiers are used.

pkg/segment/search/searchaggs.go
sathkrith: The check is done in updateSegmentStats, which PerformAggs calls to update the segment stats.

pkg/segment/search/searchaggs_test.go
sathkrith: updated.

pkg/segment/structs/segstructs.go
sathkrith: Moved the definition out of `QueryAggregators `block

AndrewHess:
This is failing our e2e tests with this error:
```
FATA[0012] RunQueryFromFile: Actual aggregate/record value: 69 is not [167 eq] for query city=Boston | eval nullField1=null() | eval nullField2=null() | eval nullStatus=nullif(http_status, 200) | fillnull value=FILLING nullField1 nullField2 | fields city, http_status, nullField1, nullField2, nullStatus | where nullStatus=null() | stats count(*)
```

Btw, our e2e tests are run like this https://github.com/siglens/siglens/blob/develop/.circleci/config.yml#L32-L43 and the one that's failing is the `query esbulk` after doing `ingest esbulk`

PR #1615, opened 2024-09-09
Check dataDiskThresholdPercent before ingest

sunitakawane:
# Description
Summarize the change.

stop ingest when used_disk_space > dataDiskThresholdPercent

# Testing

# Checklist:


pkg/server/ingest/entryHandlers.go
```diff
@@ -33,11 +37,49 @@ import (
...
+	if err != nil {
+		log.Errorf("getDiskUsagePercent: Error getting disk usage for the disk data path=%v, err=%v", config.GetDataPath(), err)
+		return 0, err
+	}
+	percentUsed := (s.Used * 100) / s.Total
+	return percentUsed, nil
+}
+
+func canIngest() bool {
+	return atomic.LoadInt32(&diskUsageExceeded) == 0
```
nkunal, line 75: there is already a bool version of this, so why do we use that instead of converting it on every ingest request: https://pkg.go.dev/sync/atomic#Bool.Store

pkg/server/ingest/entryHandlers.go
```diff
@@ -33,11 +37,49 @@ import (
...
+func MonitorDiskUsage() {
+	for {
+		usage, err := getDiskUsagePercent()
+		if err != nil {
+			log.Error("MonitorDiskUsage: Error getting disk usage:", err)
+			time.Sleep(30 * time.Second)
+			continue
+		}
+		if usage >= config.GetDataDiskThresholdPercent() {
+			log.Infof("MonitorDiskUsage: Disk usage exceeded the threshold of %v", config.GetDataDiskThresholdPercent())
```
nkunal, line 55: log this as error.
Also it would be helpful not only display the threshold but also the actual usage. Threshold anyways can be found from the config but the running usage value can be know only in this function

pkg/server/ingest/entryHandlers.go
```diff
@@ -77,7 +77,7 @@ func canIngest() bool {
 func esPostBulkHandler() func(ctx *fasthttp.RequestCtx) {
 	return func(ctx *fasthttp.RequestCtx) {
 		if !canIngest() {
-			utils.SendError(ctx, "Ingestion request rejected", "", nil)
+			utils.SendError(ctx, "Ingestion requested rejected due to no storage available", "", nil)
```
nkunal, line 80: typo in msg

nkunal (approved):

PR #1617, opened 2024-09-09
fix: Building and Pushing of SigClient image to DockerHub

Macbeth98:
# Description
- There is an issue with building of docker image for SigClient. We were copying the `go.mod` and `go.sum` files of SigLens instead of SigClient and then copying everything of SigLens and in the docker switching the worker directory to `tools/SigClient` and then we are building the image. 
- Even though the above one works we are needlessly copying the whole siglens and running `go mod` for SigLens.
- Also there is an issue with pushing of SigClient image, the existing [job fails at pushing of SigClient image](https://github.com/siglens/siglens/actions/runs/10744047913/job/29800103071#step:4:227) (not sure why it says the job is successful)
- So I have updated these two configs

# Testing
- I have tested that the docker is being built successfully, but I have not verified the Github Actions flow.

# Checklist:

### DIFF
--- a/.github/workflows/publish-prod-images.yml
+++ b/.github/workflows/publish-prod-images.yml
@@ -72,10 +72,10 @@ jobs:
           uses: docker/build-push-action@v3
           continue-on-error: true
           with:
-              context: .
-              file: tools/sigclient/Dockerfile
+              context: tools/sigclient
+              file: Dockerfile
               push: true
-              tags: sigclient/sigclient:latest
+              tags: siglens/sigclient:latest
   sigscalr-release-assets:
       runs-on: ubuntu-latest
       environment: build-environment
--- a/tools/sigclient/Dockerfile
+++ b/tools/sigclient/Dockerfile
@@ -5,7 +5,7 @@ COPY go.mod go.sum ./
 RUN go mod download
 
 COPY . .
-WORKDIR /usr/app/tools/sigclient
+WORKDIR /usr/app
 RUN go build -o sigclient main.go
 
 CMD ["./sigclient"]
--- a/tools/sigclient/Dockerfile
+++ b/tools/sigclient/Dockerfile
@@ -8,4 +8,4 @@ COPY . .
 WORKDIR /usr/app
 RUN go build -o sigclient main.go
 
-CMD ["./sigclient"]
+ENTRYPOINT ["./sigclient"]


.github/workflows/publish-prod-images.yml
```diff
@@ -72,10 +72,10 @@ jobs:
           uses: docker/build-push-action@v3
           continue-on-error: true
```
AndrewHess, line 73: > Also there is an issue with pushing of SigClient image, the existing [job fails at pushing of SigClient image](https://github.com/siglens/siglens/actions/runs/10744047913/job/29800103071#step:4:227) (not sure why it says the job is successful)

It says the job is successful because of this line:
```
continue-on-error: true
```

We probably want to keep this line, but maybe we could have another step/job that checks if any of the steps/jobs failed, and so we mark the whole workflow as failed if anything failed. Not sure if there's a better way to do this, and I'm not sure of the priority of doing this

.github/workflows/publish-prod-images.yml
```diff
@@ -72,10 +72,10 @@ jobs:
...
           with:
-              context: .
-              file: tools/sigclient/Dockerfile
+              context: tools/sigclient
+              file: Dockerfile
               push: true
-              tags: sigclient/sigclient:latest
+              tags: siglens/sigclient:latest
```
AndrewHess, line 78: I'm not sure we want this change. @nkunal what do you think?

tools/sigclient/Dockerfile
```diff
@@ -5,7 +5,7 @@ COPY go.mod go.sum ./
 RUN go mod download
 
 COPY . .
-WORKDIR /usr/app/tools/sigclient
+WORKDIR /usr/app
 RUN go build -o sigclient main.go
 
-CMD ["./sigclient"]
\ No newline at end of file
+ENTRYPOINT ["./sigclient"]
```
AndrewHess, line 11: I'm not too familiar with CMD vs ENTRYPOINT. Is this a best-practices change, or is something deeper going on here?

AndrewHess (commented):
Looks pretty good

.github/workflows/publish-prod-images.yml
Macbeth98: yeah actually the current one fails at build and push, cause it did not find the username `sigclient`. But if we want to keep `sigclient` separate from SigLens. Then probably we should create a new username.

tools/sigclient/Dockerfile
Macbeth98: Yes, so if I we keep this as `CMD`, then if we keep some command like `ingest esbulk ...` then those will override the `CMD` instead of appending to it. 

But `ENTRYPOINT` will always have `./sigclient` as the start command and whatever other arguments will be appended to it.

And since we have a lot of different variation of args, I thought this would be better.

AndrewHess: Cool, sounds good

PR #1627, opened 2024-09-11
Bump Version to 0.2.36d

Macbeth98:
# Description
- Bump version to 0.2.36d
- This also includes a change to SigClient-docker job in `publish-prod-images.yml` file. Pointing to the correct dockerfile.

### DIFF
--- a/.github/workflows/publish-prod-images.yml
+++ b/.github/workflows/publish-prod-images.yml
@@ -73,7 +73,7 @@ jobs:
           continue-on-error: true
           with:
               context: tools/sigclient
-              file: Dockerfile
+              file: tool/sigclient/Dockerfile
               push: true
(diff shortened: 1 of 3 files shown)


nkunal (approved):

.github/workflows/publish-prod-images.yml
```diff
@@ -73,7 +73,7 @@ jobs:
           continue-on-error: true
           with:
               context: tools/sigclient
-              file: Dockerfile
+              file: tool/sigclient/Dockerfile
```
AndrewHess, line 76: Why is this getting added back? I thought you just recently removed this?

Macbeth98: We need to give the docker file from the root path.

PR #1633, opened 2024-09-13
Remote Logs not showing up for API requests

vasukalariya:
# Description
Summarize the change.
Using the api requests, records from remoteLogs were not showing up in the UI.
Handled this edge case.

# Testing

# Checklist:

### DIFF
--- a/pkg/segment/query/querystatus.go
+++ b/pkg/segment/query/querystatus.go
@@ -558,20 +558,20 @@ func GetRemoteRawLogInfo(remoteID string, inrrcs []*utils.RecordResultContainer,
 	}
 	defer arqMapLock.RUnlock()
 
-	return rQuery.searchRes.GetRemoteInfo(remoteID, inrrcs)
+	return rQuery.searchRes.GetRemoteInfo(remoteID, inrrcs, false)
 }
 
-func GetAllRemoteLogs(inrrcs []*utils.RecordResultContainer, qid uint64) ([]map[string]interface{}, error) {
+func GetAllRemoteLogs(inrrcs []*utils.RecordResultContainer, qid uint64) ([]map[string]interface{}, []string, error) {
 	arqMapLock.RLock()
 	rQuery, ok := allRunningQueries[qid]
 	if !ok {
 		log.Errorf("GetAllRemoteLogs: qid %+v does not exist!", qid)
 		arqMapLock.RUnlock()
-		return nil, fmt.Errorf("qid does not exist")
+		return nil, nil, fmt.Errorf("qid does not exist")
 	}
 	defer arqMapLock.RUnlock()
 
-	return rQuery.searchRes.GetAllRemoteLogs(inrrcs), nil
+	return rQuery.searchRes.GetRemoteInfo("", inrrcs, true)
 }
 
 func round(num float64) int {
--- a/pkg/segment/query/querystatus.go
+++ b/pkg/segment/query/querystatus.go
@@ -536,13 +536,13 @@ func GetMeasureResultsForQid(qid uint64, pullGrpBucks bool, skenc uint16, limit 
 
 func GetQueryType(qid uint64) structs.QueryType {
 	arqMapLock.RLock()
+	defer arqMapLock.RUnlock()
+
 	rQuery, ok := allRunningQueries[qid]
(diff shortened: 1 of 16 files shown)


pkg/segment/results/segresults/segresults.go
```diff
@@ -498,6 +498,27 @@ func (sr *SearchResults) GetRemoteInfo(remoteID string, inrrcs []*utils.RecordRe
...
+	rawLogs := sr.remoteInfo.remoteLogs
+	count := 0
+	for i := 0; i < len(inrrcs); i++ {
+		if inrrcs[i].SegKeyInfo.IsRemote {
+			finalLogs = append(finalLogs, rawLogs[inrrcs[i].SegKeyInfo.RecordId])
+			count++
+		}
+	}
+
+	return finalLogs
```
Macbeth98, line 519: You are also supposed to return `remoteCols`.  And assign them in `searchHandler.go` at Line: 404.

pkg/segment/segexecution.go
```diff
@@ -439,6 +439,11 @@ func ExecuteQuery(root *structs.ASTNode, aggs *structs.QueryAggregators, qid uin
 		log.Errorf("qid=%d, ExecuteQuery: failed to get number of RRCs for qid! Error: %v", qid, err)
 	}
 
+	res.RemoteLogs, err = query.GetAllRemoteLogs(res.AllRecords, qid)
```
Macbeth98, line 442: This is fine. 
But, I am not sure whether we should fetch the `RawLogs` here itself, it does not exactly match with our current pattern, where we fetch the raw logs in `searchhandler` or `wsSearchHandler`.

Macbeth98 (commented):
I see that you have mentioned this as a rough Draft, so its fine for testing. But when are actually merging this, it might better to do this:
 
We could refactor: `GetRemotInfo` function so that we could use for both cases.

As of now, there is only difference I see which is the [`prefix` check happening here](https://github.com/siglens/siglens/pull/1633/files#diff-153fd9c9d5bb1a446f897390942735ad9010e6656c9fcd80494b41cd5ccc5252R477-R486).

So may be we could pass a flag that will say `fetchAllLogs`. And if that is set true we won't check this condition and add all raw logs. And you could send empty string for `remoteId`.

pkg/segment/segexecution.go
vasukalariya: This was done to reduce changes in the executeQuery since it is being used at quite a few places and we would need to get remote logs before query gets deleted.

pkg/segment/results/segresults/segresults.go
vasukalariya: marked as TODO

pkg/segment/query/querystatus.go
```diff
@@ -558,7 +558,20 @@ func GetRemoteRawLogInfo(remoteID string, inrrcs []*utils.RecordResultContainer,
...
+
+func GetAllRemoteLogs(inrrcs []*utils.RecordResultContainer, qid uint64) ([]map[string]interface{}, []string, error) {
+	arqMapLock.RLock()
+	rQuery, ok := allRunningQueries[qid]
+	if !ok {
+		log.Errorf("GetAllRemoteLogs: qid %+v does not exist!", qid)
+		arqMapLock.RUnlock()
+		return nil, nil, fmt.Errorf("qid does not exist")
+	}
+	defer arqMapLock.RUnlock()
```
AndrewHess, line 572: nit: your unlocking looks correct, but a bit messy. If you only need the lock for `rQuery, ok := allRunningQueries[qid]`, then you can unlock it (not defer unlocking it) right after you read from that map. Or if you need it for longer, you can defer unlocking it right after you lock it

AndrewHess (commented):
Looking good; one nit

pkg/segment/query/querystatus.go
vasukalariya: oh sorry. Missed it

vasukalariya: addressed in the latest commit. I have corrected this at a few more places. Later, will go over through this file and see if any more could be cleaned up.

AndrewHess (approved):
Looks good

PR #1640, opened 2024-09-16
Check rotated files when unrotated are absent

AndrewHess:
# Description
There's a concurrency bug that sometimes happens:
1. We get a query, and find the segments that match it; one of the unrotated segments match it
2. Concurrently, we receive more data, and rotate the segment that matched it
3. We continue processing the query, and try to read from the files where data is stored. But the segment got rotated, so it's in the `.../final/...` directory instead of the `.../active/...` directory. We log and return an error for that segment.

This PR fixes the issue by looking in `/final/` if we got an error looking in `/active/`

# Testing
Mostly manual testing. I decreased the size of segments via `maxSegFileSize: 200000000` in my server.yaml. I also added a 5-second sleep just above this line: https://github.com/siglens/siglens/blob/develop/pkg/segment/reader/record/recordreader.go#L138. Then I continually ingested data, ran the query `*`, and ingested more data so the segment would rotate before my sleep finished. Without this PR, I saw errors like:
```
InitSharedMultiColumnReaders: failed to open file data/sigsingle.nZgMjB6N9wZZUPAkyh6DdL/active/ind-0/0-0-16142326282426453201/290/290_7976178037195605558.csg for columns weekday. Error: open data/sigsingle.nZgMjB6N9wZZUPAkyh6DdL/active/ind-0/0-0-16142326282426453201/290/290_7976178037195605558.csg: no such file or directory. 
```
With this PR, those errors don't show up because we check `final` before logging the error, and the `final` finals are present

# Checklist:

### DIFF
--- a/pkg/segment/reader/segread/multicolreader.go
+++ b/pkg/segment/reader/segread/multicolreader.go
@@ -22,7 +22,6 @@ import (
 	"fmt"
 	"os"
 	"sort"
-	"strings"
 
 	"github.com/cespare/xxhash"
 	"github.com/siglens/siglens/pkg/blob"
@@ -192,10 +191,8 @@ func InitSharedMultiColumnReaders(segKey string, colNames map[string]bool, block
 		if err != nil {
 			// This segment may have been recently rotated; try reading the
 			// rotated segment file.
-			rotatedFName := strings.Replace(fName, "/active/", "/final/", 1)
+			rotatedFName := writer.GetRotatedVersion(fName)
 			currFd, err = os.OpenFile(rotatedFName, os.O_RDONLY, 0644)
-
-			log.Errorf("andrew failed to open a file, but final err=%v", err)
 			if err != nil {
 				log.Errorf("qid=%d, InitSharedMultiColumnReaders: failed to open file %s for columns %s. Error: %v.",
 					qid, rotatedFName, colName, err)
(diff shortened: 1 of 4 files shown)


pkg/segment/reader/segread/multicolreader.go
```diff
@@ -189,9 +189,15 @@ func InitSharedMultiColumnReaders(segKey string, colNames map[string]bool, block
...
 		if err != nil {
-			log.Errorf("qid=%d, InitSharedMultiColumnReaders: failed to open file %s for columns %s. Error: %v.",
-				qid, fName, colName, err)
-			continue
+			// This segment may have been recently rotated; try reading the
+			// rotated segment file.
+			rotatedFName := writer.GetRotatedVersion(fName)
+			currFd, err = os.OpenFile(rotatedFName, os.O_RDONLY, 0644)
+			if err != nil {
+				log.Errorf("qid=%d, InitSharedMultiColumnReaders: failed to open file %s for columns %s. Error: %v.",
```
nkunal, line 197: in case if the original err was not a active/final issue, do we want to also log that err here and maybe add some a couple of words something of : `orig fname: %v had failed with err: %v, rotated also failed with fname: %v, newErr: %v`

nkunal (commented):
looks good, one minor nit

nkunal (approved):
merging it for now. the comment can be addressed in the next PR

PR #1645, opened 2024-09-17
Included Unrotated stats in the stats table

Macbeth98:
# Description
- Added unrotated stats and updated the Ingest stats hook.
- This also includes the unrotated columns stats. Currently for the unrotated segments we are not showing the Segment count and Columns count. This PR fixes that.

# Checklist:


pkg/health/clusterStatsHandler.go
```diff
@@ -73,7 +73,14 @@ func ProcessClusterStatsHandler(ctx *fasthttp.RequestCtx, myid uint64) {
...
-			&queryCount, &totalResponseTimeSinceRestart)
+			&queryCount, &totalResponseTimeSinceRestart, &totalResponseTimeSinceInstall,
+			&queriesSinceInstall, totalColumnsSet)
+	}
+
+	logsColumnCount := len(totalColumnsSet)
+	// Remove the columns set from the index data
+	for _, idxData := range indexData.IndexToStats {
+		idxData.ColumnsSet = nil
 	}
```
AndrewHess, line 84: Why?

pkg/hooks/hooks.go
```diff
@@ -56,7 +56,8 @@ type Hooks struct {
 	AddMultinodeStatsHook      func(indexData utils.AllIndexesStats, orgId uint64,
 		logsIncomingBytes *float64, logsOnDiskBytes *float64, logsEventCount *int64,
 		metricsIncomingBytes *uint64, metricsOnDiskBytes *uint64, metricsDatapointsCount *uint64,
-		queryCount *uint64, totalResponseTime *float64)
+		queryCount *uint64, totalResponseTimeSinceRestart *float64, totalResponseTimeSinceInstall *float64,
+		totalQueryCountSinceInstall *uint64, totalColumnsSet map[string]struct{})
```
AndrewHess, line 60: nit: this is a lot of parameters. It would probably be good to make some struct types to store some of these that are semantically similar. But I don't think it's necessary to do that in this PR

AndrewHess (commented):
Looks pretty good

pkg/health/clusterStatsHandler.go
Macbeth98: I thought this data is not necessary to send back to the UI. And if there are too many indexes and each index on average has 200 or so columns. I jus thought we don't need this overhead while sending the response back.

pkg/hooks/hooks.go
Macbeth98: Yes for now I will add this as a TODO.

nkunal (approved):

PR #1654, opened 2024-09-18
Sigclient - added new genType benchmark-ingest-query,  run query after ingest

sunitakawane:
# Description
Summarize the change.
Sigclient - When ingesting with benchmark flag,  run query after ingest

When ingesting with benchmark flag,  run query after ingest

# Testing

# Checklist:


tools/sigclient/pkg/query/query.go
```diff
@@ -909,3 +910,84 @@ func RunQueryFromFileAndOutputResponseTimes(dest string, filepath string, queryR
...
+		}
+
+		readEvent := make(map[string]interface{})
+		for {
+			err = conn.ReadJSON(&readEvent)
+			if websocket.IsCloseError(err, websocket.CloseNormalClosure) {
+				//WebSocket closed normally (1000).
+				break
+			} else if err != nil {
+				log.Infof("RunBenchmarkUUIDQuery : Received error from server: %+v\n", err)
```
AndrewHess, line 951: This `Infof` should at least be `Errorf`. Should it be `Fatalf`?

tools/sigclient/pkg/query/query.go
```diff
@@ -909,3 +910,84 @@ func RunQueryFromFileAndOutputResponseTimes(dest string, filepath string, queryR
...
+				break
+			}
+			switch readEvent["state"] {
+			case "RUNNING", "QUERY_UPDATE":
+			case "COMPLETE":
+				for eKey, eValue := range readEvent {
+					if evaluationType == "total" && eKey == "totalMatched" {
+						var hits bool
+						var finalHits float64
```
AndrewHess, line 960: nit: this naming is confusing to me. What does `hits` as a bool mean? Is it more aptly `hitSomething`?

tools/sigclient/pkg/query/query.go
```diff
@@ -909,3 +910,84 @@ func RunQueryFromFileAndOutputResponseTimes(dest string, filepath string, queryR
...
+										log.Errorf("RunBenchmarkUUIDQuery: Returned total matched is not a float: %v", v)
+									}
+									hits, err = utils.VerifyInequality(finalHits, relation, expectedValue)
+								}
+							}
+						}
+						if err != nil {
+							log.Errorf("RunBenchmarkUUIDQuery: Error in verifying hits: %v", err)
+						} else if !hits {
+							log.Errorf("RunBenchmarkUUIDQuery: For Ident query: %v , Actual Hits: %v not gt than 0", data, finalHits)
```
AndrewHess, line 981: > Actual Hits: %v not gt than 0

nit: don't hardcode `gt` and `0`. I think you got these from your query, right?
```
		relation := "gt"
		expectedValue := "0"
```

AndrewHess (commented):
You should update the sigclient readme

tools/sigclient/pkg/query/query.go
sunitakawane: I will change this to Errorf. We want to continue with next uuid query so not using fatalf.

tools/sigclient/pkg/ingest/ingest.go
```diff
@@ -44,6 +46,8 @@ const (
 	OpenTSDB
 )
 
+var uuidList []string
+
```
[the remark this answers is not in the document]
sunitakawane: keeping this global, since it is used to collect ingested uuids from different goroutines here `rrdUUIDList, _ := rdr.GetUUIDList()`

tools/sigclient/pkg/ingest/ingest.go
```diff
@@ -451,4 +463,43 @@ readChannel:
...
 	}
+	//run search queries for all UUIDs
+	if generatorType == "benchmark-ingest-query" {
+		log.Info("Verifying benchmark ingestion by searching for ident column")
+		log.Info("Waiting for ingest data to be available. Sleeping for 2 minutes")
+		time.Sleep(120 * time.Second)
+		log.Info("Starting queries for benchmark ident column")
+		ticker := time.NewTicker(30 * time.Second)
+		uuidListLen := len(uuidList)
+		queriesPerBatch := uuidListLen / processCount
```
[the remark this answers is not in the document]
sunitakawane: 1. I changed the code and moved uuidList to reader.go
2. Added code to handle the extra/remainders in ingest and query code

tools/sigclient/pkg/query/query.go
sunitakawane: renamed it to uuidFound

tools/sigclient/pkg/ingest/ingest.go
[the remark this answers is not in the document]
sunitakawane: refactored the code , moved it to reader.go

tools/sigclient/pkg/utils/reader.go
```diff
@@ -36,12 +36,13 @@ import (
 )
 
 var json = jsoniter.ConfigFastest
+var uuidList []string
+var mu sync.Mutex
```
AndrewHess, line 40: nit: especially since this is a global variable, use a more descriptive name than `mu`, or at least comment what this mutex is used for

tools/sigclient/pkg/query/query.go
```diff
@@ -909,3 +910,84 @@ func RunQueryFromFileAndOutputResponseTimes(dest string, filepath string, queryR
...
+			"state":         "query",
+			"searchText":    "ident=" + uuid,
+			"startEpoch":    "now-1d",
+			"endEpoch":      "now",
+			"indexName":     "*",
+			"queryLanguage": "Splunk QL",
+		}
+		evaluationType := "total"
+		relation := "gt"
+		expectedValue := "0"
```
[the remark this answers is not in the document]
AndrewHess: Ok, then maybe this should be equal to the value of `p`. My concern is that with `> 0`, some bugs could potentially slip in unnoticed that would have been caught if we tested for equality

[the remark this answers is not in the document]
sunitakawane: For 100 queries , p=3 , the end for first batch will be adjusted to add the remainder of 1 and  will have one query with unique ident column not present in other batches. So checking for >=p fails .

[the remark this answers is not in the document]
AndrewHess: Ok. Do you think you can get it to work with equality, or should we keep this pretty relaxed constraint of `> 0`?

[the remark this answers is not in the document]
sunitakawane: we can add a check if processCount==1 then we can use `eq 1`  else >0. This also assumes we start fresh with no data otherwise subsequent ingestion will also have same uuid which will increase the hits count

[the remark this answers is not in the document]
nkunal: for this gentype subsequent ingestion should not cause any issue. This should pick a random seed so that the results are not dependent on whether previously data was present or not. If we want to pick the same seed, then a flag can be passed

tools/sigclient/pkg/ingest/ingest.go
```diff
@@ -356,7 +357,10 @@ func getReaderFromArgs(iType IngestType, nummetrics int, gentype string, str str
 		rdr = utils.InitDynamicUserGenerator(ts, seed, generatorDataConfig)
 	case "benchmark-ingest-query":
 		log.Infof("Initializing benchmark-ingest-query reader")
-		seed := int64(1001)
+		source := rand.NewSource(time.Now().UnixNano())
+		randomNumGen := rand.New(source)
+		lowerBound := 1001
+		seed := int64(lowerBound + randomNumGen.Intn(10000)) // Generate random number between 1000 and 11000
```
AndrewHess, line 363: What's the purpose of this lower bound?

tools/sigclient/pkg/query/query.go
```diff
@@ -926,15 +926,15 @@ func runBenchmarkQueryBatch(uuidList []string, wg *sync.WaitGroup, semaphore cha
 			"queryLanguage": "Splunk QL",
 		}
 		evaluationType := "total"
-		relation := "gt"
-		expectedValue := "0"
+		var relation, expectedValue string
+		relation = "eq"
+		expectedValue = "1"
```
AndrewHess, line 931: So now that we're using a random seed, we'll always get distinct UUIDs, even when run with parallelization—am I understanding that right?

sunitakawane: yes, now we can now check for equality condition

tools/sigclient/pkg/ingest/ingest.go
sunitakawane: I thought all other gentypes are using 1000 or 1001 as seed so I added that as lowerbound but it does not matter so I removed it now.

tools/sigclient/pkg/query/query.go
```diff
@@ -909,3 +910,130 @@ func RunQueryFromFileAndOutputResponseTimes(dest string, filepath string, queryR
...
+	wg := sync.WaitGroup{}
+	log.Info("Verifying benchmark ingestion by searching for ident column")
+	log.Info("Waiting for ingest data to be available. Sleeping for 15 seconds...")
+	time.Sleep(15 * time.Second)
+	log.Info("Starting queries for benchmark ident column")
+	ticker := time.NewTicker(30 * time.Second)
+	uuidListLen := len(uuidList)
+	queriesPerBatch := uuidListLen / processCount
+	remainder := uuidListLen % processCount
+	semaphore := make(chan struct{}, processCount)
```
AndrewHess, line 1006: I'm confused by this `semaphore` channel. Can you explain it to me?

You make the channel with size `processCount` and then start `processCount` goroutines that each put one item on the channel and then take one item off when they're done. So it looks like it doesn't actually do anything??

tools/sigclient/pkg/query/query.go
```diff
@@ -909,3 +910,130 @@ func RunQueryFromFileAndOutputResponseTimes(dest string, filepath string, queryR
...
+			"endEpoch":      "now",
+			"indexName":     "*",
+			"queryLanguage": "Splunk QL",
+		}
+		evaluationType := "total"
+		var relation, expectedValue string
+		relation = "eq"
+		expectedValue = "1"
+		// create websocket connection
+		conn, _, err := websocket.DefaultDialer.Dial("ws://localhost:5122/api/search/ws", nil)
```
AndrewHess, line 933: Should this read the address from a command line arg so we can test on non-localhost setups?

tools/sigclient/pkg/query/query.go
```diff
@@ -909,3 +910,130 @@ func RunQueryFromFileAndOutputResponseTimes(dest string, filepath string, queryR
...
+	done := make(chan bool)
+	startTime := time.Now()
+	go func() {
+		for {
+			select {
+			case <-done:
+				return
+			case <-ticker.C:
+				log.Infof("Benchmark UUID queries still running. Total elapsed time:%s.", time.Since(startTime).Truncate(time.Second))
+			}
```
AndrewHess, line 1016: I _think_ this `done` channel is never written to, so this goroutine runs forever. Am I missing something?

PR #1658, opened 2024-09-18
Symlink active to final dir on rotation

AndrewHess:
# Description
Instead of moving the `active` directory to `final` on segment rotation, make a symlink. Occasionally, moving the directory causes query problems because when we start the query the segment data is in `active` but by the time we're actually fetching the data it's been moved to `final`. 

# Testing
Manual testing for continuous ingestion with continuous query, and restarting siglens and verifying the data can still be queried.

# Checklist:


AndrewHess:
We're using https://github.com/siglens/siglens/pull/1661 to solve the issue instead

PR #1661, opened 2024-09-18
Merge active and final data directories

AndrewHess:
# Description
Previously we would flush WIP blocks for a segment into an `active` directory, and when we rotated the segment we'd move that `active` directory to `final`. This causes some problems during query because sometimes we obtain segkeys for `active`, but then the segment rotates, and then when we look for that data we can't find it because it's in `final`. This PR addresses this issue by removing the `active` dir; now we put all data—even unrotated data—in `final`.

With this, we do lose some ease of identifying if a segment is rotated or not. We can look in segmeta.json for the segkey; also, this PR adds a new file (segment-validity.json) that is created during segment rotation—if that file is present, then the segment is valid and rotated.

# Testing
Ran continuous ingest and query, and it seems to be working normally.

# Checklist:

### DIFF
--- a/pkg/utils/segutils.go
+++ b/pkg/utils/segutils.go
@@ -22,12 +22,17 @@ import (
 	"encoding/gob"
 	"fmt"
 	"math/rand"
+	"os"
+	"path/filepath"
+	"strings"
 
 	"github.com/cespare/xxhash"
 	"github.com/rogpeppe/fastuuid"
+	log "github.com/sirupsen/logrus"
 )
 
 const UnescapeStackBufSize = 64
+const SegmentValidityFname = "segment-validity.json"
 
 var UUID_GENERATOR *fastuuid.Generator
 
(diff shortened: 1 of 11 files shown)


pkg/utils/segutils.go
```diff
@@ -131,3 +136,41 @@ func IsSubWordPresent(haystack []byte, needle []byte, isCaseInsensitive bool) bo
...
+	}
+
+	_, err = os.Stat(filepath.Join(segBaseDir, SegmentValidityFname))
+	return err == nil
+}
+
+func getSegBaseDirFromFilename(filename string) (string, error) {
+	// Note: this is coupled to getBaseSegDir. If getBaseSegDir changes, this
+	// should change too.
+	depth := 6 // getBaseSegDir has 6 components
```
nkunal, line 160: nit: maybe add a comment on the example of what those 6 components are

nkunal (approved):

PR #1668, opened 2024-09-23
minor fix-change GetUIDomain to return hostname  or localhost

sunitakawane:
# Description
Summarize the change.

# Testing

# Checklist:

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -238,7 +238,12 @@ func SetEmailConfig(smtpHost string, smtpPort int, senderEmail string, gmailAppP
 }
 
 func GetUIDomain() string {
-	return GetQueryHostname()
+	hostname := GetQueryHostname()
+	if hostname == "" {
+		return "localhost"
+	} else {
+		return hostname
+	}
 }
 
 func GetSiglensDBConfig() (string, string, uint64, string, string, string) {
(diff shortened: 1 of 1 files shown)


AndrewHess (commented):
The change looks good, but it doesn't do what the PR title says it does

pkg/config/config.go
```diff
@@ -238,7 +239,12 @@ func SetEmailConfig(smtpHost string, smtpPort int, senderEmail string, gmailAppP
 }
 
 func GetUIDomain() string {
-	return GetQueryHostname()
+	hostname := GetQueryHostname()
+	parsedUrl, err := url.Parse(hostname)
+	if err != nil {
+		log.Fatalf("GetUIDomain: Failed to parse QueryHostname: %v, err: %v", hostname, err)
```
AndrewHess, line 245: Are you sure we want to crash if we get an error here? This function could feasibly get called at any point, not just at startup

pkg/config/config.go
```diff
@@ -238,7 +239,12 @@ func SetEmailConfig(smtpHost string, smtpPort int, senderEmail string, gmailAppP
...
+	}
+	return parsedUrl.Host
```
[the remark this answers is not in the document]
sunitakawane: It is being set to locahost:port in config.go https://github.com/siglens/siglens/blob/bf755876ade29643b4edbf47e04ba3cd94ff0e2d/pkg/config/config.go#L789

pkg/config/config.go
sunitakawane: I changed this to Errorf and setting it to localhost in case of error

PR #1685, opened 2024-09-26
Cancel the query timeout thread when deleting the query

Macbeth98:
# Description
- When a query is started, a query timeout context is created and assigned to the running query state. 
- When the query is deleted, the associated query timeout context is canceled, effectively concluding the go-routine responsible for managing the query timeout.

# Testing
- Tested by running some successful queries and verified through logs that the query timeout go-routine is completed.
- And decreased the timeout to be 5 seconds and ran a query that will take more than 5 seconds and verified that the query is being cancelled.

# Checklist:


pkg/segment/query/querystatus.go
```diff
@@ -171,8 +177,15 @@ func StartQuery(qid uint64, async bool) (*RunningQueryState, error) {
...
 	arqMapLock.Lock()
-	delete(allRunningQueries, qid)
+	rQuery, ok := allRunningQueries[qid]
+	if ok {
+		delete(allRunningQueries, qid)
+	}
 	arqMapLock.Unlock()
+
+	if ok && !rQuery.isCancelled {
+		rQuery.timeoutCancelFunc()
```
nkunal, line 187: should we just do this cancel right inside the lock, before/after we delete from map?

AndrewHess (commented):
This looks pretty good. One question: does this handle the case where a query is manually canceled (e.g., a user runs a query on the UI and then cancels it). I think we'd keep the goroutine alive until it timed out, but I might be missing something

pkg/segment/query/querystatus.go
Macbeth98: No Kunal, we do not need to do that. We can delete the qid from the map and can cancel the timeout. Calling this function is just gonna invoke `context.Done()` and then it will just exit the function completing the go routine.
Whether this `qid` is there in the map or not, it does not matter.

Macbeth98:
> This looks pretty good. One question: does this handle the case where a query is manually canceled (e.g., a user runs a query on the UI and then cancels it). I think we'd keep the goroutine alive until it timed out, but I might be missing something

Yes, the manual cancelling of query will also work. As after the query is cancelled, deleteQuery will also be called. So the query timeout will also be cancelled.

AndrewHess:
@nkunal this PR looks good to me. I'm not sure though if your question was about a correctness issue or a style issue, so I'll hold off on merging this

nkunal (approved):

nkunal:
> @nkunal this PR looks good to me. I'm not sure though if your question was about a correctness issue or a style issue, so I'll hold off on merging this

just a style issue. 
approving and merging it now

PR #1686, opened 2024-09-26
Delete segStore with Lock Enabled

Macbeth98:
# Description
- Delete the segstore with Lock Enabled while deleting the unused segstores.

# Testing
- Tested by ingesting some data and I found no issues. 

# Checklist:


pkg/segment/writer/segwriter.go
```diff
@@ -559,6 +559,10 @@
 
 func rotateSegmentOnTime() {
 	segRotateDuration := time.Duration(SEGMENT_ROTATE_DURATION_SECONDS) * time.Second
+
```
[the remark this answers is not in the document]
Macbeth98: I will be making changes with this suggestion.

pkg/segment/writer/segwriter.go
```diff
@@ -589,14 +592,22 @@ func rotateSegmentOnTime() {
...
 	}
 	wg.Wait()
 	allSegStoresLock.RUnlock()
+
+	close(segStoresToDeleteChan)
+
+	allSegStoresLock.Lock()
+	for streamid := range segStoresToDeleteChan {
+		delete(allSegStores, streamid)
+	}
```
AndrewHess, line 609: I think there's a small gap where a concurrency bug can still happen, but correct me if I'm wrong. You hold the read lock to detect which segments to delete, then release it, get the write lock, and delete those segment. But between releasing the read lock and obtaining the write lock, a segstore marked for deletion could be modified so it shouldn't be deleted—in that case we still delete it once we get the write lock. I think that would cause a loss of ingested data, is that right?

Really what you want to do is upgrade your read lock to a write lock. I don't think Go has a way to do this (but you might want to check). Your other alternatives are:
1. Hold the write lock the whole time. This might cause performance issues
2. Just before you actually delete the segment, verify that it still needs to be deleted

Macbeth98: I have addressed this: when looping through the buffered channel, the code checks again whether the streamid can be deleted before actually deleting it.

nkunal (approved):

PR #1690, opened 2024-09-28
fix concurrent cmi map write and long segmeta entry fix

nkunal:
# Description
1. Fix for large segmeta entry (>64KB), we were silently failing to read it
2. When blocks are small parse, and some columns appear only in some blocks, we were getting segread error. This commit fixes that by not logging those errors if the compressedbuflen is zero for that column for this block.
3. Add rwlock in segmicroindex, utilize read locks for readers and rwlocks for some of the nil out operations that were happening
4. Code restructuring to allowing introduction of locks in segmeta layer.
5. Reduce the number of exported functions in segmicroindex.

# Testing

# Checklist:


pkg/segment/metadata/metadata_test.go
```diff
@@ -0,0 +1,179 @@
...
+
+	for i := int64(0); i < 10; i++ {
+		assert.Equal(t, uint64(i), GetSegmentMetadataReverseIndexForTest()[strconv.FormatInt(i, 10)].LatestEpochMS)
+	}
+	nextVal := uint64(9)
+	for i := 0; i < 10; i++ {
+		assert.Equal(t, GetAllSegmentMicroIndexForTest()[i].LatestEpochMS, nextVal)
+		log.Infof("i %+v latest: %+v", i, GetAllSegmentMicroIndexForTest()[i].LatestEpochMS)
+		nextVal--
+	}
```
AndrewHess, line 55: I'm missing something. For the second loop, why is the expected value `9, 8, 7, ...` instead of `0, 1, 2, ...`?

When you made the mirco indexes, you set the `i`-th one to have a `LatestEpoch` of `i`, and that's what your first `for` tests here.

pkg/segment/metadata/metadata_test.go
```diff
@@ -0,0 +1,179 @@
...
+	for i := 0; i < fileCount; i++ {
+		assert.False(t, GetAllSegmentMicroIndexForTest()[i].AreMicroIndicesLoaded())
+	}
+
+	testBlockRebalance(t, fileCount)
+
+	err := os.RemoveAll("data/")
+	if err != nil {
+		assert.FailNow(t, "failed to remove data %+v", err)
+	}
```
AndrewHess, line 125: nit: if this test panics before here (or if the test is changed and it exits before here, e.g., with a bad `require.True()`), then this cleanup code won't run. You can instead put this at the start of the function (and add error handling if you want):
```
	t.Cleanup(func() { os.RemoveAll("data/") })
```

pkg/segment/metadata/metaupdater_test.go
```diff
@@ -0,0 +1,67 @@
...
+	BulkAddSegmentMicroIndex([]*SegmentMicroIndex{duplicate})
+	assert.Len(t, GetAllSegmentMicroIndexForTest(), fileCount)
+	assert.Len(t, GetSegmentMetadataReverseIndexForTest(), fileCount)
+	assert.Contains(t, GetTableSortedMetadataForTest(), "evts")
+	assert.Len(t, GetTableSortedMetadataForTest()["evts"], fileCount)
+
+	err := os.RemoveAll("data/")
+	if err != nil {
+		log.Fatal(err)
+	}
```
AndrewHess, line 66: Same comment

pkg/segment/metadata/metaupdater_test.go
```diff
@@ -0,0 +1,67 @@
...
+	assert.True(t, sortedByLatest, "slice is sorted with most recent at the front")
+
+	tableSorted := GetTableSortedMetadataForTest()["evts"]
+	isTableSorted := sort.SliceIsSorted(tableSorted, func(i, j int) bool {
+		return tableSorted[i].LatestEpochMS > tableSorted[j].LatestEpochMS
+	})
+	assert.True(t, isTableSorted, "slice is sorted with most recent at the front")
+
+	for i, rawCMI := range tableSorted {
+		assert.Equal(t, rawCMI, GetAllSegmentMicroIndexForTest()[i], "bc only one table exists, these sorted slices should point to the same structs")
```
AndrewHess, line 50: nit: change "bc" to "because" or "since" to avoid confusion if we do fail the unit test here

pkg/segment/metadata/segmentmicroindex.go
```diff
@@ -69,14 +72,15 @@ type SegmentSearchMetadata struct {
...
-func (ssm *SegmentSearchMetadata) IsSearchMetadataLoaded() bool {
+func (ssm *SegmentSearchMetadata) isSearchMetadataLoaded() bool {
 	return ssm.loadedSearchMetadata
 }
 
 func InitSegmentMicroIndex(segMetaInfo *structs.SegMeta) *SegmentMicroIndex {
 
 	sm := &SegmentMicroIndex{
 		SegMeta: *segMetaInfo,
+		smilock: &sync.RWMutex{},
```
AndrewHess, line 83: nit
```suggestion
		smiLock: &sync.RWMutex{},
```

pkg/segment/metadata/segmentmicroindex.go
```diff
@@ -287,14 +306,148 @@ func (sm *SegmentMicroIndex) GetColumns() map[string]bool {
...
 	return uint32(sm.SegMeta.RecordCount)
 }
+
+func GetLoadSsm(segkey string, qid uint64) (*SegmentMicroIndex, int64, error) {
+
+	smi, exists := getMicroIndex(segkey)
+	if !exists {
+		log.Errorf("qid=%d, Seg  %+v does not exist in block meta, but existed in time filtering", qid, segkey)
+		return nil, 0, fmt.Errorf("seg file %+v does not exist in block meta, but existed in time filtering", segkey)
+	}
```
AndrewHess, line 336: nit: you can use `utils.TeeErrorf()` to log and return an error. But you can keep what you have if you want

pkg/segment/metadata/segmentmicroindex.go
```diff
@@ -287,14 +306,148 @@ func (sm *SegmentMicroIndex) GetColumns() map[string]bool {
...
+
+func (smi *SegmentMicroIndex) ClearSearchTimeData() {
+
+	if !smi.AreMicroIndicesLoaded() {
+		smi.clearMicroIndices()
+	}
+	if !smi.isSearchMetadataLoaded() {
+		smi.clearSearchMetadata()
+	}
+}
```
AndrewHess, line 403: I might be missing something; this looks like a no-op. e.g., it only clears micro indices if there's no micro indices loaded

pkg/segment/reader/segread/segreader.go
```diff
@@ -75,6 +75,7 @@ type SegmentFileReader struct {
 	deTlv              [][]byte // deTlv[dWordIdx] --> []byte (the TLV byte slice)
 	deRecToTlv         []uint16 // deRecToTlv[recNum] --> dWordIdx
 	blockSummaries     []*structs.BlockSummary
+	someBlksAbsent     bool // this is used for not log some errors
```
AndrewHess, line 78: nit
```suggestion
	someBlksAbsent     bool // this is used to not log some errors
```

pkg/segment/reader/segread/segreader.go
```diff
@@ -232,19 +234,29 @@ func (sfr *SegmentFileReader) ReadRecordFromBlock(blockNum uint16, recordNum uin
...
 func (sfr *SegmentFileReader) iterateNextRecord() error {
 	nextOff := sfr.currOffset + sfr.currRecLen
 	if nextOff >= sfr.currUncompressedBlockLen {
-		log.Errorf("SegmentFileReader.iterateNextRecord: reached end of block, next Offset: %+v, curr uncompressed blklen: %+v", nextOff, sfr.currUncompressedBlockLen)
+		if !sfr.someBlksAbsent {
+			log.Errorf("SegmentFileReader.iterateNextRecord: reached end of block, next Offset: %+v, curr uncompressed blklen: %+v", nextOff, sfr.currUncompressedBlockLen)
+		}
+		// we don't log an error, but we are returning err so that, the caller does not
+		// get stuck an loop
 		return errors.New("no more records to iterate")
```
AndrewHess, line 260: Should we return `io.EOF` as the error?

pkg/segment/writer/segmetareader.go
```diff
@@ -29,6 +29,8 @@ import (
 	log "github.com/sirupsen/logrus"
 )
 
+const ONE_MB = 1024 * 1024
```
AndrewHess, line 32: nit: maybe use 1000 * 1000, or specify this is actually 1 MiB

AndrewHess (commented):
Looking good; I had some nits

pkg/segment/metadata/segmentmicroindex.go
nkunal: this function is called during search time, during search we need micro indices for all segs that we are searching for. But we need to then immediately clear them out to avoid mem overload. The rebalance layer only keeps microindices that it is allowed to keep based on the mem percentages. And it then sets the flag isLoaded to true. 
however during search time, we merely load the microindices but do not set the flag, so we are kind of overloading the use of that variable so that we can clear out this microindices. But we also want to mindful, that while we are clearing out and if the rebalance thread kicks in and it loads the micro for this segment, then we dont want to clear it.

pkg/segment/metadata/metadata_test.go
nkunal: The second for loop tests the functionality of this func `GetAllSegmentMicroIndexForTest` to make sure it returns the microindices according the latestEpoch sorted by time (most recent first), since that is what is used during query time. We want to look at the segment that is most recent

pkg/segment/metadata/metadata_test.go
nkunal: yup makes sense

pkg/segment/metadata/metaupdater_test.go
nkunal: done

pkg/segment/metadata/metaupdater_test.go
nkunal: done

AndrewHess (approved):
Looks good!

PR #1695, opened 2024-10-01
misc fix for test-data page and index.js

sunitakawane:
Code cleanup for test-data.js , fix listIndices to display test-data index

# Description
Summarize the change.

# Testing

# Checklist:


pkg/ast/pipesearch/listIndicesHandler.go
```diff
@@ -33,17 +33,16 @@ type IndexInfo struct {
...
 
 	var allIndexNames []string
 	if hooks.GlobalHooks.AddMultiNodeIndexHook != nil {
-		allIndexNames = hooks.GlobalHooks.AddMultiNodeIndexHook(myid)
-	} else {
-		allIndexNames = vtable.ExpandAndReturnIndexNames("*", myid, false)
+		allIndexNames = hooks.GlobalHooks.AddMultiNodeIndexHook(orgId)
 	}
+	allIndexNames = append(allIndexNames, vtable.ExpandAndReturnIndexNames("*", orgId, false)...)
 	sort.Strings(allIndexNames)
```
AndrewHess, line 46: Do you need some deduping, or will there never be duplicates even if it goes into the `if` block?

AndrewHess (commented):
Backend looks good; just one question

pkg/ast/pipesearch/listIndicesHandler.go
sunitakawane: added dedup code

pkg/segment/writer/segwriter.go
```diff
@@ -334,7 +334,7 @@ func (ss *SegStore) doLogEventFilling(ple *ParsedLogEvent, tsKey *string) (bool,
 		colWip, _, matchedCol = ss.initAndBackFillColumn(cname, SS_DTYPE(ctype), matchedCol)
 
 		switch ctype {
-		case VALTYPE_ENC_SMALL_STRING[0]:
+		case VALTYPE_ENC_SMALL_STRING[0], VALTYPE_ENC_BOOL[0]:
```
AndrewHess, line 337: Can we really do this? I though we encode strings as their type (string), then length, then value; and for bools just their type (bool) and value. So I think this will cause issues with ingesting booleans. Are you absolutely sure it will not?

nkunal: yeah this is not right, we would be copying some garbage for bool types.

static/js/test-data.js
```diff
@@ -32,11 +32,11 @@ $(document).ready(function () {
         credentials: 'include'
     })
         .then((_res) => {
-            let token;
+            let iToken="";
```
sonamgupta21, line 35: Remove the redundant `iToken` declaration form here since it is already declared globally.

sonamgupta21 (commented):
Frontend changes look good to me.

pkg/segment/writer/segwriter.go
sunitakawane: fixed it in a separate PR

AndrewHess (approved):

PR #1698, opened 2024-10-02
Fix total cols for unrotated data in cluster stats

vasukalariya:
# Description
Summarize the change.
We were not updating the totalCols for the case of unrotated data when there is no existing data for that [index](https://github.com/siglens/siglens/blob/f0257e9e377f71a7f28143bcead60648172b333c/pkg/health/clusterStatsHandler.go#L382). We have to update totalCols if there is any unrotated data like [here](https://github.com/siglens/siglens/blob/f0257e9e377f71a7f28143bcead60648172b333c/pkg/health/clusterStatsHandler.go#L385).
Moreover, earlier we were just using segMeta to get the column count, in the unrotated data we get the data from segstore, which has `timestamp` as a column too. Added this timestamp column for rotated data as well.

# Testing
- Clean the data directory.
- Start siglens server
- Ingest Some data
- Observe the cluster stats, earlier we were getting 0 as the total column count.
- After this change it should show correctly.
- After ingestion, restart the server (let the data be rotated). Look at the cluster stats, they should be same as what you saw before.
- Earlier, the segment count was 1 less than what we see for unrotated data. Now, they should be same.

# Checklist:

### DIFF
--- a/pkg/health/clusterStatsHandler.go
+++ b/pkg/health/clusterStatsHandler.go
@@ -27,6 +27,7 @@ import (
 
 	"github.com/dustin/go-humanize"
 	jsoniter "github.com/json-iterator/go"
+	"github.com/siglens/siglens/pkg/config"
 	"github.com/siglens/siglens/pkg/hooks"
 	"github.com/siglens/siglens/pkg/segment/structs"
 	"github.com/siglens/siglens/pkg/segment/writer"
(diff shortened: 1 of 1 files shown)


pkg/health/clusterStatsHandler.go
```diff
@@ -355,6 +356,10 @@ func getStats(myid uint64, filterFunc func(string) bool, allSegMetas []*structs.
 			allIndexCols[indexName][col] = struct{}{}
 			totalCols[col] = struct{}{}
 		}
+		allIndexCols[indexName][config.GetTimeStampKey()] = struct{}{}
```
nkunal, line 359: nit: do we want to get the tskey once at the start of this func and then use it to avoid the mulitple calls in the for loop ?

vasukalariya: addressed in the latest commit.

nkunal (approved):

PR #1700, opened 2024-10-02
Fix Test_IngestMultipleTypesIntoOneColumn  test

sunitakawane:
# Description,

Test_IngestMultipleTypesIntoOneColumn .It was using ProcessIndexRequest with nil return value and working before. When I changed the ProcessIndexRequest implementation to use ple , it is failing doLogEventFilling: unknown ctype: 1"
It was not handling bool type

# Testing

# Checklist:


pkg/segment/writer/segwriter.go
```diff
@@ -356,6 +356,23 @@ func (ss *SegStore) doLogEventFilling(ple *ParsedLogEvent, tsKey *string) (bool,
...
+			startIdx := colWip.cbufidx
+			copy(colWip.cbuf[startIdx:], ple.allCvalsTypeLen[i][0:1])
+			colWip.cbufidx += 1
+			copy(colWip.cbuf[colWip.cbufidx:], ple.allCvals[i][0:1])
+			colWip.cbufidx += 1
+			boolVal := utils.BytesToBoolLittleEndian(ple.allCvals[i][0:1])
+			var asciiBytesBuf bytes.Buffer
+			_, err := fmt.Fprintf(&asciiBytesBuf, "%v", boolVal)
+			if err != nil {
+				return false, utils.TeeErrorf("doLogEventFilling: cannot write boolVal %v: %v", boolVal, err)
```
nkunal, line 369: we do not want to log every bool that failed, will overwhem the logs

pkg/segment/writer/segwriter.go
```diff
@@ -356,6 +356,23 @@ func (ss *SegStore) doLogEventFilling(ple *ParsedLogEvent, tsKey *string) (bool,
 				ss.checkAddDictEnc(colWip, colWip.cbuf[startIdx:colWip.cbufidx], ss.wipBlock.blockSummary.RecCount, startIdx, false)
 			}
 			ss.updateColValueSizeInAllSeenColumns(cname, colWip.cbufidx-startIdx)
+		case VALTYPE_ENC_BOOL[0]:
+			startIdx := colWip.cbufidx
+			copy(colWip.cbuf[startIdx:], ple.allCvalsTypeLen[i][0:1])
+			colWip.cbufidx += 1
+			copy(colWip.cbuf[colWip.cbufidx:], ple.allCvals[i][0:1])
+			colWip.cbufidx += 1
```
AndrewHess, line 364: nit: just do one `copy()` to copy both bytes

pkg/segment/writer/segwriter.go
```diff
@@ -356,6 +356,23 @@ func (ss *SegStore) doLogEventFilling(ple *ParsedLogEvent, tsKey *string) (bool,
...
+			boolVal := utils.BytesToBoolLittleEndian(ple.allCvals[i][0:1])
+			var asciiBytesBuf bytes.Buffer
+			_, err := fmt.Fprintf(&asciiBytesBuf, "%v", boolVal)
+			if err != nil {
+				return false, utils.TeeErrorf("doLogEventFilling: cannot write boolVal %v: %v", boolVal, err)
+			}
+			addSegStatsBool(segstats, cname, asciiBytesBuf.Bytes())
```
AndrewHess: Why do you have to do so much work here to get `asciiBytesBuf.Bytes()`? Isn't it always going to be 1?

pkg/segment/writer/segwriter.go
sunitakawane: I am copying from two different sources.

pkg/segment/writer/segwriter.go
sunitakawane: bool value can be true or false , we need to get the value not just the length

pkg/segment/writer/segwriter.go
AndrewHess: Oh, my bad

AndrewHess (approved):

PR #1701, opened 2024-10-02
Limit the number of colWips

vasukalariya:
# Description
Summarize the change.
Added a strict check to prevent the number of colWips from exceeding the limit specified by `colWipsSizeLimit`.
colWips should not be allocated more than what is available. To track this change, I have used an counter which is updated atomically whenever a column is added or removed.
The only exceptions to this check are:
- [convertColumnToNumbers](https://github.com/siglens/siglens/blob/8ea6a2e09af94f7052284e91b03abaddba09f46c/pkg/segment/writer/segstore.go#L379)
- [convertColumnToStrings](https://github.com/siglens/siglens/blob/8ea6a2e09af94f7052284e91b03abaddba09f46c/pkg/segment/writer/segstore.go#L452)

These two functions create a new colWip and delete the old one, so it shouldn't be a problem. Also, this is done during the appendWipToSegFile, so we might want to loosen up on this check here cause (wipblock is filled and will be reset after flushing).

# Testing
- Modified the `colWipsSizeLimit` to 34 (number of cols in sigclient).
- Update the seg rotation time to a smaller value like 10-20s.
- Ingest using this command `go run main.go ingest esbulk -n 2 -g benchmark -d http://localhost:8081/elastic -t 4 -b 2`
- The above command will insert 2 records in each index having 34 columns each.
- Only one batch succeeded as expected and checked the responses.
- Once, the segment was rotated, this remaining batch also got ingested successfully.
- Here are some screenshots of responses and logs.

image
image

# Checklist:


pkg/es/writer/esBulkHandler.go
```diff
@@ -405,6 +423,10 @@ func ProcessIndexRequest(rawJson []byte, tsNow uint64, indexNameIn string,
 		log.Errorf("ProcessIndexRequest: failed to add entry to in mem buffer, StreamId=%v, rawJson=%v, err=%v", streamid, rawJson, err)
 		return err
 	}
+	pleResponse := ple.GetResponse()
+	if pleResponse != "" {
+		return utils.TeeErrorf("ProcessIndexRequest: failed to add entry to in mem buffer, StreamId=%v, rawJson=%v, err=%v", streamid, rawJson, pleResponse)
```
nkunal, line 428: this will overwhelm the logs, we only want to return the error to the user that there was ingestion issue

pkg/segment/writer/segwriter.go
```diff
@@ -80,6 +83,15 @@ var wipCbufPool = sync.Pool{
 	},
 }
 
+func GetNumOfActiveColWips() int64 {
+	value := atomic.LoadInt64(&activeColWips)
+	return value
+}
+
+func AddNumOfActiveColWips(count int64) {
```
nkunal, line 91: these two functions do not need to be exported, they are only used within the writer package

pkg/es/writer/esBulkHandler.go
vasukalariya: This function is now removed.

pkg/segment/writer/segwriter.go
```diff
@@ -51,10 +51,13 @@ import (
 
 // Throttle the number of indexes to help prevent excessive memory usage.
 const maxAllowedSegStores = 1000
+const MAX_ACTIVE_COL_WIPS = 1500 // TODO: make this variable based on available memory
```
nkunal, line 54: we already have a config var for it , max_open_cols or something

AndrewHess:
Closing because it's stale

PR #1702, opened 2024-10-03
Add helper functions to read one/all RRC columns

AndrewHess:
# Description
See PR title. These functions will be used when we refactor query execution, but aren't used yet.

# Testing
The new unit tests pass, but otherwise this hasn't been tested. None of the new public functions are actually called, so no functionality has changed.

# Checklist:


nkunal (dismissed):
I am not fully understanding the code, but once all the things get plugged in then will review again. 
if we want we can merge now. so approving it.

pkg/segment/reader/record/recordreader.go
```diff
@@ -37,6 +37,211 @@ import (
...
+	}
+
+	blockSummary, err := getBlockSummary(segKey)
+	if err != nil {
+		log.Errorf("readUserDefinedColForRRCs: failed to get block summary for segKey %s; err=%v", segKey, err)
+		return nil, err
+	}
+
+	consistentCValLen := map[string]uint32{cname: utils.INCONSISTENT_CVAL_SIZE} // TODO: use correct value
+	qid := uint64(0)                                                            // TODO
```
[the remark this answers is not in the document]
AndrewHess: Fixed

AndrewHess:
> I am not fully understanding the code

Hmm that's not good. Basically I wanted to add these two functions:
- ReadColForRRCs (we'll need this during query evaluation for something like `* | eval myCol=latency*10` so we can read the `latency` for each RRC)
- ReadAllColsForRRCs (we'll need this at the end of the query when we have a list of RRCs and want to show all the columns to the user on the UI)

All the other changes are so I could implement those. These functions are for taking a list of RRCs and either reading one specific column or reading all the columns. We do have an existing way to read all the columns (`getRecordsFromSegmentHelper()`), but it has a lot of bloat that we don't need for the new way we'll handle queries.

pkg/segment/reader/record/recordreader.go
```diff
@@ -36,6 +36,171 @@ import (
...
+
+	consistentCValLen := map[string]uint32{cname: utils.INCONSISTENT_CVAL_SIZE} // TODO: use correct value
+	sharedReader, err := segread.InitSharedMultiColumnReaders(segKey, map[string]bool{cname: true},
+		blockMetadata, blockSummary, 1, consistentCValLen, qid)
+	if err != nil {
+		log.Errorf("qid=%v, readUserDefinedColForRRCs: failed to initialize shared readers for segkey %v; err=%v", qid, segKey, err)
+		return nil, err
+	}
+	defer sharedReader.Close()
+	multiReader := sharedReader.MultiColReaders[0]
```
nkunal, line 149: why are we taking only the first reader (wouldn't this be just one csg file ?

pkg/segment/reader/record/recordreader.go
```diff
@@ -36,6 +36,171 @@ import (
...
+		}
+
+		return handleBlock(multiReader, rrcsInBatch[0].BlockNum, rrcsInBatch, qid)
+	}
+
+	enclosures := toputils.BatchProcess(rrcs, batchingFunc, batchKeyLess, operation)
+	return enclosures, nil
+}
+
+func handleBlock(multiReader *segread.MultiColSegmentReader, blockIdx uint16,
```
nkunal, line 170: nit: should we call this `blockNum` as we have been referring to it that way in most of the code base ? 
but also ok with keeping it `blockIdx`

pkg/segment/reader/record/recordreader.go
```diff
@@ -36,6 +36,171 @@ import (
 	log "github.com/sirupsen/logrus"
 )
 
+func ReadAllColsForRRCs(segKey string, vTable string, rrcs []*utils.RecordResultContainer,
```
nkunal, line 39: if this toplevel function gets called multiple times on multiple segkeys in during one query run time. Then we should think if passing in a array of array of RRCs, so that we can reuse the rrc array. The cval enclosures are small sizes but when done on thousands of records/columns, they can create GC pressure. But first we should calculate out how many times this function is going to get called on a query and then decide if we want to pre-alloc. 
This does not need to be done now. Just making a note of this as an optimization we can think of when we start doing perf tests.

nkunal (approved):
This is looking fantastic, the abstractions are cool. 
just have a couple of minor comments

pkg/segment/reader/record/recordreader.go
AndrewHess: I don't fully understand the readers, so you may be right. I copied this part from `getRecordsFromSegmentHelper()` where we get the first; in the call to `InitSharedMultiColumnReaders()` we pass `1` as `numReaders`, so there should only be one reader returned, which is why I'm doing this. I thought that one reader would be able to read all the necessary columns, but I might be wrong

pkg/segment/reader/record/recordreader.go
AndrewHess: Yes, passing in an array of CValEnclosures (I think that's what you meant?) could be a good optimization. I'll wait for benchmarking first though, and not do that in this PR

pkg/segment/reader/record/recordreader.go
AndrewHess: I'll change that in a different PR. It was called blockIdx other places in this file, but I agree blockNum is better

pkg/segment/reader/record/recordreader.go
nkunal: ok, lets take a note of this, and once we are in the testing phase , then we can validate this

nkunal (approved):

PR #1710, opened 2024-10-03
Add IntermediateQueryResult type

AndrewHess:
# Description
- [x] https://github.com/siglens/siglens/pull/1702 should be merged first, and then this PR should be rebased.

This adds the basic functionality for an IntermediateQueryResult (IQR). In the design, this was called RRCsWrapper. I chose a new name because for some queries we don't have RRCs at some points in the evaluation, but we still want to use this type to store the results (e.g., after evaluating `stats` we don't have RRCs)

As we flesh out the new query pipeline, we'll need to add more functionality to this IQR type, but this PR should have the functionality needed for basic queries deal with RRCs.

# Testing
New unit tests pass

# Checklist:


pkg/segment/query/iqr/intermediateQueryResult.go
```diff
@@ -0,0 +1,372 @@
...
+type IQR struct {
+	mode iqrMode
+
+	// Used if and only if the mode is withRRCs.
+	rrcs             []*utils.RecordResultContainer
+	encodingToSegKey map[uint16]string
+
+	// Used in both modes.
+	qid            uint64
+	knownValues    map[string][]utils.CValueEnclosure
```
nkunal, line 48: nit: a comment about what the map key is going to be about

pkg/segment/query/iqr/intermediateQueryResult.go
```diff
@@ -0,0 +1,372 @@
...
+	deletedColumns map[string]struct{}
+	renamedColumns map[string]string
```
nkunal, line 50: nit: a comment about what this map is going to contain

pkg/segment/query/iqr/intermediateQueryResult.go
```diff
@@ -0,0 +1,372 @@
...
+	if iqr == nil {
+		return fmt.Errorf("IQR is nil")
+	}
+
+	if iqr.mode == invalidMode {
+		return fmt.Errorf("IQR.mode is invalid")
+	}
+
+	for cname := range iqr.knownValues {
+		if len(iqr.knownValues[cname]) != len(iqr.rrcs) && len(iqr.rrcs) != 0 {
```
nkunal, line 81: nit: not sure how go compiler does this under the hood but was thinking:
we should do `for cname, cvals := range ...` 
and also `lenOfCvals := len(cvals)` so that we can reuse these two vars in below checks and avoid the second map lookup len call.

nkunal: actually nvm, the second call for map lookup and len calculations happens only in error scenario

pkg/segment/query/iqr/intermediateQueryResult.go
```diff
@@ -0,0 +1,372 @@
...
+	return nil
+}
+
+func (iqr *IQR) AppendRRCs(rrcs []*utils.RecordResultContainer, segEncToKey map[uint16]string) error {
+	if len(rrcs) == 0 {
+		return nil
+	}
+
+	if err := iqr.validate(); err != nil {
+		log.Errorf("IQR.AppendRRCs: invalid state: %v", err)
```
nkunal, line 98: nit: should we call this `... validation failed: ...`

pkg/segment/query/iqr/intermediateQueryResult.go
```diff
@@ -0,0 +1,372 @@
...
+	cValRecords := toputils.TransposeMapOfSlices(records)
+	recordsAsAny := make([]map[string]interface{}, len(cValRecords))
+	for i, record := range cValRecords {
+		recordsAsAny[i] = make(map[string]interface{})
+		for key, value := range record {
+			recordsAsAny[i][key] = value.CVal
+		}
+	}
+
+	response := &pipesearch.PipeSearchResponseOuter{
```
nkunal, line 359: do we need to add a TODO for other response handler types like es bulk response handler , pipesearchWS response ?

nkunal (commented):
Looking fantastic, added a few comments

pkg/segment/query/iqr/intermediateQueryResult.go
AndrewHess: Yes I updated this to `for cname, values := range... ` because I think that's better. I didn't make the `lendOfValues` variable though because it only happens in the error case and `len` should be an O(1) operation anyways, so I doubt there'd be much/any performance change

pkg/segment/query/iqr/intermediateQueryResult.go
AndrewHess: For websocket searches I think we do need another one, so I added a TODO. I'm not sure what you meant for ES bulk; we'll only want to convert the IQR to a result for queries, and I thought we only had API queries and WS queries

nkunal: so we do have the es query response handler, this is mostly used for the kibana and ES api compatibility. But we have de-priroritized that. For now just putting a todo comment for es resp handler is good enough

nkunal (approved):

PR #1712, opened 2024-10-04
fix: fix slice init length

cuishuang:
# Description

The intention here should be to initialize a slice with a capacity of length rather than initializing the length of this slice. 

The online demo: https://go.dev/play/p/q1BcVCmvidW

# Testing

# Checklist:

### DIFF
--- a/pkg/ast/common.go
+++ b/pkg/ast/common.go
@@ -188,7 +188,7 @@ func createMatchPhraseFilterCriteria(k, v interface{}, opr LogicalOperator, nega
 	var matchWordsOriginal [][]byte
 	if cci != nil && cci.ShouldAlsoSearchWithOriginalCase() {
 		originalRtInput = strings.TrimSpace(cci.originalColValue.(string))
-		matchWordsOriginal = make([][]byte, len(matchWords))
+		matchWordsOriginal = make([][]byte, 0, len(matchWords))
 		for _, word := range strings.Split(originalRtInput, " ") {
 			matchWordsOriginal = append(matchWordsOriginal, [][]byte{[]byte(word)}...)
 		}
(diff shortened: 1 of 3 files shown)


pkg/segment/writer/metrics/tagstree.go
```diff
@@ -489,7 +489,7 @@ func (tree *TagTree) encodeTagsTree() ([]byte, error) {
 		idx += 4
 		sizeToAdd := int64(tagBuf.Len())
 		if sizeToAdd > 0 {
-			newArr := make([]byte, sizeToAdd)
+			newArr := make([]byte, 0, sizeToAdd)
 			dataBuf = append(dataBuf, newArr...)
```
AndrewHess, line 493: Also here I think we want what we had

pkg/segment/search/segsearch.go
```diff
@@ -164,7 +164,7 @@ func writePqmrFiles(segmentSearchRecords *SegmentSearchStatus, segmentKey string
 
 	sizeToAdd := len(bufEmpty)
 	if sizeToAdd > 0 {
-		newArr := make([]byte, sizeToAdd)
+		newArr := make([]byte, 0, sizeToAdd)
 		buf = append(buf, newArr...)
```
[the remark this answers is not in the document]
cuishuang: > Here I think we actually want the original version, since we just want to expand `buf`

Modified.

pkg/segment/writer/metrics/tagstree.go
cuishuang: Modified

cuishuang:
> Thanks for these fixes. I think most of them are correct, but a few of them were correct before. Also some unit tests are failing; you can run `make test` to run them all locally

Sorry.  I am not familiar with the project. Is there a problem with the single test itself?


If necessary, you can close this PR and it will be up to you to fix the issue I have discovered.

PR #1716, opened 2024-10-04
Handle multiple input streams on a DataProcessor

AndrewHess:
# Description
Some DataProcessors require multiple input streams. This PR handles fetching input from all streams and ensuring that the combined data is properly sorted.

DataProcessors typically need input data in a specific order. When multiple input streams are involved, each stream guarantees that its data is sorted. However, the DataProcessor must merge these sorted streams into a single, ordered stream.

A complication is that fetching data from a stream does not retrieve all its data at once (i.e., subsequent Fetch() calls can yield more data from that stream). So the DataProcessor must stop merging streams once it reaches the end of one stream. Continuing to merge without fetching more data from that stream could invalidate the order of the remaining part of the combined stream.

This raises another complication: since we don't want to fetch all of the data from all of the streams (as that would use a lot of RAM), and we have to stop the merge before processing all the data, we have to save the excess data we got for the other streams. This PR handles that with the `cachedStream` type.

# Testing
New unit tests

# Checklist:


pkg/segment/query/iqr/intermediateQueryResult.go
```diff
@@ -323,6 +324,166 @@ func (iqr *IQR) readColumnWithRRCs(cname string) ([]utils.CValueEnclosure, error
...
+		}
+
+		for cname := range iqr.knownValues {
+			if _, ok := result.knownValues[cname]; !ok {
+				result.knownValues[cname] = make([]utils.CValueEnclosure, 0)
+			}
+		}
+
+		if iqr.mode != result.mode {
+			return nil, fmt.Errorf("mergeMetadata: inconsistent modes")
```
nkunal, line 431: should we include these two different values in the fmt.

pkg/segment/query/iqr/intermediateQueryResult.go
```diff
@@ -323,6 +324,166 @@ func (iqr *IQR) readColumnWithRRCs(cname string) ([]utils.CValueEnclosure, error
...
+		}
+
+		if iqr.qid != result.qid {
+			return nil, fmt.Errorf("mergeMetadata: inconsistent qids")
```
nkunal, line 435: should we include these two different values in the fmt.

pkg/segment/query/iqr/intermediateQueryResult.go
```diff
@@ -323,6 +324,166 @@ func (iqr *IQR) readColumnWithRRCs(cname string) ([]utils.CValueEnclosure, error
...
+		if !reflect.DeepEqual(iqr.renamedColumns, result.renamedColumns) {
+			return nil, fmt.Errorf("mergeMetadata: inconsistent renamed columns")
+		}
+
+		if !reflect.DeepEqual(iqr.groupbyColumns, result.groupbyColumns) {
+			return nil, fmt.Errorf("mergeMetadata: inconsistent groupby columns")
+		}
+
+		if !reflect.DeepEqual(iqr.measureColumns, result.measureColumns) {
+			return nil, fmt.Errorf("mergeMetadata: inconsistent measure columns")
```
nkunal, line 451: same comment for all of the inconsistencies

nkunal (commented):
looks a good, a few minor nits

PR #1730, opened 2024-10-09
Add an index file that contains the functional Queries to run

vasukalariya:
# Description
Summarize the change.
Added an index file that contains the paths of functional query files to run.
Added 4 types of queries
- Normal stats query for all functions (currently implemented).
- Normal group by query for all functions (currently implemented) with 2 columns.
- 2 complex search queries.

There are 2 queries that are failing for multi-index:
- 9stats_groupby_2col.json
Root cause: `http_method=POST AND (app_name=Bracecould OR app_name=R*) | regex app_name=".*could"` , this previous search condition is returning incorrect results in multi-index.
- complex_search_2.json
Root cause: This query returns no results in its last query update call and so columnsOrder does not match (columns specified by fields are not in sorted order and the default order is sorted). I am not sure if we want the test tool to capture this and ignore it or do we want to ensure that the columnsOrder is same irrespective of the number of records fetched from each segment.

# Testing

# Checklist:


tools/sigclient/pkg/query/functional.go
```diff
@@ -18,28 +18,64 @@
...
+		if !strings.HasSuffix(queryFilePath, ".json") {
+			log.Fatalf("FunctionalTest: Invalid file format: %v. Expected .json", queryFilePath)
+		}
+
+		if !path.IsAbs(queryFilePath) {
+			queryFilePath = filepath.Join(baseDir, queryFilePath)
+		}
+
+		RunQuery(queryFilePath, qid, dest)
+		qid++
```
nkunal, line 71: after `Scan` call completes (aka it returns false and you exit from the for loop), you should call the err function exposed by the scanner to see if there was any error because of which it returned false.

tools/sigclient/pkg/query/functional.go
```diff
@@ -18,28 +18,64 @@
...
 
-	// validate JSON files
-	for _, file := range queryFiles {
-		if !strings.HasSuffix(file.Name(), ".json") {
-			log.Fatalf("FunctionalTest: Invalid file format: %v. Expected .json", file.Name())
+	qid := 1
+	// Read line by line
+	for scanner.Scan() {
+		queryFilePath := strings.TrimSpace(scanner.Text())
+		if queryFilePath == "" {
```
nkunal, line 58: also add a check that if this var starts with a comment (#) then ignore this line and move to the next line

nkunal (commented):
looks good , just a couple of minor nits

tools/sigclient/pkg/query/functional.go
vasukalariya: addressed in the latest commit.

tools/sigclient/pkg/query/functional.go
vasukalariya: addressed in the latest commit.

Macbeth98 (approved):

PR #1744, opened 2024-10-11
move segmeta read/write to sep file, move out allpqsids, avoid cmi re-reads

nkunal:
# Description
This PR reduces the contention of the smrLock which is grabbed every time:
1. We add a segmeta entry
2. For every query, we add the pqsid into segmeta
3. on seg rotation
4. on seg deletion 

Changes Include:
1. Move all segmeta.json operations to a separate file
2. Move out AllPQSIDs and ColumnsNames into segment specific files.
3. Backfilling pqs ids now has been moved from segmeta.json into segment specific files
4. Introduce a new file type 
5. Don't clear cmi during search time
6. Don't re-read the cmi if the blk/cname was already present
7. Delete segment-validity file last

# Testing

# Checklist:


nkunal:
@AndrewHess this is ready for review

pkg/segment/writer/segmetarw.go
```diff
@@ -0,0 +1,552 @@
...
+	if err != nil {
+		if errors.Is(err, os.ErrNotExist) {
+			// for first time during bootup this will occur
+			_, err := os.OpenFile(localSegmetaFname, os.O_WRONLY|os.O_CREATE|os.O_TRUNC, 0644)
+			if err != nil {
+				log.Errorf("initSmr: failed to open a new filename=%v: err=%v", localSegmetaFname, err)
+				return
+			}
+		}
+		return
```
AndrewHess, line 58: Same here: line 57 should be:
```
} else {
// log the error
}

File: pkg/segment/writer/segmetarw.go [hunk 2]
```diff
@@ -0,0 +1,552 @@
...
+	smrLock.RLock()
+	retVal, err := getAllSegmetas(smFilename)
+	smrLock.RUnlock()
+	if err != nil {
+		log.Errorf("ReadSegmeta: getallsegmetas err=%v ", err)
+	}
+	return retVal
+}
+
+// read only the current nodes segmeta
```
AndrewHess, line 77: nit
```suggestion
// read only the current node's segmeta
```

pkg/segment/writer/segmetarw.go
```diff
@@ -0,0 +1,552 @@
...
+		return nil, err
+	}
+	return sfm, nil
+}
+
+func writeSfm(segkey string, sfmData *structs.SegFullMeta) {
+
+	// create a separate individual file for SegFullMeta
+	sfmFname := getSegFullMetaFnameFromSegkey(segkey)
+	sfmFd, err := os.OpenFile(sfmFname, os.O_WRONLY|os.O_CREATE|os.O_TRUNC, 0644)
```
AndrewHess, line 136: Do you need to get a write lock for this file? Is a lock not needed because this only happens on rotation and rotation only happens once? Is `readSfm` guaranteed to only happen after `writeSfm` (if it happens concurrently, `readSfm` might get partial JSON which it can't decode)

pkg/segment/query/pqs/pqs.go
```diff
@@ -23,80 +23,48 @@ import (
...
-	segKey       string // segment key
-	tableName    string // table name for segKey
-	pqid         string // persistent query id
-	pqmrFilePath string // raw file path for pqmr file
-}
-
-// segKey -> [qid -> PersistentQueryResults]
-var allPersistentQueryResults map[string]map[string]*PersistentQueryResults
+// segKey -> [qid] -> bool
+var allPersistentQueryResults map[string]map[string]bool
```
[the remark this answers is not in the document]
nkunal: I wanted this to be a struct, but this map is merged with the AllPQIDs map which uses a bool. The allPQIDs are stored as bool in segmeta.json (version prior to this PR). So in order to maintain the backwards compatibility, keeping it as  a bool. 
Eventually we could use a struct and have the backwards compatibility handled cleanly. We can do that in a separate PR

pkg/segment/query/pqs/pqs.go
```diff
@@ -23,80 +23,48 @@ import (
...
-}
-
 func GetAllPersistentQueryResults(segKey string, pqid string) (*pqmr.SegmentPQMRResults, error) {
 
-	pqResults, err := getPQResults(segKey, pqid)
-	if err != nil {
-		return nil, err
+	isPqidPresent := DoesSegKeyHavePqidResults(segKey, pqid)
+	if isPqidPresent {
+		return nil, errors.New("segKey does not have any persistent query results")
```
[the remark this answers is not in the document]
nkunal: yeah I don't like this either. But the rest of the code (in search/query) has no clean way of deciding between nopqid present and an actual error. Right now the error is treated as (either an error occured or a pqid not present), in either case there is no pqid is present and the caller just continues on.

pkg/segment/writer/segmetarw.go
nkunal: `writeSfm` only happens once when we rotate the segment, until then the readSfm cannot be called because this segment has not been rotated yet (and it is not present in segmeta.json) so the reader/populateMicroIndices code will not see this seg hence they can't call the readSfm func

nkunal:
all comments have been addressed, please re-review @AndrewHess

AndrewHess (approved):
Looks good

PR #1745, opened 2024-10-11
Functional test Validate grouping on time

Macbeth98:
# Description
- For commands, like `timechart` that groups on timestamp, the timestamp values will keep changing depending on when we ingest the data. 
- So we actually cannot verify these group by values. But the result of each of these group by values should be the same and the buckets that are created should also be same.
- So this PR made changes to verify these scenarios.

# Testing
- Tested with an existing it works sometimes.
- Currently the timechart implementation is not giving consistent results.

# Checklist:

### DIFF
--- a/tools/sigclient/functionalQueries/functionalQueries.yml
+++ b/tools/sigclient/functionalQueries/functionalQueries.yml
@@ -1,24 +1,23 @@
-3stats.json
-9stats.json
-complex_search.json
-eval_groupby_2col_2stats.json
-grpby_1col_4stats.json
-search_1col_fields.json
-rare.json
-sort_tail_eval_newField.json
-stats_groupby_with_eval.json
-streamstats_win3.json
-streamstats_win4_currF.json
-streamstats_win4_grpby.json
-streamstats_win4_reset_before_after.json
-streamstats_win2_gloF_currF_grpby.json
-streamstats_win2_gloF_grpby.json
-streamstats_multi_resets_grpby.json
-streamstats_eval.json
-streamstats_after_stats.json
-streamstats_after_grpby.json
-multi_value.json
-search_timechart_no_groupby.json
+# 3stats.json
+# 9stats.json
+# complex_search.json
+# eval_groupby_2col_2stats.json
+# grpby_1col_4stats.json
+# search_1col_fields.json
+# rare.json
+# sort_tail_eval_newField.json
+# stats_groupby_with_eval.json
+# streamstats_win3.json
+# streamstats_win4_currF.json
+# streamstats_win4_grpby.json
+# streamstats_win4_reset_before_after.json
+# streamstats_win2_gloF_currF_grpby.json
+# streamstats_win2_gloF_grpby.json
+# streamstats_multi_resets_grpby.json
+# streamstats_eval.json
+# streamstats_after_stats.json
+# streamstats_after_grpby.json
+# multi_value.json
 
 ## Queries not working for multi-index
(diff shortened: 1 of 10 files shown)


tools/sigclient/functionalQueries/functionalQueries.yml
```diff
@@ -1,23 +1,23 @@
...
+# streamstats_win4_currF.json
+# streamstats_win4_grpby.json
+# streamstats_win4_reset_before_after.json
+# streamstats_win2_gloF_currF_grpby.json
+# streamstats_win2_gloF_grpby.json
+# streamstats_multi_resets_grpby.json
+# streamstats_eval.json
+# streamstats_after_stats.json
+# streamstats_after_grpby.json
+# multi_value.json
```
vasukalariya, line 20: Why have we commented these out, are these not working?

tools/sigclient/pkg/query/validation.go
```diff
@@ -597,36 +602,60 @@ func ValidateStatsQueryResults(queryRes *Result, expRes *Result) error {
...
+	}
+
 	for idx, expMeasureRes := range expRes.MeasureResults {
 		var err error
 		queryMeasureRes := queryRes.MeasureResults[idx]
 
-		equal := reflect.DeepEqual(queryMeasureRes.GroupByValues, expMeasureRes.GroupByValues)
-		if !equal {
-			return fmt.Errorf("ValidateStatsQueryResults: GroupByCombination mismatch, expected: %+v, got: %+v", expMeasureRes.GroupByValues, queryMeasureRes.GroupByValues)
+		if len(queryMeasureRes.GroupByValues) != len(expMeasureRes.GroupByValues) {
```
vasukalariya, line 615: Do we need the length check here? Wouldn't reflect.DeepEqual check that?

tools/sigclient/pkg/query/validation.go
```diff
@@ -597,36 +602,60 @@ func ValidateStatsQueryResults(queryRes *Result, expRes *Result) error {
...
+				if !equal {
+					return fmt.Errorf("ValidateStatsQueryResults: MeasureVal mismatch for key: %v, expected: %v, got: %v", key, value, actualValue)
 				}
-			} else {
-				equal = reflect.DeepEqual(actualValue, value)
 			}
+		} else {
+			// if group by values are not verified, then the bucket keys may not be consistent.
+			// So, we will only verify the measure values.
+			expMeasureSlice := utils.MapToSlice(expMeasureRes.MeasureVal)
```
vasukalariya, line 653: So basically, we are just checking the agg/measure names in `MeasureVal`, but the actual values are skipped because of inconsistency in timestamps due to different execution times?

vasukalariya:
Looks good, just minor nit

tools/sigclient/functionalQueries/functionalQueries.yml
Macbeth98: No they are working. I forgot to uncomment them.

tools/sigclient/pkg/query/validation.go
Macbeth98: yes it will check, but I added this as we want to check the length of the `groupByvalues` slice when we do not verify them.
may be I will add that check in the else condition.

tools/sigclient/pkg/query/validation.go
Macbeth98: Yeah I changed this. We don't need to do this.

vasukalariya (approved):

PR #1750, opened 2024-10-15
Add searcher for new query pipeline

AndrewHess:
# Description
Adds a `searcher` type responsible for doing raw search and emitting RRCs in time-sorted order.

# Testing
New unit tests cover most of the new functionality. Some parts (like verifying the correct blocks are retrieved, and that each block yields the correct RRCs) were not tested

# Checklist:


pkg/segment/query/processor/searcher.go
```diff
@@ -0,0 +1,366 @@
...
+	segutils "github.com/siglens/siglens/pkg/segment/utils"
+	toputils "github.com/siglens/siglens/pkg/utils"
+	log "github.com/sirupsen/logrus"
+)
+
+type block struct {
+	*structs.BlockSummary
+	*structs.BlockMetadataHolder
+	parentSSR *structs.SegmentSearchRequest
+	filename  string
```
nkunal, line 40: nit: what is this filename for ? is it a segkey filename or a csg filename or something else ?

nkunal (commented):
Looks good, one thing that is missing is we are not logging the qid in most of the logs statements that get called during query processing. I didn't add comment for each line. 
But we should look at each error/info log statement and if qid is available then we should log it according so that debugging concurrent queries becomes easier.

pkg/segment/query/processor/searcher.go
AndrewHess: I think it's the segkey filename, but I'm not 100% sure. It's needed to build the `map[string]SegmentSearchRequest` parameter to `ApplyFilterOperatorInternal()`. The new code in this PR doesn't directly use the filename; it's computed from `GetSSRsFromQSR()`, then saved to each of the blocks I make, and then read from the blocks to rebuild the `map[string]SegmentSearchRequest`.

Do you think I should rename it?

nkunal:
Yeah maybe rename it to segkeyFname for now. Later if the meaning of this field changes then we can change it

nkunal (approved):

PR #1756, opened 2024-10-16
Listen and Process all PQS Backfill and Empty Results through one go-routine

Macbeth98:
# Description
- There is a channel that is created through which the other functions/go routines can input the PQS or Backfill requests.
- There is a function called: `listenBackFillAndEmptyPQSRequests` that will listen to this channel and process the requests every 10 seconds or when the channel size reached 100.

# Testing
- Tested by executing the queries and verified that the segmeta entry and empty pqmeta are being created/updated.

# Checklist:

### DIFF
--- a/pkg/segment/writer/segmetarw.go
+++ b/pkg/segment/writer/segmetarw.go
@@ -25,6 +25,8 @@ import (
 	"os"
 	"path"
 	"sync"
+	"sync/atomic"
+	"time"
 
 	"github.com/siglens/siglens/pkg/config"
 	"github.com/siglens/siglens/pkg/hooks"
@@ -40,6 +42,14 @@ var smrLock sync.RWMutex = sync.RWMutex{}
 var localSegmetaFname string
 
 var SegmetaFilename = "segmeta.json"
+
+type PQSChanMeta struct {
+	pqid     string
+	segKey   string
+	emptyPqs bool
+}
+
+var pqsChan = make(chan PQSChanMeta, 1000)
 
 func initSmr() {
 
@@ -60,6 +70,9 @@ func initSmr() {
 		return
 	}
 	fd.Close()
+
+	// start a go routine to listen on the channel
+	go listenBackFillAndEmptyPQSRequests()
 }
 
 func getSegFullMetaFnameFromSegkey(segkey string) string {
@@ -459,8 +472,7 @@ func removeSegmetas(segkeysToRemove map[string]struct{}, indexName string) map[s
 	return segbaseDirs
 }
 
-func BackFillPQSSegmetaEntry(segkey string, newpqid string) {
-
+func BulkBackFillPQSSegmetaEntries(segkey string, pqidMap map[string]bool) {
 	sfmData, err := readSfm(segkey)
 	if err != nil {
 		return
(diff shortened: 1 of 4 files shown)


pkg/segment/writer/segmetarw.go
```diff
@@ -480,12 +492,107 @@ func BackFillPQSSegmetaEntry(segkey string, newpqid string) {
...
+
+func listenBackFillAndEmptyPQSRequests() {
+	// Listen on the channel, every 10 seconds or if the size of the channel is 100,
+	// it would get all the data in the channel and then do the process of Backfilling PQMR files.
+	// This is to avoid multiple writes to the same file
+
+	ticker := time.NewTicker(10 * time.Second) // every 10 seconds
+	defer ticker.Stop()
+
+	buffer := make([]PQSChanMeta, 0, 100)
```
AndrewHess, line 519: 10 seconds and 100 should probably be constants at the top of this file

AndrewHess (commented):
Looks pretty good

pkg/segment/writer/segmetarw.go
```diff
@@ -41,6 +43,14 @@ var localSegmetaFname string
 
 var SegmetaFilename = "segmeta.json"
 
+type PQSChanMeta struct {
+	pqid     string
+	segKey   string
+	emptyPqs bool
+}
+
+var pqsChan = make(chan PQSChanMeta, 1000)
```
nkunal, line 52: make this a const PQS_CHAN_SIZE

pkg/segment/writer/segmetarw.go
```diff
@@ -480,12 +492,107 @@ func BackFillPQSSegmetaEntry(segkey string, newpqid string) {
...
+func listenBackFillAndEmptyPQSRequests() {
+	// Listen on the channel, every 10 seconds or if the size of the channel is 100,
+	// it would get all the data in the channel and then do the process of Backfilling PQMR files.
+	// This is to avoid multiple writes to the same file
+
+	ticker := time.NewTicker(10 * time.Second) // every 10 seconds
+	defer ticker.Stop()
+
+	buffer := make([]PQSChanMeta, 0, 100)
+	var processing int32
```
nkunal, line 520: nit: name this var a little better

pkg/segment/writer/segmetarw.go
nkunal, line 519: why is this called Meta ?

pkg/segment/writer/segmetarw.go
```diff
@@ -480,12 +492,107 @@ func BackFillPQSSegmetaEntry(segkey string, newpqid string) {
...
+	isProcessing := func() bool {
+		return atomic.LoadInt32(&processing) == 1
+	}
+
+	callProcessBackFillPQSRequests := func() {
+		atomic.StoreInt32(&processing, 1)
+		bufferCopy := make([]PQSChanMeta, len(buffer))
+		copy(bufferCopy, buffer)
+		buffer = buffer[:0]
+		go func(pqsRequests []PQSChanMeta) {
```
nkunal, line 531: we shouldn't need a go routine here, once we have accumulated 100 (btw make this a const and not a hard coded value). Then we call some function that is going to take this 100 things and call the process... function

pkg/segment/writer/segmetarw.go
```diff
@@ -480,12 +492,107 @@ func BackFillPQSSegmetaEntry(segkey string, newpqid string) {
...
+			atomic.StoreInt32(&processing, 0)
+		}(bufferCopy)
+	}
+
+	for {
+		select {
+		case pqsChanMeta := <-pqsChan:
+			buffer = append(buffer, pqsChanMeta)
+			if len(buffer) >= 100 && !isProcessing() {
+				callProcessBackFillPQSRequests()
```
nkunal, line 542: see comment above, we should accumulate upto x things and then call the process func in sync, and then return back to the select loop.

pkg/segment/writer/segmetarw.go
```diff
@@ -480,12 +492,107 @@ func BackFillPQSSegmetaEntry(segkey string, newpqid string) {
...
+		}
+	}
+}
+
+func processBackFillAndEmptyPQSRequests(pqsRequests []PQSChanMeta) {
+	if len(pqsRequests) == 0 {
+		return
+	}
+
+	segKeyPqidMap := make(map[string]map[string]bool)
```
nkunal, line 557: explain with a comment what each of this map is for e.g `map[segKey] => map[blah] => bool`

pkg/segment/writer/segmetarw.go
```diff
@@ -480,12 +492,107 @@ func BackFillPQSSegmetaEntry(segkey string, newpqid string) {
...
+		}
+	}
+
+	var wg sync.WaitGroup
+	wg.Add(2)
+
+	go func() {
+		defer wg.Done()
+
+		for segKey, pqidMap := range segKeyPqidMap {
```
nkunal, line 580: nit: `allPQIDs` is a better name for `pqidMap` , it aligns with the name that we have in `SegMeta` struct

pkg/segment/writer/segmetarw.go
Macbeth98: I named it Meta because `segkey` and `pqid`, we are adding this information segment Meta and pqid empty meta.

pkg/segment/writer/segmetarw.go
nkunal: nit: 
1)we are calling a one-liner function `callProcessBackFillPQSRequests` and resetting the buf len. We can just call the `processBackFillAndEmptyPQSRequests` directly

2) also do we need a append call here ? we could use a counter and pass the counter to the processBackfill func, so that it knows what is the length. But if you like append, you can keep that

pkg/segment/writer/segmetarw.go
```diff
@@ -480,12 +494,98 @@ func BackFillPQSSegmetaEntry(segkey string, newpqid string) {
...
+	// segKey -> pqid -> true
+	segKeyPqidMap := make(map[string]map[string]bool)
+	// pqid -> segKey -> true
+	pqidSegKeyMap := make(map[string]map[string]bool) // for empty PQS
+
+	for _, pqsRequest := range pqsRequests {
+		if _, ok := segKeyPqidMap[pqsRequest.segKey]; !ok {
+			segKeyPqidMap[pqsRequest.segKey] = make(map[string]bool)
+		}
+		segKeyPqidMap[pqsRequest.segKey][pqsRequest.pqid] = true
```
nkunal, line 557: shouldn't we be adding to the `segKeyPqidMap` only if `emptyPqs` is false ? 
also I prefer the maps to be called as what they mean, for e.g. one map means you are writing `this segment has this valid pqs data` while the other means `this pqs id contains these segments which have no-matching results or emptyPQS data`.

Macbeth98: No We would add to `segKeyPqidMap` irrespective of whether `emptyPQS` is set to `true` or `false`. As this `segKeytoPQID` map is the one which will be updated per segment in the `sfm` file. So that one will add all the `Non-empty` and `Empty` PQIDs.

So the first map contains all the PQIDs for that segment (empty or non-empty). The second map is a map from PQID to seg key which represents the empty segments for a PQID.

I will try to add some better names.

pkg/segment/writer/segmetarw.go
Macbeth98: 1. Yes, we could just call there instead of calling `callProcessBackFillPQSRequests`. I have made this function because, I was first trying to synchronize to keep the async of receiving data from the channel and made it a non-blocking one. So there was similar code in both the cases and it made sense for me to just make this `callProcessBackFillPQSRequests` internal function which both the cases will call.
   - But when I changed it to blocking one, I still kept this because, we will still be rewriting the reset of buffer line in both cases and may be if we add some other functionality then it might be better to have this single function that handles the process. 
   - But I will change it to how you have suggested for better readability.

2. Yes, that makes sense. I used append, and initially created the slice with the capacity of `100` so that the buffer could also handle more than `100` items during the non-blocking version. But since, it is a blocking one, I could change to the way you have suggested.

nkunal (approved):

PR #1759, opened 2024-10-17
fix: Writing PQMR file and Tests based on Mocking of Col Segment Files

Macbeth98:
# Description
- PR: #1756 needs to be merged before this. 
- This PR fixes the issues of not writing PQMR files and retrieving the PQMR files.
- Fixes all the test cases that depend on the `WriteMockColSegFile` function. The tests started failing as they depend on the `segBasedir` to have the format `final/index/stream-id/suffix/`

# Testing
- All uint test cases Pass

# Checklist:


pkg/segment/query/metadata/metareader_test.go
```diff
@@ -36,7 +36,7 @@ func Test_readWriteMicroIndices(t *testing.T) {
 	segKey := segDir + "test"
 	blockSummariesFile := structs.GetBsuFnameFromSegKey(segKey)
 	numBlocks := 10
-	_, blockSummaries, _, _, allBmh, allColsSizes := writer.WriteMockColSegFile(segKey, numBlocks, 30)
+	_, blockSummaries, _, _, allBmh, allColsSizes := writer.WriteMockColSegFile(segKey, segKey, numBlocks, 30)
```
nkunal, line 39: in line 34, rename `segDir` to be `dataDir`
and have a specific segBaseDir param, which is one level higher than segkey, otherwise someone else copies this code and may have wrong assumptions

nkunal (approved):
looking great, one minor comment, can be addressed in this or next PR

PR #1763, opened 2024-10-17
Add gentimes command to new query pipeline

vasukalariya:
# Description
Summarize the change.
With the new query pipeline, the following commands work
```
| gentimes start=04/04/2021 end=04/07/2021
| gentimes start=08/10/2022 end=08/11/2022 increment=5h
| gentimes start=12/01/2021:16:11:56 end=12/05/2021:12:00:01 increment=3m
| gentimes start=-2 increment=5h
| gentimes start=-30 end=-20 increment=7s
```

Gentimes is not made to use UTC because [here](https://docs.splunk.com/Documentation/Splunk/9.2.1/SearchReference/Gentimes). They have mentioned that default end time is based on local time, so my best guess is that they use local time.
Also, they are displaying the time zone diff like this, so it should be local time in my opinion. Let me know if we need to change it.
image
 I have not modified parser because we can just hook the same parsed output when creating aggNodes [here](https://github.com/siglens/siglens/blob/7d93e8629691712c88ddc570ea2863beda936cf9/pkg/ast/pipesearch/searchQueryParser.go#L539). When we migrate the changes we can update the parser cause we will need to cleanup later anyways.

I think Rewind is use to generate the rows again, so I am setting the currStartTime to what was parsed, so that we can generate the same data again. But, I am not sure how to test this, since stats, bin, etc. or commands requiring all of the data are not yet implemented.

# Testing

# Checklist:


pkg/segment/query/processor/queryprocessor.go
```diff
@@ -64,12 +64,12 @@ func NewQueryProcessor(firstAgg *structs.QueryAggregators, queryInfo *query.Quer
...
-
+		dataProcessor.qid = searcher.qid
 		dataProcessors = append(dataProcessors, dataProcessor)
 	}
 
 	// Hook up the streams (searcher -> dataProcessors[0] -> ... -> dataProcessors[n-1]).
-	if len(dataProcessors) > 0 {
+	if len(dataProcessors) > 0 && !dataProcessors[0].IsGenerateDataProcessor() {
 		dataProcessors[0].streams = append(dataProcessors[0].streams, NewCachedStream(searcher))
 	}
```
AndrewHess, line 74: Why don't you want to hook up the dataProcessor in this case?

pkg/segment/query/processor/queryprocessor.go
```diff
@@ -112,6 +112,11 @@ func newQueryProcessorHelper(queryType structs.QueryType, input streamer,
 	}, nil
 }
 
+func (dp *DataProcessor) IsGenerateDataProcessor() bool {
+	_, ok := dp.processor.(*gentimesProcessor)
+	return ok
+}
```
AndrewHess, line 118: This isn't the right file to add this method. It should be in dataprocessor.go by the other DataProcessor methods if you need this function

AndrewHess (commented):
The actual gentimes implementation looks good; I added some comments about how you're hooking it up though

pkg/segment/query/processor/dataprocessor.go
```diff
@@ -232,9 +237,23 @@ func NewFillnullDP(options *structs.FillNullExpr) *DataProcessor {
...
+	if options.Interval == nil {
+		options.Interval = &structs.SpanLength{
+			Num:       1,
+			TimeScalr: segutils.TMDay,
+		}
+	}
+	if options.Interval.Num < 0 {
+		options.StartTime, options.EndTime = options.EndTime, options.StartTime
+		options.Interval.Num = -options.Interval.Num
+	}
```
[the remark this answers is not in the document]
vasukalariya: addressed in the latest commit.

pkg/segment/query/iqr/intermediateQueryResult.go
```diff
@@ -74,6 +74,10 @@ func NewIQR(qid uint64) *IQR {
 	}
 }
 
+func (iqr *IQR) GetQid() uint64 {
+	return iqr.qid
+}
+
```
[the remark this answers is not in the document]

pkg/segment/query/processor/queryprocessor.go
vasukalariya: addressed in the latest commit.

nkunal:
I am guessing we already have a E2E test for this ? if not we should add one

pkg/segment/query/processor/dataprocessor.go
```diff
@@ -109,6 +108,15 @@ func (dp *DataProcessor) Fetch() (*iqr.IQR, error) {
 	}
 }
 
+func (dp *DataProcessor) IsGenerateDataProcessor() bool {
```
AndrewHess, line 111: nit: rename to something like `IsGenerator` or `IsDataGenerator` (sorry I missed this on my first pass)

AndrewHess (commented):
Looks good; minor nit

pkg/segment/query/processor/dataprocessor.go
vasukalariya: addressed in the latest commit.

AndrewHess (approved):

PR #1765, opened 2024-10-18
fillnull with fields command with New Query pipeline

Macbeth98:
# Description
- Implemented the `fillnull` command with fields through the new query pipeline.
- Added `createdColumns` struct in `IQR` to let the query processor know about the newly created columns.
- Made changes when retrieving the all column values, so that if `knownValues` has these values, then the data would be taken from there.
- Modified the `BatchProcess` to also return errors.

# Testing
- Tested by running fillnull command with fields (one existing column and one non-existing column: should create that column).
- Ensured that the fill values are properly reflecting along with the creation of new column.

# Checklist:


pkg/segment/query/iqr/intermediateQueryResult.go
```diff
@@ -709,3 +719,47 @@ func (iqr *IQR) AsResult() (*structs.PipeSearchResponseOuter, error) {
...
+	if iqr.mode == notSet {
+		return nil
+	}
+
+	if _, ok := iqr.deletedColumns[cname]; ok {
+		return fmt.Errorf("IQR.AddNewCreatedColumn: column %s is deleted", cname)
+	}
+
+	return nil
+}
```
AndrewHess, line 765: I don't think this is used. If not, delete it

pkg/segment/query/iqr/intermediateQueryResult.go
```diff
@@ -709,3 +719,47 @@ func (iqr *IQR) AsResult() (*structs.PipeSearchResponseOuter, error) {
 
 	return response, nil
 }
+
+func (iqr *IQR) AddColumnWithDefaultValue(cname string, value utils.CValueEnclosure) error {
```
AndrewHess, line 723: nit: this method looks correct, but probably unnecessary because `AppendKnownValues` does this logic. If you want to keep this method, I'd recommend just doing the `validate()` and then creating the slice, then having `AppendKnownValues` do the rest. That way, there's less code for this IQR implementation, so there's fewer places things can go wrong.

P.S. you can use `toputils.ResizeSliceWithDefault()` on an empty slice to fill it with a default value.
```

File: pkg/utils/sliceutils_test.go [hunk 3]
```diff
@@ -91,21 +91,21 @@ func Test_BatchProcess(t *testing.T) {
...
 
-		return result
+		return result, nil
 	}
 
 	input := []int{1, 2, 3, 20, 42, 100, 47}
 	expected := []int{4, 5, 6, 21, 44, 101, 49}
 	expectedBatchSizes := []int{1, 2, 1, 3} // Batches should be 100s, 40s, 20s, 0s
-	actual := BatchProcess(input, batchingFunc, batchOrderingFunc, operation)
+	actual, _ := BatchProcess(input, batchingFunc, batchOrderingFunc, operation)
```
AndrewHess, line 108: Don't ignore the error in a unit test. Assert it's nil

pkg/segment/query/processor/fillnullcommand.go
```diff
@@ -18,16 +18,87 @@
...
+	"github.com/siglens/siglens/pkg/segment/utils"
+	toputils "github.com/siglens/siglens/pkg/utils"
+	log "github.com/sirupsen/logrus"
 )
 
 type fillnullProcessor struct {
 	options *structs.FillNullExpr
 }
 
+func getTheFinalFillValue(fillValue string, qid uint64) (utils.CValueEnclosure, error) {
```
AndrewHess, line 34: In this function, I think you make a DtypeEnclosure just to make a CValueEnclosure, right? Why not make the CValueEnclosure directly? We have a function for this already:
```
func (e *CValueEnclosure) ConvertValue(val interface{}) error {
```

pkg/segment/query/processor/fillnullcommand.go
```diff
@@ -18,16 +18,87 @@
...
+
+func performFillNullForTheFields(iqr *iqr.IQR, fields []string, cTypeFillValue utils.CValueEnclosure) {
+	fillNullForAllRecords := make(map[string]struct{})
+
+	for _, field := range fields {
+		values, err := iqr.ReadColumn(field)
+		if err != nil {
+			fillNullForAllRecords[field] = struct{}{}
+			continue
+		}
```
AndrewHess, line 63: nit: I'm not really sure why you're adding this to a map to handle slightly later instead of handling this column right now. But if you like this way better, feel free to keep it

pkg/segment/query/processor/fillnullcommand.go
```diff
@@ -18,16 +18,87 @@
...
+
+		for i, value := range values {
+			if value.IsNull() {
+				value.CVal = cTypeFillValue.CVal
+				value.Dtype = cTypeFillValue.Dtype
+				values[i] = value
+			}
+		}
```
AndrewHess, line 71: nit: I _think_ we'll get a little better performance for basically no cost by avoiding copying to `value` like this:
```
		for i := range values {
			if values[i].IsNull() {
				values[i].CVal = cTypeFillValue.CVal
				values[i].Dtype = cTypeFillValue.Dtype
			}
		}
```
Maybe the Go compiler already does this optimization, but I don't think there's any downside to doing it explicitly

AndrewHess (commented):
Looking pretty good

pkg/segment/query/iqr/intermediateQueryResult.go
Macbeth98: Yes, deleted.

AndrewHess (approved):

PR #1766, opened 2024-10-18
FillNull cmd with no fields implementation for New Query Pipeline

Macbeth98:
# Description
- Implemented `fillnull` cmd without fields
- Added a new function for `IQR` that just fetches the Columns.

# Testing
- Tested the code by ingesting across segments and verified that the `null` values are being replaced with the correct value
```
--data '{
    "state": "query",
    "searchText": "city=boston | fillnull value=UNKNOWN",
    "startEpoch": "now-7d",
    "endEpoch": "now",
    "indexName": "*",
    "from": 0,
    "queryLanguage": "Splunk QL"
}'
```

# Checklist:


pkg/segment/query/iqr/intermediateQueryResult.go
```diff
@@ -395,6 +395,28 @@ func (iqr *IQR) Append(other *IQR) error {
...
 
+func (iqr *IQR) GetColumns() (map[string]struct{}, error) {
+	if err := iqr.validate(); err != nil {
+		return nil, toputils.TeeErrorf("IQR.GetColumns: validation failed: %v", err)
+	}
+
+	segKey, ok := iqr.encodingToSegKey[iqr.rrcs[0].SegKeyInfo.SegKeyEnc]
+	if !ok {
+		return nil, toputils.TeeErrorf("IQR.readColumnWithRRCs: unknown encoding %v", iqr.rrcs[0].SegKeyInfo.SegKeyEnc)
+	}
```
AndrewHess, line 406: This is going to panic if the IQR has no RRCs

pkg/segment/query/iqr/intermediateQueryResult.go
```diff
@@ -395,6 +395,28 @@ func (iqr *IQR) Append(other *IQR) error {
...
+
+	vTable := iqr.rrcs[0].VirtualTableName
+
+	allColumns, err := record.GetColsForSegKey(segKey, vTable)
+	if err != nil {
+		return nil, err
+	}
```
AndrewHess, line 413: You're making an assumption here that all the RRCs in the IQR belong to the same segment. That's not true in general. An IQR is just a list of records—those records can come from anywhere. You may want to check the `iqr.encodingToSegKey` to get all the segkeys. However, while that map has all the segkeys used by the `iqr.RRCs`, it may have some segkeys which are not used by any of the RRCs.

Another potential option would be to iterate through the `iqr.RRCs` and get all the segkeys and VTable names

pkg/segment/query/processor/fillnullcommand.go
AndrewHess, line 1: This looks correct. Can you also add some unit tests? It might be hard to unit test with RRCs, but you should be able to test by making two or more IQRs and using `AppendKnownValues`, then pass those to `Process()` to get an IQR, and then `iqr.ReadAllColumns()` and checking the correct things got filled

AndrewHess (commented):
Looks pretty good. A couple issues

pkg/segment/query/processor/fillnullcommand_test.go
```diff
@@ -0,0 +1,178 @@
...
+func insertColumnsWithSomeNulls(t *testing.T, iqr *iqr.IQR, fillValue utils.CValueEnclosure, columnsToInsert []string, valuesCountToInsert int) (
+	map[string][]utils.CValueEnclosure, map[string][]utils.CValueEnclosure) {
+	knownValuesToInsert := make(map[string][]utils.CValueEnclosure)
+	knownValuesToExpect := make(map[string][]utils.CValueEnclosure)
+
+	var err error
+
+	for _, column := range columnsToInsert {
+		valuesToInsert := make([]utils.CValueEnclosure, valuesCountToInsert)
+		valuesToExpect := make([]utils.CValueEnclosure, valuesCountToInsert)
```
AndrewHess, line 40: `knownValuesToInsert` and `knownValuesToExpect` look like they're always the same by the end of this function. Am I missing something? Also the `valuesToInsert` and `valuesToExpect` look the same

pkg/segment/query/processor/fillnullcommand_test.go
```diff
@@ -0,0 +1,178 @@
...
+		knownValuesToExpect[column] = valuesToExpect
+	}
+
+	err = iqr.AppendKnownValues(knownValuesToInsert)
+	assert.NoError(t, err)
+
+	return knownValuesToInsert, knownValuesToExpect
+}
+
+func Test_fillNullCommandWithFields(t *testing.T) {
```
AndrewHess, line 70: nit: I found this unit test a little hard to understand. I think it's correct, so you can keep it as is, but FYI this is what I would have done:
1. Make the fillnull processor like you did
2. Hard code some IQRs with just NewIQR and then AppendKnownValues, passing in a hard coded map so it's really obvious what the IQR has.
3. Call Process() like you did to get an IQR
4. Use ReadColumn or ReadAllColumns on the IQR, and hard code what you expect for each value.

For example with the [sort command tests](https://github.com/siglens/siglens/blob/develop/pkg/segment/query/processor/sortcommand_test.go#L30) I did basically that, and I think that makes it really obvious that the asserts in the test are correct, and makes it obvious which edge cases I'm testing. In general for me personally, I'm much more lenient towards hardcoding and duplicating code in unit tests than I am in "real code" because I want unit tests to be obvious

AndrewHess (commented):
This can probably be merged; I just had one question

pkg/segment/query/processor/fillnullcommand_test.go
Macbeth98: This the code inside the loop

```
                         cValueToInsert := utils.CValueEnclosure{}
			var cValueToExpect utils.CValueEnclosure
			if insertNull {
				err = cValueToInsert.ConvertValue(nil)
				assert.NoError(t, err)
				cValueToExpect = fillValue
			} else {
				err = cValueToInsert.ConvertValue(fmt.Sprintf("value-%v", i))
				assert.NoError(t, err)
				cValueToExpect = cValueToInsert
			}

			valuesToInsert[i] = cValueToInsert
			valuesToExpect[i] = cValueToExpect
```

If `insertNull` is true, then `cValueToExpect` is assigned to the `fillValue` and `cValueToInsert` is set to `nil`. And if `insertNull` is same, then `cValueToExpect` and `cValueToInsert` are same.

When `fillNullProcessor.Process` is called, the `nil` will be filled by the `fillValue` and that is what I am expecting.

So No they would not end up same.

pkg/segment/query/processor/fillnullcommand_test.go
Macbeth98: Got it, next Time, I will just hard code some values.

AndrewHess (approved):

PR #1768, opened 2024-10-18
Add inputlookup to new query pipeline

vasukalariya:
# Description
Summarize the change.
- Added support for merging a withoutRRC IQR to withRRCIQR.
- Handle case for nil RRCs.
- Added unit test for creating an withRRC IQR from without RRC IQR.

This PR streamlines the handling of EOF. For all the commands, we should ensure that if input returned is `nil`, it should mean that they do not have any data to send, which is ideal. Based on that, the current data processor should act accordingly. It would be easier and simpler to return EOF when we expect to return output IQR as `nil`.

# Testing

Tested on the following commands
```
| inputlookup test_lookup.csv
| inputlookup max=6 test_lookup.csv where Test > 0
| inputlookup test_lookup.csv where Test < "a*"
| inputlookup test_lookup.csv.gz where Test = "a*"
| inputlookup start=1 max=5 test_lookup.csv where Test = "abc"
| inputlookup start=1 max=4 append=true test_lookup.csv where Test < 0
app_name=Bracecould | inputlookup start=5 append=true test_lookup.csv where Test = "abc"
app_name=Bracecould | inputlookup start=3 max=3 append=true test_lookup.csv where Test="abc" OR Department="e*" | inputlookup append=true test_lookup.csv.gz where Department="s*" AND Salary > 60000
```

Also, verified simple cases for where, sort and gentimes to ensure that this does not break previously implemented commands.

# Checklist:


pkg/segment/query/iqr/intermediateQueryResult.go
```diff
@@ -252,13 +282,19 @@ func (iqr *IQR) ReadColumn(cname string) ([]utils.CValueEnclosure, error) {
 func (iqr *IQR) readAllColumnsWithRRCs() (map[string][]utils.CValueEnclosure, error) {
 	// Prepare to call BatchProcessToMap().
 	getBatchKey := func(rrc *utils.RecordResultContainer) uint16 {
+		if rrc == nil {
+			return uint16(9999) // Default Segkey for nil RRCs
```
nkunal, line 286: make this a const. also why this specific number ?

pkg/segment/query/iqr/intermediateQueryResult.go
```diff
@@ -570,6 +618,29 @@ func mergeMetadata(iqrs []*IQR) (*IQR, error) {
...
+	err := newIQR.AppendRRCs(make([]*utils.RecordResultContainer, valuesLen), nil)
+	if err != nil {
+		return nil, fmt.Errorf("IQR.mergeWithoutRRCIntoRRCIQR: error appending RRCs: %v", err)
+	}
+	err = newIQR.AppendKnownValues(iqr.knownValues)
+	if err != nil {
+		return nil, fmt.Errorf("IQR.mergeWithoutRRCIntoRRCIQR: error appending known values: %v", err)
+	}
+
+	return newIQR, nil
```
nkunal, line 641: the func names in the fmt error statements is not correct. 
also the name of this func is a bit confusing, could you simplify this a bit

pkg/segment/query/processor/inputlookupcommand.go
```diff
@@ -0,0 +1,174 @@
...
+	processed uint64
+}
+
+func checkCSVFormat(filename string) bool {
+	return strings.HasSuffix(filename, ".csv") || strings.HasSuffix(filename, ".csv.gz")
+}
+
+func createRecord(columnNames []string, record []string) (map[string]utils.CValueEnclosure, error) {
+	if len(columnNames) != len(record) {
+		return nil, fmt.Errorf("createRecord: Column and record lengths are not equal, columnNames: %v, record: %v", columnNames, record)
```
nkunal, line 50: this will be a huge log statement, only print the lengths here

pkg/segment/query/processor/inputlookupcommand.go
```diff
@@ -0,0 +1,174 @@
...
+	}
+	recordMap := make(map[string]utils.CValueEnclosure)
+	for i, col := range columnNames {
```
nkunal, line 53: nit: typically whenever you are refrerring to columnname, be explicit about the var name, col is too generic and it can mean a few different things. 
In most of the code base we use something like 
`for, i, cname := range columnNames`

pkg/segment/query/processor/inputlookupcommand.go
```diff
@@ -0,0 +1,174 @@
...
+	}
+	filename := p.options.Filename
+
+	if !checkCSVFormat(filename) {
+		return nil, fmt.Errorf("inputlookupProcessor.Process: Only .csv and .csv.gz formats are currently supported")
+	}
+
+	filePath := filepath.Join(config.GetLookupPath(), filename)
+
+	file, err := os.Open(filePath)
```
nkunal, line 79: nit: typically, the return param of `Open` call is a file descriptor, we refer to it as `fd` in most of the codebase. suggest using the same. `file` var name is too generic

nkunal (commented):
overall looking good, a few nits.

pkg/segment/query/iqr/intermediateQueryResult.go
vasukalariya: addressed in the latest commit.

pkg/segment/query/processor/inputlookupcommand.go
vasukalariya: addressed in the latest commit.

pkg/segment/query/processor/inputlookupcommand.go
vasukalariya: addressed in the latest commit.

pkg/segment/query/processor/inputlookupcommand.go
vasukalariya: addressed in the latest commit.

pkg/segment/query/iqr/intermediateQueryResult.go
```diff
@@ -701,6 +775,22 @@ func (iqr *IQR) AsResult() (*structs.PipeSearchResponseOuter, error) {
...
 		}
+		numOfRecords := 0
+		for col, values := range records {
+			if numOfRecords == 0 {
+				numOfRecords = len(values)
+			}
+			if numOfRecords != len(values) {
+				return nil, fmt.Errorf("IQR.AsResult: inconsistent number of rows for column %v, expected: %v, got: %v", col, numOfRecords, len(values))
+			}
+		}
```
AndrewHess, line 786: If we want to do this validation, I think moving it into `readAllColumnsWithRRCs()` would be better

pkg/segment/query/iqr/intermediateQueryResult.go
```diff
@@ -351,6 +389,18 @@ func (iqr *IQR) Append(other *IQR) error {
 		return err
 	}
 
+	if iqr.mode == withRRCs && other.mode == withoutRRCs {
```
AndrewHess, line 392: You're handling only one of two cases. You also need to handle when iqr.mode is withoutRRCs and other.mode is withRRCs

pkg/segment/query/iqr/intermediateQueryResult_test.go
```diff
@@ -431,6 +431,73 @@ func Test_DiscardAfter(t *testing.T) {
...
+	assert.NoError(t, err)
+	numGeneratedCol := withoutRRCIqr.NumberOfRecords()
+	assert.Equal(t, 3, numGeneratedCol)
+
+	convertedRRCIQR, err := withoutRRCIqr.getRRCIQR()
+	assert.NoError(t, err)
+	assert.Equal(t, withRRCs, convertedRRCIQR.mode)
+	assert.Equal(t, 3, len(convertedRRCIQR.rrcs))
+	assert.Equal(t, knownValues, convertedRRCIQR.knownValues)
+	assert.Equal(t, 3, convertedRRCIQR.NumberOfRecords())
```
AndrewHess, line 472: I think this chunk is unrelated to the focus of `Test_MergeWithoutRRCIQRIntoRRCIQR`—I mean, I `convertedRRCIQR` is not used lower, and I don't think `getRRCIQR` has side-effects . If it is unrelated, it might make more sense to have this in a separate unit test.

pkg/segment/query/iqr/intermediateQueryResult_test.go
```diff
@@ -431,6 +431,73 @@ func Test_DiscardAfter(t *testing.T) {
...
+			} else {
+				assert.Equal(t, knownValues[col][i-numGeneratedCol], value)
+			}
+		}
+	}
+
+	knownValues["col1"] = knownValues["col1"][1:]
+	newIQR := NewIQR(0)
+	err = newIQR.AppendKnownValues(knownValues)
+	assert.Error(t, err)
```
AndrewHess, line 494: I might be misunderstanding. But this looks like you're just validating `AppendKnownValues` since the two columns have different numbers of records. So it's really unrelated to your test `Test_MergeWithoutRRCIQRIntoRRCIQR`, so it should be moved to a different test.

pkg/segment/query/processor/inputlookupcommand.go
```diff
@@ -0,0 +1,175 @@
...
+	}
+
+	if p.options == nil {
+		return nil, fmt.Errorf("inputlookupProcessor.Process: InputLookup is nil")
+	}
+	filename := p.options.Filename
+
+	if !checkCSVFormat(filename) {
+		return nil, fmt.Errorf("inputlookupProcessor.Process: Only .csv and .csv.gz formats are currently supported")
+	}
```
AndrewHess, line 76: nit: renaming to `isCSVFormat` or similar would make this more clear

pkg/segment/query/processor/inputlookupcommand.go
```diff
@@ -0,0 +1,175 @@
...
+	columnNames, err := reader.Read()
+	if err != nil {
+		return nil, fmt.Errorf("inputlookupProcessor.Process: Error reading column names, err: %v", err)
+	}
+
+	curr := uint64(0)
+	for curr < p.start {
+		_, err := reader.Read()
+		if err != nil {
+			if err.Error() == "EOF" {
```
AndrewHess, line 108:
```suggestion
			if err == io.EOF {
```

pkg/segment/query/processor/inputlookupcommand.go
```diff
@@ -0,0 +1,175 @@
...
+	"github.com/siglens/siglens/pkg/segment/utils"
+	putils "github.com/siglens/siglens/pkg/utils"
+)
+
+type inputlookupProcessor struct {
+	options   *structs.InputLookup
+	eof       bool
+	qid       uint64
+	start     uint64
+	processed uint64
```
AndrewHess, line 41:
```suggestion
	numProcessed uint64
```

pkg/segment/query/processor/inputlookupcommand.go
```diff
@@ -0,0 +1,175 @@
...
+	count := p.processed
+	records := map[string][]utils.CValueEnclosure{}
+
+	for !p.eof && count < putils.MinUint64(p.options.Max, utils.QUERY_EARLY_EXIT_LIMIT) {
+		count++
+		curr++
+		csvRecord, err := reader.Read()
+		if err != nil {
+			// Check if we've reached the end of the file
+			if err.Error() == "EOF" {
```
AndrewHess, line 126: Same here. Use `== io.EOF`

pkg/segment/query/processor/inputlookupcommand.go
```diff
@@ -0,0 +1,175 @@
...
+
+	// read columns from first row of csv file
+	columnNames, err := reader.Read()
+	if err != nil {
+		return nil, fmt.Errorf("inputlookupProcessor.Process: Error reading column names, err: %v", err)
+	}
+
+	curr := uint64(0)
+	for curr < p.start {
+		_, err := reader.Read()
```
AndrewHess, line 106: I think you're skipping to the `p.start` in the file in case you previously read only part of the file in a previous Process(); is that right? Is it possible to directly skip to the right position in the file using something like File.Seek()? If not, I think the code would be a little more clear if you move this logic to a helper function like `skipToStartPos()`

pkg/segment/query/processor/inputlookupcommand.go
AndrewHess, line 1: It would be good to add some unit tests for inputlookup

pkg/segment/structs/segstructs.go
```diff
@@ -211,6 +211,7 @@ type GenTimes struct {
 }
 
 type InputLookup struct {
+	FirstCommand         bool
```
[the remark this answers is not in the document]
vasukalariya: addressed in the latest commit.

pkg/segment/query/processor/dataprocessor.go
```diff
@@ -79,9 +79,7 @@ func (dp *DataProcessor) Fetch() (*iqr.IQR, error) {
 	for {
 		gotEOF := false
 		input, err := dp.getStreamInput()
-		if err == io.EOF {
-			gotEOF = true
-		} else if err != nil {
+		if err != nil && err != io.EOF {
 			return nil, utils.TeeErrorf("DP.Fetch: failed to fetch input: %v", err)
```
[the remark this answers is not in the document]
vasukalariya: Because of the EOF returned from getStreamInput, even if the data processor does not return an EOF, we would exit the fetch because when we getEOF we return `output` and `io.EOF` over [here](https://github.com/siglens/siglens/blob/b1384e1783390c93b0e1a9f65efd261900dad724/pkg/segment/query/processor/dataprocessor.go#L102).

pkg/segment/query/iqr/intermediateQueryResult.go
```diff
@@ -128,6 +141,19 @@ func (iqr *IQR) AppendRRCs(rrcs []*utils.RecordResultContainer, segEncToKey map[
 	return nil
 }
 
+func ValidateKnownValues(knownValues map[string][]utils.CValueEnclosure) error {
```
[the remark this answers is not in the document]

pkg/segment/query/iqr/intermediateQueryResult.go
```diff
@@ -701,6 +775,22 @@ func (iqr *IQR) AsResult() (*structs.PipeSearchResponseOuter, error) {
...
+				return nil, fmt.Errorf("IQR.AsResult: inconsistent number of rows for column %v, expected: %v, got: %v", col, numOfRecords, len(values))
+			}
+		}
+		// Populate knownValues
+		for cname, values := range iqr.knownValues {
+			if numOfRecords != len(values) {
+				return nil, fmt.Errorf("IQR.AsResult: inconsistent number of rows for column %v, expected: %v, got: %v", cname, numOfRecords, len(values))
+			}
+			records[cname] = values
+		}
```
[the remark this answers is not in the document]

pkg/segment/query/iqr/intermediateQueryResult.go
vasukalariya: This was just an extra validation which I don't think is required since, we would validate this while appending IQRs anyways. Have removed it.

pkg/segment/query/processor/inputlookupcommand.go
vasukalariya: addressed in the latest commit.

pkg/segment/query/processor/inputlookupcommand.go
vasukalariya: addressed in the latest commit.

pkg/segment/query/processor/inputlookupcommand.go
vasukalariya: addressed in the latest commit.

pkg/segment/query/processor/inputlookupcommand.go
vasukalariya: addressed in the latest commit.

pkg/segment/query/processor/inputlookupcommand.go
vasukalariya: addressed in the latest commit. Will check for the optimization later in next PR.

pkg/segment/query/processor/inputlookupcommand.go
vasukalariya: addressed in the latest commit. Added a basic test case, will add more in the next PR.

vasukalariya: added more tests in the latest commit.

pkg/segment/query/iqr/intermediateQueryResult_test.go
```diff
@@ -505,6 +505,21 @@ func Test_DiscardAfter(t *testing.T) {
 	assert.Equal(t, 1, iqr.NumberOfRecords())
 }
 
+func getKnownValues() map[string][]utils.CValueEnclosure {
```
AndrewHess, line 508: nit: rename to indicate this is only for testing

pkg/segment/query/processor/queryprocessor.go
```diff
@@ -80,12 +80,11 @@ func NewQueryProcessor(firstAgg *structs.QueryAggregators, queryInfo *query.Quer
...
 		}
-		dataProcessor.qid = searcher.qid
 		dataProcessors = append(dataProcessors, dataProcessor)
 	}
 
 	// Hook up the streams (searcher -> dataProcessors[0] -> ... -> dataProcessors[n-1]).
-	if len(dataProcessors) > 0 && !dataProcessors[0].IsDataGenerator() {
+	if len(dataProcessors) > 0 && !dataProcessors[0].CheckAndSetQidForDataGenerator(searcher.qid, true) {
 		dataProcessors[0].streams = append(dataProcessors[0].streams, NewCachedStream(searcher))
 	}
```
AndrewHess, line 89: nit: I think this change makes the code less clear. `CheckAndSetQidForDataGenerator` is probably doing too much. You might want to keep the check for `IsDataGenerator()` and then only call `CheckAndSetQidForDataGenerator` if necessary.

You can ignore this comment if you think doing it the way you are is the best approach

pkg/segment/query/processor/inputlookupcommand_test.go
```diff
@@ -0,0 +1,306 @@
...
+	"github.com/siglens/siglens/pkg/config"
+	"github.com/siglens/siglens/pkg/segment/query/iqr"
+	"github.com/siglens/siglens/pkg/segment/structs"
+	"github.com/siglens/siglens/pkg/segment/utils"
+	"github.com/stretchr/testify/assert"
+)
+
+func initTestConfig(t *testing.T) error {
+	runningConfig := config.GetTestConfig(t.TempDir())
+	runningConfig.DataPath = "inputlookup-data/"
```
AndrewHess, line 36: Why are you setting up the test config with a temp dir, but then immediately overriding the name to `inputlookup-data`? Ideally, just use the temp dir. If you need to hardcode the data dir instead, then you should add this code to make sure the dir gets deleted even if the test panics:
```
	t.Cleanup(func() { os.RemoveAll("inputlookup-data/") })
```

pkg/segment/query/processor/inputlookupcommand_test.go
```diff
@@ -0,0 +1,306 @@
...
+		actualCols, err := iqr.GetColumns()
+		assert.Nil(t, err)
+		assert.Equal(t, expectedCols, actualCols)
+
+		// Column values from 2 to 101 would be read
+		for col := range expectedCols {
+			colValues, err := iqr.ReadColumn(col)
+			assert.Nil(t, err)
+			assert.Equal(t, getExpectedColumnValues(col, start[fetch], numRecords[fetch]), colValues)
+		}
```
AndrewHess, line 157: I'm confused. Where are 2 and 101 coming from?

pkg/segment/query/processor/inputlookupcommand_test.go
```diff
@@ -0,0 +1,306 @@
...
+
+func Test_InputLookupMultipleFetchAndRewind(t *testing.T) {
+
+	err := initTestConfig(t)
+	assert.Nil(t, err)
+
+	data := []string{"a,b,c\n"}
+	for i := 0; i < 110; i++ {
+		idx := fmt.Sprintf("%v", i)
+		row := "a" + idx + ",b" + idx + ",c" + idx + "\n"
```
AndrewHess, line 172: nit: I think `Sprintf` for the whole `row` would be a little cleaner

Same comment elsewhere

pkg/segment/query/processor/inputlookupcommand_test.go
```diff
@@ -0,0 +1,306 @@
...
+	numRecords := []int{100, 8}
+
+	validate_MultiFetch(t, dp, expectedCols, start, numRecords, 2)
+
+	// Rewind to the start
+	dp.Rewind()
+
+	validate_MultiFetch(t, dp, expectedCols, start, numRecords, 2)
+
+	os.RemoveAll(config.GetDataPath())
```
AndrewHess, line 204: nit: It's better to defer this in a `t.Cleanup()` at the start of the test; if the test panics, it will not reach but the functions in `t.Cleanup()` will still get called

See https://pkg.go.dev/testing#T.Cleanup

pkg/segment/query/iqr/intermediateQueryResult.go
```diff
@@ -43,6 +43,8 @@ const (
 	withoutRRCs
 )
 
+const NIL_RRC_SEGKEY = uint16(9999)
```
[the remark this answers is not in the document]
vasukalariya: changed it to maxUint16 in the latest commit.

pkg/segment/query/iqr/intermediateQueryResult_test.go
vasukalariya: addressed in the latest commit.

pkg/segment/query/processor/queryprocessor.go
vasukalariya: addressed in the latest commit.

pkg/segment/query/processor/inputlookupcommand_test.go
vasukalariya: forgot to remove the comment, addressed in the latest commit.

pkg/segment/query/processor/inputlookupcommand_test.go
vasukalariya: addressed in the latest commit.

pkg/segment/query/processor/inputlookupcommand_test.go
vasukalariya: Just using the tempDir, so need of this. addressed in the latest commit.

AndrewHess (approved):

PR #1770, opened 2024-10-18
Add dedup without sort to new query pipeline

AndrewHess:
# Description
Implement [dedup](https://docs.splunk.com/Documentation/Splunk/9.1.1/SearchReference/Dedup) on the new query pipeline. The only option that doesn't work yet is `sortby`.

# Testing
Unit tests!

I also ran a few queries on the API:
```
curl -X POST -d '{
    "searchText": "* | sort limit=10000000 weekday | dedup 2 weekday",
    "indexName": "*",
    "startEpoch": "now-24h",
    "endEpoch": "now",
    "queryLanguage": "Splunk QL"
}' http://localhost:5122/api/search | python3 -m json.tool | less
```
And `* | dedup consecutive=true gender` and `* | dedup consecutive=false gender`. These all gave correct results

# Checklist:


pkg/segment/utils/segconsts.go
```diff
@@ -799,6 +799,55 @@ type CValueEnclosure struct {
...
+	switch e.Dtype {
+	case SS_DT_STRING:
+		return e.CVal.(string) == other.CVal.(string)
+	case SS_DT_BOOL:
+		return e.CVal.(bool) == other.CVal.(bool)
+	case SS_DT_UNSIGNED_NUM:
+		return e.CVal.(uint64) == other.CVal.(uint64)
+	case SS_DT_SIGNED_NUM:
+		return e.CVal.(int64) == other.CVal.(int64)
+	case SS_DT_FLOAT:
```
nkunal, line 816: for float do we need an exact equal or an approximate equal ?

nkunal (dismissed):

pkg/segment/utils/segconsts.go
AndrewHess: Good catch. Probably approximate equal since in general we could do some arithmetic to create/modify the float value

nkunal (dismissed):

nkunal (approved):

PR #1771, opened 2024-10-19
Stats cmd with GroupBy for RRCs implementation for New Query Pipeline

Macbeth98:
# Description
- The PR: #1766 needs to be merged first. 
- Now `stats with groupby` commands can be run on the new query pipeline when it is not the first aggregator/processor. 
- This PR is for the commands in the form: `<search> | ... | stats measure1, measure2, ..., measuren group by col1, col2, .., coln`

# Testing
- Tested by running the below queries.
```
--data '{
    "state": "query",
    "searchText": "* | fillnull value=UNKNOWN account_status | stats sum(http_status), count(eval(account_status=\"UNKNOWN\")), avg(http_status) by http_method, gender",
    "startEpoch": "now-2d",
    "endEpoch": "now",
    "indexName": "*",
    "from": 0,
    "queryLanguage": "Splunk QL"
}'
```

```
--data '{
    "state": "query",
    "searchText": "* | fillnull value=UNKNOWN account_status | where account_status=\"UNKNOWN\" | stats sum(http_status), count(http_status), avg(http_status) by http_method, gender",
    "startEpoch": "now-2d",
    "endEpoch": "now",
    "indexName": "*",
    "from": 0,
    "queryLanguage": "Splunk QL"
}'
```

# Checklist:


pkg/segment/results/blockresults/blockresult.go
```diff
@@ -563,13 +564,19 @@ func (gb *GroupByBuckets) ConvertToAggregationResult(req *structs.GroupByRequest
...
+		if req.IsBucketKeySeparatedByDelim {
+			bucketKey = strings.Split(key, string(utils.BYTE_TILDE))
+		} else {
+			bucketKey, err = utils.ConvertGroupByKey([]byte(key))
+			if len(bucketKey.([]string)) == 1 {
+				bucketKey = bucketKey.([]string)[0]
+			}
+			if err != nil {
+				log.Errorf("GroupByBuckets.ConvertToAggregationResult: failed to convert group by key: %v, err: %v", key, err)
+			}
```
AndrewHess, line 577: nit: reverse the `if`s so that you handle the `err` first, since that's how the rest of the codebase does it

pkg/segment/query/processor/statscommand.go
```diff
@@ -18,22 +18,151 @@
...
+	numOfRecords := inputIQR.NumberOfRecords()
+
+	if p.bucketKeyWorkingBuf == nil {
+		p.bucketKeyWorkingBuf = make([]byte, len(p.options.GroupByRequest.GroupByColumns)*utils.MAX_RECORD_SIZE)
+	}
+
+	if p.searchResults == nil {
+		p.options.GroupByRequest.BucketCount = int(utils.QUERY_MAX_BUCKETS)
+		p.options.GroupByRequest.IsBucketKeySeparatedByDelim = true
+		searchResults, err := segresults.InitSearchResults(uint64(numOfRecords), &structs.QueryAggregators{GroupByRequest: p.options.GroupByRequest}, structs.GroupByCmd, inputIQR.GetQID())
```
AndrewHess, line 91: nit: wrap this long line

pkg/segment/query/processor/statscommand.go
```diff
@@ -18,22 +18,151 @@
...
+	bucketHolderArr, measureFuncs, aggGroupByCols, _, bucketCount := p.searchResults.GetGroupyByBuckets(int(utils.QUERY_MAX_BUCKETS))
+
+	err := iqr.AppendStatsResults(bucketHolderArr, measureFuncs, aggGroupByCols, bucketCount)
+	if err != nil {
+		return nil, toputils.TeeErrorf("stats.Process.extractGroupByResults: cannot append stats results; err=%v", err)
+	}
+
+	p.logErrors()
+
+	return iqr, io.EOF
```
AndrewHess, line 161: Should you return an actual error instead of `EOF` if you got an error?

pkg/segment/query/processor/statscommand.go
AndrewHess, line 1: Now that stats works on data streamed in from an IQR, it would be good to add a unit test for it, if that will be fairly easy to add

AndrewHess (commented):
Looking pretty good

pkg/segment/query/processor/statscommand.go
Macbeth98: In that case, I should return as soon as I found an error. Instead of tracking them. 

But for Group by for example `col1`, `col2`. Col2 may not exist in the first segment or batch of results but it might exist in the next batch.

So for that reason, I am just tracking the errors and logging them. And backfilling values for the column that throws an error.

pkg/segment/query/processor/statscommand.go
Macbeth98: Added tests.

pkg/segment/query/processor/statscommand_test.go
```diff
@@ -0,0 +1,302 @@
...
+			MeasureCol:  "col2",
+			MeasureFunc: utils.Sum,
+		},
+		&structs.MeasureAggregator{
+			MeasureCol:  "col2",
+			MeasureFunc: utils.Avg,
+		},
+	}
+
+	groupByCols := []string{"col1", "col3"}
```
AndrewHess, line 78: Each value in `col3` is unique, so every bucket only has one element. That's a good test, but you should also have a test where multiple rows get aggregated into the same bucket

pkg/segment/query/processor/statscommand.go
AndrewHess: Ok, then should they be logged as Warn/Info? I think currently you're logging them as errors

pkg/segment/query/processor/statscommand_test.go
Macbeth98: I also have that version of test. Where I have verified that [here](https://github.com/siglens/siglens/pull/1771/files/c52fed5a6e74f531a150d72dd24796f24211f728..8bc6199e79c86e0380e3ce60f07b6d878378350a#diff-a26b93772dc6a4edcf317c1443b994b9eb1d2c38fee43b9717bc15c15a3cbadaR222-R223)

For the second iqr, I removed `col3` and executed the Process, which would make the RRCs fall into the same bucket because `col1` is there and `col3` is filled with default backfill value.

AndrewHess (approved):

PR #1772, opened 2024-10-19
Stats cmd Measure results for RRCs implementation for New Query Pipeline

Macbeth98:
# Description
- The PR #1771  needs to be merged first.
- Now `stats cmd with measure results` can be run with the new query pipeline.
- This PR is for the commands in the form: `<search> | ... | stats measure1, measure2, ... measuren`

# Testing
- Tested by running the below query and verified the same by running an equivalent query on the UI.
- Query run on UI: `* | stats sum(eval(http_status > 0)), count(http_status), sum(http_status), avg(http_status), sum(latency)`

```
--data '{
    "state": "query",
    "searchText": "* | where http_status > 0 | stats sum(eval(http_status > 0)), count(http_status), sum(http_status), avg(http_status), sum(latency)",
    "startEpoch": "now-2d",
    "endEpoch": "now",
    "indexName": "*",
    "from": 0,
    "queryLanguage": "Splunk QL"
}'
```

# Checklist:


pkg/segment/query/iqr/intermediateQueryResult.go
```diff
@@ -384,6 +389,28 @@ func (iqr *IQR) Append(other *IQR) error {
...
 }
 
+func (iqr *IQR) GetColumns() (map[string]struct{}, error) {
+	if err := iqr.validate(); err != nil {
+		return nil, toputils.TeeErrorf("IQR.GetColumns: validation failed: %v", err)
+	}
+
+	segKey, ok := iqr.encodingToSegKey[iqr.rrcs[0].SegKeyInfo.SegKeyEnc]
+	if !ok {
+		return nil, toputils.TeeErrorf("IQR.readColumnWithRRCs: unknown encoding %v", iqr.rrcs[0].SegKeyInfo.SegKeyEnc)
```
nkunal, line 399: func name not right

pkg/segment/query/processor/fillnullcommand.go
```diff
@@ -18,24 +18,123 @@
...
+			log.Errorf("performFillNullForTheFields: failed to append known values; err=%v", err)
+		}
+	}
+
+	for field := range fillNullForAllRecords {
+		err := iqr.AddColumnWithDefaultValue(field, cTypeFillValue)
+		if err != nil {
+			log.Errorf("performFillNullForTheFields: failed to add column with default value; err=%v", err)
+		}
+	}
```
nkunal, line 86: if qid is available in this method then log it in the log lines

pkg/segment/query/processor/statscommand.go
```diff
@@ -18,22 +18,323 @@
...
+
+	bucketHolderArr, retMFuns, aggGroupByCols, _, _ := p.searchResults.GetGroupyByBuckets(int(utils.QUERY_MAX_BUCKETS))
+
+	knownValues := make(map[string][]utils.CValueEnclosure)
+
+	for _, aggGroupByCol := range aggGroupByCols {
+		knownValues[aggGroupByCol] = make([]utils.CValueEnclosure, len(bucketHolderArr))
+	}
+	for _, retMFun := range retMFuns {
+		knownValues[retMFun] = make([]utils.CValueEnclosure, len(bucketHolderArr))
```
nkunal, line 174: create var for this len once before the two for loops and then reuse it

pkg/segment/query/processor/statscommand.go
```diff
@@ -18,22 +18,323 @@
...
+			}
+		}
+	}
+
+	err := iqr.AppendKnownValues(knownValues)
+	if err != nil {
+		return nil, toputils.TeeErrorf("stats.Process: cannot append known values; err=%v", err)
+	}
+
+	p.logErrors()
```
nkunal, line 200: make sure we do not log all the error log entries, (see comment in other PR). just log the first 5 entries.

pkg/segment/query/processor/statscommand.go
```diff
@@ -18,22 +18,323 @@
...
+		}
+		return nil, io.EOF
+	}
+
+	numOfRecords := uint64(inputIQR.NumberOfRecords())
+
+	if p.searchResults == nil {
+		searchResults, err := segresults.InitSearchResults(numOfRecords, &structs.QueryAggregators{MeasureOperations: p.options.MeasureOperations}, structs.SegmentStatsCmd, inputIQR.GetQID())
+		if err != nil {
+			return nil, toputils.TeeErrorf("stats.Process.processMeasureOperations: cannot initialize search results; err=%v", err)
```
nkunal, line 219: add qid , unless the caller is going to log the qid

pkg/segment/query/processor/statscommand.go
```diff
@@ -18,22 +18,323 @@
...
+			} else {
+				p.errorData.notSupportedStatsType[colName] = struct{}{}
+				continue
+			}
+		}
+	}
+
+	err := p.searchResults.UpdateSegmentStats(segStatsMap, p.options.MeasureOperations)
+	if err != nil {
+		log.Errorf("stats.Process: cannot update segment stats; err=%v", err)
```
nkunal, line 280: add qid

pkg/segment/query/processor/statscommand.go
```diff
@@ -18,22 +18,323 @@
...
+			err := knownValues[measureFunction][i].ConvertValue(value)
+			if err != nil {
+				p.errorData.convertToCValueErr[fmt.Sprintf("%v_%v", i, measureFunction)] = value
+			}
+		}
+	}
+
+	err := iqr.AppendKnownValues(knownValues)
+	if err != nil {
+		return nil, toputils.TeeErrorf("stats.Process.extractSegmentStatsResults: cannot append known values; err=%v", err)
```
nkunal, line 312: same comment for qid

pkg/segment/query/processor/statscommand.go
```diff
@@ -18,22 +18,323 @@
...
+	}
+
+	p.logErrors()
```
nkunal, line 315: same comment about not logging too many errors

pkg/segment/query/processor/statscommand.go
```diff
@@ -18,22 +18,323 @@
...
+	if len(p.errorData.cValueGetStringErr) > 0 {
+		log.Errorf("stats.Process: failed to get string from CValue: %v", p.errorData.cValueGetStringErr)
+	}
+
+	if len(p.errorData.notSupportedStatsType) > 0 {
+		log.Errorf("stats.Process: not supported stats type: %v", p.errorData.notSupportedStatsType)
+	}
+
+	if len(p.searchResults.AllErrors) > 0 {
+		log.Errorf("stats.Process: search results errors: %v", p.searchResults.AllErrors)
```
nkunal, line 338: this is where you could log only the first 5 elements

pkg/segment/reader/segread/multicolreader.go
```diff
@@ -160,6 +161,7 @@ func InitSharedMultiColumnReaders(segKey string, colNames map[string]bool, block
 		numReaders:      numReaders,
 		numOpenFDs:      maxOpenFds,
 		allFDs:          allFDs,
+		ErrorColMap:     make(map[string]error),
```
nkunal, line 164: nit: add comment about what is the key for this map : `map[blah] ==> error`

nkunal: this is an unbounded map, not sure what the key is here, but if it is a high cardinality value, then this map could grow high. Suggest adding a limiting whereever this var is used. 
Rather prefer to keep this var unexported and add a method to log the error. That way you can limit how many things can go into this map in one place

pkg/segment/utils/segconsts.go
```diff
@@ -916,6 +943,110 @@ func (e *CValueEnclosure) GetUIntValue() (uint64, error) {
...
+			return 0, nil
+		}
+	case SS_DT_UNSIGNED_NUM:
+		return int64(e.CVal.(uint64)), nil
+	case SS_DT_SIGNED_NUM:
+		return e.CVal.(int64), nil
+	case SS_DT_FLOAT:
+		return int64(e.CVal.(float64)), nil
+	default:
+		return 0, errors.New("CValueEnclosure GetIntValue: unsupported Dtype")
```
nkunal, line 975: add the value of what the unsupported dtype was

pkg/segment/query/processor/statscommand.go
Macbeth98: The errors that would be logged here are only for the columns, that are given as `groupByColumns` and the `measureColumns`

pkg/segment/query/processor/statscommand.go
```diff
@@ -18,45 +18,52 @@
...
 	log "github.com/sirupsen/logrus"
+	bbp "github.com/valyala/bytebufferpool"
 )
 
 type ErrorData struct {
-	// columnName -> error. Tracks errors while reading the column through iqr.Record.ReadColumn()
-	// This is logged as a warning. As the column may not exist in one segment but exist in another.
-	readColumns map[string]error
+	readColumns           map[string]error       // columnName -> error. Tracks errors while reading the column through iqr.Record.ReadColumn
+	cValueGetStringErr    map[string]interface{} // columnName -> error. Tracks errors while converting CValue to string
```
AndrewHess, line 38: You're saying the value is `error`; why do you map to a `interface{}`?

pkg/segment/query/processor/statscommand.go
```diff
@@ -159,14 +169,128 @@ func (p *statsProcessor) extractGroupByResults(iqr *iqr.IQR) (*iqr.IQR, error) {
...
+				stats.AddSegStatsStr(segStatsMap, colName, values[i].CVal.(string), p.byteBuffer, aggColUsage, hasValuesFunc, hasListFunc)
+			} else if values[i].IsNumeric() {
+				stringVal, err := values[i].GetString()
+				if err != nil {
+					p.errorData.cValueGetStringErr[colName] = err
+					stringVal = fmt.Sprintf("%v", values[i].CVal)
+				}
+
+				if values[i].IsFloat() {
+					stats.AddSegStatsNums(segStatsMap, colName, utils.SS_FLOAT64, 0, 0, values[i].CVal.(float64), stringVal, p.byteBuffer, aggColUsage, hasValuesFunc, hasListFunc)
```
AndrewHess, line 234: nit: wrap this long line

AndrewHess (commented):
Looks good; just a few nits

pkg/utils/maputils.go
```diff
@@ -268,3 +268,14 @@ func (twm *TwoWayMap[T1, T2]) GetMapForReading() map[T1]T2 {
...
+// AddIfNewEntryAndIncrementSize adds the key-value pair to the map if the key does not exist in the map.
+// If the key is added, the currentSize is incremented by 1.
+func AddIfNewEntryAndIncrementSize[K comparable, V any](m map[K]V, key K, value V, currentSize int) int {
+	if _, exists := m[key]; !exists {
+		currentSize++
+		m[key] = value
+	}
+
+	return currentSize
+}
```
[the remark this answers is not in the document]
Macbeth98: I first added this with the use case where I was tracking similar errors at different places within the same map. So for that I wanted to make sure that I at least have 5 of each of those errors. 
But later I changed the code and that use case was removed. But I still thought I can keep this. 
Shall I remove this?

pkg/segment/query/processor/statscommand.go
Macbeth98: Yes, I changed it. It will be updated in my next commit.

pkg/utils/maputils.go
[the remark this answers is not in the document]
AndrewHess: I'm not seeing a good use case for this. I see why you'd want to insert the value only if the key is not already in the map, but I don't know why you need this `currentSize`; if it's always the size of the map, then use `len()` instead, which will make the code using this more obvious; if `currentSize` would sometimes not be the size of the map, I'm just not able to come up with use cases where we'd want that.

Do you think we'll want this util later? If not, I'd say remove it. If you do think we'll want it later, just let me know and I'll approve this PR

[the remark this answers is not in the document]
Macbeth98: Yeah i will remove this for now.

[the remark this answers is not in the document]
Macbeth98: @AndrewHess I have removed util.

AndrewHess (approved):

PR #1773, opened 2024-10-20
StatGroupByCmd Query Type for New Query Pipeline

Macbeth98:
# Description
- Now the stats RRC `GroupByCmd` query type is supported with the new query pipeline.
- The searcher is hooked with the old query pipeline if it is `GroupByCmd` Query Type and once this is processed it goes to the new query pipeline to process further data processors.
- Adjusted the output to reflect the `aggs-query` response.
- This PR is for the commands in the form: `<search> | stats measure1, measure2, ..., measuren by col1, col2, ..., coln`
- This PR also enables the **Timechart** queries in the form `<search> | timechart measure by col`. But in the response, need to indicate `IsTimechart: true` so that the chart can be displayed.

# Testing
- Tested by running the below query with the new query pipeline and verified response by running the same query through UI.

```
curl --location 'http://localhost:5122/api/search' \
--header 'Content-Type: application/json' \
--data '{
    "state": "query",
    "searchText": "* | stats avg(http_status), sum(http_status), count(http_status) by http_method, gender",
    "startEpoch": "now-2d",
    "endEpoch": "now",
    "indexName": "*",
    "from": 0,
    "queryLanguage": "Splunk QL"
}'
```

# Checklist:

- [x] I have self-reviewed this PR.
- [x] I have removed all print-debugging and commented-out code that should not be merged.
- [x] I have added sufficient comments in my code, particularly in hard-to-understand areas.
- [x] I have formatted the code, if applicable. For Go, I have run `goimports -w .`.

Issue #519, opened 2024-02-21
central authentication for ui


thatsk:
# Description
Authentication for UI

# Benefits
It will be good with some editor and viewer , admin permission so that people can read, modify and update graphs.

# (Optional) Possible Implementation
https://dexidp.io/docs/connectors/
with the help of this it is easily adoptable for ldap or any auth mechanism

nkunal:
This is an enterprise feature

jar-work:
@nkunal I'm just starting to look at the codes, and [I found this method](https://github.com/siglens/siglens/blob/6884e6a330476401fbcec23e45120636d83dd3af/pkg/utils/httpserverutils.go#L799) that can be used for a basic authentication. There are even unit tests (the only references I found are the tests)

We are starting to use siglens for some of our systems, and we need a simple way to block the access to the UI using username and password. I played a bit with the codes and I implemented a simple middleware that handles the authorization:

```
func BasicAuthMiddleware(next fasthttp.RequestHandler) fasthttp.RequestHandler {
	return func(ctx *fasthttp.RequestCtx) {

		var _, user, password = config.GetBasicAuthConfig()  ---> this comes from the server.yaml file
		usernameHash := xxhash.Sum64String(user)
		passwordHash := xxhash.Sum64String(password)

		if !utils.VerifyBasicAuth(ctx, usernameHash, passwordHash) {
			ctx.Response.Header.Add("WWW-Authenticate", `Basic realm="Required username and password"`)
			ctx.Response.SetStatusCode(fasthttp.StatusUnauthorized)
			return
		}

		next(ctx)

	}
}
```
and a function to wrap the routes with that middleware (for now I only tested on top of the routes that require the `tracing.TraceMiddleware`):

```
	getMiddlewares := func(next fasthttp.RequestHandler, requiresTracing bool) fasthttp.RequestHandler {
		var nextToReturn = next

		if requiresTracing {  --> for now passed as parameter
			nextToReturn = tracing.TraceMiddleware(next)
		}

		var authenticationEnabled, _, _ = config.GetBasicAuthConfig() ----> from server.yaml

		if authenticationEnabled {
			nextToReturn = basicauth.BasicAuthMiddleware(nextToReturn)
		}

		return nextToReturn
	}
```
how to use it:

```
	hs.Router.GET(server_utils.API_PREFIX+"/search/live_tail", getMiddlewares(hs.Recovery(liveTailHandler()), true))
[...]

PR #1774, opened 2024-10-20
SegmentStatsCmd Query Type for New Query Pipeline

Macbeth98:
# Description
- The PR: #1773 needs to be merged first.
- Now the stats RRC `SegmentStatsCmd` query type is supported in the new query pipeline.
- Refactored the code as `SegmentStatsCmd` and `GroupByCmd` mostly use same flow.
- Added some qid logging on `IQR` file.

# Testing
- Tested by running the below query with new query pipeline and verified the response by running the same query through UI.
```
--data '{
    "state": "query",
    "searchText": "* | stats avg(http_status), sum(http_status), count(http_status)",
    "startEpoch": "now-7d",
    "endEpoch": "now",
    "indexName": "*",
    "from": 0,
    "queryLanguage": "Splunk QL"
}'
```

# Checklist:

### DIFF
--- a/pkg/segment/query/iqr/intermediateQueryResult.go
+++ b/pkg/segment/query/iqr/intermediateQueryResult.go
@@ -62,7 +62,7 @@ type IQR struct {
 	qType          structs.QueryType
 }
 
-func NewIQR(qid uint64) *IQR {
+func NewIQR(qid uint64, qType structs.QueryType) *IQR {
 	return &IQR{
 		mode:             notSet,
 		qid:              qid,
@@ -73,7 +73,7 @@ func NewIQR(qid uint64) *IQR {
 		renamedColumns:   make(map[string]string),
 		groupbyColumns:   make([]string, 0),
 		measureColumns:   make([]string, 0),
-		qType:            structs.RRCCmd,
+		qType:            qType,
 	}
 }
 
--- a/pkg/segment/query/iqr/intermediateQueryResult.go
+++ b/pkg/segment/query/iqr/intermediateQueryResult.go
@@ -786,10 +786,10 @@ func (iqr *IQR) AsResult(qType structs.QueryType) (*structs.PipeSearchResponseOu
 			CanScrollMore:      false,
 			ColumnsOrder:       toputils.GetSortedStringKeys(records),
 		}
-	case structs.GroupByCmd:
-		bucketHolderArr, aggGroupByCols, measureFuncs, bucketCount, err := iqr.getFinalResultForGroupBy()
+	case structs.GroupByCmd, structs.SegmentStatsCmd:
+		bucketHolderArr, aggGroupByCols, measureFuncs, bucketCount, err := iqr.getFinalStatsResults()
 		if err != nil {
-			return nil, toputils.TeeErrorf("IQR.AsResult: error getting final result for GroupBy: %v", err)
+			return nil, toputils.TeeErrorf("qid=%v, IQR.AsResult: error getting final result for GroupBy: %v", iqr.qid, err)
 		}
 
 		response = &structs.PipeSearchResponseOuter{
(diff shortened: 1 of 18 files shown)


pkg/segment/results/segresults/segresults.go
```diff
@@ -517,7 +517,7 @@ func (sr *SearchResults) GetSegmentStatsResults(skEnc uint16) ([]*structs.Bucket
 		}
 	}
 	aggMeasureResult := []*structs.BucketHolder{bucketHolder}
-	return aggMeasureResult, sr.segStatsResults.measureFunctions, sr.segStatsResults.groupByCols, nil, len(sr.segStatsResults.measureResults)
+	return aggMeasureResult, sr.segStatsResults.measureFunctions, sr.segStatsResults.groupByCols, nil, 1
```
AndrewHess, line 520: Is this change to fix a bug, or was the `len()` always 1?

pkg/segment/query/segquery.go
```diff
@@ -399,6 +399,11 @@ func GetNodeResultsForSegmentStatsCmd(queryInfo *QueryInformation, sTime time.Ti
 			ErrList: []error{err},
 		}
 	}
+	for _, qsr := range sortedQSRSlice {
+		if qsr.sType == structs.RAW_SEARCH || qsr.sType == structs.PQS {
+			qsr.sType = structs.SEGMENT_STATS_SEARCH
+		}
+	}
```
AndrewHess, line 406: I don't quite understand why you're doing this. It seems like we'd expect the sType to already be `SEGMENT_STATS_SEARCH` since we're in the function `GetNodeResultsForSegmentStatsCmd()`. Why would we get here with `RAW_SEARCH` or `PQS`?

pkg/segment/query/iqr/intermediateQueryResult.go
```diff
@@ -806,18 +806,16 @@ func (iqr *IQR) AsResult(qType structs.QueryType) (*structs.PipeSearchResponseOu
...
-		panic("IQR.AsResult: SegmentStatsCmd not implemented")
 	default:
 		return nil, fmt.Errorf("IQR.AsResult: unexpected query type %v", qType)
 	}
 
 	return response, nil
 }
 
-func (iqr *IQR) AppendGroupByResults(bucketHolderArr []*structs.BucketHolder, measureFuncs []string, aggGroupByCols []string, bucketCount int) error {
+func (iqr *IQR) AppendRRCStatsResults(bucketHolderArr []*structs.BucketHolder, measureFuncs []string, aggGroupByCols []string, bucketCount int) error {
```
AndrewHess, line 816: nit: should this be renamed to `AppendStatsResults`? Does it matter that the stats results came from RRCs?

AndrewHess (commented):
Looks pretty good; a couple questions

pkg/segment/results/segresults/segresults.go
Macbeth98: Yes, the `len` will always be 1. That is the bucketCount. And For only `stats` without groupby, the bucket count is always 1.

if you check the [Line #519](https://github.com/siglens/siglens/pull/1774/files/3b46f3aa392361b0c318c597c89067443dfa60d2#diff-153fd9c9d5bb1a446f897390942735ad9010e6656c9fcd80494b41cd5ccc5252R519), We are just assigning one bucket Holder and making it a slice.

And the `len(sr.segStatsResults.measureResults)` gives the length of measure functions like `sum(col1), avg(col2), etc`

pkg/segment/query/iqr/intermediateQueryResult.go
Macbeth98: It does not I renamed.

pkg/segment/query/segquery.go
Macbeth98: Yes, actually because I am sending `SQRS` that we are fetching from `searched.initializeQSRs`, and passing it to the `segmentStats` function, it will not fetch again from the `getAllSegmentsInAggs` function, which is why I have to add this code.

But now I have removed this and passing for `SegmentStatsCmd` and also only fetching the `SQRs`, if the command type is `GroupByCmd`.

AndrewHess (approved):

PR #1782, opened 2024-10-21
Fields command for the new pipeline

vasukalariya:
# Description
Summarize the change.
- Added support for fields command.
- When multiple columns match a wildcard criteria, we return those cols in sorted order to provide consistent results.

# Testing
Tested using the following queries
```
app_name=Bracecould | fields first_name, city | fields city, first_name
app_name=Bracecould | fields -a* | fields first_name, city
app_name=Bracecould | fields first_name, city
app_name=Bracecould | fields first_name, city, b*
app_name=Bracecould | fields first_name, a*
app_name=Bracecould | fields first_name, city, batch | fields city, first_name, batch
```

# Checklist:

### DIFF
--- a/pkg/segment/query/iqr/intermediateQueryResult.go
+++ b/pkg/segment/query/iqr/intermediateQueryResult.go
@@ -55,6 +55,7 @@ type IQR struct {
 	knownValues    map[string][]utils.CValueEnclosure // column name -> value for every row
 	deletedColumns map[string]struct{}
 	renamedColumns map[string]string // old name -> new name
+	columnIndex    map[string]int
 
 	// Used only if the mode is withoutRRCs. Sometimes not used in that mode.
 	groupbyColumns []string
@@ -72,6 +73,7 @@ func NewIQR(qid uint64) *IQR {
 		renamedColumns:   make(map[string]string),
 		groupbyColumns:   make([]string, 0),
 		measureColumns:   make([]string, 0),
+		columnIndex:      make(map[string]int),
 	}
 }
 
@@ -231,9 +233,12 @@ func (iqr *IQR) GetAllColumnNames() (map[string]struct{}, error) {
 		// There's no data.
 		return nil, nil
 	case withRRCs, withoutRRCs:
-		if withRRCs == iqr.mode {
-			iqr.readAllColumnNamesWithRRCs(allCnames)
-		} 
+		if iqr.mode == withRRCs {
+			err := iqr.readAllColumnNamesWithRRCs(allCnames)
+			if err != nil {
+				return nil, fmt.Errorf("IQR.ReadAllColumns: error reading all column names: %v", err)
+			}
+		}
 		for cname := range iqr.knownValues {
 			if _, ok := iqr.deletedColumns[cname]; !ok {
 				allCnames[cname] = struct{}{}
@@ -422,6 +427,10 @@ func (iqr *IQR) Append(other *IQR) error {
 		log.Errorf("IQR.Append: validation failed on other: %v", err)
 		return err
 	}
+
+	iqr.AddColumnsToDelete(other.deletedColumns)
+
+	iqr.AddColumnIndex(other.columnIndex)
 
 	mergedIQR, err := mergeMetadata([]*IQR{iqr, other})
 	if err != nil {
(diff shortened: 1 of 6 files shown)


pkg/segment/query/iqr/intermediateQueryResult.go
```diff
@@ -218,6 +220,43 @@ func (iqr *IQR) ReadAllColumns() (map[string][]utils.CValueEnclosure, error) {
 	}
 }
 
+func (iqr *IQR) GetAllColumnNames() (map[string]struct{}, error) {
```
AndrewHess, line 223: FYI: https://github.com/siglens/siglens/pull/1766 also made a method like this. We should only keep one version of this method, so if that PR gets merged first you should use it from that (and also, maybe check if you find any bugs with that implementation)

pkg/segment/query/iqr/intermediateQueryResult.go
```diff
@@ -686,6 +774,34 @@
 	return nil
 }
 
+func (iqr *IQR) GetColumnsOrder(allCnames []string) []string {
```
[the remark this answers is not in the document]
AndrewHess: I'm a bit confused by this method. It looks like multiple columns can be at the same index and then they're sorted alphabetically? Adding good unit tests would help clarify how this function is supposed to work

pkg/segment/query/iqr/intermediateQueryResult.go
```diff
@@ -709,6 +825,9 @@ func (iqr *IQR) AsResult() (*structs.PipeSearchResponseOuter, error) {
 
 		// Append the known values to the result.
 		for cname, values := range iqr.knownValues {
+			if _, isDeleted := iqr.deletedColumns[cname]; isDeleted {
+				continue
+			}
 			records[cname] = values
 		}
```
AndrewHess, line 832: This is already done in `readAllColumnsWithRRCs()`. So instead of adding to this loop, you should delete the whole loop

pkg/segment/reader/record/recordreader.go
```diff
@@ -37,9 +37,9 @@ import (
 )
 
 func ReadAllColsForRRCs(segKey string, vTable string, rrcs []*utils.RecordResultContainer,
-	qid uint64) (map[string][]utils.CValueEnclosure, error) {
+	qid uint64, deletedCols map[string]struct{}) (map[string][]utils.CValueEnclosure, error) {
```
AndrewHess, line 40: nit: I think renaming `deletedCols` to `ignoredCols` would be slightly better. This function doesn't need to know why it's skipping some columns; it just needs to know that it should skip some columns

pkg/segment/query/processor/fieldscommand.go
```diff
@@ -18,22 +18,87 @@
 package processor
```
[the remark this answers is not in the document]
AndrewHess: Yes. You can reference sortcommand_test.go if you want

AndrewHess (commented):
Looking decent; a few comments

pkg/segment/query/iqr/intermediateQueryResult.go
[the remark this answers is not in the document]
vasukalariya: addressed in the latest commit.

pkg/segment/query/processor/fieldscommand.go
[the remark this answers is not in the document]
vasukalariya: addressed in the latest commit.

pkg/segment/query/iqr/intermediateQueryResult.go

pkg/segment/query/iqr/intermediateQueryResult.go

pkg/segment/reader/record/recordreader.go
vasukalariya: addressed in the latest commit.

pkg/segment/query/iqr/intermediateQueryResult.go
```diff
@@ -356,6 +358,10 @@ func (iqr *IQR) Append(other *IQR) error {
 		return err
 	}
 
+	iqr.AddColumnsToDelete(other.deletedColumns)
+
+	iqr.AddColumnIndex(other.columnIndex)
+
 	mergedIQR, err := mergeMetadata([]*IQR{iqr, other})
```
AndrewHess, line 365: This seems like it should be part of `mergeMetadata`. And do you need to do any validation between the two IQRs before you merge them (e.g., they should already have the same `deletedColumns`, or should you blindly accept the `other.deletedColumns`)?

pkg/segment/query/iqr/intermediateQueryResult.go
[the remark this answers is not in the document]
AndrewHess: Ok, so here's what I think this function does:
1. For each column in `allCnames`, find out which group it's in based on the `iqr.columnIndex`.
2. The final result has all columns in group 1, then all in group 2, ..., and finally all columns without a group.
3. Within a group, all the columns are sorted lexicographically . The last columns (those without a group) are also sorted lexicographically

Is that right, or did I miss something? Maybe add a comment similar to that to this function to explain what it's supposed to do.

pkg/segment/query/iqr/intermediateQueryResult.go
vasukalariya: Yes you can accept other deleted columns, since some IQRs might have different columns based on the rrcs/segments they have.

vasukalariya: For example: you do an -a*, IQR_1 has `address`, IQR_2 has `application`. In these cases, you won't have same deleted columns and there is no way to validate it. The best approach is to add both of them.

AndrewHess (dismissed):
Looks good

AndrewHess (approved):

PR #1783, opened 2024-10-22
write empty pqs meta only through channel

Macbeth98:
# Description
- All the writes to empty PQS meta is only done through the `PQSChan` channel.

# Testing
- Tested by ingesting and making some queries and ensuring those were written to the empty pqid meta file and sfm file.

# Checklist:

### DIFF
--- a/pkg/retention/retention.go
+++ b/pkg/retention/retention.go
@@ -194,9 +194,9 @@ func GetRetentionTimeMs(retentionHours int, currTime time.Time) uint64 {
 func deleteSegmentsFromEmptyPqMetaFiles(segmentsToDelete map[string]*structs.SegMeta) {
 	for _, segmetaEntry := range segmentsToDelete {
 		for pqid := range segmetaEntry.AllPQIDs {
-			updatedEmptyPQS := pqsmeta.GetUpdatedPQSMapAfterSegmentRemoval(pqid, segmetaEntry.SegmentKey)
-			if updatedEmptyPQS != nil {
-				go writer.AddEmptyPQIDMapToChan(pqid, updatedEmptyPQS)
+			updatedEmptyPQSMap := pqsmeta.GetUpdatedPQSMapAfterSegmentRemoval(pqid, segmetaEntry.SegmentKey)
+			if updatedEmptyPQSMap != nil {
+				go writer.AddEmptyPQIDMapToChan(pqid, updatedEmptyPQSMap)
 			}
 		}
 	}
--- a/pkg/retention/retention.go
+++ b/pkg/retention/retention.go
@@ -194,7 +194,10 @@ func GetRetentionTimeMs(retentionHours int, currTime time.Time) uint64 {
 func deleteSegmentsFromEmptyPqMetaFiles(segmentsToDelete map[string]*structs.SegMeta) {
 	for _, segmetaEntry := range segmentsToDelete {
 		for pqid := range segmetaEntry.AllPQIDs {
-			pqsmeta.DeleteSegmentFromPqid(pqid, segmetaEntry.SegmentKey)
+			updatedEmptyPQS := pqsmeta.GetUpdatedPQSMapAfterSegmentRemoval(pqid, segmetaEntry.SegmentKey)
+			if updatedEmptyPQS != nil {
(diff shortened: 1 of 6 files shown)


pkg/segment/writer/segmetarw.go
```diff
@@ -505,9 +506,22 @@ func BackFillPQSSegmetaEntry(segkey string, newpqid string) {
...
+// AddEmptyPQIDToChan adds a new pqid to the channel
+// if writeToEmptyPQIDMeta is true, then it will write the EmptyResults for this pqid
+func AddEmptyPQIDToChan(pqid string, segKey string) {
+	pqsChan <- PQSChanMeta{pqid: pqid, segKey: segKey, writeToEmptyPQIDMeta: true}
+}
+
+func AddEmptyPQIDMapToChan(pqid string, pqidToEmptySegMap map[string]bool) {
+	for segKey := range pqidToEmptySegMap {
+		AddEmptyPQIDToChan(pqid, segKey)
+	}
```
Macbeth98, line 524: This function is being called from `deleteSegmentsFromEmptyPqMetaFiles` in `retention.go` file. Previously function is directly calling`DeleteSegmentFromPqid` and now it calls this with the updated PQID Map.

Is this fine? Or this needs to be immediate?

pkg/retention/retention.go
```diff
@@ -194,7 +194,10 @@ func GetRetentionTimeMs(retentionHours int, currTime time.Time) uint64 {
 func deleteSegmentsFromEmptyPqMetaFiles(segmentsToDelete map[string]*structs.SegMeta) {
 	for _, segmetaEntry := range segmentsToDelete {
 		for pqid := range segmetaEntry.AllPQIDs {
-			pqsmeta.DeleteSegmentFromPqid(pqid, segmetaEntry.SegmentKey)
+			updatedEmptyPQS := pqsmeta.GetUpdatedPQSMapAfterSegmentRemoval(pqid, segmetaEntry.SegmentKey)
+			if updatedEmptyPQS != nil {
```
Macbeth98, line 198: This function removes the PQMR files, if the `len(updatedEmptyPQS) == 0`. Which also deletes the `pqid.meta` file. 

Is this fine? As this also a kind of write?

pkg/retention/retention.go
```diff
@@ -196,7 +195,7 @@ func GetRetentionTimeMs(retentionHours int, currTime time.Time) uint64 {
 func deleteSegmentsFromEmptyPqMetaFiles(segmentsToDelete map[string]*structs.SegMeta) {
 	for _, segmetaEntry := range segmentsToDelete {
 		for pqid := range segmetaEntry.AllPQIDs {
-			pqsmeta.DeleteSegmentFromPqid(pqid, segmetaEntry.SegmentKey)
+			go writer.RemoveSegmentFromEmptyPqmeta(pqid, segmetaEntry.SegmentKey)
```
nkunal, line 198: we don't want a go routine here, as it will spawn 1000*NumSegsToDelete gorutines. 
It is ok to have this as a blocking call.

pkg/segment/query/pqs/meta/pqsmeta.go
```diff
@@ -160,17 +156,22 @@ func removePqmrFilesAndDirectory(pqid string) error {
...
 	if err != nil {
-		log.Errorf("DeleteSegmentFromPqid: Failed to get empty PQS data from file at %s: Error=%v", pqFname, err)
+		log.Errorf("BulkDeleteSegKeysFromPqid: Failed to get empty PQS data from file at %s: Error=%v", pqFname, err)
 	}
-	delete(emptyPQS, segKey)
+
+	for segKey := range segKeyMap {
+		delete(emptyPQS, segKey)
+	}
+
```
nkunal, line 170: if the `len of emptyPQS` is not zero then don't we have to write the .pqmeta file with the leftover entries ?

Macbeth98: We are writing there is a call for `writeEmptyPqsMapToFile` function at Line: 178

nkunal: I see

nkunal (approved):

PR #1786, opened 2024-10-22
Add eval to the new query pipeline

vasukalariya:
# Description
Summarize the change.
- Added eval command for new query pipeline support.

# Testing

Ran this query containing multiple eval statements and verified the results.
```
app_name=Bracecould | eval abs_sum=abs(latitude) | eval concated=abs_sum.latitude."_".ltrim(gender, "fem") | eval round_sum=round(latitude, 2) | eval avg_sum_min_max= (latitude + longitude + latitude)/3 | eval cnt=latency%123 | eval len_gender=len(gender) | eval request_status=if(http_status >= 400, "Error", "Success") | eval bool_success=if(in(http_status, 400, 404, 500), "0", 1) | eval is_string=if(isstr(bool_success), bool_success, null()) | eval decoded_url=urldecode("http%3A%2F%2Fwww.splunk.com%2Fdownload%3Fr%3Dheader") | eval splitVal=split(concated, ".") | eval max_cnt=max(162, cnt) | eval exact_per=exact(cnt/max_cnt) | eval gender_short=if(match(gender, "^m"), "A", "F") | eval toNumber=tonumber(bool_success) | eval status_hex=tostring(http_status, "hex") | eval substr_res=substr("splendid", 1, 3) . substr("chunk", -3) | eval logValue=log(http_status, 2) | eval formatted_time=strftime(1719415380000, "%Y") | eval timeType=typeof(formatted_time) | eval short_gen=if(searchmatch("gender=m*e"), "M", "F") | eval x="hi" | eval y="a" | eval z="great" | eval match=if(searchmatch("x=hi y=a? z=*"), "yes", "no") | eval nowResult=relative_time(now(),"-1h") | eval timeResult=relative_time(time()/1000,"-1h") | eval finalResult=if(nowResult=timeResult, "Pass", "Fail")
```

# Checklist:


pkg/ast/pipesearch/searchQueryParser.go
```diff
@@ -450,6 +450,7 @@ func searchPipeCommandsToASTnode(node *QueryAggregators, qid uint64) (*QueryAggr
 			log.Errorf("qid=%d, searchPipeCommandsToASTnode: parseColumnsCmd error: %v", qid, err)
 			return nil, err
 		}
+		pipeCommands.EvalExpr = node.EvalExpr
```
AndrewHess, line 453: Is this useful? I thought the current query pipeline doesn't use `EvalExpr` (I might be wrong). For the new query pipeline, we'll return on line 431 so won't get here

pkg/segment/query/processor/evalcommand_test.go
```diff
@@ -0,0 +1,164 @@
...
+	"github.com/stretchr/testify/assert"
+)
+
+func Test_EvalCommand(t *testing.T) {
+
+	boolExpr := &structs.BoolExpr{
+		IsTerminal: true,
+		LeftValue: &structs.ValueExpr{
+			ValueExprMode: structs.VEMNumericExpr,
+			NumericExpr: &structs.NumericExpr{
```
AndrewHess, line 36: Hmm this large BoolExpr is very long and a bit hard to read. Maybe we could use our SPL parser to parse an `eval` query and extract the BoolExpr, so this unit test is more obvious. For that we might also have to have this unit test in a separate `tests` package to avoid an import cycle.

I think you can keep this though; I'm not sure if trying to extract the BoolExpr from the parser output will be easy. You should add a comment above this though giving the equivalent SQL command

pkg/segment/query/processor/evalcommand_test.go
```diff
@@ -0,0 +1,164 @@
...
+			utils.CValueEnclosure{Dtype: utils.SS_DT_STRING, CVal: "India"},
+			utils.CValueEnclosure{Dtype: utils.SS_DT_STRING, CVal: "China"},
+			utils.CValueEnclosure{Dtype: utils.SS_DT_STRING, CVal: "Japan"},
+		},
+	}
+
+	evalField := []utils.CValueEnclosure{
+		{Dtype: utils.SS_DT_STRING, CVal: "1"},
+		{Dtype: utils.SS_DT_STRING, CVal: "India"},
+		{Dtype: utils.SS_DT_STRING, CVal: "China"},
```
AndrewHess, line 140: I think these are your expected values, right? Renaming to `expected` or `expectedEvalField` or similar would be helpful

pkg/segment/query/processor/evalcommand.go
```diff
@@ -18,22 +18,130 @@
...
 	"github.com/siglens/siglens/pkg/segment/structs"
+	"github.com/siglens/siglens/pkg/segment/utils"
+	toputils "github.com/siglens/siglens/pkg/utils"
 )
 
 type evalProcessor struct {
 	options *structs.EvalExpr
 }
 
+func evaluateValueExpr(valueExpr *structs.ValueExpr, fieldToValue map[string]utils.CValueEnclosure) (interface{}, error) {
```
AndrewHess, line 34: Can we use the existing `ValueExpr.EvaluateValueExpr()` method instead of making this? If not, would it make sense to make this function a method on `ValueExpr` and move it to that file?

pkg/segment/query/processor/evalcommand.go
```diff
@@ -18,22 +18,130 @@
...
+	allCnames, err := iqr.GetColumns()
+	if err != nil {
+		return nil, fmt.Errorf("evalProcessor.Process: failed to get all column names, err: %v", err)
+	}
+	if len(fieldsInExpr) == 1 && fieldsInExpr[0] == "*" {
+		fieldsInExpr = []string{}
+		for cname := range allCnames {
+			fieldsInExpr = append(fieldsInExpr, cname)
+		}
+	}
```
AndrewHess, line 97: I think allCnames is only used inside the `if`, so you should move `allCnames, err := iqr.GetColumns()` inside the `if` as well, so we don't do that computation when we don't need it

AndrewHess (commented):
A couple nits; looks good overall

pkg/ast/pipesearch/searchQueryParser.go
vasukalariya: I just added it to test the parser, I have also added the test for it [here](https://github.com/siglens/siglens/pull/1786/files#diff-d9d1af5bbb801ffabc34bcc85856933ec8f56a6c629e36bd0ee5bf6ce49e7c7cR11394). Later when we remove the older flow we can change this.

vasukalariya: I added it to test the parser, have removed it and modified my [test](https://github.com/siglens/siglens/pull/1786/files#diff-d9d1af5bbb801ffabc34bcc85856933ec8f56a6c629e36bd0ee5bf6ce49e7c7cR11394).

pkg/segment/query/processor/evalcommand_test.go
vasukalariya: addressed in the latest commit.

pkg/segment/query/processor/evalcommand.go
vasukalariya: addressed in the latest commit.

AndrewHess (approved):

PR #1787, opened 2024-10-22
fix the segment deletion sequence during retention

nkunal:
# Description
When retention based deletion is kicked in, we have to delete a bunch of things. The sequence on how we delete those things matter. This PR fixes a bug wherein an incorrect sequence caused `allSegmentMicroIndex` to have entries for segments that were already deleted and removed from segmeta.json.

When a deletion is kicked off, we were
1. Deleting the segkey from metadata.allSegmentMicroIndex
2. deleting pqmr files
3. Removing entry from segmeta.json
4. Deleting all files as part of the segment

After 1 was executed and before 3, some other segment got rotated and caused the segmeta.json file's timestamp to get modified. The populateMicroIndex thread would come in and read the segmeta.json. It would also read the segkey that was just deleted in step 1, and then it would add it back to `allSegmentMicroIndex`.
This segkey would then never get deleted from `allSegmentMicroIndex` and the list of the array keep growing, and bunch of errors kept popping about (could not read bsum and other files for these orphan segkeys). 
This PR fixes the sequence so that this does not happen

# Testing

# Checklist:

### DIFF
--- a/pkg/retention/retention.go
+++ b/pkg/retention/retention.go
@@ -324,10 +324,17 @@ func DeleteSegmentData(segmentsToDelete map[string]*structs.SegMeta, updateBlob 
 	if len(segmentsToDelete) == 0 {
 		return
 	}
-	deleteSegmentsFromEmptyPqMetaFiles(segmentsToDelete)
-	// Delete segment key from all SiglensMetadata structs
+
+	// 1) First delete from segmeta.json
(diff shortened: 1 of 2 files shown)


AndrewHess (commented):
Looks good; nice PR description

AndrewHess (approved):

PR #1789, opened 2024-10-23
Regex cmd: change to Gobbable regex

Macbeth98:
# Description
- Modified the regex to use Gobbable regex

# Testing
- Tested by running the regex queries
- unit test cases pass

# Checklist:

### DIFF
--- a/pkg/segment/query/processor/regexcommand.go
+++ b/pkg/segment/query/processor/regexcommand.go
@@ -20,6 +20,7 @@ package processor
 import (
 	"fmt"
 	"io"
+	"regexp"
 
 	"github.com/siglens/siglens/pkg/segment/query/iqr"
 	"github.com/siglens/siglens/pkg/segment/structs"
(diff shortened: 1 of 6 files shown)


pkg/segment/query/processor/regexcommand.go
```diff
@@ -47,6 +49,10 @@ func (p *regexProcessor) Process(iqr *iqr.IQR) (*iqr.IQR, error) {
 		return nil, toputils.TeeErrorf("qid=%v, regex.Process: unknown operator; op=%s", iqr.GetQID(), p.options.Op)
 	}
 
+	if p.compiledRegex == nil {
+		p.compiledRegex = p.options.GobRegexp.GetCompiledRegex()
+	}
```
AndrewHess, line 54: It could still be nil after this right? Probably return an error in that case

AndrewHess (approved):

PR #1791, opened 2024-10-23
fix: Bin CMD renaming of columns plus Bin CM without bins and without span

Macbeth98:
# Description
- There is a `bin` command version without bins and span and for that new column name and parser output is not being set to the `BinExpr`.
- This is fixed and the below query will work now

```
--data '{
    "state": "query",
    "searchText": "app_name=Bracecould | bin http_status as bin_http | bin bins=2 http_status as bin_http_2 | bin span=5s http_status as bin_http_3 | bin span=2log3 http_status as bin_http_4 | bin minspan=500 http_status as bin_http_5 | bin bins=50000 http_status as bin_http_6 | bin bins=50000 minspan=2 http_status as bin_http_7 | bin end=100000 http_status as bin_http_8 | bin start=-10000 http_status as bin_http_9",
    "startEpoch": "now-7d",
    "endEpoch": "now",
    "indexName": "*",
    "from": 0,
    "queryLanguage": "Splunk QL"
}'
```

# Testing
- Tested by running the above query

# Checklist:

### DIFF
--- a/pkg/segment/query/processor/bincommand.go
+++ b/pkg/segment/query/processor/bincommand.go
@@ -124,8 +124,8 @@ func (p *binProcessor) Process(iqr *iqr.IQR) (*iqr.IQR, error) {
 			}
 			err = p.performBinWithSpan(value)
 			if err != nil {
-				return nil, utils.TeeErrorf("qid=%v, bin.Process: cannot bin value %v; err=%v",
-					qid, value, err)
+				return nil, utils.TeeErrorf("qid=%v, bin.Process: cannot bin value %v, field=%v; err=%v",
+					qid, value, p.options.Field, err)
 			}
 		}
 	}
--- a/pkg/segment/query/processor/bincommand.go
+++ b/pkg/segment/query/processor/bincommand.go
@@ -88,16 +88,18 @@ func (p *binProcessor) Process(iqr *iqr.IQR) (*iqr.IQR, error) {
 		newNameExists = false
 	}
 
-	if p.options.Field == config.GetTimeStampKey() {
-		for i := range values {
-			var value *segutils.CValueEnclosure
-			if newNameExists {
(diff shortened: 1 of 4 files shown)


pkg/segment/query/processor/bincommand.go
```diff
@@ -60,54 +60,83 @@ func (p *binProcessor) Process(iqr *iqr.IQR) (*iqr.IQR, error) {
...
+	}
+
+	var newColResultValues []segutils.CValueEnclosure
+
+	newName, newNameExists := p.options.NewFieldName.Get()
+
+	// If the new field name is different from the current field name,
+	// we need to create a new slice to store the results
+	if newNameExists && newName != p.options.Field {
+		newColResultValues = make([]segutils.CValueEnclosure, len(values))
```
AndrewHess, line 86: Is this because we want to keep the old field and also have the new field with the computed values?

pkg/segment/query/processor/bincommand.go
```diff
@@ -60,54 +60,83 @@ func (p *binProcessor) Process(iqr *iqr.IQR) (*iqr.IQR, error) {
...
 	if p.options.Field == config.GetTimeStampKey() {
 		for i := range values {
-			value := &values[i]
+			var value *segutils.CValueEnclosure
+			if newNameExists {
+				newColResultValues[i] = values[i]
+				value = &newColResultValues[i]
+			} else {
+				value = &values[i]
+			}
```
AndrewHess, line 99: nit: You're duplicating this code. It would probably be better to do this just once, when we know we need to make a copy

pkg/segment/query/processor/bincommand.go
Macbeth98: Yes

AndrewHess (approved):
Looks good

PR #1792, opened 2024-10-23
Fix eval after stats

vasukalariya:
# Description
Summarize the change.
- Eval/Generated fields were not getting displayed, now they will be displayed as expected.
- Some misc fix for sending updates/signals via websocket for stats/grpBy queries

# Testing

# Checklist:


pkg/segment/query/iqr/intermediateQueryResult.go
```diff
@@ -912,8 +915,8 @@ func (iqr *IQR) AppendStatsResults(bucketHolderArr []*structs.BucketHolder, meas
 		return err
 	}
 
-	iqr.groupbyColumns = append(iqr.groupbyColumns, aggGroupByCols...)
-	iqr.measureColumns = append(iqr.measureColumns, measureFuncs...)
+	iqr.groupbyColumns = aggGroupByCols
+	iqr.measureColumns = measureFuncs
```
AndrewHess, line 919: It looks like you're changing this function to replace the stats results instead of appending them. In that case, you should probably rename the function

pkg/segment/query/iqr/intermediateQueryResult_test.go
```diff
@@ -642,7 +642,7 @@ func Test_getFinalStatsResults(t *testing.T) {
 	assert.NoError(t, err)
 	assert.Equal(t, bucketCount, actualBucketCount)
 	assert.Equal(t, groupByCols, actualGroupByCols)
-	assert.Equal(t, measureFuncs, actualMeasureFuncs)
+	assert.ElementsMatch(t, measureFuncs, actualMeasureFuncs)
```
AndrewHess, line 645: Why don't we necessarily expect them in order?

Same comment below

tools/sigclient/functionalQueries/search_with_eval.json
```diff
@@ -2,7 +2,7 @@
...
     "expectedResult": {
         "totalMatched": {
-            "value": 13,
+            "value": 7,
```
AndrewHess, line 5: Why is this getting changed? Was the old pipeline giving the wrong result?

AndrewHess (commented):
Looks pretty good; a couple small questions

tools/sigclient/functionalQueries/search_with_eval.json
vasukalariya: I think this was a typo earlier, the correct answer is 7 so just updated it.

pkg/segment/query/iqr/intermediateQueryResult_test.go
vasukalariya: We are looping over the [knownValues](https://github.com/siglens/siglens/pull/1792/files#diff-24e928c23b70a5764f8ad7cb3869c8a7f7114e941ca26c895dd7d94dc9af025cR967) map to gather the measureCols since it could include created columns from other commands like eval, bin, etc. Map iteration order is not deterministic so used the ElementsMatch instead of Equal.

pkg/segment/query/iqr/intermediateQueryResult.go
vasukalariya: addressed in the latest commit.

pkg/segment/query/iqr/intermediateQueryResult_test.go
AndrewHess: `measureCols` can contain things that aren't measure columns from a stats? That seems odd. What's an example of when that would happen?

vasukalariya: `* | stats count as cnt | eval n = 1`, here the `n = 1` has to be returned as a measureCol. So, for segmentStatsCmd and groupByCmd, we have a fixed way of sending data that is in form of measureValues, for a query like this if there is a create column like n in this case, it has to be sent as a measureCol.

AndrewHess (approved):

PR #1795, opened 2024-10-24
Add rename to the new pipeline

vasukalariya:
# Description
Summarize the change.
- Adds rename command and aggs rename to the new pipeline
- Added tests for renamecommand as well as more tests for iqr.RenameColumn method

# Testing
- Added tests
- Furthermore test end to end using these SPL commands

```
app_name=Bracecould | rename app_name AS app | rename app AS new_app | rename new_app AS latest_app | rename city as newCity

city=Boston http_method != "POST" gender=ma* | stats avg(latency) as avg, max(http_status) as max, min(latitude) as min, sum(longitude) as sum, range(latitude), count as cnt, values(http_method), dc(first_name) as distinct_count, list(gender)

city=Boston http_method != "POST" gender=ma* | stats avg(latency) as avg, max(http_status) as max, min(latitude) as min, sum(longitude) as sum, range(latitude), count as cnt, values(http_method), dc(first_name) as distinct_count, list(gender) BY http_status

app_name=Bracecould | eval newName = 1, newName2 = 2 | rename app_name AS newName | rename newName as newName2

app_name=Bracecould | rename a* AS b*
```

# Checklist:


pkg/segment/query/iqr/intermediateQueryResult.go
```diff
@@ -289,9 +289,12 @@ func (iqr *IQR) readAllColumnsWithRRCs() (map[string][]utils.CValueEnclosure, er
...
-		// TODO: don't read these columns from the RRCs, instead of reading and
-		// then deleting them.
+	for oldName, newName := range iqr.renamedColumns {
+		_, exists := results[oldName]
+		if !exists {
+			continue
+		}
+		results[newName] = results[oldName]
 		delete(results, oldName)
 	}
```
AndrewHess, line 299: Do we need this? I see that now on line 272 we're not reading the `deletedColumns` from RRCs. I thought that when we made a call to rename columns we'd already do this, and so doing this here would be a no-op. Am I missing something?

I'm also concerned: I don't think we need this here because we should do it elsewhere, but if we keep this change because it does actually have an effect in some cases, it's going to get harder and harder to remove this code

pkg/segment/query/iqr/intermediateQueryResult.go
```diff
@@ -743,10 +746,45 @@ func (iqr *IQR) RenameColumn(oldName, newName string) error {
...
+	// delete newName since it would be overwritten
+	delete(iqr.knownValues, newName)
+
+	values, ok := iqr.knownValues[oldName]
+	if ok {
+		// if old name is present it means that it must be a created column
+		// so we will rename it and update the knownValues map
+		iqr.renamedColumns[oldName] = newName
 		iqr.knownValues[newName] = values
-		delete(iqr.knownValues, oldName)
```
AndrewHess, line 749: > if old name is present it means that it must be a created column so we will rename it and update the knownValues map

I don't think that's true. I think when we read RRCs for a certain column, we also cache that in `knownValues`. Maybe I misunderstand what you mean by "created column"

pkg/segment/query/iqr/intermediateQueryResult.go
```diff
@@ -743,10 +746,45 @@ func (iqr *IQR) RenameColumn(oldName, newName string) error {
 		return err
 	}
 
-	iqr.renamedColumns[oldName] = newName
-	if values, ok := iqr.knownValues[oldName]; ok {
+	// delete newName since it would be overwritten
+	delete(iqr.knownValues, newName)
```
AndrewHess, line 750: I don't understand why you replaced line 746 with line 750. Can you explain it to me?

AndrewHess (commented):
Looks pretty good; just a few questions. Nice job on the unit tests

pkg/segment/query/iqr/intermediateQueryResult.go
vasukalariya: But reading the column is not necessary for all the commands. 
For example consider `* | rename first_name as firstName`, here `first_name` would never be read because its not required you just need to mark it for renaming. And there might be more types of commands where there is a similar case.

pkg/segment/query/iqr/intermediateQueryResult.go
vasukalariya: We do not mark them as `deletedColumns`, cause they are not deleted, they are just renamed.

pkg/segment/query/iqr/intermediateQueryResult.go
vasukalariya: Oh I forgot to mention that cached columns can also be present over here. But, I think this is fine, since caching would probably be made optional and we would not want to mark this as a `deletedColumn` because we would need to read it for the case mentioned above.

pkg/segment/query/iqr/intermediateQueryResult.go
vasukalariya: It was just to separate things out to easily get the flow.
- If we rename `fieldA` to `fieldB`, it means that `fieldB` if present needs to be deleted.
- If `fieldA` was a created/read column, we can just populate it.
- If not, we need to update the chain for renaming since, it could be a column in RRC. There are 2 cases:
  - Was old name previously renamed to something if so mark it as renamed to the latest newname.
  - If not, just add the rename mapping to the renamed columns.

AndrewHess (approved):

AndrewHess (approved):

PR #1797, opened 2024-10-24
fix: issue with multiple bottleneck commands

Macbeth98:
# Description
- The issue was: 
       - When it is a bottleneck command, the bottleneck processor will return `iqr, err` as `nil, nil` until the bottleneck processor receives a `nil` iqr. It will then return its processed IQR results and `io.EOF`
      - This works fine for one bottleneck command,.
      - But when there is another bottleneck command, it will again send `nil, nil` as its output the first time and the current flow has already gotten `EOF` from the previous bottleneck command, so it will just return the current result `(iqr, err) as (nil, nil)`
      - So this will just return the `nil` results
  
- Now this PR, checks that if the output is `nil`, error is `nil`, and `isBottleneckCmd`, and if it is then, it will mark `gotEOF` to `false` making it go back to the loop again. 

**Update**
- This PR for every fetch, it will not consider the `EOF`, that it got from the previous stream input, the `EOF` will only be considered when the current processor returns an `EOF`
- Along with that this PR now checks that if final result exists for a processor before fetching it from the previous input stream.
- All the bottleneck commands are expected to return the final IQR, true when `GetFinalResultIfExists`. This is utilized when we have `bottleneck-cmd | twopass-cmd`. So when the twopass-cmd calls for rewind, it is going to fetch from the previous input stream which is the bottleneck-cmd, but since the bottleneck-cmd has all the data, when `GetFinalResultIfExists` in the second pass, it can just return the cached data.

# Testing
- Added unit tests for multiple bottleneck, bottleneck-twopass, and multi-twopass processors.

# Checklist:


pkg/segment/query/iqr/intermediateQueryResult.go
```diff
@@ -165,7 +165,7 @@ func (iqr *IQR) AppendKnownValues(knownValues map[string][]utils.CValueEnclosure
 
 func (iqr *IQR) NumberOfRecords() int {
 	if err := iqr.validate(); err != nil {
-		log.Errorf("qid=%v, IQR.NumberOfRecords: validation failed: %v", iqr.qid, err)
+		log.Errorf("IQR.NumberOfRecords: validation failed: %v", err)
```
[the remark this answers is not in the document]
AndrewHess: Why are you removing this?

pkg/segment/query/segquery.go
```diff
@@ -429,7 +429,9 @@ func GetNodeResultsForSegmentStatsCmd(queryInfo *QueryInformation, sTime time.Ti
 			ErrList: []error{err},
 		}
 	}
-	aggMeasureRes, aggMeasureFunctions, aggGroupByCols, _, bucketCount := allSegFileResults.GetSegmentStatsResults(0, true)
+
+	humanizeValues := !config.IsNewQueryPipelineEnabled()
+	aggMeasureRes, aggMeasureFunctions, aggGroupByCols, _, bucketCount := allSegFileResults.GetSegmentStatsResults(0, humanizeValues)
```
AndrewHess, line 434: Why don't we want to humanize?

Side note: it's a little odd that this change is in this PR, but that's ok

pkg/segment/query/processor/dataprocessor.go
```diff
@@ -90,6 +90,8 @@ func (dp *DataProcessor) Fetch() (*iqr.IQR, error) {
 			gotEOF = true
 		} else if err != nil {
 			return nil, utils.TeeErrorf("DP.Fetch: failed to process input: %v", err)
+		} else if output == nil && dp.isBottleneckCmd {
+			gotEOF = false
```
AndrewHess, line 94: I'm concerned about this change; it feels more like a hack than a proper solution. Why do we need this? Also I think this will fail (i.e., cause an infinite loop) if we get to a situation where the bottleneck command _should_ return no output (e.g., a sort where no data was passed in)

pkg/segment/query/processor/dataprocessor_test.go
```diff
@@ -256,3 +277,251 @@ func Test_Fetch_multipleStreams(t *testing.T) {
...
+		streams:         []*cachedStream{NewCachedStream(dp1)},
+		processor:       &mockBottleneckProcessor{name: "dp2"},
+		isBottleneckCmd: true,
+	}
+
+	dp3 := &DataProcessor{
+		streams:      []*cachedStream{NewCachedStream(dp2)},
+		processor:    &mockTwoPassProcessor{name: "dp3"},
+		isTwoPassCmd: true,
+	}
```
AndrewHess, line 381: You should probably also have `isBottleneckCmd: true`, since a two-pass command is always a bottleneck

pkg/segment/query/segquery.go
Macbeth98: This is when first stats is there like: `* | stats ...`. And you have something after that: `* | stats count | eval | ...` and more pipes, then if the values are humanized then we won't be able to perform subsequent command operations properly.

pkg/segment/query/processor/dataprocessor.go
Macbeth98: But wouldn't the bottleneck command should return `io.EOF` at some point? Then it does not matter what the output is, the `EOF` will be considered and it will break out of loop.

Also I am changing this part a bit.

pkg/segment/query/iqr/intermediateQueryResult.go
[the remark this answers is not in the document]
Macbeth98: I removed this because if we get an error from `validate` there is a chance that `iqr` is `nil` and if we access the `qid`, then it will panic.

pkg/segment/query/processor/sortcommand.go
```diff
@@ -28,15 +28,20 @@ import (
...
 }
 
 func (p *sortProcessor) Process(inputIQR *iqr.IQR) (*iqr.IQR, error) {
 	if inputIQR == nil {
 		// There's no more input, so we can send the results.
+		p.isResultFinal = true
 		return p.resultsSoFar, io.EOF
+	} else if p.isResultFinal {
+		// The result is already final, so this is being called for the second Pass.
+		return nil, nil
```
AndrewHess, line 44: It kind of seems like `isResultFinal` is being used to indicate if its sent all the results. Maybe rename it?

Also shouldn't line 44 by `return nil, io.EOF`?

pkg/segment/query/processor/statscommand.go
```diff
@@ -39,13 +39,23 @@ type ErrorData struct {
 	notSupportedStatsType map[string]struct{} // columnName -> struct{}. Tracks unsupported stats types
 }
 
+type statsType uint8
+
+const (
+	InvalidCmd statsType = iota
+	SegmentStatsCmd
+	GroupByCmd
+)
```
AndrewHess, line 48: We already have this in segstructs.go; probably just reuse that instead of making a new enum
```
const (
	InvalidCmd QueryType = iota
	SegmentStatsCmd
	GroupByCmd
	RRCCmd
)
```

pkg/segment/query/processor/statscommand.go
```diff
@@ -58,9 +68,27 @@ func (p *statsProcessor) Process(inputIQR *iqr.IQR) (*iqr.IQR, error) {
...
+			return p.extractGroupByResults(iqr)
+		} else if p.processorType == SegmentStatsCmd {
+			return p.extractSegmentStatsResults(iqr)
+		}
+
+		return nil, toputils.TeeErrorf("qid=%v, statsProcessor.Process: invalid processor type", p.qid)
+	} else if p.finalResultComputed {
+		// Result is already computed but the inputIQR is not nil.
+		// This means, it is the second pass, so there is no need to compute the result again.
+		return nil, nil
```
AndrewHess, line 84: Shouldn't this return `nil, io.EOF`? Also now in `Rewind()` don't you need to update some state indicating you should send the stats again when you get a `nil` input?

pkg/segment/query/processor/statscommand.go
Macbeth98: Yeah I created a new enum because `RRCCmd` is not a valid option. So I guess, I will just use this.

AndrewHess (commented):
These updates look good!

AndrewHess (approved):

PR #1798, opened 2024-10-24
Compute bloom sizes to reduce disk usage

AndrewHess:
# Description
Previously for our bloom filters, we'd give a generous estimate for how many unique values are needed, then insert all the keys, and write it to disk. For some datasets that caused unnecessarily large bloom files. With this PR, we now add all the values to the bloom, use that to check how many unique values exist, then discard that bloom and insert the values into a new, properly sized bloom. 

Additionally, if we're ingesting multiple columns, we'll use just one bloom to do that initial computation of the number of unique values; so it gets computed for one column, then the bloom is reset and used to calculate the number of unique values the next column.

This change only affects non-DE columns.

This reduces memory usage and disk usage, but the improvement depends on the dataset being ingested.

# Testing
Ingested a dataset with 330 columns and lots of nil values. Checked the RAM usage during ingestion, and the disk usage after rotation.

# Checklist:


nkunal (dismissed):
looking good. In the next version of the PR, we should change the `addToBlockBloomBothCasesWithBuf` such that on the second pass, we use the `bloom.Add` instead of `bloom.TestAndAdd` , the testadd function is slightly slower because it does some extra work to find out if the asked entry is already present

nkunal (approved):

PR #1801, opened 2024-10-25
Fix stats groupBy values format

Macbeth98:
# Description
- Maintain the format of group by values and do not convert them to strings.
- This PR will not convert the GroupByValues to String.

For Queries in the form : `* | stats group by`
- The results will include new value called `IGroupByValues` representing group by values not converted to string.
- I have to do this way, as there are lot of places where group by values were assumes `[]string` and operations were performed in the old pipeline.
- And also cicd tests (ingest.csv, restart.csv) also checks for string converted group by values.
- So to not break that, I have created the new field `IGroupByValues`, which will be reflected in the result. But this is only when we are finally fetching the results and while processing remaining commands after stats the values that are in the IQR will have their own type. 

For Queries in the form: `* | ... | stats`
- There are no discrepancies with the result. And all the values will maintain their own type and will not converted to string.

# Testing
- Updated some of the unit tests with the new type.
- All unit test cases passed
- All functional and cicd ingest.csv queries passed (on my local machine)

# Checklist:


pkg/segment/structs/segstructs.go
```diff
@@ -452,9 +452,15 @@ type AggregationResult struct {
...
 
+// TODO: Retain either IGroupByValues or GroupByValues, as having both is unnecessary.
+// The goal is to preserve the group-by value type as interface{} to avoid issues
+// when processing subsequent commands.
+// Ideally, we should update GroupByValues to have a type of []interface{}
+// and eliminate IGroupByValues.
 type BucketHolder struct {
-	GroupByValues []string
-	MeasureVal    map[string]interface{}
+	IGroupByValues []interface{} // each group-by value is stored as interface{}
```
AndrewHess, line 461: nit: maybe use `CValueEnclosure` or `DTypeEnclosure` instead of `interface{}`, so that it's more clear which types are valid

pkg/segment/utils/segconsts.go
```diff
@@ -1032,10 +1051,10 @@ func (e *CValueEnclosure) AsBytes() []byte {
 			return []byte("true")
 		}
 		return []byte("false")
-	case SS_DT_SIGNED_NUM, SS_DT_SIGNED_32_NUM, SS_DT_SIGNED_16_NUM, SS_DT_SIGNED_8_NUM:
+	case SS_DT_SIGNED_NUM:
 		buf := make([]byte, 0, 20)
 		return strconv.AppendInt(buf, e.CVal.(int64), 10)
-	case SS_DT_UNSIGNED_NUM, SS_DT_USIGNED_32_NUM, SS_DT_USIGNED_16_NUM, SS_DT_USIGNED_8_NUM:
+	case SS_DT_UNSIGNED_NUM:
```
AndrewHess, line 1057: Why do we want this change?

tools/sigclient/functionalQueries/eval_groupby_2col_2stats.json
```diff
@@ -2,7 +2,7 @@
     "queryText": "city=Boston | eval temp_value = http_status%2 | stats count, sum(latitude) BY gender, temp_value",
     "expectedResult": {
          "totalMatched": {
-             "value": 984,
+             "value": 4,
              "relation": "eq"
```
AndrewHess, line 6: Is this the desired behavior? 4 is the number of rows returned, but I assume 984 is the number of records matching the search `city=Boston` (or maybe 984 came from somewhere else?)

cc: @nkunal

pkg/segment/utils/segconsts.go
AndrewHess: Ok that makes sense why you'd want that, but the function is just `AsBytes()`, so it should work for any valid `CValueEnclosure` type. I think with your latest commits you're not using this anymore, so you can revert your changes. Actually I think no code is using this function, so you could even delete the function if you don't think we'll need it in the future

Macbeth98: Yes, I have removed the function.

PR #1802, opened 2024-10-27
Add progress updates for Websocket for new pipeline

vasukalariya:
# Description
Summarize the change.
Progress updates would be sent as follows for different types of queries:
**RRCCmd**
- Sends updates about records searched and complete percentage during the search stage. (This is done so that stats/group by commands as a RRCCmd will send at least complete percentage updates while fetching all records)
- Sends records when available during each fetches.

**SegStatsCmd/GroupByCmd**
- Here we are using the older flow so percentage complete updates are sent from [IncrementNumFinishedSegments](https://github.com/siglens/siglens/blob/79628869224e49ad5adb67ef3e328b452ecf272e/pkg/segment/query/querystatus.go#L224C6-L224C34).
- Also, the buckets and measure data is sent during the complete stage since, for these commands, there is no use to send any data in updates.

For percentage completion updates, we are using blocks for RRCCmd and Segments for SegstatsCmd/GroupByCmd (because its using an older flow), however the number of records searched and totalRecords would be displayed correctly.

With this PR, user will be able to see some update when query is executing and get an idea about how much it is complete, also for RRCCmd queries they will see the partial results during each update call.

Also, added a check to send the sortByTimestamp default based on the queryAggs.

Fixed a small bug where statscommand was not using correct qid. 

# Testing

Tested using all types of queries
- RRCCmd query without stats/groupby
- RRCCmd query with stats/groupby
- SegstatsCmd query
- GroupByCmd query 

# Checklist:

### DIFF
--- a/pkg/ast/pipesearch/wsSearchHandler.go
+++ b/pkg/ast/pipesearch/wsSearchHandler.go
@@ -197,7 +197,8 @@ func ProcessPipeSearchWebsocket(conn *websocket.Conn, orgid uint64, ctx *fasthtt
 	}
 }
 
-func RunAsyncQueryForNewPipeline(conn *websocket.Conn, qid uint64, simpleNode *structs.ASTNode, aggs *structs.QueryAggregators, qc *structs.QueryContext, sizeLimit uint64, scrollFrom int) {
+func RunAsyncQueryForNewPipeline(conn *websocket.Conn, qid uint64, simpleNode *structs.ASTNode, aggs *structs.QueryAggregators,
+	qc *structs.QueryContext, sizeLimit uint64, scrollFrom int) {
 	websocketR := make(chan map[string]interface{})
 	eventC, err := segment.ExecuteAsyncQueryForNewPipeline(simpleNode, aggs, qid, qc)
 	if err != nil {
@@ -212,32 +213,32 @@ func RunAsyncQueryForNewPipeline(conn *websocket.Conn, qid uint64, simpleNode *s
 	go listenToConnection(qid, websocketR, conn)
 	for {
 		select {
-		case qscd, ok := <-eventC:
+		case queryStateChanData, ok := <-eventC:
 			if !ok {
-				log.Errorf("qid=%v, RunAsyncQueryForNewPipeline: Got non ok, state: %v", qid, qscd.StateName)
+				log.Errorf("qid=%v, RunAsyncQueryForNewPipeline: Got non ok, state: %v", qid, queryStateChanData.StateName)
 				query.LogGlobalSearchErrors(qid)
 				return
 			}
-			switch qscd.StateName {
+			switch queryStateChanData.StateName {
 			case query.RUNNING:
 				processRunningUpdate(conn, qid)
 			case query.TIMEOUT:
 				processTimeoutUpdate(conn, qid)
 				return
 			case query.QUERY_UPDATE:
-				wErr := conn.WriteJSON(qscd.UpdateWSResp)
+				wErr := conn.WriteJSON(queryStateChanData.UpdateWSResp)
 				if wErr != nil {
 					log.Errorf("qid=%d, RunAsyncQueryForNewPipeline: failed to write update response to websocket! err: %+v", qid, wErr)
 				}
 			case query.COMPLETE:
-				wErr := conn.WriteJSON(qscd.CompleteWSResp)
+				wErr := conn.WriteJSON(queryStateChanData.CompleteWSResp)
 				if wErr != nil {
 					log.Errorf("qid=%d, RunAsyncQueryForNewPipeline: failed to write complete response to websocket! err: %+v", qid, wErr)
 				}
 				query.DeleteQuery(qid)
 				return
 			default:
-				log.Errorf("qid=%v, RunAsyncQueryForNewPipeline: Got unknown state: %v", qid, qscd.StateName)
+				log.Errorf("qid=%v, RunAsyncQueryForNewPipeline: Got unknown state: %v", qid, queryStateChanData.StateName)
 			}
 		case readMsg := <-websocketR:
 			log.Infof("qid=%d, RunAsyncQueryForNewPipeline: Got message from websocket: %+v", qid, readMsg)
--- a/pkg/ast/pipesearch/wsSearchHandler.go
+++ b/pkg/ast/pipesearch/wsSearchHandler.go
@@ -147,6 +147,11 @@ func ProcessPipeSearchWebsocket(conn *websocket.Conn, orgid uint64, ctx *fasthtt
 
 	qc := structs.InitQueryContextWithTableInfo(ti, sizeLimit, scrollFrom, orgid, false)
 	qc.RawQuery = searchText
+	if config.IsNewQueryPipelineEnabled() {
+		RunAsyncQueryForNewPipeline(conn, qid, simpleNode, aggs, qc, sizeLimit, scrollFrom)
+		return
+	}
+
 	eventC, err := segment.ExecuteAsyncQuery(simpleNode, aggs, qid, qc)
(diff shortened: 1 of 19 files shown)


pkg/segment/query/processor/queryprocessor.go
```diff
@@ -74,8 +76,12 @@ func NewQueryProcessor(firstAgg *structs.QueryAggregators, queryInfo *query.Quer
 		firstProcessorAgg = firstProcessorAgg.Next
 	}
 
+	timeOrdered := true
 	dataProcessors := make([]*DataProcessor, 0)
 	for curAgg := firstProcessorAgg; curAgg != nil; curAgg = curAgg.Next {
+		if curAgg.SortExpr != nil {
+			timeOrdered = false
+		}
```
AndrewHess, line 84: I don't think you need this. Eventually you use this in the websocket response, but I think you can always pass `false` for that since we always give results to the frontend in the correct order (and eventually we could remove this parameter from the frontend). The frontend is only using it [here](https://github.com/siglens/siglens/blob/develop/static/js/search.js#L1195-L1200) to set the sort order.

Also, if we do have this calculation for whether it's sorted, we'd have to handle other cases too (like `dedup` with the `sortby` option specified). It will be much simpler if we don't need this.

@sonamgupta21 am I missing any reason we'd still need to set this even if the backend always gives the frontend the records in the correct order?

pkg/segment/query/processor/queryprocessor.go
```diff
@@ -216,8 +223,77 @@ func (qp *QueryProcessor) GetFullResult() (*structs.PipeSearchResponseOuter, err
...
+	if err != nil {
+		return utils.TeeErrorf("GetStreamedResult: failed to get progress; err: %v", err)
+	}
+
+	relation := "eq"
+	if progress.UnitsSearched < progress.TotalUnits {
+		relation = "gte"
+	}
+
+	completeResp.TotalMatched = utils.HitsCount{Value: uint64(totalRecords), Relation: relation}
```
AndrewHess, line 288: I don't think this is right. When the full search is done you'd be using `eq`, but the query may have finished early (i.e., sent enough results to the UI to mark the query as completed, without actually searching all the segments that might have a match), and in that case the number of records matching the query could be greater than the number of results returned. But if the query didn't exit early, then we can (and should) use `eq`

pkg/segment/query/querystatus.go
```diff
@@ -995,3 +1017,92 @@ func SetQidAsFinishedForPipeRespQuery(qid uint64) {
...
+		TotalRecords:    rQuery.Progress.TotalRecords,
+		RecordsSearched: rQuery.Progress.RecordsSearched,
+	}, nil
+}
+
+func CreateWSUpdateResponseWithProgress(qid uint64, qType structs.QueryType, progress *structs.Progress) *structs.PipeSearchWSUpdateResponse {
+	percComplete := float64(0)
+	if progress.TotalUnits > 0 {
+		percComplete = (float64(progress.UnitsSearched) * 100) / float64(progress.TotalUnits)
+	}
```
AndrewHess, line 1100: I think this is going to cause the percent complete value to be inaccurate when the query exits a long time before searching all blocks/segments. For example, after time-filtering and CMI, we may have 100 segments to search, but we stop the query once we get enough results (100 on the UI), and that could happen in the first 5 segments. Then the percent complete would be <=5% and then jump to 100%; it would be better if e.g., when we returned the first 30 results to the UI, we mark the query as 30% complete. Or maybe we need to track both the percent complete we are on searching segments and the percent complete on returning enough results to the UI; since we'll end the query once one of those reaches 100%, we take the max of them and use that as the percent complete

pkg/segment/segexecution.go
```diff
@@ -452,6 +451,25 @@ func ExecuteQuery(root *structs.ASTNode, aggs *structs.QueryAggregators, qid uin
...
+
+	queryProcessor, err := SetupPipeResQuery(root, aggs, qid, qc)
+	if err != nil {
+		log.Errorf("qid=%v, ExecuteAsyncQueryForNewPipeline: failed to SetupPipeResQuery, err: %v", qid, err)
+		return nil, err
+	}
+
+	go func() {
+		_ = queryProcessor.GetStreamedResult(rQuery.StateChan)
+	}()
```
AndrewHess, line 469: Don't ignore this `error` return value

pkg/ast/pipesearch/wsSearchHandler.go
```diff
@@ -192,6 +197,59 @@ func ProcessPipeSearchWebsocket(conn *websocket.Conn, orgid uint64, ctx *fasthtt
...
+				if wErr != nil {
+					log.Errorf("qid=%d, RunAsyncQueryForNewPipeline: failed to write update response to websocket! err: %+v", qid, wErr)
+				}
+			case query.COMPLETE:
+				wErr := conn.WriteJSON(qscd.CompleteWSResp)
+				if wErr != nil {
+					log.Errorf("qid=%d, RunAsyncQueryForNewPipeline: failed to write complete response to websocket! err: %+v", qid, wErr)
+				}
+				query.DeleteQuery(qid)
+				return
```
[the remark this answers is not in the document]
vasukalariya: We are [cancelling](https://github.com/siglens/siglens/blob/79628869224e49ad5adb67ef3e328b452ecf272e/pkg/segment/query/querystatus.go#L168) the query before sending the timeout signal on the channel

pkg/segment/query/processor/queryprocessor.go
vasukalariya: We will only returned the number of records that we computed by the query and not the matched results similar to the older pipeline. I have talked about this with Kunal, that we want to show the final number of records instead of showing the totalMatched of the initial search. 
I will only send `eq` if we have gone through all the segments/blocks which is correct. In case of query exits early we can safely say that results could be more by sending `gte`.
@nkunal can confirm.

nkunal: 1. when we are exiting early we must send `gte` . 
2. When we have searched all the segments and we are sending the exact number of records back to the user (for e.g. only 34 records matched out of 5 million) then we should send `eq`
3. When we have searched all the segments and we are sending a subset of records back to the user (for e.g. sending only 500 records although matched count was 100K out of 5 million searched) then we should send `gte`

pkg/ast/pipesearch/wsSearchHandler.go
```diff
@@ -192,6 +197,59 @@ func ProcessPipeSearchWebsocket(conn *websocket.Conn, orgid uint64, ctx *fasthtt
...
+		if wErr != nil {
+			log.Errorf("qid=%d, RunAsyncQueryForNewPipeline: failed to write error response to websocket! err: %+v", qid, wErr)
+		}
+		return
+	}
+
+	go listenToConnection(qid, websocketR, conn)
+	for {
+		select {
+		case qscd, ok := <-eventC:
```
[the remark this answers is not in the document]
vasukalariya: addressed in the latest commit.

pkg/ast/pipesearch/wsSearchHandler.go
```diff
@@ -192,6 +197,59 @@ func ProcessPipeSearchWebsocket(conn *websocket.Conn, orgid uint64, ctx *fasthtt
 	}
 }
 
+func RunAsyncQueryForNewPipeline(conn *websocket.Conn, qid uint64, simpleNode *structs.ASTNode, aggs *structs.QueryAggregators, qc *structs.QueryContext, sizeLimit uint64, scrollFrom int) {
```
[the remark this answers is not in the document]
vasukalariya: addressed in the latest commit.

pkg/segment/query/querystatus.go

AndrewHess (approved):

PR #1803, opened 2024-10-28
Check If SegKey is Rotated for Segment Stats Search

Macbeth98:
# Description
- Checks if the SegKey is rotated for Segment Stats Search and assigns the proper search type

# Testing
- Tested by running the log bench tool and verified that there were no errors.
- All unit test cases pass.

# Checklist:

### DIFF
--- a/pkg/segment/query/segquery.go
+++ b/pkg/segment/query/segquery.go
@@ -813,6 +813,12 @@ func applyAggOpOnSegments(sortedQSRSlice []*QuerySegmentRequest, allSegFileResul
 		}
 		isSegmentFullyEnclosed := segReq.segKeyTsRange.AreTimesFullyEnclosed(segReq.segKeyTsRange.StartEpochMs, segReq.segKeyTsRange.EndEpochMs)
 
+		if segReq.sType != structs.SEGMENT_STATS_SEARCH {
+			if utils.IsFileForRotatedSegment(segReq.segKey) {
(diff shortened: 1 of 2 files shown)


pkg/segment/query/segquery.go
```diff
@@ -813,6 +813,13 @@ func applyAggOpOnSegments(sortedQSRSlice []*QuerySegmentRequest, allSegFileResul
 		}
 		isSegmentFullyEnclosed := segReq.segKeyTsRange.AreTimesFullyEnclosed(segReq.segKeyTsRange.StartEpochMs, segReq.segKeyTsRange.EndEpochMs)
 
+		// For Unrotated search, Check if the segment is rotated and update the search type accordingly
+		if segReq.sType != structs.SEGMENT_STATS_SEARCH {
+			if utils.IsFileForRotatedSegment(segReq.segKey) {
+				segReq.sType = structs.SEGMENT_STATS_SEARCH
+			}
+		}
```
AndrewHess, line 821: nit: I think it would be a little more clear if you change `if segReq.sType != structs.SEGMENT_STATS_SEARCH` to `if segReq.sType = structs.UNROTATED_SEGMENT_STATS_SEARCH`, since I think that's the only case you want to handle. Or are there also other values that would be valid at this point in the flow?

AndrewHess (commented):
Looks pretty good

PR #1804, opened 2024-10-28
Add support for expression for head in new pipeline

vasukalariya:
# Description
Summarize the change.
- Added expression support for head in the new pipeline.
- Fixed the parser, earlier we were not setting HeadExpr in queryAgg for simple head (without expression).
- Added unit tests for head with expression.
- Added functional test for head.

# Testing
- Added unit tests and functional test for head with expressions.
- Also verified that queries like the ones below are working properly.
   - `* | head`
   - `* | head limit=2`
   - `app_name = Bracecould | head gender="female" OR gender="male" limit=2 keeplast=true`.

# Checklist:


pkg/segment/query/processor/headcommand.go
```diff
@@ -30,11 +32,83 @@ type headProcessor struct {
...
+func (p *headProcessor) processHeadExpr(iqr *iqr.IQR) (*iqr.IQR, error) {
+	if iqr == nil || p.options.Done {
+		return nil, io.EOF
+	}
+
+	requiredFields := p.options.BoolExpr.GetFields()
+	records, err := iqr.ReadColumns(requiredFields)
+	if err != nil {
+		return nil, toputils.TeeErrorf("headProcessor.processHeadExpr: failed to read columns, requiredFields: %v, err: %v", requiredFields, err)
+	}
```
AndrewHess, line 44: I think if the column doesn't exist in this IQR (which is a valid case), then you'll get an error. But you'll have the check head's `null` option for what to do with these records then. We might want to change `ReadColumns()`—and probably `ReadColumn()` as well—distinguish between an `error` because the field doesn't exist and an actual error of something unexpected happening

pkg/segment/query/processor/headcommand.go
```diff
@@ -30,11 +32,83 @@ type headProcessor struct {
...
+				}
+			}
+		}
+		row++
+		p.numRecordsSent++
+	}
+
+	if p.numRecordsSent == p.options.MaxRows {
+		p.options.Done = true
+	}
```
AndrewHess, line 90: Shouldn't this `if` be inside the loop?

pkg/segment/query/processor/headcommand_test.go
AndrewHess, line 1: These look good; you should also add a test where one of the fields in the head expression doesn't exist

pkg/segment/query/processor/headcommand.go
vasukalariya: No we can just check this after exit, because the loop invariant has the condition `p.numRecordsSent < p.options.MaxRows`

pkg/segment/query/processor/headcommand.go
vasukalariya: Rather than changing the existent methods, we can keep those as is and use the newly created method (ReadColumsnWithBackfill) in this PR wherever necessary.

pkg/segment/query/processor/headcommand_test.go
vasukalariya: Addressed in the latest commit.

pkg/segment/query/processor/headcommand.go
vasukalariya: Addressed in the latest commit.

pkg/segment/query/iqr/intermediateQueryResult.go
```diff
@@ -248,20 +248,34 @@ func (iqr *IQR) ReadColumn(cname string) ([]utils.CValueEnclosure, error) {
...
-		values, err := iqr.ReadColumn(cname)
-		if err != nil {
-			return nil, toputils.TeeErrorf("IQR.ReadColumns: cannot get values for cname: %s; err: %v", cname, err)
+		var values []utils.CValueEnclosure
+		if _, exist := allColumns[cname]; !exist {
+			values = make([]utils.CValueEnclosure, iqr.NumberOfRecords())
+			for i := range values {
+				values[i] = *backfillCVal
+			}
+		} else {
```
AndrewHess, line 274: I don't think we really need to backfill—that's going to use more memory, and the caller probably doesn't need the backfilled values. Just adding another return value listing the columns that don't exist should be enough info for the caller.

pkg/segment/query/processor/headcommand_test.go
```diff
@@ -19,13 +19,36 @@ package processor
...
+	"math"
 	"testing"
 
+	"github.com/siglens/siglens/pkg/segment/query/iqr"
 	"github.com/siglens/siglens/pkg/segment/structs"
 	"github.com/siglens/siglens/pkg/segment/utils"
 	"github.com/stretchr/testify/assert"
 )
 
+var BoolExpr = &structs.BoolExpr{
```
AndrewHess, line 31: nit: make this private

pkg/segment/query/processor/headcommand.go
```diff
@@ -30,11 +32,83 @@ type headProcessor struct {
...
+					p.options.Done = true
+				} else {
+					p.options.Done = true
+					break
+				}
+			}
+		}
+		row++
+		p.numRecordsSent++
+	}
```
nkunal, line 86: is there a way to simplify this multi-level if/else blocks

pkg/segment/query/processor/headcommand_test.go
vasukalariya: addressed in the latest commit.

pkg/segment/query/iqr/intermediateQueryResult.go
vasukalariya: Actually `null` option is not treated include the record if it has any column as NULL. It rather is like if an expression evaluates to NULL we would want to include it, if a `null` option is `true`. [Doc](https://docs.splunk.com/Documentation/Splunk/9.1.0/SearchReference/Head)

AndrewHess (dismissed):
Looks good; needs some conflicts resolved

AndrewHess (approved):

PR #1806, opened 2024-10-28
Timechart cmd to the New Query Pipeline

Macbeth98:
# Description
- Adds the support for timechart command when it is not a first Aggregator.
- Queries in the format will work now: `<search>  | ... | timechart ...`

# Testing
- All unit test cases pass
- Tested by running the below query
```
--data '{
    "state": "query",
    "searchText": "search group=\"group 0\" city=Boston http_method=P* | eval lat_http = round((latitude*100)/http_status,2) | eval long_http = round((longitude*100)/http_status,2) | eval latency_http = round((latency*100)/http_status,2) | fields group, http_method, lat_http, long_http, latency_http | timechart span=10m avg(lat_http) as avg_lat_http by http_method limit=100",
    "startEpoch": "now-7d",
    "endEpoch": "now",
    "indexName": "*",
    "from": 0,
    "queryLanguage": "Splunk QL"
}'
```

# Checklist:


pkg/segment/query/processor/timechartcommand.go
```diff
@@ -18,26 +18,249 @@
...
+	measureResults := make([]utils.CValueEnclosure, len(internalMops))
+
+	for i := 0; i < numOfRecords; i++ {
+		bucketKeyBufIdx := 0
+		var groupByColVal string
+
+		record := inputIQR.GetRecord(i)
+		ts, err := record.GetTimestamp()
+		if err != nil {
+			p.errorData.readColumns["timestamp"] = err
```
AndrewHess, line 145: Shouldn't you read the timestamp name from config instead of hardcoding?

AndrewHess (commented):
Looks good; just a couple nits

pkg/segment/query/processor/dataprocessor.go
```diff
@@ -357,10 +358,10 @@ func NewStreamstatsDP(options *structs.StreamStatsOptions) *DataProcessor {
 	}
 }
 
-func NewTimechartDP(options *structs.TimechartExpr) *DataProcessor {
+func NewTimechartDP(options *structs.QueryAggregators, queryTimeRange *dtypeutils.TimeRange, qid uint64) *DataProcessor {
 	return &DataProcessor{
```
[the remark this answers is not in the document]
Macbeth98: Yes, I need to pass the `qid` to the constructor, so that it does some pre-populating. But it is mainly for logging purposes.

I have pushed all these params as `timechartOptions` struct field and I am passing just options to `NewTimechartDP`

pkg/segment/query/iqr/record.go
```diff
@@ -43,3 +43,22 @@ func (record *Record) ReadColumn(cname string) (*utils.CValueEnclosure, error) {
 
 	return &values[record.index], nil
 }
+
+func (record *Record) GetTimestamp() (uint64, error) {
```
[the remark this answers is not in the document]
Macbeth98: Yeah, no I do not need this. I have removed it.

pkg/segment/aggregations/timechartagg.go
```diff
@@ -262,32 +264,49 @@ func SortTimechartRes(timechart *structs.TimechartExpr, results *[]*structs.Buck
...
+			log.Errorf("SortTimechartRes: bucketKey is invalid for index %d: %v", i, err1)
 			return false
 		}
-
-		bucketKey2, ok := (*results)[j].BucketKey.(string)
-		if !ok {
-			log.Errorf("SortTimechartRes: cannot convert bucketKey to string: %v", (*results)[j].BucketKey)
+		timestamp2, err2 := extractTimestamp((*results)[j].BucketKey)
+		if err2 != nil {
+			log.Errorf("SortTimechartRes: bucketKey is invalid for index %d: %v", j, err2)
```
AndrewHess, line 274: nit: I think you could have just used `err` twice instead of `err1` and `err2`, but that's pretty minor, so this is fine

pkg/segment/query/iqr/intermediateQueryResult.go
```diff
@@ -642,7 +643,11 @@ func mergeMetadata(iqrs []*IQR) (*IQR, error) {
 }
 
 func (iqr *IQR) AddColumnsToDelete(cnames map[string]struct{}) {
+	timestampKey := config.GetTimeStampKey()
 	for cname := range cnames {
+		if cname == timestampKey {
+			continue
+		}
```
AndrewHess, line 650: We don't ever want to delete the timestamp field from an IQR? That doesn't seem right. E.g., a `fields -timestamp` command should delete it, right? If we want to enforce this for certain commands like timechart, it would probably be better to enforce it in that command processor instead of in the IQR

Macbeth98: Ahh, yes got it. I will remove this change.

PR #1818, opened 2024-10-30
Exclude Null Value Columns in Search Response

sonamgupta21:
# Description
Added an `includeNulls` flag in the search response. By default, it is set to false to exclude columns with only null values from the response.

# Testing

# Checklist:


pkg/ast/pipesearch/searchHandler.go
```diff
@@ -479,10 +497,30 @@ func convertRRCsToJSONResponse(rrcs []*sutils.RecordResultContainer, sizeLimit u
 		return allJsons, allCols, err
 	}
 
+	// Filter out null values from each record
+	filteredJsons := make([]map[string]interface{}, len(allJsons))
+	for i, record := range allJsons {
```
nkunal, line 502: creating a copy of every record and removing the null values from the copy is going to be expensive (memory and cpu wise). 
we should just modify the allJsons map, removing things that are null

nkunal (commented):
have we taken care of both the API and WS based searches  for this new flag ?

pkg/ast/pipesearch/searchHandler.go
```diff
@@ -204,9 +205,24 @@ func ParseSearchBody(jsonSource map[string]interface{}, nowTs uint64) (string, u
...
+		includeNulls = false
+	} else {
+		switch val := includeNullsVal.(type) {
+		case bool:
+			includeNulls = val
+		case string:
+			includeNulls = val == "true"
+		default:
+			includeNulls = false
+		}
```
AndrewHess, line 220: You should also have a log message here, saying we got an unexpected type, and specifying what type it is

pkg/ast/pipesearch/searchHandler.go
```diff
@@ -482,6 +500,17 @@ func convertRRCsToJSONResponse(rrcs []*sutils.RecordResultContainer, sizeLimit u
...
 		allJsons = allJsons[:sizeLimit]
 	}
+
+	if !includeNulls {
+		for _, record := range allJsons {
+			for key, value := range record {
+				if value == nil {
+					delete(record, key)
+				}
+			}
```
AndrewHess, line 510: This should work for the current pipeline, but I don't think the new pipeline calls this function (I might be wrong). Can you check?

AndrewHess (commented):
Looks pretty good

pkg/ast/pipesearch/searchHandler.go
sonamgupta21: It does work for the new pipeline. Could you please let me know where I need to make changes for the new pipeline?

pkg/segment/query/processor/queryprocessor.go
```diff
@@ -265,7 +267,7 @@ func (qp *QueryProcessor) GetFullResult() (*structs.PipeSearchResponseOuter, err
 // 3. Read from the update channel and the final result channel.
 //
 // Once the final result is sent, no more updates will be sent.
-func (qp *QueryProcessor) GetStreamedResult(stateChan chan *query.QueryStateChanData) error {
+func (qp *QueryProcessor) GetStreamedResult(stateChan chan *query.QueryStateChanData, includeNulls bool) error {
```
AndrewHess, line 270: nit: you already added `includeNulls` as a field on `QueryProcessor`, so you probably shouldn't pass it as a parameter here; use `qp.includeNulls` instead

pkg/segment/query/iqr/intermediateQueryResult.go
```diff
@@ -964,6 +964,9 @@ func (iqr *IQR) AsResult(qType structs.QueryType) (*structs.PipeSearchResponseOu
 	for i, record := range cValRecords {
 		recordsAsAny[i] = make(map[string]interface{})
 		for key, value := range record {
+			if !includeNulls && value.CVal == nil {
```
AndrewHess, line 967: nit: 
```suggestion
			if !includeNulls && value.IsNull() {
```
This handles a few more edge cases

AndrewHess (commented):
Looking good

nkunal (approved):

nkunal (approved):

PR #1827, opened 2024-10-31
fix: change timechart test query to large enough span and add Time modifier to snap back to hour

Macbeth98:
# Description
- Fixed an issue with Time modifier parsing.
- Added time modifier to the command to snap back to hour.

# Testing
- All test cases pass

# Checklist:


tools/sigclient/functionalQueries/eval_fields_timechart.json
```diff
@@ -1,5 +1,5 @@
 {
-    "queryText": "search group=\"group 0\" city=Boston http_method=P* | eval lat_http = round((latitude*100)/http_status,2) | eval long_http = round((longitude*100)/http_status,2) | eval latency_http = round((latency*100)/http_status,2) | fields group, http_method, lat_http, long_http, latency_http | timechart span=10m avg(lat_http) as avg_lat_http by http_method limit=100",
+    "queryText": "search group=\"group 0\" city=Boston http_method=P* | earliest=-12h@h | eval lat_http = round((latitude*100)/http_status,2) | eval long_http = round((longitude*100)/http_status,2) | eval latency_http = round((latency*100)/http_status,2) | fields group, http_method, lat_http, long_http, latency_http | timechart span=30m avg(lat_http) as avg_lat_http by http_method limit=100",
```
AndrewHess, line 2: I might be misunderstanding how this works. It looks like you set the query start time to 12 hours ago, and snap to the top of the hour, so then timechart will start with the first bucket from that time until 30 minutes after that, and the next bucket will be the next 30 minutes, and so on. And the goal of this is to have just one bucket contain all the ingested data. It seems to me that with snapping, you could still have data get into two buckets, but if you don't snap (so just `earliest=-12h`) and assume all the data gets ingested within 30 minutes, then you'll always get the data in one bucket. So why do you snap?

AndrewHess (approved):

PR #1836, opened 2024-11-01
Fix totalMatched for segStatsCmd and groupByCmd.

vasukalariya:
# Description
Summarize the change.
- totalMatched returned for segStatsCmd and groupByCmd should be the value of initial search
- In the old pipeline the totalMatched was correct for groupByCmd but for segStatsCmd it was 0, now it should return the count of initial search match.

# Testing
Enabled the 4 queries that are testing this scenario in ingest.csv

# Checklist:

### DIFF
--- a/pkg/segment/query/querystatus.go
+++ b/pkg/segment/query/querystatus.go
@@ -245,6 +245,21 @@ func AssociateSearchInfoWithQid(qid uint64, result *segresults.SearchResults, ag
 	return nil
 }
 
+func AssociateSearchResult(qid uint64, result *segresults.SearchResults) error {
+	arqMapLock.RLock()
+	rQuery, ok := allRunningQueries[qid]
+	arqMapLock.RUnlock()
+	if !ok {
+		return putils.TeeErrorf("AssociateSearchResult: qid %+v does not exist!", qid)
+	}
+
+	rQuery.rqsLock.Lock()
+	rQuery.searchRes = result
+	rQuery.rqsLock.Unlock()
+
+	return nil
+}
+
 // increments the finished segments. If incr is 0, then the current query is finished and a histogram will be flushed
 func IncrementNumFinishedSegments(incr int, qid uint64, recsSearched uint64,
 	skEnc uint16, remoteId string, doBuckPull bool, sstMap map[string]*structs.SegStats) {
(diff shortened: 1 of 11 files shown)


pkg/segment/query/querystatus.go
```diff
@@ -1164,3 +1179,15 @@ func InitScrollFrom(qid uint64, scrollFrom uint64) error {
 
 	return nil
 }
+
+func ConvertQueryCountToTotalResponse(qc *structs.QueryCount) interface{} {
```
AndrewHess, line 1183: Looks like when you moved this function, you also changed it to always return a `HitsCount` type. I think that's a good change, but you should also change the return type from `interface{}` now

pkg/segment/query/querystatus.go
```diff
@@ -245,6 +245,21 @@ func AssociateSearchInfoWithQid(qid uint64, result *segresults.SearchResults, ag
 	return nil
 }
 
+func AssociateSearchResult(qid uint64, result *segresults.SearchResults) error {
```
AndrewHess, line 248: There's already `AssociateSearchInfoWithQid()`. Do you need to add this function?

pkg/segment/results/segresults/segresults.go
```diff
@@ -221,6 +221,12 @@ func (sr *SearchResults) AddSSTMap(sstMap map[string]*structs.SegStats, skEnc ui
 	sr.updateLock.Unlock()
 }
 
+func (sr *SearchResults) AddResultCount(count uint64) {
```
AndrewHess, line 224: There's already `AddBlockResults()` that does this and also merges the block results with the existing results. Shouldn't that function be called? And then I think you don't need to add this function?

pkg/segment/query/querystatus.go
vasukalariya: This is used for groupByCmd and not segStatsCmd, we do not need block results in this case. We just need the initial count.

pkg/segment/results/segresults/segresults.go
vasukalariya: This is used for groupByCmd and not segStatsCmd, we do not need block results in this case. We just need the initial count.

pkg/segment/query/querystatus.go

AndrewHess (approved):

PR #1849, opened 2024-11-05
Fix where crash for a non existent column

vasukalariya:
# Description
Summarize the change.
- If a column does not exist and its present in the where clause we will just return empty result.

TODO: When we correct the evaluation to use different data types we can return the backfilled value and evaluate the conditions properly.

# Testing
Earlier a query like this was crashing, now it should not
`* | where invalidCol="abc"`

# Checklist:

### DIFF
--- a/pkg/segment/query/processor/wherecommand.go
+++ b/pkg/segment/query/processor/wherecommand.go
@@ -23,6 +23,7 @@ import (
 	"github.com/siglens/siglens/pkg/segment/query/iqr"
 	"github.com/siglens/siglens/pkg/segment/structs"
 	segutils "github.com/siglens/siglens/pkg/segment/utils"
+	"github.com/siglens/siglens/pkg/utils"
 	log "github.com/sirupsen/logrus"
 )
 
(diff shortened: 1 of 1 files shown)


pkg/segment/query/processor/wherecommand.go
```diff
@@ -42,8 +43,16 @@ func (p *whereProcessor) Process(iqr *iqr.IQR) (*iqr.IQR, error) {
...
-			log.Errorf("where.Process: cannot get field values; field=%s; err=%v", field, err)
-			return nil, err
+			return nil, utils.TeeErrorf("where.Process: cannot get field values; field: %s; err: %v", field, err)
+		}
+		if values == nil {
+			// The field is not present, discard all the rows
+			// TODO: Fix evaluation for non existing fields
+			err := iqr.Discard(iqr.NumberOfRecords())
+			if err != nil {
+				return nil, utils.TeeErrorf("where.Process: Error while discarding rows due to col %v not found, err: %v", field, err)
```
AndrewHess, line 53: nit: you could also just return `nil, nil`, or make a new empty IQR and return that, but this also works

AndrewHess (approved):

PR #1862, opened 2024-11-06
Limit number of running queries

vasukalariya:
# Description
Summarize the change.
- Limit the number of running queries to 100.

# Testing
- Tested by running parallel queries greater than the limit.
- Verified that siglens only accepts 100 queries and will reject the rest of them.
- Added delay to check the scenario where more than 100 queries are sent but with delay to ensure that only running queries are considered.

# Checklist:

### DIFF
--- a/pkg/segment/query/querystatus.go
+++ b/pkg/segment/query/querystatus.go
@@ -170,7 +170,7 @@ func StartQuery(qid uint64, async bool, cleanupCallback func()) (*RunningQuerySt
 		return nil, fmt.Errorf("qid has already been started")
 	}
 
-	if len(allRunningQueries) == MAX_RUNNING_QUERIES {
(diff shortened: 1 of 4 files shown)


pkg/segment/query/querystatus.go
```diff
@@ -169,6 +170,10 @@ func StartQuery(qid uint64, async bool, cleanupCallback func()) (*RunningQuerySt
 		return nil, fmt.Errorf("qid has already been started")
 	}
 
+	if len(allRunningQueries) == MAX_RUNNING_QUERIES {
```
nkunal, line 173: nit: this should also work, but typically in such situations I prefer to use `>` instead of exact comparison

nkunal (commented):
great job cleaning up the leaked queries. 
one small nit

pkg/segment/query/querystatus.go

nkunal (approved):

PR #1864, opened 2024-11-06
Dedup with sortby

vasukalariya:
# Description
Summarize the change.
- Current implementation in the old pipeline for dedup with sortby is incorrect because it accounts for the existing record order and not the sort order to perform dedup.
- If a sortby clause is present in dedup, the dedup should be performed as if dedup is getting performed on the data that is sorted according to the sortby clause.
- Look at the unit test for more details.
- Have also added functional and longer functional test for this.

# Testing

# Checklist:

### DIFF
--- a/pkg/ast/spl/spl.peg
+++ b/pkg/ast/spl/spl.peg
@@ -1977,8 +1977,10 @@ ByClause <- BY fieldList:FieldNameList {
 
 DedupBlock <- PIPE CMD_DEDUP dedupExpr:DedupExpr {
 
+    dedupExp := dedupExpr.(*structs.DedupExpr)
+
     letColReq := &structs.LetColumnsRequest {
-        DedupColRequest: dedupExpr.(*structs.DedupExpr),
+        DedupColRequest: dedupExp,
     }
 
     root := &structs.QueryAggregators {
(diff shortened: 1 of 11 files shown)


pkg/ast/spl/spl.peg
```diff
@@ -1977,16 +1977,41 @@ ByClause <- BY fieldList:FieldNameList {
...
         },
-        DedupExpr: dedupExpr.(*structs.DedupExpr),
+        DedupExpr: dedupExp,
+    }
+
+    if len(dedupExp.DedupSortEles) > 0 {
+        sortExpr := &structs.SortExpr {
+            SortEles: dedupExp.DedupSortEles,
+            SortAscending: dedupExp.DedupSortAscending,
+            Limit: 1000000,
```
AndrewHess, line 1998: nit: this is 1 million, but your PR description said 100k. Which do we want?

AndrewHess: Also, we can handle `dedup sortby` more efficiently than converting it to `sort | dedup`; we may want to do that optimization later. With the way you're doing this, we could end up with too few results (e.g., all the top 1 million rows have the same value, so then only 1 row is returned from dedup, but if the dataset has other rows with different values we should actually return more than 1 row)

AndrewHess (commented):
Looks pretty good

pkg/ast/spl/spl.peg
vasukalariya: Yeah I think 1M is fine, yeah we can do the optimization later.

AndrewHess (approved):

PR #1870, opened 2024-11-07
Add Streamstats CMD for the New Query Pipeline

sonamgupta21:
# Description
Summarize the change.

# Testing

# Checklist:


pkg/ast/spl/spl.peg
```diff
@@ -1112,15 +1112,21 @@ StreamStatsOptionList <- first:StreamStatsOption rest:(SPACE StreamStatsOption)*
...
+    aggNode.StreamstatsExpr = aggNode.StreamStatsOptions
     return aggNode, nil
 }
 / PIPE CMD_STREAMSTATS aggs:(CommonAggregatorBlock) {
     aggNode := aggs.(*structs.QueryAggregators)
+    ssOptions := initializeStreamStatsOptions()
 
-    aggNode.StreamStatsOptions = initializeStreamStatsOptions()
-
+   ssOptions.MeasureOperations = aggNode.MeasureOperations
```
vasukalariya, line 1126: Don't we also need to initialize groupByRequest in the ssOptions ?

pkg/segment/query/processor/streamstatscommand.go
```diff
@@ -18,26 +18,942 @@
...
+	measureAggs := p.options.MeasureOperations
+	if p.options.GroupByRequest != nil {
+		measureAggs = p.options.GroupByRequest.MeasureOperations
+	}
+
+	if p.options.RunningStreamStats == nil {
+		p.options.RunningStreamStats = make(map[int]map[string]*structs.RunningStreamStatsResults)
+	}
+
+	allColumns, err := iqr.ReadAllColumns()
```
vasukalariya, line 64: Why do we need to read all columns ? The idea behind this design is to use the inputIQR and add the extra fields created accordingly using `AppendKnownValues`.

pkg/segment/query/processor/streamstatscommand.go
```diff
@@ -18,26 +18,942 @@
...
+		}
+
+		shouldResetBefore, err := evaluateResetCondition(p.options.ResetBefore, record)
+		if err != nil {
+			return nil, err
+		}
+		if shouldResetBefore {
+			resetAccumulatedStreamStats(p.options)
+			currIndex = 0
+			numPrevSegmentProcessedRecords = 0
```
vasukalariya, line 121: `numPrevSegmentProcessedRecords` is not required to be maintained, command should be oblivious to the concept of segments and how data is stored. You can assume that data will come in correct order. This was used earlier because in the older pipeline data use to come per segment and the ordering of records had to be done by the command itself. With the new pipeline IQR will have the correct order of records.

pkg/segment/query/processor/streamstatscommand.go
```diff
@@ -18,26 +18,942 @@
...
+		if err != nil {
+			return nil, err
+		}
+		if shouldResetAfter {
+			resetAccumulatedStreamStats(p.options)
+			currIndex = 0
+			numPrevSegmentProcessedRecords = 0
+		}
+
+		p.options.NumProcessedRecords++
```
vasukalariya, line 214: These parameters should be updated before reset. If reset happens we shouldn't increment them. Refer the older implementation.

pkg/segment/query/processor/streamstatscommand.go
```diff
@@ -18,26 +18,942 @@
...
+		currIndex++
+	}
+
+	// Add results to output IQR
+	if len(knownValues) > 0 {
```
vasukalariya, line 219: This would be inefficient, rather than doing append for each record, it is better to do it once all the records are processed. Also there is no need to append columns that are not part of the streamstats command, as mentioned before only create required fields and append those.

pkg/segment/query/processor/streamstatscommand.go
```diff
@@ -18,26 +18,942 @@
...
+	}
+
+	if !ssOption.Current {
+		return result, exist, nil
+	}
+
+	return latestResult, true, nil
+}
+
+func UpdateRangeStat(floatValue float64, rangeStat *structs.RangeStat) {
```
vasukalariya, line 852: This function is already present in evalaggs.go

vasukalariya:
Also consider adding unit tests for streamstats. You can refer to tests present for other commands to get an idea. Also, refer streamstats queries in functionalQueries.yml to get ideas on what scenarios to test. There are also a bunch of queries in ingest.csv.

pkg/segment/query/processor/queryprocessor.go
```diff
@@ -431,3 +436,34 @@ func createEmptyResponse(queryType structs.QueryType) *structs.PipeSearchRespons
...
+
+	for curAgg := firstAgg; curAgg != nil; curAgg = curAgg.Next {
+		if curAgg.StreamstatsExpr != nil && curAgg.StreamstatsExpr.TimeWindow != nil {
+			// Check all previous commands up to this point
+			hasSort := false
+			timeSort := false
+			timeSortAsc := true
+
+			// Check commands that come before streamstats
+			for prevAgg := firstAgg; prevAgg != curAgg; prevAgg = prevAgg.Next {
```
vasukalariya, line 450: Why do we want to iterate twice? can't we just start from firstAgg and maintain whether the sort is present or not, if so is it on timestamp.

tools/sigclient/functionalQueries/functionalQueries.yml
```diff
@@ -35,19 +35,19 @@ sort_tail_eval_newField.json
...
+streamstats_win3.json
+streamstats_win4_currF.json
+streamstats_win4_grpby.json
+streamstats_win4_reset_before_after.json
+streamstats_win2_gloF_currF_grpby.json
+streamstats_win2_gloF_grpby.json
+streamstats_multi_resets_grpby.json
+streamstats_eval.json
+streamstats_after_stats.json
+streamstats_after_grpby.json
```
vasukalariya, line 48: do consider adding results for `streamstats_without_sort.json` for both functional and longerFunctional

vasukalariya (approved):
Looks good!

PR #1874, opened 2024-11-07
fix: panic on bucket key buffer for Stats Group By CMD

Macbeth98:
# Description
- The bucket key on the group by is paniced because the buffer allocated for the bucket key is not enough.
- In this PR, it will check the size being written and if that size is not enough then it will increment the buffer.

# Testing
- Tested by making a new unit test that checked whether the buffer size is being incremented according to column sizes.
- All unit and functional test cases pass.

# Checklist:

### DIFF
--- a/pkg/segment/query/processor/statscommand_test.go
+++ b/pkg/segment/query/processor/statscommand_test.go
@@ -408,7 +408,7 @@ func Test_ProcessGroupByRequestFullBuffer(t *testing.T) {
 	_, err = processor.Process(iqr1)
 	assert.NoError(t, err)
 
-	// value of col1 at eaach record is utils.MAX_RECORD_SIZE bytes
+	// value of col1 at each record is utils.MAX_RECORD_SIZE * factor bytes
 	// +
 	// value of col3 at each record is 1 byte
 	estimatedSizeOfBuffer := (utils.MAX_RECORD_SIZE * factorSize) + 1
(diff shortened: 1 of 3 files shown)


pkg/segment/query/processor/statscommand_test.go
```diff
@@ -379,3 +379,40 @@ func Test_ProcessSegmentStats(t *testing.T) {
...
+
+	knownValues := getTestData()
+	processor := getGroupByProcessor()
+	assert.Equal(t, 0, len(processor.bucketKeyWorkingBuf))
+
+	col1Values := knownValues["col1"]
+	factorSize := 5
+
+	for i := 0; i < len(col1Values); i++ {
+		byteVal := make([]byte, utils.MAX_RECORD_SIZE*3)
```
AndrewHess, line 395: Where is this 3 coming from? Is it based off the `factorSize`? Using `factorSize-1` could make that more clear

Also, is this `factorSize` arbitrary, or is it based on some other property of the test?

AndrewHess (commented):
Looks pretty good

pkg/segment/query/processor/statscommand_test.go
Macbeth98: oh my bad, that 3 was supposed to be `factorSize` when I was testing I created the variable and set it to 5 but forgot to replace it here.

Macbeth98: yes, factorSize is arbitary, but basically any value greater than the len(groupbyColumns). i will add that as a comment

AndrewHess (approved):

PR #1878, opened 2024-11-08
Wait queue to manage query workload dynamically

vasukalariya:
# Description
Summarize the change.
- Currently, we are running the queries using a fix limit say 32, which can cause starvation of queries. We should only have active running queries to be based on the number of processors available.
- Other queries can remain in the wait queue with some fixed size (currently set to 100).
- Other queries would be dropped.
- We should also consider the amount of memory available to run these queries (we shouldn't run it even if processor is available).
- Currently, assuming the memory required for a query to be 100MB. (This is a casual check, since all queries won't be heavy and for random workload this shouldn't be a problem).
- I have used channel to avoid explicit locking and make the implementation cleaner.

# Testing
- Ran multiple queries concurrently to verify that only max waiting queries are taken.
- Also monitored active query count never to be greater than available cores.

# Checklist:


pkg/segment/query/querystatus.go
```diff
@@ -42,7 +44,10 @@ var numStates = 4
 
 const MAX_GRP_BUCKS = 3000
 const CANCEL_QUERY_AFTER_SECONDS = 5 * 60 // If 0, the query will never timeout
-const MAX_RUNNING_QUERIES = 32
+var MAX_RUNNING_QUERIES = runtime.GOMAXPROCS(0)
+
+const MAX_WAITING_QUERIES = 100
+const QUERY_MEM = 200 * 1024 * 1024 // 200MB
```
nkunal, line 50: make this mem configurable, so that operators can tune this for differently sized systems

pkg/segment/query/querystatus.go
```diff
@@ -176,18 +190,65 @@ func StartQuery(qid uint64, async bool, cleanupCallback func()) (*RunningQuerySt
...
+		rQuery.timeoutCancelFunc()
+
+		if rQuery.cleanupCallback != nil {
+			rQuery.cleanupCallback()
+		}
+	}
+}
+
+func canRunQuery() bool {
+	activeQueries := GetActiveQueryCount()
```
nkunal, line 233: we should log some error here in this scenario:
If there are zero active queries, and the total mem is less than what is needed and we are not able to return true from this function. So that the operator can tune the query_mem on systems that are smaller and have lesser data

pkg/segment/query/querystatus.go
```diff
@@ -176,18 +190,65 @@ func StartQuery(qid uint64, async bool, cleanupCallback func()) (*RunningQuerySt
...
+func PullQueriesToRun() {
+	for {
+		if canRunQuery() {
+			select {
+			case wsData := <-waitingQueries:
+				runQuery(wsData)
+			default:
+			}
+		}
+		time.Sleep(10 * time.Millisecond)
```
AndrewHess, line 246: nit: make this a constant at the top of the file

pkg/segment/query/querystatus.go
```diff
@@ -176,18 +190,65 @@ func StartQuery(qid uint64, async bool, cleanupCallback func()) (*RunningQuerySt
...
+		startTime:         time.Now(),
+		StateChan:         stateChan,
+		cleanupCallback:   cleanupCallback,
+		rqsLock:           &sync.Mutex{},
+		isAsync:           async,
+		timeoutCancelFunc: nil,
+	}
+
+	select {
+	case waitingQueries <- WaitStateData{qid, runningState}:
```
AndrewHess, line 205: Using a channel works for now. But soon we'll want to show how many queries are ahead of this one, and be able to remove a query from the queue if the user cancels the query. How do you plan on implementing that while using a channel here?

pkg/segment/query/querystatus.go

pkg/segment/query/querystatus.go

pkg/segment/query/querystatus.go

pkg/segment/query/querystatus.go
vasukalariya: Have changed the implementation to use slices with locks instead of channels. Could've used list but this implementation is much cleaner and copying a small slice (100 in this case) wouldn't affect performance.

pkg/config/common/common.go
```diff
@@ -89,6 +89,7 @@ type MemoryConfig struct {
 	CMIPercent      uint64 `yaml:"microIndexPercent"`
 	MetadataPercent uint64 `yaml:"metadataPercent"`
 	MetricsPercent  uint64 `yaml:"metricsPercent"`
+	MemoryPerQuery  uint64 `yaml:"memPerQuery"`
```
AndrewHess, line 92: nit: specify units (bytes, MB, etc.)

vasukalariya: addressed in the latest commit.

pkg/config/config.go
```diff
@@ -55,6 +55,7 @@ const DEFAULT_ROTATED_CMI_MEM_PERCENT = 63
 const DEFAULT_METADATA_MEM_PERCENT = 20
 const DEFAULT_SEG_SEARCH_MEM_PERCENT = 15 // minimum percent allocated for segsearch
 const DEFAULT_METRICS_MEM_PERCENT = 2
+const DEFAULT_MEM_PER_QUERY_BYTES = 100 * 1024 * 1024 // 100MB
```
nkunal, line 58: lets make this 200 MB based on recent tests

pkg/segment/query/querystatus.go
```diff
@@ -66,17 +70,40 @@ type QueryStateChanData struct {
...
+func InitMaxRunningQueries() {
+	memConfig := config.GetMemoryConfig()
+	totalMemoryInBytes := config.GetTotalMemoryAvailable()
+	searchMemoryInBytes := (totalMemoryInBytes * memConfig.SearchPercent) / 100
+	maxConcurrentQueries := searchMemoryInBytes / memConfig.MemoryPerQueryInBytes
+	if maxConcurrentQueries < 1 {
+		log.Warnf("System does not have sufficient memory for query. Setting maxConcurrentQueries to 1, searchMemoryInBytes: %v, MemoryPerQueryInBytes: %v",
+			searchMemoryInBytes, memConfig.MemoryPerQueryInBytes)
+		maxConcurrentQueries = 1
+	}
```
nkunal, line 97: we should remove this check/error and if the maxConcurreQuery num comes out to be less than 2, then set it 2

pkg/segment/query/querystatus.go
```diff
@@ -176,18 +206,70 @@ func StartQuery(qid uint64, async bool, cleanupCallback func()) (*RunningQuerySt
...
+	return activeQueries < MAX_RUNNING_QUERIES
+}
+
+func PullQueriesToRun() {
+	for {
+		if canRunQuery() {
+			waitingQueriesLock.Lock()
+			if len(waitingQueries) == 0 {
+				waitingQueriesLock.Unlock()
+				continue
```
nkunal, line 260: when this go routine is lanuched, if there are no queries in the system, then there is spiining loop happening here, and 1 cpu core will be at 100% even though no queries are running at all.

nkunal: we should add a sleep even here before doing continue. 
@AndrewHess thoughts ?

pkg/segment/query/querystatus.go
```diff
@@ -42,7 +43,10 @@ var numStates = 4
 
 const MAX_GRP_BUCKS = 3000
 const CANCEL_QUERY_AFTER_SECONDS = 5 * 60 // If 0, the query will never timeout
-const MAX_RUNNING_QUERIES = 32
+var MAX_RUNNING_QUERIES = uint64(runtime.GOMAXPROCS(0))
+
+const PULL_QUERY_INTERVAL = 10 * time.Millisecond
+const MAX_WAITING_QUERIES = 100
```
nkunal, line 49: we can set this to 500

pkg/config/config.go
vasukalariya: addressed in the latest commit.

pkg/segment/query/querystatus.go
vasukalariya: addressed in the latest commit.

pkg/segment/query/querystatus.go
vasukalariya: addressed in the latest commit.

pkg/segment/query/querystatus.go

pkg/config/common/common.go
```diff
@@ -85,10 +85,11 @@ type MemoryConfig struct {
...
 	// These should sum to 100.
-	SearchPercent   uint64 `yaml:"searchPercent"`
-	CMIPercent      uint64 `yaml:"microIndexPercent"`
-	MetadataPercent uint64 `yaml:"metadataPercent"`
-	MetricsPercent  uint64 `yaml:"metricsPercent"`
+	SearchPercent         uint64 `yaml:"searchPercent"`
+	CMIPercent            uint64 `yaml:"microIndexPercent"`
+	MetadataPercent       uint64 `yaml:"metadataPercent"`
+	MetricsPercent        uint64 `yaml:"metricsPercent"`
+	MemoryPerQueryInBytes uint64 `yaml:"memPerQueryInBytes"`
```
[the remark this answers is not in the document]
vasukalariya: addressed in the latest commit.

pkg/segment/query/querystatus.go
```diff
@@ -206,35 +287,23 @@ func StartQuery(qid uint64, async bool, cleanupCallback func()) (*RunningQuerySt
...
 
-		if rQuery.cleanupCallback != nil {
-			rQuery.cleanupCallback()
-		}
-	}
+	wsData.rQuery.timeoutCancelFunc = setupTimeoutCancelFunc(wsData.qid)
+	wsData.rQuery.startTime = time.Now()
+
+	wsData.rQuery.StateChan <- &QueryStateChanData{StateName: READY}
+	wsData.rQuery.StateChan <- &QueryStateChanData{StateName: RUNNING}
```
AndrewHess, line 306: What's the purpose of this READY state if we immediately send RUNNING afterwards? Is READY sent somewhere else as well?

AndrewHess (commented):
Looks good; one question

pkg/segment/query/querystatus.go
vasukalariya: No, READY state is only meant to signal that query can now proceed. RUNNING state update is actually send over the websocket.

AndrewHess: Ah it looks like READY is supposed to be received by one function to actually start the query and RUNNING is received by the websocket handler (to send to the UI like you said); is that right? In that case, could there be a race condition where the websocket handler receives READY instead of the segexecution function receiving it, and then the query is never started?

vasukalariya: No it would never happen since websocket is only initialized after we call `ExecuteAsyncQueryForNewPipeline` we are blocked until we receive a READY state in this method.

AndrewHess (approved):

PR #1881, opened 2024-11-08
Equality Test tool

vasukalariya:
# Description
Summarize the change.

Use the following command to run
```
go run main.go equalityTest -d localhost:5122
```
This command will do the following
- Get the total number of records
- Get all columns
- Check the cardinality if the cardinality is less than 300, it would validate those values.
- It would get the count for equality and not-equality and validate whether the sum is equal to total records or not.
- At the end it would print the skipped cols, success cols, error cols.

If required we can work more on this.

# Testing

# Checklist:

### DIFF
--- a/tools/sigclient/cmd/sigclient.go
+++ b/tools/sigclient/cmd/sigclient.go
@@ -153,6 +153,20 @@ var functionalTestCmd = &cobra.Command{
 	},
 }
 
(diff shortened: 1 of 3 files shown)


vasukalariya:
This PR for just for testing. Not required anymore.

PR #1883, opened 2024-11-11
Fix not equals

vasukalariya:
# Description
Summarize the change.
For the operator `NotEquals`, we cannot do a rangeCheck directly on minVal and maxVal, for example if a block has different minVal and maxVal, we cannot return that block as false, since there are records in that block which still match for operator `NotEquals`. The only case where we can reject the block is when all minVal, maxVal and lookupVal are same.

# Testing
`http_status != 200 | stats count` was not giving any results before for benchmark dataset.
Now it works as expected.

# Checklist:


pkg/segment/query/metadata/metautils/metacheckers.go
```diff
@@ -106,7 +106,7 @@ func doesUintPassRangeFilter(op utils.FilterOperator, lookupValue uint64, minVal
 	case utils.Equals:
 		return lookupValue >= minVal && lookupValue <= maxVal
 	case utils.NotEquals:
-		if lookupValue == minVal || lookupValue == maxVal {
+		if minVal == maxVal && lookupValue == minVal {
```
AndrewHess, line 109: I think your change is correct, but not as restrictive as it could be. I think `utils.NotEquals` should be the same as `not utils.Equals`, which is
```
not (lookupValue >= minVal && lookupValue <= maxVal)
```
which is
```
lookupValue < minVal || lookupValue > maxVal
```
Am I missing something?

AndrewHess: Also, can you check the other inequalities? They look odd to me but I'm not quite convinced if they're right/wrong. There's a basic unit test for this function, but it would be good to add some more tests for it to catch edge cases like the one you found

vasukalariya: Wouldn't this condition `lookupValue < minVal || lookupValue > maxVal` give incorrect results for the case where lookupValue is 300, minVal is 200 and maxVal is 500. It would give `false` for that block even though there are records satisfying this condition.

AndrewHess: Ah, I think I was misunderstanding exactly how this function is supposed to behave. Now I think this is the expected behavior:

Returns true if there exists an `actualValue` with `minValue <= actualValue <= maxValue` such that `actualValue op lookupValue` is true

Is that what these functions are supposed to do? Can you please add a comment to the code explaining what the function does, and then I'll approve this PR

vasukalariya: Yes that is correct.

AndrewHess (approved):

PR #1885, opened 2024-11-13
Fix ingestion (col values gets overwritten due to parsing buffer)

vasukalariya:
# Description
Summarize the change.
- Do not use parsing buffer if there is an escape character.

# Testing

# Checklist:

### DIFF
--- a/pkg/segment/writer/logpacker.go
+++ b/pkg/segment/writer/logpacker.go
@@ -18,6 +18,7 @@
 package writer
 
 import (
+	"bytes"
 	"fmt"
 
 	jp "github.com/buger/jsonparser"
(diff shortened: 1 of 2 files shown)


pkg/segment/writer/logpacker.go
```diff
@@ -167,7 +166,8 @@ func parseSingleString(key string, tsKey *string, valBytes []byte, ple *ParsedLo
 	cbufidx += 1
 	n := uint16(len(valBytes))
 	utils.Uint16ToBytesLittleEndianInplace(n, ple.allCvalsTypeLen[ple.numCols][cbufidx:])
-	ple.allCvals[ple.numCols] = valBytes
+	ple.allCvals[ple.numCols] = make([]byte, n)
+	copy(ple.allCvals[ple.numCols], valBytes)
```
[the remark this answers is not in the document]
AndrewHess: I agree. Is it relatively easy to add a unit test or some other automated test for this that fails without this PR but passes with this PR?

AndrewHess (approved):

pkg/segment/writer/logpacker.go
```diff
@@ -113,7 +119,6 @@ func parseNonJaegerRawJsonArray(currKey string, data []byte, tsKey *string,
 				return
 			}
 		case jp.String:
-
 			valUnescaped, encErr := jp.Unescape(value, jsParsingStackbuf[:])
```
AndrewHess, line 122: Do you need a similar fix here like you added in `parseRawJsonObject()`?

pkg/segment/writer/logpacker.go
```diff
@@ -51,12 +52,17 @@ func ParseRawJsonObject(currKey string, data []byte, tsKey *string,
 				return fmt.Errorf("parseRawJsonObject: arr currKey: %v, err: %v", currKey, err)
 			}
 		case jp.String:
-			valUnescaped, err := jp.Unescape(value, jsParsingStackbuf[:])
-			if err != nil {
-				return fmt.Errorf("parseRawJsonObject: failed to unescape currKey: %v, err: %v",
-					currKey, err)
+			firstBackslash := bytes.IndexByte(value, '\\')
+			if firstBackslash != -1 {
+				valUnescaped, err := jp.Unescape(value, nil)
```
AndrewHess, line 57: I'm not seeing how the new code will ever give a different result than the old code, and I'm not really sure why we're doing this. Maybe add a comment about the purpose of this code (i.e., what issue you're fixing). Also, are you able to add some automated test so we don't regress on this issue? It seems like a pretty subtle bug

pkg/segment/writer/logpacker.go
vasukalariya: addressed in the latest commit.

nkunal (approved):

PR #1889, opened 2024-11-14
fix: processing panic on cleanup

Macbeth98:
# Description
- The stats command was panicing because: when we call cancel when a query is in still execution, it would call cleanup and the cleanup would remove the resources, but the actual command processing is still using those resources, which causes the panic.
- And this type of panic can happen to any command data processor during execution.
- Now I have made a flag to know whether to remove the resources or wait and remove the resources.

**UPDATE:**
- I have reverted my previous commit and added a new fields called `processorLock` in the `DataProcessor` struct and call `processorLock.Lock()` before calling `dp.Process()` and also again acquire lock before calling `dp.Cleanup()`.

# Testing
- Tested by executing this command which will panic on develop branch but will not on this PR.
```
CounterID = 38 AND IsRefresh = 0 | where strptime(EventDate,"%Y-%m-%d") >= strptime("2013-07-01", "%Y-%m-%d") AND strptime(EventDate,"%Y-%m-%d") <= strptime("2013-07-31", "%Y-%m-%d") | eval Src=if(SearchEngineID=0 AND AdvEngineID=0, Referer, "") | rename URL as Dst | stats count as PageViews by TrafficSourceID, SearchEngineID, AdvEngineID, Src, Dst | sort -PageViews | head 1010 | tail 10 | tail 10
```

# Checklist:

### DIFF
--- a/pkg/segment/query/iqr/intermediateQueryResult.go
+++ b/pkg/segment/query/iqr/intermediateQueryResult.go
@@ -303,6 +303,14 @@ func (iqr *IQR) readColumnInternal(cname string) ([]utils.CValueEnclosure, error
 
 	switch iqr.mode {
 	case withRRCs:
+		// Check if the column was renamed and use the old name
+		// for reading it from the RRCs.
+		for oldName, newName := range iqr.renamedColumns {
+			if newName == cname {
+				cname = oldName
+				break
+			}
+		}
 		return iqr.readColumnWithRRCs(cname)
 	case withoutRRCs:
 		// We don't have RRCs, so we can't read the column. Since we got here
@@ -455,8 +463,14 @@ func (iqr *IQR) readColumnWithRRCs(cname string) ([]utils.CValueEnclosure, error
 			len(iqr.rrcs), len(results))
 	}
 
+	finalCname := cname
+	// Check if the column was renamed and use the new Cname for the results.
+	if newColName, ok := iqr.renamedColumns[cname]; ok {
+		finalCname = newColName
+	}
+
 	// TODO: should we have an option to disable this caching?
-	iqr.knownValues[cname] = results
+	iqr.knownValues[finalCname] = results
 
 	return results, nil
 }
--- a/pkg/segment/query/processor/dataprocessor.go
+++ b/pkg/segment/query/processor/dataprocessor.go
@@ -115,11 +115,10 @@ func (dp *DataProcessor) Fetch() (*iqr.IQR, error) {
 				return nil, utils.TeeErrorf("DP.Fetch: failed to fetch input: %v", err)
 			}
 
+			dp.processorLock.Lock()
 			if dp.isCleanupCalled {
 				return nil, io.EOF
 			}
-
-			dp.processorLock.Lock()
 			output, err = dp.processor.Process(input)
 			dp.processorLock.Unlock()
 
--- a/pkg/segment/query/processor/dataprocessor.go
+++ b/pkg/segment/query/processor/dataprocessor.go
@@ -45,7 +45,8 @@ type DataProcessor struct {
 	isTwoPassCmd      bool // A subset of bottleneck commands.
 	finishedFirstPass bool // Only used for two-pass commands.
 
-	processorLock *sync.Mutex
+	processorLock   *sync.Mutex
+	isCleanupCalled bool
 }
 
 func (dp *DataProcessor) DoesInputOrderMatter() bool {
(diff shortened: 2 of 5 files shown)


pkg/segment/query/processor/statscommand.go
AndrewHess, line 1: It seems like you're using some bool fields to emulate a lock; wouldn't using a lock be cleaner and more robust (e.g., in your cleanup code, there's a race condition where `isExecuting==false` so you call `cleanupResources()`, but it's possible that by the time you actually make that call (or sometime inside that call) `isExecuting` becomes true again)?

Also, if you're going to have to make similar changes for all of the processors, I'd recommend doing as much as possible in the `DataProcessor.Fetch()` to avoid duplication. For example, you can take the lock just before calling `Process()` and release it just after. You may still need to take the lock in `Cleanup()` (or maybe there's a good way to have the caller of `Cleanup()` take the lock?)

pkg/segment/query/processor/statscommand.go
```diff
@@ -250,6 +282,10 @@ func (p *statsProcessor) processMeasureOperations(inputIQR *iqr.IQR) (*iqr.IQR,
 		}
 
 		for i := range values {
+			if !p.shouldContinueProcessing {
+				return nil, nil
+			}
```
AndrewHess, line 287: nit: is this necessary? I'd rather not include this in every processor command, and it seems like this would only rarely save CPU (because cancelling a query is probably rare, and the batch sizes aren't too huge)

pkg/segment/query/processor/dataprocessor.go
```diff
@@ -69,7 +72,9 @@ func (dp *DataProcessor) SetStreams(streams []*CachedStream) {
 }
 
 func (dp *DataProcessor) Cleanup() {
+	dp.processorLock.Lock()
 	dp.processor.Cleanup()
+	dp.processorLock.Unlock()
```
AndrewHess, line 77: I think in addition to the lock, you'll also need to set a flag after Cleanup so we don't call Process after we've cleaned up. (Sorry I know I just told you to remove the bools; I didn't think about this case before)

pkg/segment/query/processor/dataprocessor.go
```diff
@@ -250,6 +258,7 @@ func NewBinDP(options *structs.BinCmdOptions) *DataProcessor {
 		isPermutingCmd:    false,
 		isBottleneckCmd:   !hasSpan,
 		isTwoPassCmd:      !hasSpan,
+		processorLock:     &sync.Mutex{},
```
AndrewHess, line 261: nit: you might be able to change the type from `*Mutex` to `Mutex` and then you don't need to add this line to every constructor. But feel free to keep what you have

AndrewHess (commented):
Looks pretty good

pkg/segment/query/processor/dataprocessor.go
Macbeth98: Yeah no worries. I thought about this case, but I thought that searcher would not send any data if the query is cancelled or deleted. 

But, I think the fetch would still continue calling processor.Process after it get the input from the streams. I have added this check.

pkg/segment/query/processor/dataprocessor.go
Macbeth98: if I change the mutex to `Mutex`, then there is an issue with `copylocks`. This is the error `copylocks: literal copies lock value from *scrollerDP`

So apparently it creates copy of the lock when we pass or assign it another reference making the locks not work properly.

pkg/segment/query/processor/dataprocessor.go
```diff
@@ -107,6 +115,10 @@ func (dp *DataProcessor) Fetch() (*iqr.IQR, error) {
 				return nil, utils.TeeErrorf("DP.Fetch: failed to fetch input: %v", err)
 			}
 
+			if dp.isCleanupCalled {
+				return nil, io.EOF
+			}
+
 			dp.processorLock.Lock()
 			output, err = dp.processor.Process(input)
```
AndrewHess, line 123: Once you have the lock, you have to check `isCleanupCalled` again (it may have changed while we were getting the lock). So move the check to just above calling Process(). The check near the top of this function (before getting stream input) is fine as is without being locked

pkg/segment/query/iqr/intermediateQueryResult.go
AndrewHess, line 1: nit: these changes are a different issue, so in the future I'd prefer a separate PR for it. Since you already have the changes, it can stay in this PR if you want

Macbeth98: For now, I will keep in this PR. As this PR was made partly because of this issue.  
But I got your point and from next time, I will make a separate PR for issues like this.

AndrewHess (approved):
Looks good

PR #1894, opened 2024-11-14
Custom Query Timeout Setting

sonamgupta21:
# Description

- Added `queryTimeoutSecs` as a configurable parameter, with a default value of 5 minutes.
- Users can now adjust this setting via the Org Settings page UI, with a minimum limit of 1 minute and a maximum of 30 minutes.
image

# Testing

# Checklist:


pkg/cfghandler/pqsupdate.go
```diff
@@ -84,7 +84,18 @@ func SavePQSConfigToRunMod(filepath string, pqsEnabled bool) error {
...
+
+	var configData struct {
+		PQSEnabled       bool `json:"pqsEnabled"`
+		QueryTimeoutSecs int  `json:"queryTimeoutSecs"`
+	}
+
+	existingConfig, err := config.ReadRunModConfig(filepath)
+	if err == nil {
+		configData.QueryTimeoutSecs = existingConfig.QueryTimeoutSecs
+	}
```
nkunal, line 96: if there is an err we should default to the default timeoutsecs

pkg/config/config.go
```diff
@@ -71,6 +71,12 @@ var maxWaitWipFlushRange = ValuesRangeConfig{Min: 5, Max: 60, Default: 30}
 
 var tracingEnabled bool // flag to enable/disable tracing; Set to true if TracingConfig.Endpoint != ""
 
+const (
+	MIN_QUERY_TIMEOUT = 60   // 1 minute
+	MAX_QUERY_TIMEOUT = 1800 // 30 minutes
+	DEFAULT_TIMEOUT   = 300  // 5 minutes
```
AndrewHess, line 77: nit: specify units in the names

pkg/config/config.go
```diff
@@ -604,8 +623,10 @@ func ReadRunModConfig(fileName string) (common.RunModConfig, error) {
 func ExtractReadRunModConfig(jsonData []byte) (common.RunModConfig, error) {
 	var runModConfig common.RunModConfig
 	if len(strings.TrimSpace(string(jsonData))) == 0 {
-		log.Error("ExtractReadRunModConfig: Could not parse runmod config as it is empty")
-		return runModConfig, fmt.Errorf("ExtractReadRunModConfig: No data to parse")
+		log.Infof("ExtractReadRunModConfig: Empty or no runmod config, using defaults from server.yaml")
+		runModConfig.PQSEnabled = IsPQSEnabled()
+		runModConfig.QueryTimeoutSecs = GetQueryTimeoutSecs()
```
AndrewHess, line 628: nit: now there's some duplication here with lines 637-638, and that duplication will grow as we add more things to the runmod config. It might be worth moving these two lines to a helper method to avoid the duplication

pkg/cfghandler/querytimeoutupdate.go
```diff
@@ -0,0 +1,123 @@
...
+func SaveQueryTimeoutToRunMod(filepath string, timeoutSecs int) error {
+	existingConfig, err := config.ReadRunModConfig(filepath)
+	if err != nil {
+		log.Errorf("SaveQueryTimeoutToRunMod: Could not read existing config from %s: %v", filepath, err)
+	}
+
+	var configData struct {
+		PQSEnabled       bool `json:"pqsEnabled"`
+		QueryTimeoutSecs int  `json:"queryTimeoutSecs"`
+	}
```
AndrewHess, line 67: nit: you've duplicated this struct also in pqsupdate.go. If we add another configurable thing to the runmod config, we'll have to add a field to this struct and the one in pqsupdate.go and copy the struct in the new `func Save...ToRunMod()`. It would be better to define this type outside the functions; then you can define it just once, and if we add a new configurable item we'll have to only update the single type definition (and add the functions to read/write the new field)

pkg/cfghandler/querytimeoutupdate.go
```diff
@@ -0,0 +1,123 @@
...
+
+	if err == nil {
+		configData.PQSEnabled = existingConfig.PQSEnabled
+	} else {
+		configData.PQSEnabled = true
+	}
```
AndrewHess, line 73: In this `else`, we should probably call a function that sets all the default runmod values. Then we can call that same function when we get an error trying to update pqsEnabled

pkg/config/common/common.go
```diff
@@ -146,8 +146,10 @@ type Configuration struct {
 	MaxOpenColumns              uint64         `yaml:"maxOpenColumns"`
 	UseNewPipelineConverted     bool
 	UseNewQueryPipeline         string `yaml:"isNewQueryPipelineEnabled"`
+	QueryTimeoutSecs            int    `yaml:"queryTimeoutSecs"` // query timeout in seconds
```
AndrewHess, line 149: nit: the name is already clear, so you don't really need this comment. You can keep it if you want though

pkg/cfghandler/querytimeoutupdate.go
AndrewHess: Actually, there's already the `RunModConfig` type in config/common.go. I think you can use that here

pkg/health/clusterStatsHandler.go
```diff
@@ -175,8 +175,6 @@ func convertDataToSlice(allIndexStats utils.AllIndexesStats, volumeField, countF
 		nextVal[index][columnCountField] = humanize.Comma(int64(indexStats.NumColumns))
 		nextVal[index][earliestEpochField] = time.Unix(int64(indexStats.EarliestTimestamp/1000), 0).UTC().Format("2006-01-02 15:04:05") + " UTC"
 		nextVal[index][latestEpochField] = time.Unix(int64(indexStats.LatestTimestamp/1000), 0).UTC().Format("2006-01-02 15:04:05") + " UTC"
-		nextVal[index][recordCountField] = humanize.Comma(int64(indexStats.NumRecords))
-		nextVal[index][bytesReceivedField] = float64(indexStats.NumBytesIngested)
```
AndrewHess, line 179: Just curious: why are you removing this?

AndrewHess (commented):
Looks pretty good; mostly nits.

I didn't check the frontend

pkg/health/clusterStatsHandler.go
sonamgupta21: There was some duplicate information such as `bytes received` and `incoming volume`, as well as `num of records` and `event count`.

pkg/cfghandler/pqsupdate.go
```diff
@@ -79,18 +80,17 @@ func PostPqsUpdate(ctx *fasthttp.RequestCtx) {
...
+	var configData common.RunModConfig
+
 	existingConfig, err := config.ReadRunModConfig(filepath)
 	if err != nil {
-		log.Errorf("SavePQSConfigToRunMod: Unable to read existing config from %s, using defaults", filepath)
-		existingConfig.QueryTimeoutSecs = config.DEFAULT_TIMEOUT
+		log.Errorf("SavePQSConfigToRunMod: Using defaults as couldn't read config: %v", err)
+		configData = config.GetDefaultRunModConfig()
+	} else {
+		configData = existingConfig
```
AndrewHess, line 90: nit: this will be slightly cleaner (you don't need the `existingConfig` or the `else` block)
```
	configData, err := config.ReadRunModConfig(filepath)
	if err != nil {
		log.Errorf("SavePQSConfigToRunMod: Using defaults as couldn't read config: %v", err)
		configData = config.GetDefaultRunModConfig()
	}
```

AndrewHess (approved):
Looks good. I added a minor nit, but you can merge it as is if you want

AndrewHess (approved):
Nice!

PR #1895, opened 2024-11-14
Add Min/Max support for strings

vasukalariya:
# Description
Summarize the change.
- Added support for min/max for strings.

# Testing
- Added unit tests for important functions.
- Tested on various group by and stats queries for both matchAll and filter queries.

# Checklist:

### DIFF
--- a/pkg/segment/query/segquery.go
+++ b/pkg/segment/query/segquery.go
@@ -831,7 +831,7 @@ func applyAggOpOnSegments(sortedQSRSlice []*QuerySegmentRequest, allSegFileResul
 		aggHasEvalFunc := segReq.aggs.HasValueColRequest()
 		aggHasValuesFunc := segReq.aggs.HasValuesFunc()
 		aggHasListFunc := segReq.aggs.HasListFunc()
-		aggHasMinMaxFunc := segReq.aggs.HasMinMaxFunc()
+		aggHasMinMaxFunc := segReq.aggs.HasMinMaxFunc() // TODO: Add support for min/max strings in sst
 		var sstMap map[string]*structs.SegStats
 		if searchType == structs.MatchAllQuery && isSegmentFullyEnclosed && !aggHasEvalFunc && !aggHasValuesFunc && !aggHasListFunc && !aggHasMinMaxFunc {
 			sstMap, err = segread.ReadSegStats(segReq.segKey, segReq.qid)
--- a/pkg/segment/query/segquery.go
+++ b/pkg/segment/query/segquery.go
@@ -831,8 +831,9 @@ func applyAggOpOnSegments(sortedQSRSlice []*QuerySegmentRequest, allSegFileResul
 		aggHasEvalFunc := segReq.aggs.HasValueColRequest()
 		aggHasValuesFunc := segReq.aggs.HasValuesFunc()
 		aggHasListFunc := segReq.aggs.HasListFunc()
+		aggHasMinMaxFunc := segReq.aggs.HasMinMaxFunc()
 		var sstMap map[string]*structs.SegStats
-		if searchType == structs.MatchAllQuery && isSegmentFullyEnclosed && !aggHasEvalFunc && !aggHasValuesFunc && !aggHasListFunc {
+		if searchType == structs.MatchAllQuery && isSegmentFullyEnclosed && !aggHasEvalFunc && !aggHasValuesFunc && !aggHasListFunc && !aggHasMinMaxFunc {
 			sstMap, err = segread.ReadSegStats(segReq.segKey, segReq.qid)
 			if err != nil {
 				log.Errorf("qid=%d,  applyAggOpOnSegments : ReadSegStats: Failed to get segment level stats for segKey %+v! Error: %v", qid, segReq.segKey, err)
--- a/pkg/segment/reader/segread/segstatsreader.go
+++ b/pkg/segment/reader/segread/segstatsreader.go
@@ -199,10 +199,10 @@ func GetSegMin(runningSegStat *structs.SegStats,
 
 	// if this is the first segment, then running will be nil, and we return the first seg's stats
 	if runningSegStat == nil {
-		if !currSegStat.IsNumeric && currSegStat.StringStats.MinSet {
+		if !currSegStat.IsNumeric && currSegStat.StringStats.Min.Dtype == utils.SS_DT_STRING {
 			return &utils.CValueEnclosure{
 				Dtype: utils.SS_DT_STRING,
-				CVal:  currSegStat.StringStats.Min,
+				CVal:  currSegStat.StringStats.Min.CVal,
 			}, nil
 		}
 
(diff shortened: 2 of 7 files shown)


pkg/segment/query/segquery.go
```diff
@@ -831,8 +831,9 @@ func applyAggOpOnSegments(sortedQSRSlice []*QuerySegmentRequest, allSegFileResul
 		aggHasEvalFunc := segReq.aggs.HasValueColRequest()
 		aggHasValuesFunc := segReq.aggs.HasValuesFunc()
 		aggHasListFunc := segReq.aggs.HasListFunc()
+		aggHasMinMaxFunc := segReq.aggs.HasMinMaxFunc() // TODO: Add support for min/max strings in sst
 		var sstMap map[string]*structs.SegStats
-		if searchType == structs.MatchAllQuery && isSegmentFullyEnclosed && !aggHasEvalFunc && !aggHasValuesFunc && !aggHasListFunc {
+		if searchType == structs.MatchAllQuery && isSegmentFullyEnclosed && !aggHasEvalFunc && !aggHasValuesFunc && !aggHasListFunc && !aggHasMinMaxFunc {
```
AndrewHess, line 836: nit: this condition is getting long. Maybe extract it into a function (so the function makes the purpose clear) or add a comment saying what this condition means

pkg/segment/reader/segread/segstatsreader.go
```diff
@@ -188,113 +188,139 @@ func readSingleSst(fdata []byte, qid uint64) (*structs.SegStats, error) {
...
 	// if this is the first segment, then running will be nil, and we return the first seg's stats
 	if runningSegStat == nil {
+		if !currSegStat.IsNumeric && currSegStat.StringStats.Min.Dtype == utils.SS_DT_STRING {
+			return &utils.CValueEnclosure{
+				Dtype: utils.SS_DT_STRING,
+				CVal:  currSegStat.StringStats.Min.CVal,
+			}, nil
+		}
+
 		switch currSegStat.NumStats.Min.Ntype {
```
AndrewHess, line 209: nit: only do this switch statement if `IsNumeric`

pkg/segment/reader/segread/segstatsreader.go
```diff
@@ -188,113 +188,139 @@ func readSingleSst(fdata []byte, qid uint64) (*structs.SegStats, error) {
...
 		}
-		return &rSst, nil
+
+		return &result, nil
+	}
+
+	if !currSegStat.IsNumeric {
+		runningSegStat.StringStats.MergeMinStrStats(currSegStat.StringStats)
+
+		if runningSegStat.StringStats.Min.Dtype == utils.SS_DT_STRING {
```
AndrewHess, line 224: Will we ever get to line 224 and have the condition be false?

pkg/segment/reader/segread/segstatsreader.go
AndrewHess, line 1: nit: at some point it would be good to remove the duplication between GetSegMin and GetSegMax (probably not in this PR unless you actually want to remove the duplication, since it would be low priority)

pkg/segment/results/segresults/segresults.go
```diff
@@ -268,9 +268,25 @@ func (sr *SearchResults) UpdateNonEvalSegStats(runningSegStat *structs.SegStats,
...
-		sstResult, err = segread.GetSegMin(runningSegStat, incomingSegStat)
+		res, err := segread.GetSegMin(runningSegStat, incomingSegStat)
+		if err != nil {
+			return nil, fmt.Errorf("UpdateSegmentStats: error getting segment level stats for %v, err: %v, qid=%v", measureAgg.String(), err, sr.qid)
+		}
+		sr.segStatsResults.measureResults[measureAgg.String()] = *res
+		if runningSegStat == nil {
+			return incomingSegStat, nil
+		}
+		return runningSegStat, nil
```
AndrewHess, line 279: nit: instead of
```
		if runningSegStat == nil {
			return incomingSegStat, nil
		}
		return runningSegStat, nil
```
can you do `return res, nil`? Or will that not be the same return value?

pkg/segment/structs/segstructs.go
```diff
@@ -721,6 +723,34 @@ func GetMeasureAggregatorStrEncColumns(measureAggs []*MeasureAggregator) []strin
 	return columns
 }
 
+func (ss *StringStats) MergeMinStrStats(other *StringStats) {
+	if ss.Min.Dtype == utils.SS_INVALID {
+		ss.Min = other.Min
+	}
```
AndrewHess, line 729: nit: It might make sense to add
```
if ss == nil {
    return other
}
```
at the top of this function. I think that could also simplify some of the logic in GetSegMin (and GetSegMax)

But you can ignore this if you like your way better

pkg/segment/writer/stats/segstats.go
```diff
@@ -272,6 +278,29 @@ func AddSegStatsStr(segstats map[string]*SegStats, cname string, strVal string,
...
+		if err != nil {
+			log.Errorf("AddSegStatsStr: Error while reducing min: %v", err)
+		} else {
+			stats.StringStats.Min = res
+		}
+	} else {
+		stats.StringStats.Min = CVal
+	}
+
+	if stats.StringStats.Max.Dtype == SS_DT_STRING {
```
AndrewHess, line 293: nit: it would be nice if you can reduce this duplication with the block above for Min, but it's not necessary if you can't find a good way to do that while keeping the code clear

AndrewHess (commented):
Looking good; mostly nits

One question: we can have strings and numbers in one column; I think within each block we convert everything a string if any value is a string, but I think we still could have a block of numbers and then a block of strings. How will this work in that case? Will it give errors?

pkg/segment/results/segresults/segresults.go
vasukalariya: No, res and runningSegStat are different things. res is just the finalMeasureValue.

pkg/segment/structs/segstructs.go
vasukalariya: Wouldn't it give panic if you call function on a nil value ?

AndrewHess: No (see https://go.dev/play/p/kUK7OcqIjQN)

Golang is a little odd in that regard; this is basically how Go handles what other functions call static methods

pkg/segment/reader/segread/segstatsreader.go
vasukalariya: It might if there is no data in the currSegStat and runningSegStat or there was an error which caused these fields to be not initialized and we would just want to return INVALID CValEnclosure.

pkg/segment/structs/segstructs.go
vasukalariya: Well this is something new :), will update it.

vasukalariya: I just checked, we are not returning anything over here and this condition already handled in the merging of segstats [here](https://github.com/siglens/siglens/blob/b70392af3b4e6c892cb43e51447a8c10dd8f8cb7/pkg/segment/structs/segstructs.go#L740).

pkg/segment/writer/stats/segstats.go
vasukalariya: I am not able to find a good way to do this and I think we can do it in future, since, the Min/Max of SegStats is anyways going to be changed when we add min/max for strings and we would just combine those as a single CVal instead of Min in both NumStats and StringStats.

AndrewHess: Sounds good

pkg/segment/structs/segstructs.go
vasukalariya: I have kept is as it is, if needed we can cleanup in the future.

pkg/segment/query/segquery.go
vasukalariya: addressed in the latest commit.

pkg/segment/reader/segread/segstatsreader.go
vasukalariya: addressed in the latest commit.

pkg/segment/reader/segread/segstatsreader_test.go
AndrewHess, line 1: Nice

AndrewHess (approved):
Looks good

PR #1905, opened 2024-11-16
fix: Group By Nil Columns panic

Macbeth98:
# Description
- Fixed Stats group by `nil` column values panic. if for a bucket a value is `nil`, the code is not properly incrementing the idx. Due to which for a function like `List`, `Values`, we might be accessing `sum`, `Range`, or any other function measure result instead of the specific functions measure results.
- Now with this PR through `defer` we can be sure that the idx is incremented properly.

# Testing
- Ran this query and this passed without any issues
```
variable_col_3=male AND (app_name=Bracecould OR variable_col_7=C*) | regex variable_col_8=".*s" | stats avg(variable_col_4), max(variable_col_19) as max, min(variable_col_18) as min, sum(variable_col_17) as sum, range(variable_col_15), count as cnt, values(variable_col_9), dc(variable_col_6) as distinct_count, list(variable_col_2) BY variable_col_16, bool_col
```

# Checklist:

### DIFF
--- a/pkg/segment/results/blockresults/blockresult.go
+++ b/pkg/segment/results/blockresults/blockresult.go
@@ -625,206 +625,255 @@ func (gb *GroupByBuckets) AddResultToStatRes(req *structs.GroupByRequest, bucket
 		var hllToMerge *toputils.GobbableHll
 		var strSetToMerge map[string]struct{}
 		var eVal utils.CValueEnclosure
-		switch mInfo.MeasureFunc {
-		case utils.Count:
-			if mInfo.ValueColRequest != nil || usedByTimechart {
-				if !usedByTimechart && len(mInfo.ValueColRequest.GetFields()) == 0 {
-					log.Errorf("GroupByBuckets.AddResultToStatRes: Zero fields of ValueColRequest for count: %v", mInfoStr)
-					continue
-				}
 
-				countIdx := gb.reverseMeasureIndex[idx]
-				countVal, err := runningStats[countIdx].rawVal.GetUIntValue()
-				if err != nil {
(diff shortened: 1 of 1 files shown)


pkg/segment/results/blockresults/blockresult.go
```diff
@@ -625,206 +625,255 @@ func (gb *GroupByBuckets) AddResultToStatRes(req *structs.GroupByRequest, bucket
...
+		}
+	}
+}
+
+func (gb *GroupByBuckets) updateEValFromRunningBuckets(mInfo *structs.MeasureAggregator, runningStats []runningStats, usedByTimechart bool, mInfoStr string,
+	currRes map[string]utils.CValueEnclosure, bucket *RunningBucketResults, idxPtr *int, eVal *utils.CValueEnclosure, hllToMerge **toputils.GobbableHll,
+	strSetToMerge *map[string]struct{}) {
+	if hllToMerge == nil || strSetToMerge == nil {
+		// This should never happen
+		log.Errorf("GroupByBuckets.AddResultToStatRes: hllToMerge or strSetToMerge is nil")
```
nkunal, line 643: while debugging we woldn;t know which one was nil, so better to add the %v for them or have seperate checks and separate error msgs

pkg/segment/results/blockresults/blockresult.go
```diff
@@ -625,206 +625,255 @@ func (gb *GroupByBuckets) AddResultToStatRes(req *structs.GroupByRequest, bucket
...
+			}
+
+			eVal.CVal = avg
+			eVal.Dtype = utils.SS_DT_FLOAT
+		} else {
+			if usedByTimechart {
+				incrementIdxBy = 2
+			} else {
+				incrementIdxBy = 1
+			}
```
AndrewHess, line 708: Why? (maybe add a comment)

PR #1906, opened 2024-11-16
Added New err utils to manage Errors and fixed some log flooding with Stats Group By Queries

Macbeth98:
# Description
- PR: #1905 Needs to be Merged First.
- Added New Error utils, that is initialized at the start of the query and logs the errors at the end. 
- The Error use a `sync.Map` to avoid locking for every update for performance, but also safe from concurrent map write errors.
- Added `ErrorWithCode` type that is compatible with `error` interface. 

# Testing
- All unit test cases pass
- And ran the below query and verified that the there is no any log flooding
```
variable_col_3=male AND (app_name=Bracecould OR variable_col_7=C*) | regex variable_col_8=".*s" | stats avg(variable_col_4), max(variable_col_19) as max, min(variable_col_18) as min, sum(variable_col_17) as sum, range(variable_col_15), count as cnt, values(variable_col_9), dc(variable_col_6) as distinct_count, list(variable_col_2) BY variable_col_16, bool_col
```

# Checklist:

### DIFF
--- a/pkg/segment/structs/segstructs.go
+++ b/pkg/segment/structs/segstructs.go
@@ -1657,7 +1657,7 @@ func (qa *QueryAggregators) IsStatsAggPresentInChain() bool {
 	return qa.HasInChain(statsAggPresentInCur)
 }
 
-// use toputils.BatchError instead
+// TODO: use toputils.BatchError instead
 func (nodeRes *NodeResult) StoreGlobalSearchError(errMsg string, logLevel log.Level, err error) {
 	nodeRes.GlobalSearchErrors = StoreError(nodeRes.GlobalSearchErrors, errMsg, logLevel, err)
 }
--- a/pkg/segment/structs/segstructs.go
+++ b/pkg/segment/structs/segstructs.go
@@ -485,7 +485,7 @@ type Progress struct {
 type NodeResult struct {
 	AllRecords                  []*utils.RecordResultContainer
 	ErrList                     []error                     // Need to eventually replace ErrList with GlobalSearchErrors to prevent duplicate errors
-	GlobalSearchErrors          map[string]*SearchErrorInfo // maps global error from error message -> error info
+	GlobalSearchErrors          map[string]*SearchErrorInfo // maps global error from error message -> error info  (Deprecated: use toputils.BatchError)
 	Histogram                   map[string]*AggregationResult
 	TotalResults                *QueryCount
 	VectorResultValue           float64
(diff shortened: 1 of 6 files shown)


pkg/segment/structs/segstructs.go
```diff
@@ -485,7 +485,7 @@ type Progress struct {
 type NodeResult struct {
 	AllRecords                  []*utils.RecordResultContainer
 	ErrList                     []error                     // Need to eventually replace ErrList with GlobalSearchErrors to prevent duplicate errors
-	GlobalSearchErrors          map[string]*SearchErrorInfo // maps global error from error message -> error info
+	GlobalSearchErrors          map[string]*SearchErrorInfo // maps global error from error message -> error info  (Deprecated: use toputils.BatchError)
```
AndrewHess, line 488: Are you going to delete this soon in a different PR? Or do we need to wait for something before we can fully remove this?

pkg/segment/structs/segstructs.go
```diff
@@ -1657,6 +1657,7 @@ func (qa *QueryAggregators) IsStatsAggPresentInChain() bool {
 	return qa.HasInChain(statsAggPresentInCur)
 }
 
+// use toputils.BatchError instead
```
AndrewHess, line 1660: nit: add TODO

pkg/utils/errorutils.go
```diff
@@ -0,0 +1,180 @@
...
+	err  error
+}
+
+func init() {
+	qidToBatchErrorMap = make(map[uint64]*BatchError)
+	qidToBatchErrorMapLock = &sync.RWMutex{}
+}
+
+// NewBatchError creates a new BatchError instance
+func NewBatchError() *BatchError {
```
AndrewHess, line 57: nit: you don't really need this comment. You can keep it if you want though

pkg/utils/errorutils.go
```diff
@@ -0,0 +1,180 @@
...
+			"samples": data.errors,
+		}).Errorf("%s ErrorKey=%v", qidMsg, key)
+		return true
+	})
+}
+
+// Reset clears all stored errors
+func (be *BatchError) Reset() {
+	be.beMap = &sync.Map{}
+}
```
AndrewHess, line 109: Should this also clear the qid?

pkg/utils/errorutils.go
```diff
@@ -0,0 +1,180 @@
...
+
+// HasErrors returns true if there are any stored errors
+func (be *BatchError) HasErrors() bool {
```
AndrewHess, line 112: nit: you chose a good name for this function, so you don't really need this comment either. But you can keep it if you want

AndrewHess (commented):
Looking pretty good

pkg/segment/structs/segstructs.go
Macbeth98: Yeah to delete this, we need to do more work. I will do that in a later PR.

pkg/utils/errorutils.go
Macbeth98: Ahh yes!

nkunal (approved):

PR #1910, opened 2024-11-17
fix: evaluation to not return on nil and return correct data types

Macbeth98:
# Description
- Now the evaluations will not return an `error` on `nil` columns. But they will keep returning `nil` error across the chain and then finally before returning to the caller, it will check if it is an error caused by `nil` value, then it will just return (`nil`, `nil`).
- Also fixed some of the functions or cases return types and default values based on how splunk does.
- Fixed an issue with makemv command to not return error on encountering `nil` columns
- Fixed some of the unit tests that are broken

# Testing
- Run the queries listed on the sheet [here](https://docs.google.com/spreadsheets/d/1OBjgTPwTcVSE_kwn8ig-3bndeBH1HCHDLW2Hy4JUtmY/edit?gid=0#gid=0) and all of them are working without any issues.  
- All unit tests passed

**NOTE**: Some of the functional tests and cicd queries may broke, I will fix them. So as of now I am making this as a draft PR.

# Checklist:

### DIFF
--- a/pkg/common/dtypeutils/dtypeutils.go
+++ b/pkg/common/dtypeutils/dtypeutils.go
@@ -624,19 +624,27 @@ func AlmostEquals(left, right float64) bool {
 	}
 }
 
-func ConvertToSameType(leftType, rightType interface{}) (interface{}, interface{}, error) {
+func ConvertToSameType(leftType, rightType interface{}) (interface{}, interface{}) {
 
 	if fmt.Sprintf("%T", leftType) == fmt.Sprintf("%T", rightType) {
-		return leftType, rightType, nil
+		return leftType, rightType
 	}
 
+	var err error
+
 	if unsafe.Sizeof(leftType) > unsafe.Sizeof(rightType) {
-		rightType, err := ConvertExpToType(rightType, leftType)
-		return leftType, rightType, err
+		rightType, err = ConvertExpToType(rightType, leftType)
 	} else {
-		leftType, err := ConvertExpToType(leftType, rightType)
-		return rightType, leftType, err
+		leftType, err = ConvertExpToType(leftType, rightType)
 	}
+
+	if err != nil {
+		// convert both to strings
+		leftType = fmt.Sprint(leftType)
+		rightType = fmt.Sprint(rightType)
+	}
+
+	return leftType, rightType
 }
 
 // If you add a new field here or change the order of LogFileData, update the columnNames in logfileutils.go
--- a/pkg/segment/query/processor/headcommand_test.go
+++ b/pkg/segment/query/processor/headcommand_test.go
@@ -178,13 +178,13 @@ func Test_Head_Expr_Null(t *testing.T) {
 
 	iqr1, err = headProcessor1.Process(iqr1)
 	assert.Equal(t, io.EOF, err)
-	assert.Equal(t, 3, iqr1.NumberOfRecords())
+	assert.Equal(t, 1, iqr1.NumberOfRecords())
 
 	columnValues, err := iqr1.ReadAllColumns()
 	assert.NoError(t, err)
 	assert.Equal(t, 2, len(columnValues))
-	assert.Equal(t, knownValues["gender"][:3], columnValues["gender"])
-	assert.Equal(t, knownValues["ident"][:3], columnValues["ident"])
+	assert.Equal(t, knownValues["gender"][:1], columnValues["gender"])
+	assert.Equal(t, knownValues["ident"][:1], columnValues["ident"])
 }
 
 func Test_Head_Expr_Keeplast(t *testing.T) {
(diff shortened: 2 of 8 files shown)


pkg/segment/structs/evaluationstructs.go
```diff
@@ -1038,32 +1068,39 @@ func (self *ValueExpr) EvaluateToString(fieldToValue map[string]utils.CValueEncl
 		if err != nil {
 			//Because parsing is successful and it can not evaluate as a float in here,
 			//There is one possibility: field name may not be float
+			if !self.NumericExpr.IsTerminal {
+				// But this is not a terminal case, so we just return 0
+				return "0", err
+			}
```
AndrewHess, line 1074: I don't understand why we should return `"0"` instead of `""` in this error case. Also, since we return an error, does it matter what the string is?

pkg/segment/structs/evaluationstructs.go
```diff
@@ -2017,7 +2062,9 @@ func (self *NumericExpr) Evaluate(fieldToValue map[string]utils.CValueEnclosure)
 			}
 			number, err := strconv.ParseInt(strValue, base, 64)
 			if err != nil {
-				return 0, fmt.Errorf("NumericExpr.Evaluate: cannot convert '%v' to number with base %d", strValue, base)
+				// DO NOT return an error if the string is not a number
+				// just return 0
+				return 0, nil
```
AndrewHess, line 2067: This seems a little odd. Why should we return a nil error?

pkg/segment/query/processor/headcommand_test.go
```diff
@@ -178,13 +178,13 @@ func Test_Head_Expr_Null(t *testing.T) {
 
 	iqr1, err = headProcessor1.Process(iqr1)
 	assert.Equal(t, io.EOF, err)
-	assert.Equal(t, 3, iqr1.NumberOfRecords())
+	assert.Equal(t, 1, iqr1.NumberOfRecords())
```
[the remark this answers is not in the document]
Macbeth98: No, I assumed wrong, I fixed the tests.

tools/sigclient/functionalQueries/head.json
```diff
@@ -2,7 +2,7 @@
     "queryText": "app_name = Bracecould | head (gender=\"female\" OR variable_col_9=\"Director\") null=true keeplast=true | fields country, app_name, variable_col_9, gender, http_method",
     "expectedResult": {
         "totalMatched": {
-            "value": 4,
+            "value": 1,
```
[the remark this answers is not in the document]
Macbeth98: Yes, I changed them back.

tools/sigclient/longerFunctionalQueries/head.json
```diff
@@ -2,7 +2,7 @@
     "queryText": "app_name = Bracecould | head (gender=\"female\" OR variable_col_9=\"Director\") null=true keeplast=true | fields country, app_name, variable_col_9, gender, http_method",
     "expectedResult": {
         "totalMatched": {
-            "value": 4,
+            "value": 1,
```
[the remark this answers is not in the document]
Macbeth98: Addressed in the latest commit

pkg/segment/structs/evaluationstructs.go
```diff
@@ -574,13 +574,13 @@ func (self *BoolExpr) Evaluate(fieldToValue map[string]utils.CValueEnclosure) (b
 		case "in":
 			inFlag, err := isInValueList(fieldToValue, self.LeftValue, self.ValueList)
 			if err != nil {
-				return false, fmt.Errorf("BoolExpr.Evaluate: can not evaluate Eval In function: %v", err)
+				return false, toputils.FormatErrorWithTracef(err, "BoolExpr.Evaluate: can not evaluate Eval In function: %v", err)
```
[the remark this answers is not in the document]
Macbeth98: So actually, it is not necessary to pass the `error` again the second time, the first one is important which will preserve the main `ErrorCode`, and the second is message however the caller wants to frame. Send values, error, etc. Which will be converted to error type.

pkg/segment/structs/evaluationstructs.go
Macbeth98: Same as above, if we `tonumber(string)`, splunk returns `0`.

pkg/common/dtypeutils/dtypeutils.go
```diff
@@ -657,3 +657,42 @@ type LogFileData struct {
...
+		return leftVal != rightVal
+	}
+
+	leftFloatVal, errLeft := ConvertToFloat(leftVal, 64)
+	rightFloatVal, errRight := ConvertToFloat(rightVal, 64)
+
+	if errLeft == nil || errRight == nil {
+		switch Op {
+		case "<":
+			return leftFloatVal < rightFloatVal
```
AndrewHess, line 675: Should `errLeft == nil || errRight == nil` be `errLeft == nil && errRight == nil`? Or if we keep `||` and e.g., `errLeft != nil`, should we set `leftFloatVal = 0`?

pkg/segment/structs/evaluationstructs.go
```diff
@@ -717,50 +733,77 @@ func (self *BoolExpr) Evaluate(fieldToValue map[string]utils.CValueEnclosure) (b
...
+					return getBoolCValueEnclosure(true), nil
+				} else if self.ValueOp == "!=" {
+					// if either value is null, then the evaluation should result in true for "!="
+					return getBoolCValueEnclosure(true), nil
 				}
-				return self.ValueOp == "!=", nil
+
+				// if either value is null, then the evaluation should result in false for "="
+				// But since one of the values is null, the evaluation should send a nil value
+				return &utils.CValueEnclosure{Dtype: utils.SS_DT_BACKFILL, CVal: nil}, nil
```
AndrewHess, line 752: I'm confused. You specify a situation where we should return `false`, but then you always return null?

pkg/common/dtypeutils/dtypeutils.go
Macbeth98: ohh, good catch, it is `&&` not `||`

pkg/segment/structs/evaluationstructs.go
Macbeth98: Yes, i have updated the code, to actually return `NULL` if left or right value is `null`. As that is how splunk is executing.

AndrewHess (dismissed):

AndrewHess (approved):

PR #1923, opened 2024-11-19
Fix sort to use operators

vasukalariya:
# Description
Summarize the change.
- If an operator is specified sort has to be performed based on that.
- Consider operator for fetching the ranks.

# Testing

# Checklist:


pkg/segment/query/processor/sortcommand.go
```diff
@@ -139,19 +139,33 @@ func compareString(a, b string) compare {
...
-		return NUMERIC
-	case segutils.SS_DT_STRING:
-		_, err := strconv.ParseFloat(CValEnc.CVal.(string), 64)
-		if err == nil {
-			// If floatValue is possible then it is considered as a number
+		switch op {
+		case "num", "auto", "":
+			return NUMERIC
+		case "str":
+			return STRING
```
nkunal, line 151: calling this enum as `STRING` is too generic, we should typically prefix some descriptor for such enums. for e.g. we call the `SS_DT_STRING` (sigscalr datatype string). 
Similarly we could call these things `RANK_STRING` etc.

pkg/segment/query/processor/sortcommand.go
```diff
@@ -162,10 +176,10 @@ func getRank(CValEnc *segutils.CValueEnclosure) dTypeRank {
 }
 
 // Returns A comparison B
-func compareValues(valueA, valueB *segutils.CValueEnclosure, asc bool) compare {
+func compareValues(valueA, valueB *segutils.CValueEnclosure, asc bool, op string) compare {
 	var result compare
-	rankA := getRank(valueA)
-	rankB := getRank(valueB)
+	rankA := getRank(valueA, op)
```
nkunal, line 181: the `op` is specified during query time (or defaulted to some value). This `getRank` is getting called 2 x numRecords. The switch/case statement inside it also gets executed numRecords times.  When a explicit rank like str/num is passed in the query, then we should not waste time to find out the rank per record.

pkg/segment/query/processor/sortcommand.go
vasukalariya: addressed in the latest commit.

AndrewHess (approved):
Nice tests

PR #1925, opened 2024-11-19
Fix metadata size calculation

AndrewHess:
# Description
We have configurable limits to reduce how much memory siglens can use for various tasks. The `rebalanceSsm()` function was using too much memory because it's estimate for the memory of one segment was too low. This PR fixes that.

# Testing
Similar to https://github.com/siglens/siglens/pull/1924, but I also reduced the memory allowed for SSM down to 17 MB

Without this fix, `ReadBlockSummaries()` was taking more memory that it should.

# Checklist:

### DIFF
--- a/pkg/segment/metadata/segmentmicroindex.go
+++ b/pkg/segment/metadata/segmentmicroindex.go
@@ -106,8 +106,8 @@ func (sm *SegmentMicroIndex) initMetadataSize() {
 		sumColSize += uint64(len(cname))
 	}
 
-	blockHolderSize := sumColSize * (8 + 4)
-	blockHolderSize += 2 + 6 + 8 + 8 // blockNum, padding, 2 map pointers
+	blockHolderSize := sumColSize * (8 + 4) // int64 value + uint32 value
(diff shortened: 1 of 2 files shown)


pkg/segment/metadata/segmentmicroindex.go
```diff
@@ -91,12 +91,24 @@ func InitSegmentMicroIndex(segMetaInfo *structs.SegMeta, loadSsm bool) *SegmentM
...
+	// 	BlkNum            uint16
+	// 	ColumnBlockOffset map[string]int64
+	// 	ColumnBlockLen    map[string]uint32
+	// }
+	sumColSize := uint64(0)
+	for cname := range sm.ColumnNames {
+		sumColSize += uint64(len(cname))
+	}
+
+	blockHolderSize := sumColSize * (8 + 4)
```
nkunal, line 109: nit: comment on what the 8 and 4 is for

nkunal (commented):
looks good, one minor nit

nkunal (approved):

PR #1929, opened 2024-11-20
Spl perf improvements

nkunal:
# Description
1. Use the xxhash.Sum64String version and avoid getting mem from bufpool.
2. Reuse sfmData struct when doing cluster stats handling. When there are lot of columns (7800+), this would cause a mem spike of 1.5 GB

with #1 this query now `* | stats dc(UserID) as u BY RegionID | sort -u | head 10` now completes in `8 s` as opposed to `13 s`

# Testing

# Checklist:

### DIFF
--- a/pkg/segment/writer/segmetarw.go
+++ b/pkg/segment/writer/segmetarw.go
@@ -108,27 +108,31 @@ func ReadLocalSegmeta(readFullMeta bool) []*structs.SegMeta {
 		return retVal
 	}
 
+	workSfm := &structs.SegFullMeta{}
+
 	// continue reading/merging from individual segfiles
 	for _, smentry := range retVal {
-		sfmData, _ := readSfm(smentry.SegmentKey)
-		if sfmData == nil {
+		err := readSfm(smentry.SegmentKey, workSfm)
+		if err != nil {
+			// error is logged in the func
 			continue
 		}
+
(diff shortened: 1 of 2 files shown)


pkg/segment/writer/segmetarw.go
```diff
@@ -108,27 +108,31 @@ func ReadLocalSegmeta(readFullMeta bool) []*structs.SegMeta {
...
+			utils.MergeMapsRetainingFirst(smentry.ColumnNames, workSfm.ColumnNames)
 		}
 	}
 	return retVal
 }
 
-func readSfm(segkey string) (*structs.SegFullMeta, error) {
+func readSfm(segkey string, sfm *structs.SegFullMeta) error {
 
 	sfmFname := getSegFullMetaFnameFromSegkey(segkey)
```
AndrewHess, line 137: nit: maybe check if `sfm == nil`. I think your code still works because the unmarshalling will return an error in that case (see https://go.dev/play/p/7pxm6eble-j)

AndrewHess (approved):

pkg/segment/writer/segmetarw.go
nkunal: marshalling is going to return error if sfm is nil. So I think I will skip checking for nil.

PR #1934, opened 2024-11-20
Enable Base Value for Seg Key Encodings

Macbeth98:
# Description
- Enable setting of base value for seg key encoding.
- Migrate seg key encoding from uint16 to uint32
- Add gob encoding for map of CValueEnclosures

# Testing
- All unit test cases pass

# Checklist:


pkg/hooks/hooks.go
```diff
@@ -110,7 +110,7 @@ type Hooks struct {
 	ShouldAddDistributedInfoHook func() bool
 
 	// Distributed query
-	InitDistributedQueryServiceHook func(querySummary interface{}, allSegFileResults interface{}, distQueryId string) interface{}
+	InitDistributedQueryServiceHook func(querySummary interface{}, allSegFileResults interface{}, distQueryId string, segEncToKey uint32) interface{}
```
AndrewHess, line 113: Since this is not a map, maybe rename to something like `segKeyEnc`. Or maybe you meant to pass a map?

pkg/segment/results/segresults/segresults.go
```diff
@@ -88,11 +88,11 @@ type SearchResults struct {
...
 	segStatsResults        *segStatsResults
 	convertedBuckets       map[string]*structs.AggregationResult
-	allSSTS                map[uint16]map[string]*structs.SegStats // maps segKeyEnc to a map of segstats
+	allSSTS                map[uint32]map[string]*structs.SegStats // maps segKeyEnc to a map of segstats
 	AllErrors              []error
-	SegKeyToEnc            map[string]uint16
-	SegEncToKey            map[uint16]string
-	NextSegKeyEnc          uint16
+	SegKeyToEnc            map[string]uint32
+	SegEncToKey            map[uint32]string
```
AndrewHess, line 94: FYI there's a `utils.TwoWayMap` type to handle cases where we want both forward and reverse maps. (I don't think you should change these to that in this PR, just wanted to let you know)

pkg/utils/maputils.go
```diff
@@ -180,6 +180,16 @@ func AddToSet[K comparable](set map[K]struct{}, key K) {
 	set[key] = struct{}{}
 }
 
+func MapKeysToSlice[K comparable, V any](m map[K]V) []K {
```
AndrewHess, line 183: There's already `GetKeysOfMap()`, which does the same thing

AndrewHess (commented):
A few small nits, but it's looking good

pkg/utils/maputils.go
Macbeth98: ohh got it!

AndrewHess (approved):

PR #1935, opened 2024-11-21
Dashboard creation time

sonamgupta21:
# Description
Backend Update:
- Previously, three separate API calls were made to fetch dashboards, favorites, and default dashboards.
- Now consolidated into a single API call, with metadata updates to include createdAt, isFavorite, and isDefault, simplifying the dashboard data retrieval process.

Frontend Update:

- Dashboards are now displayed with favorites at the top, sorted by createdAt (most recent first).
- Other dashboards are also sorted by creation time.
- For legacy dashboards (created before this change), the createdAt will default to the last modification time.
- For default dashboards, the createdAt is set to 0.

image

# Testing

# Checklist:

### DIFF
--- a/pkg/dashboards/dashboards.go
+++ b/pkg/dashboards/dashboards.go
@@ -47,6 +47,13 @@ var latestDashboardReadTimeMillis map[uint64]uint64
 var allDashboardsIds map[uint64]map[string]string = make(map[uint64]map[string]string)
 var allDashboardsIdsLock *sync.RWMutex = &sync.RWMutex{}
 
+type DashboardMetadata struct {
+	Name       string `json:"name"`
+	IsFavorite bool   `json:"isFavorite"`
+	IsDefault  bool   `json:"isDefault"`
+	CreatedAt  int64  `json:"createdAt"`
+}
+
 func readSavedDashboards(orgid uint64) ([]byte, error) {
 	var dashboardData []byte
 	allidsFname := getAllIdsFileName(orgid)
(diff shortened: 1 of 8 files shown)


pkg/dashboards/dashboards.go
```diff
@@ -432,6 +387,10 @@ func updateDashboard(id string, dName string, dashboardDetails map[string]interf
 		return errors.New("updateDashboard: Error fetching dashboard details")
 	}
 
+	if currentCreatedAt, exists := currentDashboardDetails["createdAt"]; exists {
+		dashboardDetails["createdAt"] = currentCreatedAt
+	}
```
AndrewHess, line 392: nit: It would probably be a little better to make `currentDashboardDetails` an instance of a struct type and then have `createdAt` be a field. That way, if there's a typo in "createdAt" we'd get a compile-time error instead of it silently failing

pkg/dashboards/dashboards.go
```diff
@@ -845,3 +730,66 @@ func ProcessDeleteDashboardsByOrgId(orgid uint64) error {
...
+
+	return dashboardDetails, nil
+}
+
+func getDashboardMetadata(id string, name string) (map[string]interface{}, error) {
+	dashboardInfo := map[string]interface{}{
+		"name":       name,
+		"isFavorite": false,
+		"isDefault":  isDefaultDashboard(id),
+	}
```
AndrewHess, line 758: Same here: it would be good to make this a struct instead of a `map[string]interface{}`

AndrewHess (commented):
Looks pretty good; just some nits.

I didn't check the frontend

pkg/dashboards/dashboards.go
sonamgupta21: If we use a struct, won't it require updating every time the UI adds a new field?

AndrewHess: Yes. If you think that cost outweighs the benefit of detecting typos easier, then you can keep what you have. You might have a better judgement on this than me since you're more familiar with how the frontend works and how often new fields would get added.

If you do change it to a struct, I think there's some way to make the json unmarshalling strict so it will error if there's a field it doesn't expect, which might be useful. But I'll approve this and you can merge it if you want

AndrewHess (approved):

PR #1940, opened 2024-11-21
Refactoring Min/Max in SegStats

vasukalariya:
# Description
Summarize the change.
- Refactoring Min/Max in SegStats, after this change SST format will need to be modified to incorporate this design.

# Testing
- Ran unit tests
- Ran functional test
- Ran CICD

# Checklist:

- [] I have self-reviewed this PR.


pkg/segment/reader/segread/segstatsreader.go
```diff
@@ -190,216 +189,107 @@ func readSingleSst(fdata []byte, qid uint64) (*structs.SegStats, error) {
...
-		return &result, nil
+	result, err := utils.ReduceMinMax(runningSegStat.Min, currSegStat.Min, true)
+	if err != nil {
+		return &utils.CValueEnclosure{}, fmt.Errorf("GetSegMin: error in ReduceMinMax, err: %v", err)
 	}
-
-	if !runningSegStat.IsNumeric {
-		runningSegStat.NumStats = currSegStat.NumStats
+	runningSegStat.Min = result
+	if !runningSegStat.IsNumeric && runningSegStat.Min.IsNumeric() {
```
AndrewHess, line 206: Why would `IsNumeric` be false and `Min.IsNumeric` be true?

pkg/segment/structs/segstructs.go
```diff
@@ -765,6 +752,9 @@ func (ss *SegStats) Merge(other *SegStats) {
 		}
 	}
 
+	UpdateMinMax(ss, other.Min)
+	UpdateMinMax(ss, other.Max)
```
AndrewHess, line 756: nit: there was somewhere else you also called these together. If they're always called together, I think updating the min and max should be done in one function call. But maybe there's cases where we call it just for one of them?

pkg/segment/structs/segstructs.go
```diff
@@ -812,26 +799,19 @@ func (ss *NumericStats) Merge(other *NumericStats) {
...
 		return
 	}
-	switch ss.Min.Ntype {
+
+	switch ss.Sum.Ntype {
 	case utils.SS_DT_FLOAT:
 		if other.Dtype == utils.SS_DT_FLOAT {
-			ss.Min.FloatVal = math.Min(ss.Min.FloatVal, other.Min.FloatVal)
-			ss.Max.FloatVal = math.Max(ss.Max.FloatVal, other.Max.FloatVal)
 			ss.Sum.FloatVal = ss.Sum.FloatVal + other.Sum.FloatVal
```
AndrewHess, line 806: Why is it correct to remove these lines?

pkg/segment/writer/stats/segstats.go
```diff
@@ -247,6 +222,13 @@ func AddSegStatsStr(segstats map[string]*SegStats, cname string, strVal string,
...
 		segstats[cname] = stats
 	}
+
+	floatVal, err := strconv.ParseFloat(strVal, 64)
+	if err == nil {
+		AddSegStatsNums(segstats, cname, SS_FLOAT64, 0, 0, floatVal, strVal, bb, aggColUsage, hasValuesFunc, hasListFunc)
+		return
+	}
+
 	stats.Count++
```
AndrewHess, line 232: Why do you increment `stats.Count` only if `strVal` can't be parsed as a float? Does it get incremented in `AddSegStatsNums`?

AndrewHess (commented):
Looks pretty good. A few questions

pkg/segment/reader/segread/segstatsreader.go
vasukalariya: The case where currSegStat is numeric but runnningSegStat is not. After we update the min/max we should update isNumeric parameter too.

pkg/segment/structs/segstructs.go
vasukalariya: Yeah, we do not always need to call both of these always. For example when we would build the SST we would only update these based on the record value that we see and that would be one at a time.

pkg/segment/structs/segstructs.go
vasukalariya: We would be deprecating this in the next PR when we move to new SST format. Have kept it in segStats to keep this PR simpler, or else would need to update the SST writer and reader along with this.

pkg/segment/writer/stats/segstats.go
vasukalariya: The increment of Count will be handled in AddSegStatsNums. So no need to count twice.

AndrewHess (approved):

PR #1941, opened 2024-11-22
Enable IQR to read from Multiple readers

Macbeth98:
# Description
- Now the IQR reader can read from Multiple readers, each tracked with the readerId.
- `IQRReader` has two modes: `ReaderModeSingleReader` and `ReaderModeMultiReader`. In `ReaderModeSingleReader`, the flow is exactly same as now.
- For `ReaderModeMultiReader`, the `IQRReader` will group the RRCs based on the `readerId` and call respective readers with these RRCs and then finally merge the results maintaining the order.

# Testing
- Added unit tests for the IQR reader

# Checklist:

### DIFF
--- a/pkg/segment/query/iqr/iqrreader.go
+++ b/pkg/segment/query/iqr/iqrreader.go
@@ -20,11 +20,9 @@ package iqr
 import (
 	"fmt"
 	"reflect"
-	"sync"
 
 	"github.com/siglens/siglens/pkg/segment/reader/record"
 	"github.com/siglens/siglens/pkg/segment/utils"
-	log "github.com/sirupsen/logrus"
 )
 
 type ReaderMode uint8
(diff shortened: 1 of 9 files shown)


pkg/segment/query/iqr/iqrreader.go
```diff
@@ -0,0 +1,412 @@
...
+
+		return nil
+	}
+
+	if iqrRdr.IsSingleReader() {
+		iqrRdr.readerMode = ReaderModeMultiReader
+		if err := iqrRdr.AddSingleIQRReader(otherIQR); err != nil {
+			return fmt.Errorf("iqrReader.MergeIQRReaders: cannot add otherIQR to iqr; err=%v", err)
+		}
+	}
```
AndrewHess, line 196: Is this right? Above, you have 
```
if iqrRdr.IsSingleReader() && otherRdr.IsSingleReader() {
    ...
    return
}
```

So on 191, the `if iqrRdr.IsSingleReader()` is the same as `if iqrRdr.IsSingleReader() && !otherRdr.IsSingleReader()`. So should you be calling `AddSingleIQRReader()` on `otherIQR`, or instead call something like `AddMultiIQRReader()`?

pkg/segment/query/iqr/iqrreader.go
```diff
@@ -0,0 +1,412 @@
...
+
+	if otherRdr.IsSingleReader() {
+		if err := iqrRdr.AddSingleIQRReader(otherIQR); err != nil {
+			return fmt.Errorf("iqrReader.MergeIQRReaders: cannot add otherIQR to iqr; err=%v", err)
+		}
```
AndrewHess, line 201: Same here (just the other way around)

pkg/segment/query/iqr/iqrreader.go
```diff
@@ -0,0 +1,412 @@
...
+
+	resultChan := make(chan readerResult, len(rrcReaderGroups))
+
+	wg := sync.WaitGroup{}
+
+	for readerId, rrcsWithIndex := range rrcReaderGroups {
+		reader := iqrRdr.readerIdToReader[readerId]
+
+		wg.Add(1)
+		go func(readerId utils.T_SegReaderId, rrcsWithIndex []*rrcWithIndex, reader record.RRCsReaderI) {
```
AndrewHess, line 265: I was thinking there'd only be one reader per segkey. Is that incorrect? The method you're in takes a segkey and a list of RRCs, so all the RRCs should be for the same segkey, right?

If there will only be one reader with any `rrcsWithIndex`, then using goroutines and a waitgroup seems unnecessary

Macbeth98: Ohh yes, you are right, for some reason, I was under the impression that the IQR can have RRCs from multiple readers (which is true) and is also same for this function (RRCs contain multiple readers). But since here we are filtering by `SegKey`, yes we don't need that code. I will refactor

AndrewHess (approved):

PR #1945, opened 2024-11-22
New SST format with min max for strings.

vasukalariya:
# Description
Summarize the change.
- New SST format for supporting min/max in strings using SST files.

# Testing
- Wrote unit tests for some of the functions.
- Ran simple queries like `* | stats min(city), max(country), range(http_status)` on benchmark dataset.

# Checklist:


pkg/segment/utils/segconsts.go
```diff
@@ -139,9 +139,10 @@ var VERSION_MBLOCKSUMMARY = []byte{0x01}
 var VERSION_SEGSTATS = []byte{2} // version of the Segment Stats file.
 var VERSION_SEGSTATS_LEGACY = []byte{1}
 
-var VERSION_SEGSTATS_BUF = []byte{3} // version of the single column Seg Stats in a Segment
-var VERSION_SEGSTATS_BUF_LEGACY_1 = []byte{1}
-var VERSION_SEGSTATS_BUF_LEGACY_2 = []byte{2}
+var VERSION_SEGSTATS_BUF = []byte{4} // version of the single column Seg Stats in a Segment
```
nkunal, line 142: we should call this V4, because when v5 comes it becomes easier ?

pkg/segment/reader/segread/segstatsreader.go
```diff
@@ -183,7 +197,49 @@ func readSingleSst(fdata []byte, qid uint64) (*structs.SegStats, error) {
...
-	return &sst, nil
+	if readNumCount {
+		idx += 8
+
+		// read NumCount
+		sst.NumStats.NumCount = toputils.BytesToUint64LittleEndian(fdata[idx : idx+8])
+	}
+}
+
+func readCurrentSST(sst *structs.SegStats, fdata []byte, idx uint32) error {
```
AndrewHess, line 208: What does the `f` mean in `fdata`?

pkg/segment/writer/packer.go
```diff
@@ -1556,6 +1556,20 @@ func addSegStatsStrIngestion(segstats map[string]*SegStats, cname string, valByt
 		segstats[cname] = stats
 	}
 
+	floatVal, err := strconv.ParseFloat(string(valBytes), 64)
+	if err == nil {
+		if !stats.IsNumeric {
+			stats.IsNumeric = true
+		}
+		addSegStatsNums(segstats, cname, SS_FLOAT64, 0, 0, floatVal, valBytes)
+		return
```
AndrewHess, line 1565: Wouldn't we only want to set `stats.IsNumeric = true` if we get here and we're not sure if it's numeric or not (i.e., this is the first entry in `stats`)? With the current way I think you could add a string stat, get `IsNumeric = false`, then add a numeric stat here and set `IsNumeric = true` but then you'd probably (?) have issues later? Or do we always want to use numeric stats if we have any?

nkunal (changes requested):
Sst format is being rethought

pkg/segment/writer/stats/segstats_test.go
```diff
@@ -62,6 +62,31 @@ func Test_addSegStatsNums(t *testing.T) {
 
 }
 
+func Test_addSegStatsNumsMixed(t *testing.T) {
```
AndrewHess, line 65: You added basically the same test in packer_test.go. Is it useful to have this test in both places?

pkg/segment/writer/packer.go
vasukalariya: We are currently not adding any stringStat in sst.

pkg/segment/writer/stats/segstats_test.go
vasukalariya: These are 2 similar functions that are used one is on the packer/writer side and the other is one the reader side. Notice [AddSegStatsNums](https://github.com/siglens/siglens/blob/49fe395f2be32cb1a3321c136492afa4ab6f53ed/pkg/segment/writer/stats/segstats.go#L42) and [addSegStatsNums](https://github.com/siglens/siglens/blob/49fe395f2be32cb1a3321c136492afa4ab6f53ed/pkg/segment/writer/packer.go#L1581). I think we have copies because of ingestion performance where we do not want checks like hasEvalFunc etc. which are present in the AddSegStatsNums but not in addSegStatsNums. But not sure on this.

pkg/segment/reader/segread/segstatsreader.go
vasukalariya: fdata refers to file data.

pkg/segment/utils/segconsts.go
vasukalariya: addressed in the latest commit.

pkg/segment/query/segquery.go
```diff
@@ -857,18 +856,8 @@ func applyAggOpOnSegments(sortedQSRSlice []*QuerySegmentRequest, allSegFileResul
...
-				allSegFileResults.AddError(err)
-				continue
-			}
-			sstMap["*"] = &structs.SegStats{
-				Count: uint64(segReq.TotalRecords),
-			}
-			allSegFileResults.AddResultCount(uint64(segReq.TotalRecords))
-		} else {
+
+		computeSegStatsFunc := func() {
```
nkunal, line 860: nit: why not just have this as a standalone func instead of a closure ? makes reading code easier/modular as well func length smaller (the parent's)

pkg/segment/reader/segread/segstatsreader.go
```diff
@@ -143,37 +143,52 @@ func readSingleSst(fdata []byte, qid uint64) (*structs.SegStats, error) {
...
 	idx += hllSize
 
+	if version == utils.VERSION_SEGSTATS_BUF_V4[0] {
+		err := readCurrentSST(&sst, fdata, idx)
+		if err != nil {
+			return nil, fmt.Errorf("readSingleSst: error reading current sst, err: %v", err)
+		}
+		return &sst, nil
+	}
+
```
nkunal, line 153: nit: add a comment, versions prior to V4 only had numeric stats.
so that we know the code below this if block is for older versions

pkg/segment/reader/segread/segstatsreader.go
```diff
@@ -143,37 +143,52 @@ func readSingleSst(fdata []byte, qid uint64) (*structs.SegStats, error) {
...
-		max.IntgrVal = toputils.BytesToInt64LittleEndian(fdata[idx : idx+8])
+		max.CVal = toputils.BytesToInt64LittleEndian(fdata[idx : idx+8])
 	}
-	sst.NumStats.Max = max
+	sst.Max = max
 	idx += 8
 
-	// read Sum Ntype
 	sum := utils.NumTypeEnclosure{}
+	// read Sum Ntype
```
nkunal, line 191: update/remove this comment or clarify why Ntype means

pkg/segment/reader/segread/segstatsreader.go
nkunal: what does `current` mean ? is tied to the V4 or something else ? when does current become past ?

pkg/segment/reader/segread/segstatsreader.go
```diff
@@ -183,7 +198,52 @@ func readSingleSst(fdata []byte, qid uint64) (*structs.SegStats, error) {
...
+		return nil
+	}
+
+	dType := utils.SS_DTYPE(fdata[idx : idx+1][0])
+	idx += 1
+	if dType != utils.SS_DT_STRING {
+		return fmt.Errorf("readCurrentSST: invalid dtype: %v", dType)
+	}
+
+	// read Min
```
nkunal, line 220: nit: remove this comment, this line does not read mean, but there is a below comment that actually reads various things related to min

pkg/segment/reader/segread/segstatsreader.go
```diff
@@ -183,7 +198,52 @@ func readSingleSst(fdata []byte, qid uint64) (*structs.SegStats, error) {
...
+	// read Min length
+	minlen := toputils.BytesToUint16LittleEndian(fdata[idx : idx+2])
+	idx += 2
+
+	// read Min string
+	min.CVal = string(fdata[idx : idx+uint32(minlen)])
+	sst.Min = min
+	idx += uint32(minlen)
+
+	// read Max
```
nkunal, line 233: same comment here

pkg/segment/writer/segstore.go
```diff
@@ -1548,14 +1548,15 @@ func (ss *SegStore) FlushSegStats() error {
 /*
 Encoding Schema for SegStats Single Column Data
 [Version 1B] [isNumeric 1B] [Count 8B] [HLL_Size 4B] [HLL_Data xB]
-[N_type 1B] [Min 8B] [N_type 1B] [Max 8B] [N_type 1B] [Sum 8B]
+Numeric [DType 1B] [Min 8B] [DType 1B] [Max 8B] [NType 1B] [Sum 8B] [NumericCount 8B]
+NonNumeric [DType 1B] [Min_Size 2B] [Min_Data xB] [Max_Size 2B] [Max_Data xB]
```
nkunal, line 1552: Put a OR here. It will be either numeric or NonNumeric and not both

pkg/segment/query/segquery.go
vasukalariya: addressed in the latest commit.

pkg/segment/reader/segread/segstatsreader.go
vasukalariya: addressed in the latest commit.

pkg/segment/reader/segread/segstatsreader.go
vasukalariya: `NType` is already a field in NumTypeEnclosure.

pkg/segment/reader/segread/segstatsreader.go
vasukalariya: addressed in the latest commit.

pkg/segment/reader/segread/segstatsreader.go
vasukalariya: addressed in the latest commit.

pkg/segment/writer/segstore.go
vasukalariya: addressed in the latest commit.

pkg/segment/query/segquery.go
```diff
@@ -857,50 +889,31 @@ func applyAggOpOnSegments(sortedQSRSlice []*QuerySegmentRequest, allSegFileResul
...
+				sstMap["*"] = &structs.SegStats{
+					Count: uint64(segReq.TotalRecords),
 				}
+				allSegFileResults.AddResultCount(uint64(segReq.TotalRecords))
 			}
+		} else {
+			sstMap, err = computeSegStatsFromRawRecords(segReq, qs, allSegFileResults, qid, nodeRes)
 		}
+		if err != nil {
+			allSegFileResults.AddError(err)
```
nkunal, line 911: should this AddError be associated with the place where specifically the error occurred. Right now it could get added twice: line # 898 and 910

nkunal (commented):
looks good to me. Once Andrew approves, it can be merged

pkg/segment/reader/segread/segstatsreader.go
```diff
@@ -120,50 +118,38 @@ func readSingleSst(fdata []byte, qid uint64) (*structs.SegStats, error) {
...
+	case utils.VERSION_SEGSTATS_BUF_V4[0]:
 		hllSize = toputils.BytesToUint32LittleEndian(fdata[idx : idx+4])
 		idx += 4
-	case utils.VERSION_SEGSTATS_BUF_V2[0], utils.VERSION_SEGSTATS_BUF_V1[0]:
-		hllSize = uint32(toputils.BytesToUint16LittleEndian(fdata[idx : idx+2]))
-		idx += 2
 	default:
-		log.Errorf("qid=%d, readSingleSst: unknown version: %v", qid, version)
-		return nil, errors.New("readSingleSst: unknown version")
+		return nil, fmt.Errorf("qid=%d, readSingleSst: unknown version: %v", qid, version)
```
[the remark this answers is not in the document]
vasukalariya: No, because ReadSegStats would give an error and we would resort to computing the stats manually by reading records.

pkg/segment/query/segquery.go
vasukalariya: addressed in the latest commit.

pkg/segment/structs/segstructs.go
```diff
@@ -534,8 +534,8 @@ type SegStats struct {
 }
 
 type NumericStats struct {
-	NumCount uint64                 `json:"numCount,omitempty"`
-	Sum      utils.NumTypeEnclosure `json:"sum,omitempty"`
+	NumericCount uint64                 `json:"numericCount,omitempty"`
+	Sum          utils.NumTypeEnclosure `json:"sum,omitempty"`
```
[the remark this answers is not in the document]
AndrewHess: Ok, sounds good. What about my first question in this thread?

PR #1950, opened 2024-11-25
Enabled IQR to store stats in Aggregated format

Macbeth98:
# Description
- Now the IQR will also have `statsResults` that contains Group By and Segment Stats cmd results in Aggregated format.
- Updated the Gob encoding
- MergeIQRs now also merge these stats from multiple IQR streams.
- Added tests to test the merging of IQR stats.
- Refactored the code for `stats`, `timechart`, `top`, and `rare` commands.

# Testing
- Added new tests to test the merging of IQR and updated gob encoding/decoding test of IQR.
- All test cases are passed.
- Tested by running some of stats and top/rare queries

# Checklist:

### DIFF
--- a/pkg/segment/results/blockresults/runningstats.go
+++ b/pkg/segment/results/blockresults/runningstats.go
@@ -710,7 +710,7 @@ func (rs *runningStats) ToSerializedRunningStats() *SerializedRunningStats {
 	}
 }
 
-func (srb *SerializedRunningBucketResults) ToRunningBucketResults() *RunningBucketResults {
+func (srb *SerializedRunningBucketResults) ToRunningBucketResults(qid uint64) *RunningBucketResults {
 	runStats := make([]runningStats, len(srb.RunningStats))
 	for i := 0; i < len(srb.RunningStats); i++ {
 		runStats[i] = *srb.RunningStats[i].ToRunningStats()
@@ -727,6 +727,7 @@ func (srb *SerializedRunningBucketResults) ToRunningBucketResults() *RunningBuck
 		currStats:           srb.CurrStats,
 		groupedRunningStats: groupedRunningStats,
 		count:               srb.Count,
+		qid:                 qid,
 	}
 }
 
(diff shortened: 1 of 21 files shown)


pkg/segment/results/blockresults/runningstats.go
```diff
@@ -665,3 +679,176 @@ func (rj RunningStatsJSON) GetRunningStats() (runningStats, error) {
...
+func (srs *SerializedRunningStats) ToRunningStats() *runningStats {
+	return &runningStats{
+		rawVal:    srs.RawVal,
+		hll:       srs.Hll,
+		rangeStat: srs.RangeStat,
+		avgStat:   srs.AvgStat,
+	}
+}
+
+func GetRunningBucketResultsSliceForTest() []*RunningBucketResults {
```
AndrewHess, line 743: nit: this should probably be in a `..._test.go` file

pkg/utils/sliceutils.go
```diff
@@ -399,3 +399,11 @@ func SelectFromSlice[T any](slice []T, shouldKeep func(T) bool) []T {
...
 	return result
 }
+
+func NormalizeSlice[T any](slice []T) []T {
+	if slice == nil {
+		return []T{}
+	}
+
+	return slice
+}
```
AndrewHess, line 409: Just curious: in what situations is having an empty slice instead of a nil slice beneficial? Go has some nice properties with nil maps, but I'm not sure about nil slices

AndrewHess (commented):
Looks good; a few small nits

pkg/segment/results/blockresults/runningstats.go
Macbeth98: if I move this to a test file, I won't be able to call this function from `intermediateQueryResult_test.go`. I need the function in the same `pkg` so that it could also set the private fields.

pkg/utils/sliceutils.go
Macbeth98: So I had to do this because, when we do `reflect.DeepEqual`, this is causing an issue when there is a `nil` slice and you are comparing with slice empty slice.

So I created this for that purpose.

AndrewHess (approved):

PR #1952, opened 2024-11-26
Added Additional Metrics to  Index Summary

sonamgupta21:
# Description
Updated the index summary on the My Org page to include the following new fields:

- CMI Size: Calculated by summing up all cmi sizes from the .sfm file (full segment metadata) for each segment in the index.
- CSG Size: Calculated by summing up all csg sizes from the .sfm file for each segment in the index.
- Number of Segment Files: Total count of cmi + csg files across all segments in the index.
- Number of Blocks: Extracted from segmeta.json to calculate the total number of blocks in the index.

image

# Testing

# Checklist:

### DIFF
--- a/pkg/segment/writer/segmetarw.go
+++ b/pkg/segment/writer/segmetarw.go
@@ -53,6 +53,13 @@ type PQSChanMeta struct {
 	deleteFromEmptyPqMeta bool
 }
 
+type SegmentSizeStats struct {
+	TotalCmiSize uint64
+	TotalCsgSize uint64
+	NumSegFiles  int
+	NumBlocks    int64
+}
+
 var pqsChan = make(chan PQSChanMeta, PQS_CHAN_SIZE)
 
 func initSmr() {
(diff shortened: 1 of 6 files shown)


pkg/segment/writer/segmetarw.go
```diff
@@ -682,3 +689,90 @@ func writeOverSegMeta(segMetaEntries []*structs.SegMeta) error {
...
+	sfm, err := readSfm(segmentKey)
+	if err != nil {
+		return nil, fmt.Errorf("calculateSegmentSizes: failed to read sfm for segment %s: %v", segmentKey, err)
+	}
+
+	stats := &SegmentSizeStats{}
+	for _, colInfo := range sfm.ColumnNames {
+		if colInfo.CmiSize == 0 && colInfo.CsgSize == 0 {
+			continue
+		}
```
nkunal, line 703: nit: if these two params are zero, then thats ok, 0 will get added to the below two lines, which is no -op. By removing them we have less code to maintain

sonamgupta21: I used continue here so that stats.NumIndexFiles is not incremented if both CmiSize and CsgSize are zero.

nkunal (approved):

PR #1961, opened 2024-11-27
Flush SegFullMeta every block and sync Segmeta with SegFullmeta on bootup

Macbeth98:
# Description
- Flushes the SegFullMeta with SegMeta + columnsInfo + AllPqids every block for a segment.
- removed the use of `.sid` and replaced it with `.sfm`.
- On bootup, the system will load the `SegMetas` from the `segemeta.json` file and then it will check for `each .sfm` file and verify whether the segkey is present in the globalMeta or not. If not present, then it will read the SFM, and update Segment Micro Index and will update the `segmeta.json` with the new segments.
- Updated to use `IsSegkeyUnrotated` instead of checking for `segment-validity.json` file.
- Removed writing of validity file for logs.
- Fixed flaky tests.

# Testing
- Ingested data across 5 indexes and verified that the block if flushed and then after 5 block flushes for 5 indexes, caused a panic.
- Verified that the `.sfm` files were created for all 5 indexes
- Then restarted the system and verified that the segment.json is updated and verified the data by doing some queries.
- Repeated the process again for the second time.

- All unit test cases passed
- Functional suite passed.

# Checklist:


benchmark_test.go
```diff
@@ -368,7 +368,7 @@ func Benchmark_logSummarySegKey(b *testing.B) {
 
 	segKey := "a"
 
-	sidName := segKey + ".sid"
+	sidName := segKey + ".sfm"
```
AndrewHess, line 371: nit: rename `sidName`

pkg/common/fileutils/fileutils.go
```diff
@@ -150,3 +150,10 @@ func ReadLineByLine(filePath string, processLine func(line []byte) error) error
 
 	return nil
 }
+
+func TrimLeadingSlash(s string) string {
+	if len(s) > 0 && s[0] == '/' {
+		return s[1:]
+	}
+	return s
+}
```
AndrewHess, line 159: When would we want to do this? Changing `/foo/bar/` to `foo/bar/` seems odd to me since it changes where the file is

pkg/segment/query/segquery_test.go
```diff
@@ -408,19 +409,22 @@ func Test_segQueryFilter(t *testing.T) {
...
-	metadata.InitMockColumnarMetadataStore("data/", fileCount, numBuffers, numEntriesForBuffer)
+	metadata.InitMockColumnarMetadataStore(segBaseDir, fileCount, numBuffers, numEntriesForBuffer)
 
 	bloomMetadataFilter(t, numBuffers, numEntriesForBuffer, fileCount)
 	rangeMetadataFilter(t, numBuffers, numEntriesForBuffer, fileCount)
 	// pqsSegQuery(t, numBuffers, numEntriesForBuffer, fileCount)
 	// add more simple, complex, and nested metadata checking
 	time.Sleep(1 * time.Second) // sleep to give some time for background pqs threads to write out dirs
-	err = os.RemoveAll("data/")
+	err = os.RemoveAll(dir)
```
AndrewHess, line 428: nit: you don't really need to remove this dir since it's a `TempDir()` on line 412. The testing framework should take care of deleting the dir. But it's fine to keep this

pkg/segment/query/queryrefresh.go
```diff
@@ -250,6 +253,76 @@ func populateMicroIndices(smFile string) error {
...
+
+		filesInDir, err := os.ReadDir(vTableBaseDir)
+		if err != nil {
+			log.Errorf("syncSegMetaWithSegFullMeta: Error in reading directory, vTableBaseDir:%v , err:%v", vTableBaseDir, err)
+			continue
+		}
+
+		for _, file := range filesInDir {
+			fileName := file.Name()
+			segkey := filepath.Join(vTableBaseDir, fileName, fileName)
```
AndrewHess, line 277: nit: This seems a little odd. I think you're doing this because if the baseDir ends in `/suffix` then the segkey ends in `/suffix/suffix`. Hopefully at some point we'll fix that. Here, I think it would be good to call some function that computes the segkey based on the baseDir

AndrewHess (commented):
Looking good; just some nits

AndrewHess (approved):
Nice

PR #1971, opened 2024-12-03
Misc utils

vasukalariya:
# Description
Summarize the change.

# Testing

# Checklist:

### DIFF
--- a/pkg/common/fileutils/fileutils.go
+++ b/pkg/common/fileutils/fileutils.go
@@ -22,6 +22,7 @@ import (
 	"io"
 	"os"
 	"path"
+	"path/filepath"
 	"syscall"
 
 	"github.com/siglens/siglens/pkg/config"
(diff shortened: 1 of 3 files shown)


pkg/common/fileutils/fileutils.go
```diff
@@ -124,6 +125,29 @@ func GetAllFilesInDirectory(path string) []string {
 	return retVal
 }
 
+func GetAllFilesWithSameNameInDirectory(path string, fname string) []string {
```
nkunal, line 128: Where is this function going to be used. Be careful where this func is going to get called from since there could be thounsands of files in dir. 
also maybe there is a better os utility package available that can do exact/wildcard match of whatever file you are looking for : for e.g. ls myfile*

AndrewHess: You should be able to use `filepath.Walk` to do what a bit cleaner; I'm not sure if there's some other function that does exactly what you want

vasukalariya: Addressed in the latest commit.

AndrewHess (approved):
Looks good. Anything else @nkunal?

nkunal (approved):

PR #1972, opened 2024-12-03
Fix for incorrect timestamps being read when PQS is enabled

vasukalariya:
# Description
Summarize the change.
Due to a bug in Go 1.19, scope of the loop variable is per loop not per iteration and thus inserting the same value multiple times in the pool which can cause the same pointer to a slice being returned multiple times.
Refer https://go.dev/blog/loopvar-preview and https://github.com/golang/go/discussions/56010 for more details.

# Testing
After this fix longer functional should pass for github actions

# Checklist:

### DIFF
--- a/.github/workflows/nightly-longer-functional.yml
+++ b/.github/workflows/nightly-longer-functional.yml
@@ -2,7 +2,7 @@ name: Nightly-Longer-Functional-Test
 
 on:
   schedule:
-    - cron: '0 5 * * *'  # Every day at midnight EST (UTC-5)
+    - cron: '0 0 * * *'  # TODO: Every day at midnight EST (UTC-5)
   workflow_dispatch:  # Allows manual trigger as well
 
 jobs:
--- a/pkg/segment/reader/segread/timereader.go
+++ b/pkg/segment/reader/segread/timereader.go
@@ -422,12 +422,11 @@ func ReadAllTimestampsForBlock(blks map[uint16]*structs.BlockMetadataHolder, seg
 
 func ReturnTimeBuffers(og map[uint16][]uint64) {
 	for k := range og {
-		// Due to a bug in Go 1.19, when looping over the values of a map using a range loop,
-		// the scope of the loop variable is per loop and not iteration which was causing the
-		// same slices to be inserted multiple times causing the pool to return same slices for multiple
-		// subsequent Get calls. This is a workaround to fix the issue.
+		// Due to a bug in Go 1.19, scope of the loop variable is per loop not per iteration
+		// and thus inserting the same value multiple times in the pool which can cause the same pointer
+		// to a slice being returned multiple times.
 		// Refer https://go.dev/blog/loopvar-preview and https://github.com/golang/go/discussions/56010 for more details.
-		timestamps := og[k]
-		rawTimestampsBufferPool.Put(&timestamps)
+		timeBuffer := og[k]
+		rawTimestampsBufferPool.Put(&timeBuffer)
 	}
 }
--- a/pkg/segment/reader/segread/timereader.go
+++ b/pkg/segment/reader/segread/timereader.go
@@ -421,7 +421,13 @@ func ReadAllTimestampsForBlock(blks map[uint16]*structs.BlockMetadataHolder, seg
 }
 
 func ReturnTimeBuffers(og map[uint16][]uint64) {
-	for _, v := range og {
-		rawTimestampsBufferPool.Put(&v)
+	for k := range og {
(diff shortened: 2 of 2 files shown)


.github/workflows/nightly-longer-functional.yml
```diff
@@ -2,7 +2,7 @@ name: Nightly-Longer-Functional-Test
 
 on:
   schedule:
-    - cron: '0 5 * * *'  # Every day at midnight EST (UTC-5)
+    - cron: '0 0 * * *'  # TODO: Every day at midnight EST (UTC-5)
```
AndrewHess, line 5: Why are you changing this?

pkg/segment/reader/segread/timereader.go
```diff
@@ -421,7 +421,12 @@ func ReadAllTimestampsForBlock(blks map[uint16]*structs.BlockMetadataHolder, seg
...
-	for _, v := range og {
-		rawTimestampsBufferPool.Put(&v)
+	for k := range og {
+		// Due to a bug in Go 1.19, scope of the loop variable is per loop not per iteration
+		// and thus inserting the same value multiple times in the pool which can cause the same pointer
+		// to a slice being returned multiple times.
+		// Refer https://go.dev/blog/loopvar-preview and https://github.com/golang/go/discussions/56010 for more details.
+		timeBuffer := og[k]
+		rawTimestampsBufferPool.Put(&timeBuffer)
 	}
```
AndrewHess, line 431: I think this will still have issues; maybe I'm missing something. This is a bad example I'm thinking of:
1. You do this for loop, putting the local stack address into the pool
2. Eventually Go reclaims this stack space
3. Eventually, Go reuses this stack space (e.g., for a different function call)

After step 2, things would still work fine because only the pool is using that memory, but after step 3 happens, then if the pool hands out that memory address and someone uses it, then you've corrupted the stack which would be really bad (probably a panic?) or the stack has corrupted the data that someone else got by calling pool.Get(). Either way, you'd end up with a pretty nasty bug that I think will be quite difficult to track down

Am I missing something that makes it so the above situation can't happen? If it can happen, I think it would be best to fix it properly now even if the likelihood of it happening is low because it will probably be really annoying to figure out later if we do see this bug

.github/workflows/nightly-longer-functional.yml
vasukalariya: Just to know that we have to reset it to nightly after we are done monitoring this fix.

pkg/segment/reader/segread/timereader.go
vasukalariya: I ran an escape analysis on a smaller code mentioned below which does the same thing.
```
package main

import (
	"fmt"
	"sync"
)

func main() {
	arrLen := 10
	bufferPool := sync.Pool{
		New: func() interface{} {
			slice := make([]int64, arrLen)
			return &slice
		},
	}

	mp := make(map[string][]int64)

	arr1 := *bufferPool.Get().(*[]int64)
	for i := 0; i < arrLen; i++ {
		arr1[i] = int64(i + 1)
	}

	mp["key1"] = arr1

	arr2 := *bufferPool.Get().(*[]int64)
	for i := 0; i < arrLen; i++ {
		arr2[i] = -int64(i + 1)
	}

	mp["key2"] = arr2

	for _, v := range mp {
		buffer := v
		bufferPool.Put(&buffer)
	}

	poolArr1 := *bufferPool.Get().(*[]int64)

	poolArr2 := *bufferPool.Get().(*[]int64)

	fmt.Println(poolArr1)
	fmt.Println(poolArr2)

}
```


Here is the escape analysis on this code.

```
# main
./main.go:11:8: can inline main.func1
./main.go:42:13: inlining call to fmt.Println
./main.go:43:13: inlining call to fmt.Println
./main.go:10:2: moved to heap: bufferPool
./main.go:34:3: moved to heap: buffer
./main.go:12:4: moved to heap: slice
./main.go:11:8: func literal escapes to heap
./main.go:12:17: make([]int64, arrLen) escapes to heap
./main.go:17:12: make(map[string][]int64) does not escape
./main.go:42:13: ... argument does not escape
./main.go:42:14: poolArr1 escapes to heap
./main.go:43:13: ... argument does not escape
./main.go:43:14: poolArr2 escapes to heap
```

We can observe that `buffer` is moved to heap, meaning the pointers stored in pool refer to variables stored on heap.

AndrewHess: Oh nice, that's pretty cool

nkunal (approved):

PR #1974, opened 2024-12-04
Add sort index basic implementation

AndrewHess:
# Description
This adds the basic features of a sort index. There's still quite a bit to do before this feature is complete, so this is hidden beneath a new feature flag.

The sort index is a file per segment for one column that lists the sorted order of all the values for that column, along with the block and record number for each record with that value. Then for queries like `search ... | sort ...` we can use the sort index to efficiently run the query.

# Testing
Mostly manual testing by ingesting data and running queries. There are some new unit tests though

# Checklist:


pkg/segment/query/processor/searcher.go
```diff
@@ -149,6 +158,113 @@ func (s *Searcher) Fetch() (*iqr.IQR, error) {
 	}
 }
 
+func (s *Searcher) fetchColumnSortedRRCs() (*iqr.IQR, error) {
+	qsrs, err := query.GetSortedQSRs(s.queryInfo, s.startTime, s.querySummary)
+	if err != nil {
+		log.Errorf("qid=%v, searcher.getBlocks: failed to get sorted QSRs: %v", s.qid, err)
```
nkunal, line 164: nit: incorrect func name in log

pkg/segment/query/processor/searcher.go
```diff
@@ -149,6 +158,113 @@ func (s *Searcher) Fetch() (*iqr.IQR, error) {
...
+	const recordLimit = 1000 // TODO: find a better way to limit.
+	cname := s.sortExpr.SortEles[0].Field
+	lines, err := sortindex.ReadSortIndex(qsr.GetSegKey(), cname, recordLimit)
+	if err != nil {
+		return nil, err
+	}
+
+	if len(lines) == 0 {
+		// TODO: raw search this segment if it has the cname
+		panic("got no results in sortindex")
```
nkunal, line 205: I guess this panic will get replaced with something else in another PR very soon ?

pkg/segment/reader/segread/segreader/segreader.go
```diff
@@ -79,18 +82,74 @@ type SegmentFileReader struct {
...
+
+		numRecs := blockSummaries[blockNum].RecCount
+		blockToRecords[blockNum] = make([][]byte, 0, numRecs)
+		for i := uint16(0); i < numRecs; i++ {
+			bytes, err := fileReader.ReadRecord(i)
+			if err != nil {
+				return nil, fmt.Errorf("ReadAllRecords: error reading record %v in block %v; err=%+v", i, blockNum, err)
+			}
+
+			// TODO: don't copy so much; without copying, there's a data
```
nkunal, line 122: good todo.

pkg/segment/sortindex/sortindex.go
```diff
@@ -0,0 +1,235 @@
...
+
+	"github.com/siglens/siglens/pkg/segment/reader/segread/segreader"
+	segutils "github.com/siglens/siglens/pkg/segment/utils"
+	"github.com/siglens/siglens/pkg/utils"
+	toputils "github.com/siglens/siglens/pkg/utils"
+)
+
+type Block struct {
+	BlockNum uint16   `json:"blockNum"`
+	Records  []uint16 `json:"records"`
```
nkunal, line 37: nit: should we call this `recNums` ?

pkg/segment/sortindex/sortindex.go
```diff
@@ -0,0 +1,235 @@
...
+}
+
+func getFilename(segkey string, cname string) string {
+	return filepath.Join(segkey, cname+".sort")
```
nkunal, line 41: nit: have been sticking with 3 char extension (no reason), so maybe we can name this `.srt`

pkg/segment/sortindex/sortindex.go
```diff
@@ -0,0 +1,235 @@
...
+}
+
+func WriteSortIndex(segkey string, cname string) error {
```
nkunal, line 44: nit: once we finalize the file format, we should write a comment above this func about how the file format looks like

pkg/segment/writer/segstore.go
```diff
@@ -825,6 +827,10 @@ func (segstore *SegStore) checkAndRotateColFiles(streamid string, forceRotate bo
 		updateRecentlyRotatedSegmentFiles(segstore.SegmentKey)
 		metadata.AddSegMetaToMetadata(&segmeta)
 
+		// TODO: make this a background job; somehow handle siglens being
+		// gracefully shutdown.
+		writeSortIndexes(segstore.SegmentKey)
```
nkunal, line 832: for now should we just make this a goroutine, atleast the seg rotation will not be stuck here waiting for the sorted index to be created ?

pkg/utils/withdefault.go
```diff
@@ -0,0 +1,66 @@
...
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+package utils
+
+import (
+	"gopkg.in/yaml.v3"
+)
+
+type WithDefault[T comparable] struct {
```
nkunal, line 24: sweet !

nkunal (commented):
looks fantastic! 
Some minor nits

pkg/segment/query/processor/searcher.go
AndrewHess: Oops, I'll remove that panic. Yes, this will get handled properly in a few days

pkg/segment/writer/segstore.go
AndrewHess: I'd slightly prefer to keep it as is until it's properly fixed. This function is a no-op unless the sort index feature flag is enabled, btw.

The downside of keeping this as is is that ingestion stalls for a little after rotation, like you said. The downside of changing this to a goroutine without doing the proper fix, is that if someone gracefully shuts down and that causes segment rotation, that could cause a .srt file to be only half written. When siglens is restarted, queries sorting on that column will give bad results if they need data from the unwritten part of that .srt file.

Either way, there's no downside when the feature flag is disabled

nkunal: sounds good

nkunal (approved):

PR #1976, opened 2024-12-04
Cleanup refreshGlobalMetadata

vasukalariya:
# Description
Summarize the change.

# Testing

# Checklist:

### DIFF
--- a/pkg/segment/query/queryrefresh.go
+++ b/pkg/segment/query/queryrefresh.go
@@ -147,7 +147,6 @@ func initGlobalMetadataRefresh(getMyIds func() []uint64) {
 	if err != nil {
 		log.Errorf("initGlobalMetadataRefresh: Error in refreshing global metadata, err:%v", err)
 	}
-	go refreshGlobalMetadataLoop(getMyIds)
 }
 
 func RefreshGlobalMetadata(fnMyids func() []uint64, ownedSegments map[string]struct{}) error {
@@ -211,17 +210,6 @@ func getOwnedSegments() map[string]struct{} {
 		return nil
 	} else {
 		return hook()
-	}
-}
-
-func refreshGlobalMetadataLoop(getMyIds func() []uint64) {
-	for {
-		ownedSegments := getOwnedSegments()
-		err := RefreshGlobalMetadata(getMyIds, ownedSegments)
-		if err != nil {
-			log.Errorf("refreshGlobalMetadataLoop: Error in refreshing global metadata, err:%v", err)
-		}
-		time.Sleep(SECONDS_REFRESH_GLOBAL_METADATA * time.Second)
 	}
 }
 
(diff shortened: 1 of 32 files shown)


pkg/segment/query/queryrefresh.go
```diff
@@ -131,35 +135,29 @@ func initGlobalMetadataRefresh(getMyIds func() []uint64) {
 	if !config.IsQueryNode() || !config.IsS3Enabled() {
 		return
 	}
-
-	err := refreshGlobalMetadata(getMyIds)
+	err := blob.DownloadAllIngestNodesDir()
 	if err != nil {
-		log.Errorf("initGlobalMetadataRefresh: Error in refreshing global metadata, err:%v", err)
+		log.Errorf("downloadIngestNodesAndGetOwnedSegments: Error in downloading ingest nodes dir, err:%v", err)
```
AndrewHess, line 140: nit: wrong function name in log

pkg/segment/structs/segmeta.go
```diff
@@ -35,6 +35,7 @@ type VtableCounts struct {
 	OnDiskBytesCount uint64
 }
 
+// Ensure that when SegMeta is read, the corresponding SegFullMeta is also read
 type SegMeta struct {
```
AndrewHess, line 39: Why? I thought the point of splitting SegFullMeta into a different struct/file is so that when we only need info that's in the SegMeta, we can just read that info and not read a lot more info that's in SegFullMeta

pkg/segment/writer/segmetarw.go
```diff
@@ -93,31 +93,20 @@ func getSegFullMetaFnameFromSegkey(segkey string) string {
...
-func ReadLocalSegmeta(readFullMeta bool) []*structs.SegMeta {
 
-	smrLock.RLock()
-	retVal, err := getAllSegmetas(localSegmetaFname)
-	smrLock.RUnlock()
-	if err != nil {
-		log.Errorf("ReadLocalSegmeta: getallsegmetas err=%v ", err)
-		return retVal
-	}
+	readSfmForSegMetas(segMetas)
```
AndrewHess, line 102: Why do we need this inside `ReadSegmeta()`? (continuation of my previous comment)

vasukalariya: This is [present](https://github.com/siglens/siglens/pull/1744/files#diff-7f61f72d8e97fc004dc47f9ac98903e19cd28a4f4088c8d922cb86cfb76f9520) since the beginning. We need to set column names, PQIDs etc from sfm file into the current segMeta entries.

pkg/segment/structs/segmeta.go
vasukalariya: See the comment below.

pkg/segment/query/queryrefresh.go
vasukalariya: addressed in the latest commit.

AndrewHess (approved):

PR #1978, opened 2024-12-04
Misc file utils

vasukalariya:
# Description
Summarize the change.

# Testing

# Checklist:


pkg/common/fileutils/fileutils.go
```diff
@@ -150,6 +150,34 @@ func GetAllFilesWithSameNameInDirectory(dir string, fname string) []string {
 	return result
 }
 
+func GetAllFilesWithSpecificExtensions[T any](dir string, ext map[string]T) []string {
```
AndrewHess, line 153: nit: It looks like the value `T` is never used. Is there a reason you're using `map[string]T` instead of `map[string]struct{}`?

pkg/common/fileutils/fileutils_test.go
```diff
@@ -50,3 +50,59 @@ func Test_GetAllFilesWithSameNameInDirectory(t *testing.T) {
...
+	err = os.Mkdir(temp+"/dir2", 0755)
+	assert.Nil(t, err)
+	err = os.Mkdir(temp+"/dir1/dir3", 0755)
+	assert.Nil(t, err)
+	err = os.Mkdir(temp+"/dir2/dir4", 0755)
+	assert.Nil(t, err)
+	_, err = os.Create(temp + "/dir1/abc.txt")
+	assert.Nil(t, err)
+	_, err = os.Create(temp + "/dir1/abc.mno")
+	assert.Nil(t, err)
```
AndrewHess, line 69: nit: this is getting repetitive (along with your above function `Test_GetAllFilesWithSameNameInDirectory`). You can keep it like this if you want, but you might want to make a small helper function like `forceCreate(filename)` that makes the dir and creates the file and fails the test if either of those gave an error

pkg/utils/segutils.go
```diff
@@ -194,3 +194,21 @@ func WriteValidityFile(segBaseDir string) error {
 
 	return nil
 }
+
+func GetSegDirFromSegKey(segKey string) string {
```
AndrewHess, line 198: Can you reuse `getSegBaseDirFromFilename()` (making it public if needed), or does that function not handle this case?

pkg/common/fileutils/fileutils.go
vasukalariya: Just used T since, the caller's doesn't have to be send this as a specific type. I just made it for generalization. But yeah can change it to `map[string]struct{}`

pkg/utils/segutils.go
vasukalariya: addressed in the latest commit.

AndrewHess (approved):

PR #1980, opened 2024-12-05
Add hook for Blocking the processing of query and node stats on changes

Macbeth98:
# Description
- Block Processing of query and node stats on Changes

# Checklist:

### DIFF
--- a/pkg/segment/query/querystatus.go
+++ b/pkg/segment/query/querystatus.go
@@ -275,26 +275,35 @@ func DeleteQuery(qid uint64) {
 }
 
 func canRunQuery() bool {
-	if hook := hooks.GlobalHooks.CheckForOwnedSegmentsProcessingHook; hook != nil {
-		hook()
-	}
 	activeQueries := uint64(GetActiveQueryCount())
 	return activeQueries < MAX_RUNNING_QUERIES
 }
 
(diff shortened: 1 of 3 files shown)


pkg/segment/query/querystatus.go
```diff
@@ -278,19 +279,32 @@ func canRunQuery() bool {
 	return activeQueries < MAX_RUNNING_QUERIES
 }
 
+func initiateRunQuery(segsRLockFunc, segsRUnlockFunc func()) {
+	waitingQueriesLock.Lock()
+	if len(waitingQueries) == 0 {
+		waitingQueriesLock.Unlock()
+		time.Sleep(PULL_QUERY_INTERVAL)
+		return
```
AndrewHess, line 287: nit: sleeping before a return seems really odd to me. I see you got this from refactoring, where the sleep was before a `continue`. Here's what I would suggest:
1. Make this function return `*wsData` instead of nothing, and remove the Sleep() here
2. If the caller of this function gets nil, then sleep and continue
3. Call `runQuery()` in the caller, so that in this function you can make the locking a bit easier by deferring the unlock to avoid multiple unlock paths

AndrewHess (commented):
Looks good, just a nit

AndrewHess (approved):

PR #1982, opened 2024-12-07
Process segments in batches

vasukalariya:
# Description
Summarize the change.
- Currently all the blocks from all the segments are being processed together which could lead to high memory usage.
- With this PR, we should fetch segments in batches and process them. One caveat with this is that segments that overlap needs to be processed together since we do not know the order of records.
- Blocks inside these segments would be processed similar as before.

# Testing

# Checklist:


pkg/segment/query/processor/searcher.go
```diff
@@ -40,6 +40,8 @@ import (
 	log "github.com/sirupsen/logrus"
 )
 
+const segmentBatchOffset = 2 // TODO: Find an optimal value
```
AndrewHess, line 43: What does batch offset mean?

pkg/segment/query/processor/searcher.go
```diff
@@ -63,6 +65,11 @@ type Searcher struct {
 	gotBlocks             bool
 	remainingBlocksSorted []*block // Sorted by time as specified by sortMode.
 	qsrs                  []*query.QuerySegmentRequest
+	segmentIdx            int
+	minTimestamp          uint64
+	processedBlocks       map[string]map[uint16]struct{}
+	processedSegment      map[string]struct{}
```
AndrewHess, line 71: nit: use the same pluralization for these two fields

pkg/segment/query/processor/searcher.go
```diff
@@ -171,6 +204,10 @@ func (s *Searcher) fetchRRCs() (*iqr.IQR, error) {
 	// we always take blocks from the front of the list.
 	s.remainingBlocksSorted = s.remainingBlocksSorted[len(nextBlocks):]
 
+	if len(s.remainingBlocksSorted) == 0 {
+		s.gotBlocks = false
+	}
```
AndrewHess, line 209: Why should we do this?

pkg/segment/query/processor/searcher.go
```diff
@@ -387,13 +424,66 @@ func (s *Searcher) initializeQSRs() error {
...
+func (s *Searcher) getQSRSToProcess() []*query.QuerySegmentRequest {
+	qsrs := make([]*query.QuerySegmentRequest, 0)
+
+	prevMinTimestamp := s.minTimestamp
+
+	s.segmentIdx += segmentBatchOffset
+	if s.segmentIdx >= len(s.qsrs) {
+		s.segmentIdx = len(s.qsrs)
+		s.minTimestamp = 0
+		s.gotAllSegments = true
```
AndrewHess, line 456: Is this necessary? `s.minTimestamp = 0`

pkg/segment/query/processor/searcher.go
```diff
@@ -387,13 +424,66 @@ func (s *Searcher) initializeQSRs() error {
...
+	// Batch all the segments having records within the startTime and are not yet processed
+	for i := 0; i < len(s.qsrs); i++ {
+		_, processed := s.processedSegment[s.qsrs[i].GetSegKey()]
+		if processed {
+			continue
+		}
+		if s.qsrs[i].GetEndEpochMs() > prevMinTimestamp {
+			s.processedSegment[s.qsrs[i].GetSegKey()] = struct{}{}
+			continue
+		}
```
AndrewHess, line 470: I think you're assuming the `searcher.sortMode` is `recentFirst`, but you should also handle `recentLast`

pkg/segment/query/processor/searcher.go
```diff
@@ -387,13 +424,66 @@ func (s *Searcher) initializeQSRs() error {
...
+		s.gotAllSegments = true
+	} else {
+		s.minTimestamp = s.qsrs[s.segmentIdx].GetStartEpochMs()
 	}
 
+	// Batch all the segments having records within the startTime and are not yet processed
+	for i := 0; i < len(s.qsrs); i++ {
+		_, processed := s.processedSegment[s.qsrs[i].GetSegKey()]
+		if processed {
+			continue
```
AndrewHess, line 465: nit: it might be slightly more performant to have a list of unprocessed QSRs instead of a list of all QSRs and a map of which QSRs have already been processed. Is there a reason you did it this way?

pkg/segment/query/segquery.go
```diff
@@ -516,7 +516,7 @@ func getSortedQSRResult(aggs *structs.QueryAggregators, allQSRs []*QuerySegmentR
 		} else {
 			// index 0 should have the earliest time
 			sort.Slice(allQSRs, func(i, j int) bool {
-				return allQSRs[i].segKeyTsRange.StartEpochMs > allQSRs[j].segKeyTsRange.StartEpochMs
+				return allQSRs[i].segKeyTsRange.EndEpochMs > allQSRs[j].segKeyTsRange.EndEpochMs
```
AndrewHess, line 519: Is this going to break other things? If other code relies on what it was, then you might want to do this custom sort inside searcher.go (also, you might want to sort in searcher.go anyways to handle `recentFirst` and `recentLast` sorting modes)

pkg/segment/query/segqueryhelpers.go
```diff
@@ -65,6 +65,14 @@ type QuerySegmentRequest struct {
...
 }
 
+func (qsr *QuerySegmentRequest) GetStartEpochMs() uint64 {
+	return qsr.segKeyTsRange.StartEpochMs
+}
+
+func (qsr *QuerySegmentRequest) GetEndEpochMs() uint64 {
+	return qsr.segKeyTsRange.EndEpochMs
+}
+
```
AndrewHess, line 75: nit: move these down near the toerh QuerySegmentRequest getters

pkg/segment/query/processor/searcher.go
vasukalariya: addressed in the latest commit. Removed this offset to select cutOff timestamps. We are now getting cutOffs one by one from segments.

pkg/segment/query/processor/searcher.go
vasukalariya: addressed in the latest commit.

pkg/segment/query/processor/searcher.go
vasukalariya: addressed in the latest commit. Removed this and cleaned up some code.

pkg/segment/query/processor/searcher.go
vasukalariya: addressed in the latest commit.

pkg/segment/query/processor/searcher.go
vasukalariya: addressed in the latest commit.

pkg/segment/query/segquery.go
vasukalariya: Currently, all tests seemed to have passed, but I think we should update it just so that we don't assume the current code and ground truth and build tests which might be incorrect.

pkg/segment/query/segqueryhelpers.go
vasukalariya: addressed in the latest commit.

pkg/segment/query/processor/searcher.go
```diff
@@ -557,45 +594,76 @@ func (s *Searcher) getFilteredBlocks(blocks []*block) []*block {
 	return filteredBlocks
 }
 
-func (s *Searcher) getQSRSToProcess() []*query.QuerySegmentRequest {
-	qsrs := make([]*query.QuerySegmentRequest, 0)
+func (s *Searcher) shouldProcessQSR(cutOff uint64, qsr *query.QuerySegmentRequest) bool {
+	switch s.sortMode {
+	case recentFirst:
+		return qsr.GetEndEpochMs() >= cutOff
```
AndrewHess, line 600: nit: here and in `willProcessQSRCompletely()` you pass `cutOff` as a parameter, but in `shouldProcessBlock()` you use `s.cutOffTimestamp`. I'd suggest doing it the same way in each of these functions, unless there's a good reason not to

pkg/segment/query/processor/searcher.go
```diff
@@ -69,10 +68,9 @@ type Searcher struct {
 	gotBlocks             bool
 	remainingBlocksSorted []*block // Sorted by time as specified by sortMode.
 	qsrs                  []*query.QuerySegmentRequest
-	segmentIdx            int
-	minTimestamp          uint64
+	cutOffTimestamp       uint64
```
AndrewHess, line 71: nit: specify the units of the timestamp

pkg/segment/query/processor/searcher.go
AndrewHess: I see. I think saying that in a comment would be good

pkg/segment/query/segquery.go
AndrewHess: I'm not quite following. I think with your most recent commits, you don't rely on this change anymore; do you think there's a reason to keep this change? I'm not sure if by "current code" you mean the the unmodified version of this or the modified version

vasukalariya: I meant that the existing code (unmodified version) is incorrect and if someone uses this method, this can cause issues. Although it is not being used right now with my change. I thought since, its a public method, it might be better to correct it here as well.

pkg/segment/query/processor/searcher.go
vasukalariya: addressed in the latest commit.

pkg/segment/query/processor/searcher.go
vasukalariya: addressed in the latest commit.

pkg/segment/query/processor/searcher.go
vasukalariya: addressed in the latest commit.

AndrewHess (approved):

PR #1984, opened 2024-12-09
New format for sort index files

vasukalariya:
# Description
Summarize the change.
- JSON marshaling and unmarshaling takes a lot of time for large srt files.
- With this new format we should be able to read information quickly.
- Currently, we are only supporting strings so `writeCValEnc` method is only handling strings. Later we should add support for numbers. The format for the same is specified in the comments.
- Morevoer, when maxRecords are specified, we would want to restrict the number of records to this value and not read anything extra, in the earlier format we were reading all the records in a block. In this new format we would skip reading once we reach `maxRecords`.

# Testing
- Ran query `SearchPhrase != "" | sort -str(SearchPhrase) | head 10 | fields SearchPhrase` clickbench dataset (10M), earlier it was taking around 1400ms. Now, it only takes 350ms.

# Checklist:

### DIFF
--- a/pkg/segment/sortindex/sortindex.go
+++ b/pkg/segment/sortindex/sortindex.go
@@ -161,13 +161,13 @@ func writeSortIndex(segkey string, cname string, valToBlockToRecords map[string]
 			// Write blockNum
 			_, err = writer.Write(utils.Uint16ToBytesLittleEndian(block.BlockNum))
 			if err != nil {
-				return fmt.Errorf("writeSortIndex: failed writing blockNum: %v", err)
+				return fmt.Errorf("writeSortIndex: failed writing blockNum, blockNum: %v, err: %v", block.BlockNum, err)
 			}
 
 			// Write number of records
 			_, err = writer.Write(utils.Uint32ToBytesLittleEndian(uint32(len(block.RecNums))))
 			if err != nil {
-				return fmt.Errorf("writeSortIndex: failed writing number of records: %v", err)
+				return fmt.Errorf("writeSortIndex: failed writing number of records: %v, err: %v", len(block.RecNums), err)
 			}
 
 			// Write sortedRecords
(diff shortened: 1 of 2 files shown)


AndrewHess (approved):

pkg/segment/sortindex/sortindex.go
```diff
@@ -177,53 +219,122 @@ func AsRRCs(lines []Line, segKeyEncoding uint32) ([]*segutils.RecordResultContai
...
+	for numColValues < totalUniqueColValues && !done {
+		// Read DType
+		var Dtype segutils.SS_DTYPE
+		err = binary.Read(file, binary.LittleEndian, &Dtype)
 		if err == io.EOF {
 			break
 		}
 		if err != nil {
-			return nil, err
+			return nil, fmt.Errorf("ReadSortIndex: failed reading DType: %v", err)
```
nkunal, line 260: the EOF error in line 256 should also be considered as an error and we should return an error to the caller. This is Because we entered the loop since we still had some more colValues to read, but encountered a EOF, and that means the file was prematurely cut.

nkunal: the error statement should say above, and also include the numColValues and totalUnieueColVals var in it

pkg/segment/sortindex/sortindex.go
```diff
@@ -177,53 +219,122 @@ func AsRRCs(lines []Line, segKeyEncoding uint32) ([]*segutils.RecordResultContai
...
-		var blocks []Block
-		err = json.Unmarshal(blocksBytes, &blocks)
-		if err != nil {
-			return nil, err
+		for numBlocks < totalBlocks && !done {
+			// Read blockNum
+			var blockNum uint16
+			err = binary.Read(file, binary.LittleEndian, &blockNum)
+			if err == io.EOF {
+				break
```
nkunal, line 292: same comment as above

pkg/segment/sortindex/sortindex.go
```diff
@@ -177,53 +219,122 @@ func AsRRCs(lines []Line, segKeyEncoding uint32) ([]*segutils.RecordResultContai
...
+			}
+
+			numRecords := uint32(0)
+			recNums := make([]uint16, 0)
+			for numRecords < totalRecords && totalRecordsRead < uint64(maxRecordsToRead) {
+				// Read recNum
+				var recNum uint16
+				err = binary.Read(file, binary.LittleEndian, &recNum)
+				if err == io.EOF {
+					break
```
nkunal, line 312: same comment as above

PR #1998, opened 2024-12-16
API endpoint to set sort columns & non-API function to retrieve columns name

sonamgupta21:
# Description
API endpoint to set sort columns & non-API function to retrieve those columns name.

```
curl -X POST \
  http://localhost:5122/api/sort-columns \ 
  -H 'Content-Type: application/json' \
  -d '["timestamp", "app_name"]'
  ```
  
# Testing

# Checklist:

### DIFF
--- a/pkg/segment/sortindex/sortindex.go
+++ b/pkg/segment/sortindex/sortindex.go
@@ -20,21 +20,34 @@ package sortindex
 import (
 	"bufio"
 	"encoding/binary"
+	"encoding/json"
 	"fmt"
 	"io"
 	"os"
 	"path/filepath"
 	"sort"
+	"sync"
 
 	"github.com/siglens/siglens/pkg/segment/reader/segread/segreader"
 	segutils "github.com/siglens/siglens/pkg/segment/utils"
 	"github.com/siglens/siglens/pkg/utils"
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: pkg/segment/sortindex/sortindex.go [hunk 1]
```diff
@@ -547,3 +559,47 @@ func IsEOF(checkpoint *Checkpoint) bool {
 
 	return checkpoint.eof
 }
+
+func SetSortColumns(columnNames []string) error {
+	if columnNames == nil {
+		columnNames = make([]string, 0)
+	}
```
AndrewHess, line 566: nit: you don't really need this. I think you're doing this to avoid an issue when iterating over a nil slice, but Go already handles that

AndrewHess (commented):
Looks pretty good. I think this PR would also be a good time to remove the current hardcoding we're doing [here](https://github.com/siglens/siglens/blob/develop/pkg/segment/writer/segstore.go#L66) and use your new function [here](https://github.com/siglens/siglens/blob/develop/pkg/segment/writer/segstore.go#L855) instead of the hardcoded values

AndrewHess (approved):

Macbeth98 (approved):

PR #2012, opened 2024-12-20
Log memory fetch source

Macbeth98:
# Description
- Logs whether memory is from `os` or from `cgroups`

# Testing
- Tested by running the siglens server and ensured that the logs were being displayed

# Checklist:

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -170,9 +170,12 @@ func GetMemoryMax() uint64 {
 	// try to get the memory from the cgroup
 	memoryMax, err := getContainerMemory(memoryValueMax)
 	if err != nil {
-		log.Debugf("GetMemoryMax: Error while getting memory from cgroup: %v", err)
+		log.Warnf("GetMemoryMax: Error while getting memory from cgroup: %v", err)
 		// if we can't get the memory from the cgroup, get it from the OS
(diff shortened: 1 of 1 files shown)


pkg/config/config.go
```diff
@@ -170,9 +170,12 @@ func GetMemoryMax() uint64 {
 	// try to get the memory from the cgroup
 	memoryMax, err := getContainerMemory(memoryValueMax)
 	if err != nil {
-		log.Debugf("GetMemoryMax: Error while getting memory from cgroup: %v", err)
+		log.Warnf("GetMemoryMax: Error while getting memory from cgroup: %v", err)
 		// if we can't get the memory from the cgroup, get it from the OS
 		memoryMax = memory.TotalMemory()
+		log.Infof("GetMemoryMax: Memory from the OS in bytes: %v", memoryMax)
```
nkunal, line 176: nit: `...from the OS ...`  should be `...from the Host ..`

nkunal: also log the memoryMax in humanize.Comma format

Macbeth98: I changed it to log in MB, similar to how we do in other places.

nkunal (approved):

PR #2017, opened 2024-12-26
Inode-based Segment Deletion

sonamgupta21:
# Description

- Checks the current inode usage percentage of the filesystem
- If usage is above the threshold (85%) and we've passed the warning period:
      -   Gets all segments sorted by time (oldest first)
      -   Deletes oldest segments until inode usage is back under threshold

# Testing

# Checklist:


pkg/retention/retention.go
```diff
@@ -416,3 +419,108 @@ func DeleteMetricsSegmentData(mmetaFile string, metricSegmentsToDelete map[strin
...
+		log.Errorf("doInodeBasedDeletion: Failed to get inode stats: %v", err)
+		return
+	}
+
+	totalInodes := stat.Files
+	freeInodes := stat.Ffree
+	usedInodes := totalInodes - freeInodes
+
+	if totalInodes == 0 {
+		log.Errorf("doInodeBasedDeletion: Invalid total inodes count")
```
nkunal, line 438: Add the invalid value (0) in the log msg

pkg/retention/retention.go
```diff
@@ -416,3 +419,108 @@ func DeleteMetricsSegmentData(mmetaFile string, metricSegmentsToDelete map[strin
...
+		return
+	}
+
+	currentUsagePercent := (usedInodes * 100) / totalInodes
+
+	log.Infof("doInodeBasedDeletion: Current inode usage: %v%%, Max allowed: %v%%, IngestNodeDir: %v",
```
nkunal, line 444: Maybe add the actual values here since that would be more helpful in debugging rather than just percentages

pkg/retention/retention.go
```diff
@@ -416,3 +419,108 @@ func DeleteMetricsSegmentData(mmetaFile string, metricSegmentsToDelete map[strin
...
+		if currentUsagePercent <= maxInodeUsagePercent {
+			break
+		}
+
+		switch entry := metaEntry.(type) {
+		case *structs.MetricsMeta:
+			metricSegmentsToDelete[entry.MSegmentDir] = entry
+		case *structs.SegMeta:
+			segmentsToDelete[entry.SegmentKey] = entry
+		}
```
Macbeth98, line 518: We are not actually deleting any files here. We are just adding them to the map of segments we want to delete.
So the below line
```
// Check if we've deleted enough by getting fresh inode stats
		err := syscall.Statfs(filepath.Clean(dataPath), &stat)
```
Checking for the latest `inodes` state on each iteration of the loop might not be necessary? Let me know if I miss something here.

pkg/retention/retention.go
```diff
@@ -416,3 +419,108 @@ func DeleteMetricsSegmentData(mmetaFile string, metricSegmentsToDelete map[strin
...
+		allEntries = append(allEntries, allSegMetas[i])
+	}
+
+	sort.Slice(allEntries, func(i, j int) bool {
+		var timeI uint64
+		if segMeta, ok := allEntries[i].(*structs.SegMeta); ok {
+			timeI = segMeta.LatestEpochMS
+		} else if metricMeta, ok := allEntries[i].(*structs.MetricsMeta); ok {
+			timeI = uint64(metricMeta.LatestEpochSec * 1000)
+		} else {
```
Macbeth98, line 481: Probably log an Error in the `else` case. Although it should never come here.

pkg/retention/retention.go
```diff
@@ -416,3 +419,108 @@ func DeleteMetricsSegmentData(mmetaFile string, metricSegmentsToDelete map[strin
 		return
 	}
 }
+
+func doInodeBasedDeletion(ingestNodeDir string, maxInodeUsagePercent uint64, deletionWarningCounter int) {
+	var stat syscall.Statfs_t
```
Macbeth98, line 424: nit: rename to `statfs` or `fsStats`
I think this is more clear.

pkg/retention/retention.go
```diff
@@ -416,3 +419,100 @@ func DeleteMetricsSegmentData(mmetaFile string, metricSegmentsToDelete map[strin
...
+			return false
+		}
+		return timeI < timeJ
+	})
+
+	// Delete oldest segments until we get below the threshold
+	segmentsToDelete := make(map[string]*structs.SegMeta)
+	metricSegmentsToDelete := make(map[string]*structs.MetricsMeta)
+
+	for _, metaEntry := range allEntries {
```
Macbeth98, line 502: You are deleting all the entries here. Regardless of the threshold, is that what we want?

pkg/retention/retention.go
```diff
@@ -416,3 +419,107 @@ func DeleteMetricsSegmentData(mmetaFile string, metricSegmentsToDelete map[strin
...
+	segmentsMarked := uint64(0)
+
+	for _, metaEntry := range allEntries {
+		if segmentsMarked >= inodesToFree {
+			break
+		}
+		switch entry := metaEntry.(type) {
+		case *structs.MetricsMeta:
+			metricSegmentsToDelete[entry.MSegmentDir] = entry
+			segmentsMarked++
```
Macbeth98, line 511: Deleting a segment is currently being treated as removing a single `inode`, but that's not accurate. The `Delete` function actually deletes the entire segment directory, which may contain multiple inodes. It would be better to calculate the inode count for each segment directory and increment the required inode deletions by that amount when marking a segment as deleted.

pkg/retention/retention.go
```diff
@@ -71,6 +73,7 @@ func internalRetentionCleaner() {
 		} else {
 			DoRetentionBasedDeletion(config.GetCurrentNodeIngestDir(), config.GetRetentionHours(), 0)
 			doVolumeBasedDeletion(config.GetCurrentNodeIngestDir(), 60000, deletionWarningCounter)
+			doInodeBasedDeletion(config.GetCurrentNodeIngestDir(), 85, deletionWarningCounter) // 85% max inode usage
```
AndrewHess, line 76: nit:
1. make 85 a constant near the top of this file (or maybe we want it in server.yaml, but I think a constant is probably fine for now)
2. if we'll only ever call `doInodeBasedDeletion` with this value of 85, then maybe it shouldn't even be a parameter, and the function should use the constant directly. This is largely personal preference though, so you can keep it as a parameter if you like that better

pkg/retention/retention.go
```diff
@@ -416,3 +419,107 @@ func DeleteMetricsSegmentData(mmetaFile string, metricSegmentsToDelete map[strin
 		return
 	}
 }
+
+func doInodeBasedDeletion(ingestNodeDir string, maxInodeUsagePercent uint64, deletionWarningCounter int) {
```
AndrewHess, line 423: It would be good to test this function (or part of it). I think it's going to be hard to test like it is though, because a test would have to use a lot of inodes; to make it easier to test, you might want to change this `maxInodeUsagePercent` to a total number of inodes (and perhaps do the calculation of that number in the caller of this function); then to test this you can ingest some data and run this function with a low number of allowed inodes, and make sure the correct segments get deleted

sonamgupta21: I tested the function as you suggested, and it correctly deletes segments in chronological order based on latestEpochMs.

AndrewHess (approved):

PR #2021, opened 2024-12-30
Read SigLens Memory from cgroups

Macbeth98:
# Description
- Read memory from cgroups by detecting cgroup version and falling back to host memory.

# Testing
- Tested on my local docker with Kubernetes enabled and the it worked fine with cgroup v2
- Tested by creating a gcp instance with docker image set to cgroupv1 and verified that the memory is being read properly

# Checklist:


pkg/config/config.go
```diff
@@ -100,23 +112,134 @@ func init() {
...
+		memoryFilePaths.UsagePaths = []string{cgroupV2UsageMemoryPath}
+	} else {
+		memoryFilePaths.MaxPaths = []string{cgroupV1MaxMemoryPath}
+		memoryFilePaths.UsagePaths = []string{cgroupV1UsageMemoryPath}
+	}
+
+	// Add Kubernetes-specific paths only when inside Kubernetes
+	memoryFilePaths.MaxPaths = append(memoryFilePaths.MaxPaths,
+		"/sys/fs/cgroup/kubepods/memory.max",                 // Kubernetes cgroup v2
+		"/sys/fs/cgroup/kubepods.slice/memory.max",           // Kubernetes cgroup v2 with systemd
```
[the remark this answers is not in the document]
Macbeth98: I wanted to add them regardless of kubernetes env, but I added the change to add only when inside a kubernetes env.

pkg/config/common/common.go
```diff
@@ -84,6 +84,8 @@ type DatabaseConfig struct {
 }
 
 type MemoryConfig struct {
+	MaxMemoryAllowedToUseInBytes uint64 `yaml:"maxMemoryAllowedToUseInBytes"`
```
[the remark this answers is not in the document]
Macbeth98: If it is `0`, it is ignored. 
I have added the test cases.
But I want to keep it simple as `uint64` only. I added a comment mentioning that a `0` value will be ignored.

pkg/config/config.go
```diff
@@ -100,23 +112,134 @@ func init() {
...
+
+	detectCgroupVersion()
+	initMemoryPaths()
+}
+
+func detectCgroupVersion() {
+	data, err := os.ReadFile("/proc/mounts")
+	if err != nil {
+		log.Warnf("detectCgroupVersion: Error reading /proc/mounts: %v", err)
+	} else {
```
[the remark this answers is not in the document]
Macbeth98: So this is assuming that the docker containers will be on linux. I have added the code to skip the function if the OS is not `linux`.

pkg/config/config.go
```diff
@@ -84,8 +84,7 @@ const (
 type cgroupVersion uint8
 
 const (
-	noCgroupVersion cgroupVersion = iota
-	cgroupVersion1
+	cgroupVersion1 cgroupVersion = iota
```
AndrewHess, line 87: In `detectCgroupVersion` you return 0 to indicate there's no cgroup. But that's actually `cgroupVersion1`, so maybe you need `cgroupVersion1 cgroupVersion = iota + 1`

pkg/config/config.go
```diff
@@ -216,7 +247,7 @@ func detectCgroupPath(basePath string) string {
 }
 
 func getContainerMemory(memoryValueType memoryValueType) (uint64, error) {
-	if cgroupVersionDetected == noCgroupVersion {
+	if cgroupVersionDetected == 0 {
```
AndrewHess, line 250: Hmm I actually think the way you had it before was more readable. So I'd suggest changing it back

Sorry for my bad advice to remove `noCgroupVersion`

pkg/config/config_test.go
```diff
@@ -241,13 +246,14 @@ func Test_ExtractConfigData(t *testing.T) {
...
 				EnableSortIndex:             utils.DefaultValue(false),
 				MemoryConfig: common.MemoryConfig{
-					MaxUsagePercent: 80,
-					LowMemoryMode:   utils.DefaultValue(false),
-					SearchPercent:   DEFAULT_SEG_SEARCH_MEM_PERCENT,
-					CMIPercent:      DEFAULT_ROTATED_CMI_MEM_PERCENT,
-					MetadataPercent: DEFAULT_METADATA_MEM_PERCENT,
-					MetricsPercent:  DEFAULT_METRICS_MEM_PERCENT,
-					BytesPerQuery:   DEFAULT_BYTES_PER_QUERY,
+					MaxMemoryAllowedToUseInBytes: uint64(math.Min(10_000_000_000, float64(memory.TotalMemory()))),
```
AndrewHess, line 249: Nice

AndrewHess (commented):
Looks pretty good; just a few minor things

pkg/config/config.go
Macbeth98: Ahh yes!

pkg/config/config.go
Macbeth98: No worries!

AndrewHess (approved):

PR #2025, opened 2024-12-31
Implement Folder Structure for Dashboards

sonamgupta21:
# Description

1. Updated the dashboard structure to support hierarchical folders for better organization.
2. Added API endpoints for managing dashboards and folders.
3. Updated the API documentation to include all dashboard and folder-related APIs.
4. Removed the outdated dashboard_test.go file. Will create new test file will be created in a separate PR.
5. Some outstanding features:
    -   Update the folder for a dashboard.
    -   Update folder name
    -   Select multiple items - delete and move them.
    -   Duplicate dashboard option.
    -   Fix Playwright Test
    
#### Breaking changes:
  - Removed `/api/dashboards/listall` endpoint. Use `/api/dashboards/list` endpoint instead, which provides the same functionality with additional filtering and sorting options.

image
image
image
image
image
image

# Testing

# Checklist:


pkg/dashboards/folders.go
```diff
@@ -0,0 +1,883 @@
...
+func wouldCreateCircularReference(folderID, newParentID string, structure *FolderStructure) bool {
+	current := newParentID
+	for current != "" {
+		if current == folderID {
+			return true
+		}
+		if item, exists := structure.Items[current]; exists {
+			current = item.ParentID
+		} else {
+			break
```
AndrewHess, line 459: I think this will cause an infinite loop if there's already a circular reference. That might be ok, since I think we'll never get into a state where there's already a circular reference when we call this function though, right? If we do need to be more robust and handle this case, you can lookup the classic solution to checking if a linked list has a cycle

pkg/dashboards/folders.go
```diff
@@ -0,0 +1,883 @@
...
+			} else {
+				// For specific folder, show only immediate children
+				if item.ParentID != req.FolderID {
+					continue
+				}
+			}
+		}
+
+		// Apply type filter
+		if req.Type != "" && req.Type != "all" && item.Type != req.Type {
```
AndrewHess, line 593: Same here: use a constant or enum instead of `"all"`

pkg/dashboards/folders.go
```diff
@@ -0,0 +1,883 @@
...
+		} else if item.Type == "folder" {
+			if item.CreatedAt > 0 {
+				createdAt = time.UnixMilli(item.CreatedAt)
+			}
+		}
+
+		// Apply starred filter
+		if req.Starred && !isStarred {
+			continue
+		}
```
AndrewHess, line 625: You might already be handling this, but I have a question. It seems like there's three things users might want to search for: only starred, only non-starred, and all. Is there a way for them to make searches for each of these cases?

pkg/dashboards/folders.go
AndrewHess, line 1: You're going to add some unit tests for this in a different PR, right? It would be good to have unit tests at some point

pkg/dashboards/dashboards.go
```diff
@@ -579,228 +524,96 @@ func ProcessFavoriteRequest(ctx *fasthttp.RequestCtx) {
...
 			}
 		}
 	}
+
+	// Update the dashboard
 	err = updateDashboard(dId, dName, dashboardDetails, myid)
 	if err != nil {
-		if err.Error() == "dashboard name already exists" {
+		switch err.Error() {
+		case "dashboard name already exists in this folder":
```
AndrewHess, line 563: This looks like something you're returning from javascript probably? I'd recommend adding a comment to make that clear (e.g., `// keep this synced with dashboardClashMsg in foo.js`). Also, if you're using this exact string more than once in the backend, I'd recommend making a constant for it

API_README.md
```diff
@@ -1,104 +1,207 @@
...
 
+### List All Items
+    endpoints: /api/dashboards/list?sort=created-desc&type=dashboard&starred=true&folderId=folder-id&query=search
+    method: GET
+    query parameters:
+        sort: alpha-asc, alpha-desc, created-asc, created-desc
+        type: dashboard, folder, all
+        starred: true/false
+        folderId: filter by folder
+        query: search term
```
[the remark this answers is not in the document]
sonamgupta21: query: Filter items by name. For example, `query=sales` will return dashboards and folders with "sales" in their name.
I updated the documentation with a clear example.

pkg/dashboards/folders.go
```diff
@@ -0,0 +1,883 @@
...
+	ParentID string `json:"parentId"`
+}
+
+type FolderContentResponse struct {
+	Folder      FolderItemInfo   `json:"folder"`
+	Items       []FolderItemInfo `json:"items"`
+	Breadcrumbs []Breadcrumb     `json:"breadcrumbs"`
+}
+
+type FolderItemInfo struct {
```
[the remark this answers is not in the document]
sonamgupta21: They serve different purposes:

`FolderItem` is for internal storage of item in `folder_struture.json`
`FolderItemInfo` is for API responses with additional view-specific fields

Updated the name to:
`FolderItem` → `StoredFolderItem`
`FolderItemInfo` → `FolderItemResponse`

pkg/dashboards/folders.go
sonamgupta21: Currently, we can:

Search all items with `query=searchterm`
Search starred items with `query=searchterm&starred=true`
On the UI, we have a checkbox to toggle between starred and all items, so I haven't implemented the "only non-starred" filter.

pkg/dashboards/dashboards.go
sonamgupta21: The error message 'dashboard name already exists in this folder' actually originates from `updateDashboard` and `createDashboard` functions.
I've moved it to a constant `ErrDashboardNameExists`

pkg/dashboards/folders.go
sonamgupta21: I've updated the function to use Floyd's cycle-finding algorithm (tortoise and hare) to detect circular references.

pkg/dashboards/folders.go
sonamgupta21: Yes, I will add the unit tests for this in a separate PR.

playwright-tests/tests/dashboard-1.test.js
AndrewHess, line 1: Why did you comment out these tests (and a few other files)? Should you delete these files instead, or do you expect to reenable the tests soon?

pkg/dashboards/folders.go
AndrewHess: Ok, handling those two cases is probably good enough. But maybe rename "starred" to "starred only" or similar, so it's more clear that both starred and unstarred match when the value is false

AndrewHess (commented):
Looks good; just a few minor things

AndrewHess (approved):
Looks good!

PR #2034, opened 2025-01-06
Dashboard Fix

sonamgupta21:
# Description
Added `ProcessDeleteDashboardsByOrgId` function.

# Testing

# Checklist:

### DIFF
--- a/pkg/dashboards/dashboards.go
+++ b/pkg/dashboards/dashboards.go
@@ -615,3 +615,23 @@ func ProcessDeleteDashboardRequest(ctx *fasthttp.RequestCtx, myid uint64) {
 	})
 	ctx.SetStatusCode(fasthttp.StatusOK)
 }
+
+func ProcessDeleteDashboardsByOrgId(orgid uint64) error {
+	dashboardsDir := config.GetDataPath() + "querynodes/" + config.GetHostID() + "/dashboards"
+
+	// Delete dashboard details directory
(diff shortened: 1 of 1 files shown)


pkg/dashboards/dashboards.go
```diff
@@ -615,3 +615,23 @@ func ProcessDeleteDashboardRequest(ctx *fasthttp.RequestCtx, myid uint64) {
...
+
+	// Delete dashboard details directory
+	if err := os.RemoveAll(dashboardsDir + "/details"); err != nil {
+		log.Errorf("ProcessDeleteOrgData: Failed to delete dashboard details directory: %v", err)
+	}
+
+	// Delete folder structure file
+	if err := os.Remove(dashboardsDir + "/folder_structure.json"); err != nil && !os.IsNotExist(err) {
+		log.Errorf("ProcessDeleteOrgData: Failed to delete folder structure file: %v", err)
+	}
```
AndrewHess, line 630: 1. Should we return an error if either of these fail? Currently you still return a nil error unless the third check (below) fails
2. You check `IsNotExist` in the second `if` but not in the first. Should you check it in the first? Or does `RemoveAll()` never give that error?

sonamgupta21: `RemoveAll()` doesn't return ErrNotExist - it returns nil if path doesn't exist

PR #2036, opened 2025-01-07
Export metrics about metric ingestion

AndrewHess:
# Description
This adds a gauge for each of the following; the gauges are updated each minute.
- Export number of metric names
- Export metric for number of time series
- Number of datapoints per minute
- Number of tag keys (across all metric names)
- Number of tag values 
- Number of bytes on disk for metric data

# Testing
Ingested some data with `go run main.go ingest metrics -d http://localhost:8081/otsdb -t 25_000  -m 5 -p 1` and manually  check the output metrics using `curl localhost:2222/metrics`

# Checklist:

### DIFF
--- a/pkg/common/fileutils/fileutils.go
+++ b/pkg/common/fileutils/fileutils.go
@@ -209,3 +209,17 @@ func DoesFileExist(filePath string) bool {
 	_, err := os.Stat(filePath)
 	return err == nil
 }
+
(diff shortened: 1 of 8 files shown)


nkunal (approved):
Changes look good. 
Only concern I have is, every 1 minute we will be running these queries, would it cause memory/cpu overhead ?

nkunal (approved):

PR #2039, opened 2025-01-09
Metrics for logs

sonamgupta21:
# Description
Added few additional metrics 
- event count per minute
- event volume per minute
- number of indexes
- total segment count
- total column count
- total CMI size
- total CSG size
- Number of files (per index; key=indexName, value=(e.g., ind-0))
- Number of blocks (per index)

# Testing

# Checklist:

### DIFF
--- a/pkg/ingest/ingestionStats.go
+++ b/pkg/ingest/ingestionStats.go
@@ -55,8 +55,7 @@ func ingestionMetricsLooper() {
 		var totalCmiSize uint64 = 0
 		var totalCsgSize uint64 = 0
 		uniqueIndexes := make(map[string]struct{})
-		totalColumns := make(map[string]struct{})
-		segmentCounts := make(map[string]int)
+		uniqueColumns := make(map[string]struct{})
 
 		allSegmetas := segwriter.ReadGlobalSegmetas()
 
@@ -67,7 +66,7 @@ func ingestionMetricsLooper() {
 			uniqueIndexes[segmeta.VirtualTableName] = struct{}{}
 
 			for col := range segmeta.ColumnNames {
-				totalColumns[col] = struct{}{}
+				uniqueColumns[col] = struct{}{}
 			}
 
 			totalSegments++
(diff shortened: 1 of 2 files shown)


pkg/ingest/ingestionStats.go
```diff
@@ -44,13 +51,76 @@ func ingestionMetricsLooper() {
 		currentEventCount := int64(0)
 		currentBytesReceived := int64(0)
 		currentOnDiskBytes := int64(0)
+		totalSegments := int64(0)
+		var totalCmiSize uint64 = 0
+		var totalCsgSize uint64 = 0
+		uniqueIndexes := make(map[string]struct{})
+		totalColumns := make(map[string]struct{})
+		segmentCounts := make(map[string]int)
```
AndrewHess, line 59: nit:
1. Unless there's a reason not to, probably use name the index set and column set similarly. Like `uniqueIndexes` and `uniqueColumns`
2. `segmentCounts` isn't immediately obvious to me. Maybe make it more descriptive like `indexToNumSegs` or something
3. This function is getting long. It might be worthwhile to move your new code into a different function (or multiple functions), and just call that function in this function. But do whatever makes the code easier to follow

I might be missing something, but I only see `segmentCounts` being read but don't see values being set. Can this get removed?

pkg/instrumentation/ssgauges.go
```diff
@@ -143,6 +150,41 @@ var allSimpleGauges = map[Gauge]*simpleInt64Gauge{
 		unit:        "count",
 		description: "Total number of tag values",
 	},
+	TotalIndexCount: {
+		name:        "ss.total.index.count",
+		unit:        "count",
+		description: "Total number of indexes",
+	},
+	EventCountPerMinute: {
+		name:        "ss.event.count.per.minute",
```
AndrewHess, line 159: nit: there's a different metric called `ss.past.minute.num.data.points`; we should name all these per-minute metrics similarly. I think either `past.minute` or `per.minute` is good, so choose whichever you like better, and then let's stick with that from now on

AndrewHess (commented):
Looks pretty good; just a few nits

pkg/ingest/ingestionStats.go
sonamgupta21: 1. Update `totalColumns` to `uniqueColumns`
2. Removed `segmentCounts`.

AndrewHess (approved):

PR #2040, opened 2025-01-10
Fix: Reset segment search memory to Initial Allocated Value during rebalance

Macbeth98:
# Description
- Resets the segment search bytes value to the initial allocated search bytes during rebalance. 
- While rebalancing the memory assigned to segment search can be lower than the initial assignment and it needs to be reset to the initial value in the next rebalancing.

# Testing
- Tested by running siglens on a docker container with `2GB` of ram.
- Set the memory threshold to `10%` and ran logs and metrics ingestion.
- Quickly the allocated memory for search became `0` and through logs I have verified it is being reset and may become `0` again depending on the ingestion or writer memory.

# Checklist:

### DIFF
--- a/pkg/segment/memory/limit/memorylimit.go
+++ b/pkg/segment/memory/limit/memorylimit.go
@@ -33,7 +33,7 @@ import (
 
 const MINUTES_UPDATE_METADATA_MEM_ALLOC = 1
 
-var AllocatedSegSearchBytes uint64 // Should not be changed after initialization
+var SegSearchAllocatedBytes uint64 // Should not be changed after initialization
 
 var LOG_GLOBAL_MEM_FREQUENCY = 5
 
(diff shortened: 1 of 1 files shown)


pkg/segment/memory/limit/memorylimit.go
```diff
@@ -52,7 +54,7 @@ func InitMemoryLimiter() {
 	memory.GlobalMemoryTracker = &structs.MemoryTracker{
 		TotalAllocatableBytes:   totalAvailableSizeBytes,
 		RotatedCMIBytesInMemory: rotatedCMIBytes,
-		SegSearchRequestedBytes: segSearchBytes,
+		SegSearchRequestedBytes: AllocatedSegSearchBytes,
```
nkunal, line 57: If the first one is called SegSearchRequestedBytes then we should call this SegSearchAllocatedBytes

nkunal (approved):

PR #2041, opened 2025-01-10
refac: Info and error log lines

Macbeth98:
# Description
- Refactor Info and error log lines
- Updated `server.yaml` and `playground.yaml` with new memory usage percent example value.

# Checklist:

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -1158,7 +1158,7 @@ func ExtractConfigData(yamlData []byte) (common.Configuration, error) {
 
 	if segutils.ConvertUintBytesToMB(totalMemory) < SIZE_8GB_IN_MB {
 		if memoryLimits.MaxUsagePercent > 50 {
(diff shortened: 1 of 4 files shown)


pkg/config/config.go
```diff
@@ -1158,7 +1158,7 @@ func ExtractConfigData(yamlData []byte) (common.Configuration, error) {
 
 	if segutils.ConvertUintBytesToMB(totalMemory) < SIZE_8GB_IN_MB {
 		if memoryLimits.MaxUsagePercent > 50 {
-			log.Infof("ExtractConfigData: MemoryThresholdPercent is set to %v%% but bringing it down to 50%%", memoryLimits.MaxUsagePercent)
+			log.Infof("ExtractConfigData: Max Memory Usage Percent is set to %v%% but bringing it down to 50%%", memoryLimits.MaxUsagePercent)
```
nkunal, line 1161: Use the exact name in the log line as it is defined in config so that it becomes easier for user to debug

nkunal (approved):

PR #2046, opened 2025-01-10
fix: Global Unrotated Metadata Size Tracker

Macbeth98:
# Description
- This is not being tracked correctly before because when rebalancing, we set the `cmiLoaded` to false, and then from next rebalancing the allocated memory for that segment is not being deleted.
- And even when at the final stage when the segment is being removed, since the `cmiLoaded` is set to `false`, the cmi size won't be included.
- This made the `TotalUnrotatedMetadataSizeBytes` increase in value.

# Testing
- Tested by keeping the memory usage to 10% and memory size to `2GB`
- Ran the ingestion. And verified that the `TotalUnrotatedMetadataSizeBytes` was being removed properly during rebalance and removal of segment in memory after rotation

# Checklist:

### DIFF
--- a/pkg/segment/writer/unrotatedquery.go
+++ b/pkg/segment/writer/unrotatedquery.go
@@ -43,6 +43,7 @@ type UnrotatedSegmentInfo struct {
 	TableName           string
 	searchMetadataSize  uint64 // size of blockSummaries & blockInfo
 	cmiSize             uint64 // size of UnrotatedBlockCmis
+	removedCmiSize      uint64 // size of removed CMI due to memory rebalance
 	isCmiLoaded         bool   // is UnrotatedBlockCmis loaded?
 	RecordCount         int
 	orgid               uint64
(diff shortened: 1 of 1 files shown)


pkg/segment/writer/unrotatedquery.go
```diff
@@ -76,17 +77,17 @@ func RebalanceUnrotatedMetadata(totalAvailableSize uint64) uint64 {
...
 	removedSize := uint64(0)
 	count := 0
 	for i := 0; i < len(ss); i++ {
 		if removedSize >= sizeToRemove {
 			break
 		}
-
-		if ss[i].isCmiLoaded {
-			removedSize += ss[i].removeInMemoryMetadata()
+		sizeToRemove = ss[i].removeInMemoryMetadata()
```
nkunal, line 88: are we reusing this `sizeToRemove` var to save on variable names OR do we infact need this var which was defined in line 71.  Because if reuse this var gets new value in each iteration of the for loop and we are then using it in the if comparison at line 85. Which may change the existing flow logic. If this is indeed the intended behavior then you can keep the code as is (and add comments as to how these reassigning of values is affecting) OR if that is not the intention then better to create a new far for the removal here.

Macbeth98: Yes, kunal, actually we need to use new variable here. I fixed this with the latest commit!

AndrewHess (approved):

PR #2057, opened 2025-01-16
Pause and rerun Queries

Macbeth98:
# Description
- This PR needs to be merged before: #2053 
- The Queries can be paused and they can be restarted again with different `qid`, but the WS Client connection is maintained.  

# Testing
- All unit test cases passed.

# Checklist:


pkg/utils/chanutils.go
```diff
@@ -0,0 +1,27 @@
...
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+package utils
+
+func IsChannelClosed[T any](ch <-chan T) bool {
+	select {
+	case _, ok := <-ch:
+		return !ok // If `ok` is false, the channel is closed
```
AndrewHess, line 23: I think this is going to cause issues. This changes the state of the channel (removes a message) if the channel is not empty

pkg/segment/query/querystatus.go
```diff
@@ -242,39 +254,99 @@ func StartQuery(qid uint64, async bool, cleanupCallback func()) (*RunningQuerySt
...
+				log.Infof("qid=%v, Query is cancelled, not running the query again.", qid)
+				DeleteQuery(qid)
+				return nil, nil, fmt.Errorf("qid=%v, query is cancelled", qid)
+			}
+		} else {
+			// close the new channel if old one exists
+			close(rQuery.StateChan)
+			rQuery.StateChan = StateChan
+
+			rQuery.StateChan <- &QueryStateChanData{StateName: RESTARTED, Qid: qid}
```
AndrewHess, line 307: This seems convoluted to me:
- `if StateChan != nil && putils.IsChannelClosed(StateChan) && async`, then (for some reason, that's not obvious to me), this means the user canceled the query
- `if StateChan != nil && !putils.IsChannelClosed(StateChan)`, then we have to restart the query, even though we're in the function `StartQueryAsCoordinator()` so I thought we were just starting the query for the first time?

Can you clean this up somehow?

pkg/segment/query/querystatus.go
```diff
@@ -242,39 +254,99 @@ func StartQuery(qid uint64, async bool, cleanupCallback func()) (*RunningQuerySt
...
+	rQuery, err := StartQuery(qid, async, cleanupCallback, forceRun)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	var wsData *WaitStateData
+
+	defer func() {
+		if forceRun && wsData != nil {
+			go RunQuery(*wsData)
```
AndrewHess, line 282: If `forceRun == true`, won't you run the query twice (line 273 and line 282)?

pkg/segment/query/querystatus.go
```diff
@@ -86,6 +87,8 @@ const (
 	CANCELLED
 	TIMEOUT
 	ERROR
+	PAUSED
+	RESTARTED
```
AndrewHess, line 91: 1. Do we need both of these? I thought we never pause a query, we'd just have to restart it?
2. Will the query state ever be RESTARTED for a meaningful amount of time, or would the RESTARTED signal get sent and then someone listening on that channel immediately move it back to something like RUNNING?

pkg/segment/query/querystatus.go
```diff
@@ -359,8 +431,57 @@ func runQuery(wsData WaitStateData) {
...
+			astNode:       rQuery.astNode,
+			qc:            rQuery.qc,
+			aggs:          rQuery.aggs,
+			StateChan:     rQuery.StateChan,
+			rqsLock:       rQuery.rqsLock,
+		}
+	}
+
+	// remove any queries in waiting queue that are do not have a coordinator set
+	newWaitingQueries := make([]*WaitStateData, 0, len(waitingQueries))
```
AndrewHess, line 467: If you need to do this, I think the code should be restructured and/or some things should be renamed. We're inside `PauseAllRunningQueries()`, and I don't immediately see why we'd need to update which queries are waiting if we're just trying to pause the running queries

pkg/segment/segexecution.go
```diff
@@ -507,10 +542,37 @@ func ExecuteAsyncQueryForNewPipeline(root *structs.ASTNode, aggs *structs.QueryA
 	return rQuery.StateChan, nil
 }
 
+func ResubmitPausedQueries(queries map[uint64]*query.RunningQueryState) {
+	for oldQid, rQuery := range queries {
+		isAsync, isCoordinator, astNode, qc, aggs := rQuery.GetQueryInitiationInfo()
+		if !isCoordinator {
+			// This case should not happen.
```
AndrewHess, line 549: This seems like a valid code path—I think what you mean is that this function isn't intended to be used so it gets here. Maybe you should add a comment above this function for how it's intended to be used (or maybe I'm missing something)

pkg/segment/segexecution.go
```diff
@@ -507,10 +542,37 @@ func ExecuteAsyncQueryForNewPipeline(root *structs.ASTNode, aggs *structs.QueryA
...
+			qid := rutils.GetNextQid()
+			log.Infof("qid=%v, ResubmitPausedQueries: Resubmitting query with new qid=%v", oldQid, qid)
+			go func(rq *query.RunningQueryState) {
+				_, err := ExecuteAsyncQueryForNewPipeline(astNode, aggs, qid, qc, qc.Scroll, rq.StateChan, true)
+				if err != nil {
+					log.Errorf("qid=%v, ResubmitPausedQueries: error while running query, err: %v", qid, err)
+				}
+			}(rQuery)
+		}
+		// TODO: Implement for Http (non-async) queries
```
AndrewHess, line 565: I think resubmitting queries adds some potentially unneeded complexity, especially because then we have to worry about whether the query came from API or websocket. Would it be feasible to instead do something like `restartQuery(qid)`—perhaps somewhere else, as I'm not sure we'd need this `ResubmitPausedQueries()` function at all then—and then we don't have to worry about where the query initially came from, and we don't have to stop queries, mess with the waiting query queue, and re-add the stopped queries?

pkg/segment/query/querystatus.go
Macbeth98: Hmm I think, I will move this code block to `ResubmitAllPausedQueries` functions.  There I will check whether to restart a query or not. Thanks for pointing this out, this was confusing, now that I look at it.

> if StateChan != nil && putils.IsChannelClosed(StateChan) && async, then (for some reason, that's not obvious to me), this means the user canceled the query

So I said, the user has canceled the query, because when a user cancels the query, we stop listening on this channel, and the corresponding `WSConnection` also won't be there. So there would be no need to run the query again. 

I actually forgot to add a code block at `RunAsyncQueryForNewPipeline` where I would defer `close(channel)`, but I also think the channel would be close once, there are no references for it.

pkg/segment/query/querystatus.go
Macbeth98: 1. I am calling it `PAUSE`, the query running is stopped/deleted but since it will be restarted again. And also the query execution will be stopped/paused until owning of segments based on new state is completed.
2. `RESTARTED` is just a signal, it would be moved back to `RUNNING` immediately.

pkg/ast/pipesearch/searchHandler.go
```diff
@@ -590,3 +566,142 @@ func processMaxScrollCount(ctx *fasthttp.RequestCtx, qid uint64) {
...
+			} else {
+				return nil, false, root.TimeRange, fmt.Errorf("qid=%v, RunQueryForNewPipeline: query timed out", qid)
+			}
+		case query.CANCELLED:
+			log.Infof("qid=%v, RunQueryForNewPipeline: query cancelled", qid)
+		}
+	}
+}
+
+func listenToRestartQuery(qidPtr *uint64, rQueryPtr **query.RunningQueryState, isAsync bool, conn *websocket.Conn) bool {
```
AndrewHess, line 655: nit: IMO, removing one * on both the qid and rQuery, and instead returning the new qid and rQuery seems a little easier to understand (more so, since you don't actually change them in this function, but pass them to handleRestartQuery() which changes them). You can keep what you have if you prefer it

pkg/ast/pipesearch/searchHandler.go
```diff
@@ -590,3 +566,142 @@ func processMaxScrollCount(ctx *fasthttp.RequestCtx, qid uint64) {
...
+
+			if queryStateData.Qid != *qidPtr {
+				continue
+			}
+
+			if queryStateData.StateName == query.QUERY_RESTART {
+				handleRestartQuery(qidPtr, rQueryPtr, isAsync, conn)
+				return true
+			}
+		case <-timeout:
```
AndrewHess, line 681: Do you need to do anything if you get a different StateName?

pkg/segment/query/querystatus.go
```diff
@@ -42,7 +43,7 @@ import (
 
 type QueryState int
 
-var numStates = 4
+var numStates = 9
```
AndrewHess, line 46: I'm confused about this.
1. Is this for the enum values on line 84? If so, you should move it near there.
2. There's 8 states there, not 9.
3. I think this is only used for `stateChan = make(chan *QueryStateChanData, numStates)`. But I'm confused by this too, since we could get multiple update messages on the channel. I think that shouldn't be an issue—because the goroutine adding to the channel would just block until there's space—but I don't see why we'd want to base the channel size on the number of states

Perhaps I've misunderstood something.

pkg/segment/query/querystatus.go
```diff
@@ -242,39 +250,146 @@ func StartQuery(qid uint64, async bool, cleanupCallback func()) (*RunningQuerySt
...
-		return nil, putils.TeeErrorf("StartQuery: qid=%v cannot be started, Max number of waiting queries reached", qid)
+		return fmt.Errorf("addToWaitingQueriesQueue: qid=%v cannot be started, Max number of waiting queries reached", wsData.qid)
+	}
+	waitingQueries = append(waitingQueries, wsData)
+
+	return nil
+}
+
+// Starts tracking the query state. If async is true, the RunningQueryState.StateChan will be defined & will be sent updates
+// If async, updates will be sent for any update to RunningQueryState. Caller is responsible to call DeleteQuery
```
AndrewHess, line 268: nit: there's duplicate info here
> If async is true, the RunningQueryState.StateChan will be defined & will be sent updates
// If async, updates will be sent for any update to RunningQueryState.

pkg/ast/pipesearch/searchHandler.go
Macbeth98: So no, even if we get a different state, we ca ignore. We are just waiting for the query restart signal to happen. If it does not happen within given timeout it will just error out, which would have been the original flow, if restart flow is not there.

pkg/segment/query/querystatus.go
Macbeth98: Hmm, that was what I was also thinking, as it was there currently in the develop. But I changed it to `9` as before I had a `PAUSE` state, which I have removed now.

So I was thinking should we have like some predefined default size of `100`?

Yeah and also may be we also do not want to block the query processing, as this channel is mostly used for updates. So would it better to define some value that is not too large?

pkg/segment/query/querystatus.go
```diff
@@ -264,8 +270,9 @@ func addToWaitingQueriesQueue(wsData *WaitStateData) error {
 	return nil
 }
 
-// Starts tracking the query state. If async is true, the RunningQueryState.StateChan will be defined & will be sent updates
-// If async, updates will be sent for any update to RunningQueryState. Caller is responsible to call DeleteQuery
+// Starts tracking the query state. RunningQueryState.StateChan will be defined & can be used to be sent query updates
+// if forceRun is true, the query will be run immediately, otherwise it will be added to the waiting queue
+// Caller is responsible to call DeleteQuery
```
AndrewHess, line 275: nits:
1. Did you mean to remove `If async is true`?
2. I think the end of lines 273 and 275 are the end of sentences. If so, add punctuation/capitalization to make that clear
3. `can be used to be sent query updates` should change to something like `can be used to send query updates` or `can be used so the caller can be sent query updates`

pkg/segment/query/querystatus.go
```diff
@@ -289,6 +296,10 @@ func StartQuery(qid uint64, async bool, cleanupCallback func(), forceRun bool) (
 	return runningState, nil
 }
 
+// Starts tracking the query state and sets the query as a coordinator.
+// If StateChan is nil, a new channel will be created, otherwise the provided channel will be used for sending query updates.
+// if forceRun is true, the query will be run immediately, otherwise it will be added to the waiting queue
```
AndrewHess, line 301: nit: add capitalization and punctuation to make it clear line 301 is its own sentence

pkg/segment/query/querystatus.go
AndrewHess: Ok I think you should rename this then, and setting it to something like 10 is probably fine. Generally I'd expect more time to be spent actually computing the update than processing the update (i.e., sending it to the frontend), so even if we get a lot of UPDATE messages, I doubt the backend will be blocked waiting for there to be more space in the channel

pkg/segment/query/querystatus.go
Macbeth98: 1. Now Irrespective of `async` is true or not, the channel will be created that will be used to monitor the query updates.

AndrewHess (dismissed):

AndrewHess (approved):

PR #2067, opened 2025-01-22
perf: implemented a memory efficient version of GetString

### LINKED ISSUE #1538: [STORY] Implement an efficient version of GetString

Psingle20:
# Description

- Implemented the` ExtractJsonValueBuffered` function for efficient extraction of JSON values, utilizing a dynamically resizable work buffer `workBuf` to avoid unnecessary memory allocations.

- This change optimizes memory usage and improves performance by extending the buffer based on the extracted value's size, ensuring minimal overhead.

- The function handles JSON string, number, and boolean values, as well as nested structures and arrays, making it versatile for different use cases.

Fixes #1538  

# Testing
The code was tested using the following steps:

1. Verified the function behavior with various JSON inputs, including nested arrays and objects.
2. Used unit tests to confirm that the extracted values matched the expected results.
3. Ensured that the buffer (workBuf) was appropriately extended without erasing old data.
4. Confirmed no memory leaks or inefficiencies by testing with large JSON inputs.
To reproduce the tests:

Run the existing test suite with go test ./....
Use the provided JSON test cases in jsonutils_test.go.

# Checklist

### DIFF
--- a/pkg/utils/jsonutils.go
+++ b/pkg/utils/jsonutils.go
@@ -19,6 +19,7 @@ package utils
 
 import (
 	"fmt"
+	"github.com/buger/jsonparser"
 )
 
 // Flatten takes a map and returns a new one where nested maps are replaced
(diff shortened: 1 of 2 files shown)


nkunal:
most recently (after the issue #1538 was filed), we discovered there could be some issues reusing the buffer , which resulted in corrupted data being stored. Hence we will not be doing this. I will be closing the issue as well as this ticket

PR #2097, opened 2025-01-31
Maintain Virtual table names and avoid duplicate appends to virtual table files.

Macbeth98:
# Description
- Currently there is an issue that the we update the virtual table file with the same table names when the server is started.
- This PR fixes that issue.
- And also uses go routines to update the data all at once as bulk, instead of doing one at a time.
- Refactors and removed some of the other code.

# Testing
- Tested by starting the server with no data and ingested across multiple indexes.
- Kill the server, verify the local file with proper index names.
- Restart the server and verify that the indexnames are not duplicated.

# Checklist:

### DIFF
--- a/pkg/segment/query/queryrefresh.go
+++ b/pkg/segment/query/queryrefresh.go
@@ -126,16 +126,7 @@ func RefreshGlobalMetadata(fnMyids func() []int64, ownedSegments map[string]stru
 		}
 	}
 
-	default_id := int64(0)
-	// myids should also include the default_id
 	myids := fnMyids()
-	if len(myids) == 0 {
-		myids = append(myids, default_id)
-	}
-
-	if myids[0] != default_id {
-		myids = append(myids, default_id)
-	}
 
 	allSfmFiles := make([]string, len(ingestNodes))
 
--- a/pkg/segment/query/queryrefresh.go
+++ b/pkg/segment/query/queryrefresh.go
@@ -95,7 +95,7 @@ func initMetadataRefresh() {
 	initMetricsMetaRefresh()
 }
 
-func updateVTable(vfname string, orgid uint64) error {
+func UpdateVTable(vfname string, orgid uint64) error {
 	vtableFd, err := os.OpenFile(vfname, os.O_RDONLY, 0644)
 	if err != nil {
 		if errors.Is(err, os.ErrNotExist) {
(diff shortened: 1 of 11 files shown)


pkg/segment/query/queryrefresh.go
```diff
@@ -162,40 +125,88 @@ func RefreshGlobalMetadata(fnMyids func() []int64, ownedSegments map[string]stru
...
+				log.Errorf("RefreshGlobalMetadata: Error in getting vtable names for default org, err:%v", err)
+				return
 			}
+
 			for _, myid := range myids {
-				vfname := virtualtable.GetFilePathForRemoteNode(node, myid)
-				err := updateVTable(vfname, myid)
+				vTableFileName = virtualtable.GetFilePathForRemoteNode(ingestNode, myid)
+				vTableNamesMap[myid] = make(map[string]bool)
+				err := virtualtable.LoadVirtualTableNamesFromFile(vTableFileName, vTableNamesMap[myid])
```
AndrewHess, line 165: nit: instead of duplicating this code, do just the `for` block and either:
1. Add a comment at the top of the function saying `fnMyIds` should include the default ID of 0
2. After `myids := fnMyids()`, insert the default ID into `myids` if not present

pkg/virtualtable/virtualtable.go
```diff
@@ -151,54 +172,125 @@ func CreateVirtTableBaseDirs(vTableBaseDir string, vTableMappingsDir string,
...
 	}
+
+	if vTableFileUpdated {
+		go func() {
+			err := blob.UploadIngestNodeDir()
+			if err != nil {
+				log.Errorf("AddVirtualTable: Failed to upload vtable data to blob store, err=%v", err)
+			}
+		}()
+	}
```
AndrewHess, line 246: Do you need to be holding a write lock for the duration of this? Is it possible that we'd have an async download call at the same time that could corrupt the ingest node dir?

pkg/segment/query/queryrefresh.go
```diff
@@ -124,8 +125,17 @@ func RefreshGlobalMetadata(fnMyids func() []int64, ownedSegments map[string]stru
...
+	// myids should also include the default_id
 	myids := fnMyids()
-	defaultMyId := int64(0)
+	if len(myids) == 0 {
+		myids = append(myids, default_id)
+	}
+
+	if myids[0] != default_id {
+		myids = append(myids, default_id)
+	}
```
AndrewHess, line 138: Now you have an odd requirement: if `myids` contains the default ID, then it's supposed to contain it as the first element. I think you should get rid of that requirement. Just iterate `myids` and check if it's present (or maybe there's a util for that?); also then you won't need this edge case of `len(myids) == 0`

Macbeth98: I am not exactly sure why the code was written like this. But I think I can remove this addition of the default index.
But based on the code, we do not need this at all.

AndrewHess (approved):

PR #2098, opened 2025-01-31
Fix incorrect type conversion

AndrewHess:
# Description
We have two ways to store types: as `VALTYPE_XXX` and as `SS_DT_XXX`. We were incorrectly casting from one type to the other. Now we do the conversion correctly.

**Update:** The original version of this had some failing tests in the functional suite. This happened because `initMicroIndices()` was passed a VALTYPE instead of an SS_DTYPE and the string VALTYPE is the signed integer SS_DTYPE. `initMicroIndices()` is called when a column needs to be backfilled, so when we were supposed to backfill a field as a string column, we'd actually backfill it as a numeric column. The functional suite tests failed because the tests were created on ingested data that had this bug. In particular, some columns like SSN and Zip Code have numeric data that's actually ingested as a string. In our variable column functional tests, some of the rows don't have data for certain columns, so for example a column that had Zip Code data was previously incorrectly backfilled as a numeric column and then some functional tests did numeric calculations on that column. With the fix in this PR, the column is correctly backfilled as a string column, so doing numeric calculations on the string column gave "incorrect" results; so I fixed the tests to explicitly convert the strings to numbers.

**Note**: In the functional suite tests, there's some columns like `variable_col_<i>`. The data in those columns is the same type of data as the `i`-th element in [this slice](https://github.com/siglens/siglens/blob/develop/tools/sigclient/pkg/utils/dynamicuserutils.go#L27) (zero-indexed). So `variable_col_4` has fake SSN info.

# Testing
1. Start siglens
2. `git clone https://github.com/open-telemetry/opentelemetry-demo.git` and `cd` into it
3. Set the existing `src/otel-collector/otelcol-config-extras.yml` to
```
exporters:
  otlphttp/siglens:
    endpoint: "http://host.docker.internal:8081/otlp"
  prometheusremotewrite:
    endpoint: "http://host.docker.internal:8081/promql/api/v1/write"

processors:
  batch:
    send_batch_size: 5000
    timeout: 10s

service:
  pipelines:
    traces:
    metrics:
      processors: [batch]
      exporters: [prometheusremotewrite]
```
4. `make start`
5. Check the siglens log file
Without this PR, there we a lot of errors like `initMicroIndices: unknown valType: 16` and a few with `initMicroIndices: unknown valType: 17`. With this PR, there's just a few with `initMicroIndices: unknown ssType: 4` (this is because valtype 17 is sstype 4 which is float64, which initMicroIndices doesn't currently handle; the valType 16 though is a signed num sstype, which initMicroIndices does handle though).
 
# Checklist:

### DIFF
--- a/tools/sigclient/functionalQueries/functionalQueries.yml
+++ b/tools/sigclient/functionalQueries/functionalQueries.yml
@@ -73,7 +73,7 @@ variable_col_tests/var_multi_value.json
 variable_col_tests/var_mvexpand.json
 variable_col_tests/var_rex_stats_groupby_eval_stats.json
 variable_col_tests/var_stats_eval.json
-variable_col_tests/var_eval_timechart_sort.json
+# variable_col_tests/var_eval_timechart_sort.json this test is wrong; the expected values need to be updated
 variable_col_tests/var_search_timechart_no_groupby.json
 variable_col_tests/var_tail.json
 variable_col_tests/var_sort_tail_eval_newField.json
--- a/pkg/segment/utils/segconsts.go
+++ b/pkg/segment/utils/segconsts.go
@@ -176,6 +176,42 @@ const (
 	SS_DT_RAW_JSON
 )
 
+func ValTypeToSSDType(valtype byte) SS_DTYPE {
+	switch valtype {
+	case VALTYPE_ENC_BOOL[0]:
+		return SS_DT_BOOL
+	case VALTYPE_ENC_UINT8[0]:
+		return SS_DT_USIGNED_8_NUM
+	case VALTYPE_ENC_UINT16[0]:
+		return SS_DT_USIGNED_16_NUM
+	case VALTYPE_ENC_UINT32[0]:
+		return SS_DT_USIGNED_32_NUM
+	case VALTYPE_ENC_UINT64[0]:
+		return SS_DT_UNSIGNED_NUM
+	case VALTYPE_ENC_INT8[0]:
+		return SS_DT_SIGNED_8_NUM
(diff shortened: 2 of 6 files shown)


nkunal:
looks good to me, but not sure why the functional test failed on the return value comparison

AndrewHess:
> not sure why the functional test failed on the return value comparison

Fixed. See the update in the PR description

tools/sigclient/functionalQueries/functionalQueries.yml
```diff
@@ -73,7 +73,7 @@ variable_col_tests/var_multi_value.json
 variable_col_tests/var_mvexpand.json
 variable_col_tests/var_rex_stats_groupby_eval_stats.json
 variable_col_tests/var_stats_eval.json
-variable_col_tests/var_eval_timechart_sort.json
+# variable_col_tests/var_eval_timechart_sort.json this test is wrong; the expected values need to be updated
```
nkunal, line 76: should we upload this as part of this PR or we do that in a separate PR

AndrewHess: This test fails with this PR fix. I guess I could make a PR just to disable this test, merge that first, and then merge this PR. Is there a benefit to that?

nkunal: no thats fine. I just realized i mis-typed, i meant to say should we update the test results as part of this pr or make that as a seperate pr.
I guess we can leave the test disabled and in another pr we enable the test and fix the expected results part

nkunal (approved):

PR #2109, opened 2025-02-06
fix: sync SFM With Segmeta on Bootup

Macbeth98:
# Description
- `syncSegMetaWithSegFullMeta` function will check for any residual pending `sfm` files that are not in the `segmeta.json`.
- Currently, on bootup, the code expects these `sfm` segkeys to be in present in the `ownedSegments`, but they will not as they are not existed at all.
- With this PR, check for the segkeys in `AllSegKeys` metadata and if exists, we do not need to sync even if the segment is owned or not.
- If it does not exist on `AllSegKeys`, then it needs to be synced. This PR does that.
- Also refactored some part with rotation of logs segment.

# Testing
- Tested by running some ingestion and killed the server.
- Restarted the server and ensured the segmeta is updated and able to run queries.

# Checklist:


pkg/segment/query/queryrefresh.go
```diff
@@ -356,7 +354,9 @@ func syncSegMetaWithSegFullMeta(myId int64) {
 	}
 
 	writer.BulkAddRotatedSegmetas(segMetaSlice, false)
-	log.Infof("syncSegMetaWithSegFullMeta: Added %d segmeta entries", smiCount)
+	log.Infof("syncSegMetaWithSegFullMeta: myid=%v, Added %d segmeta entries", myId, smiCount)
+
+	return smiCount
```
AndrewHess, line 359: nit: from the log message, `smiCount` is the number of SMIs added, not the total. But just from `smiCount` it's not clear to me which it is. Since you're returning the number now, I'd recommend renaming it to clarify it's the number added

AndrewHess (commented):
Looks good. One nit

AndrewHess (approved):

PR #2119, opened 2025-02-06
Refactor OTLP trace ingestion

AndrewHess:
# Description
This shouldn't change any behavior. It's intended to make it easier to add OTLP log ingestion in the next PR. Then the package structure will be something like
- utils.go
- traces.go
- logs.go
- (eventually) metrics.go

# Testing
I was still able to manually ingest traces using the OpenTelemetry Demo with this config
```
exporters:
  otlphttp/siglens:
    endpoint: "http://host.docker.internal:8081/otlp"
  prometheusremotewrite:
    endpoint: "http://host.docker.internal:8081/promql/api/v1/write"

processors:
  batch:
    send_batch_size: 5000
    timeout: 10s

service:
  pipelines:
    traces:
    metrics:
      processors: [batch]
      exporters: [prometheusremotewrite]
```

# Checklist:

### DIFF
--- a/pkg/otlp/utils.go
+++ b/pkg/otlp/utils.go
@@ -6,11 +6,32 @@ import (
 	"fmt"
 	"io"
 
+	"github.com/siglens/siglens/pkg/utils"
 	log "github.com/sirupsen/logrus"
 	"github.com/valyala/fasthttp"
 	"google.golang.org/genproto/googleapis/rpc/status"
 	"google.golang.org/protobuf/proto"
 )
+
+func getDataToUnmarshal(ctx *fasthttp.RequestCtx) ([]byte, error) {
+	// We don't support JSON requests yet.
+	if contentType := utils.GetContentType(ctx); contentType != utils.ContentProtobuf {
+		log.Errorf("getDataToUnmarshal: got a non-protobuf request. Got Content-Type: %s", contentType)
+		return nil, fmt.Errorf("Expected a protobuf request")
+	}
+
+	// From https://opentelemetry.io/docs/specs/otlp/#otlphttp-response:
+	// The server MUST use the same “Content-Type” in the response as it received in the request.
+	utils.SetContentType(ctx, utils.ContentProtobuf)
+
+	data, err := getUncompressedData(ctx)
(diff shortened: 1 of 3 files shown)


pkg/otlp/utils.go
```diff
@@ -0,0 +1,82 @@
...
+	"github.com/valyala/fasthttp"
+	"google.golang.org/genproto/googleapis/rpc/status"
+	"google.golang.org/protobuf/proto"
+)
+
+func getDataToUnmarshal(ctx *fasthttp.RequestCtx) ([]byte, error) {
+	// We don't support JSON requests yet.
+	if contentType := utils.GetContentType(ctx); contentType != utils.ContentProtobuf {
+		log.Errorf("getDataToUnmarshal: got a non-protobuf request. Got Content-Type: %s", contentType)
+		return nil, fmt.Errorf("Expected a protobuf request")
```
nkunal, line 20: we should probably do:
```
if ... not protobuf {
        return nil, fmt.Errorf(getDataToUnmarshal: got a non-protobuf request. Got Content-Type: %s", contentType)
}

Comment (nkunal) [reply pkg/otlp/utils.go:20]: and have the caller log the error msg, so that 
1) we get only one log line
2) when we implement logs and metrics, we can differentiate easily at the caller for which type this error was

File: pkg/otlp/utils.go [hunk 2]
```diff
@@ -0,0 +1,82 @@
...
+	}
+
+	// From https://opentelemetry.io/docs/specs/otlp/#otlphttp-response:
+	// The server MUST use the same “Content-Type” in the response as it received in the request.
+	utils.SetContentType(ctx, utils.ContentProtobuf)
+
+	data, err := getUncompressedData(ctx)
+	if err != nil {
+		log.Errorf("getDataToUnmarshal: failed to uncompress data: %v", err)
+		return nil, fmt.Errorf("Unable to gzip decompress the data")
```
nkunal, line 30: same comment as above

nkunal (commented):
looks good, just one minor comment

nkunal (approved):

PR #2160, opened 2025-02-12
Generate Gantt Chart by SpanID

Tominatoo:
# Description
    endpoint: api/traces/span/ganttChart
    method: POST

    Example:
    request: http://localhost:5122/api/traces/span/ganttChart
    body:
        {
            "startEpoch": "now-1h",
            "endEpoch": "now",
            "searchText": "span_id=a98166653c6f7f44"
        }

Finds the corresponding trace_id from the span_id and generates a gantt chart according to it. If the span_id does not have a corresponding trace_id, it returns an error of an orphaned span_id state. 

Right now it only logs a comment if there are more than 1 associated traces to a span_id, but later on this case should be handled properly. 

# Testing

# Checklist:

### DIFF
--- a/pkg/segment/tracing/handler/tracehandler.go
+++ b/pkg/segment/tracing/handler/tracehandler.go
@@ -276,6 +276,23 @@ func ExtractTraceID(searchText string) (bool, string) {
 	return true, matches[1]
 }
 
+// Check if searchText only contains spanId as query condition
+func ExtractSpanID(searchText string) (bool, string) {
+	pattern := `^span_id=([a-zA-Z0-9]+)$`
+
+	regex, err := regexp.Compile(pattern)
+	if err != nil {
+		return false, ""
+	}
+
+	matches := regex.FindStringSubmatch(searchText)
+	if len(matches) != 2 {
+		return false, ""
+	}
+
+	return true, matches[1]
+}
+
 func AddTrace(pipeSearchResponseOuter *segstructs.PipeSearchResponseOuter, traces *[]*structs.Trace, traceId string, traceStartTime uint64,
 	traceEndTime uint64, serviceName string, operationName string) {
 	spanCnt := 0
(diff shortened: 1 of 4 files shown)


nkunal (approved):

pkg/segment/tracing/handler/tracehandler.go
```diff
@@ -276,6 +276,23 @@ func ExtractTraceID(searchText string) (bool, string) {
...
 }
 
+// Check if searchText only contains spanId as query condition
+func ExtractSpanID(searchText string) (bool, string) {
+	pattern := `^span_id=([a-zA-Z0-9]+)$`
+
+	regex, err := regexp.Compile(pattern)
+	if err != nil {
+		return false, ""
+	}
```
AndrewHess, line 286: FYI: you can use `regexp.MustCompile()` here. `MustCompile` will panic if it can't compile, so we definitely don't want to do it on user-provided regex, but for hardcoded regex like this it's ok and avoids this error handling

But, you can keep `Compile()` if you want

pkg/segment/tracing/handler/tracehandler.go
```diff
@@ -988,3 +1005,83 @@ func ProcessSearchTraceRelatedLogsRequest(ctx *fasthttp.RequestCtx, myid int64)
...
+	if !isOnlySpanID {
+		writeErrMsg(ctx, "ProcessSpanGanttChartRequest", "only provide 1 span ID", nil)
+		return
+	}
+
+	page := 1
+	pageVal, ok := readJSON["page"]
+	if !ok || pageVal == 0 {
+		page = 1
+	} else {
```
AndrewHess, line 1030: nit: the inside of the `if` block doesn't do anything, so you can restructure to delete it. But you can also keep this if you want

AndrewHess (approved):

pkg/segment/tracing/handler/tracehandler.go
Tominatoo: Changed!

AndrewHess (approved):

PR #2182, opened 2025-02-14
Count function in PromQL matches expected behavior - now no longer op…

Tominatoo:
…erates on all time series by default

# Description
Count now operates as an aggregate function the same way as sum and avg do. 
count(group by ({groups}) ({metric})) 
Will now count all of the groups and return a single value, instead of returning {groups} number of counts of value 1. 
This now matches how `count` is handled in PromQL. 
Unit test on count aggregation has been changed to account for this change.

# Testing

# Checklist:

### DIFF
--- a/pkg/segment/results/mresults/seriesresult.go
+++ b/pkg/segment/results/mresults/seriesresult.go
@@ -820,6 +820,8 @@ func reduceEntries(entries []Entry, fn utils.AggregateFunctions, fnConstant floa
 			}
 		}
 	case utils.Count:
+		// Maybe do something here? return len(entries)?
+		ret += float64(len(entries))
 		// Count is to calculate the number of time series, we do not care about the entry value
 	case utils.Quantile: //valid range for fnConstant is 0 <= fnConstant <= 1
 		// TODO: calculate the quantile without needing to sort the elements.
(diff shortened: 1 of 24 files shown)


pkg/segment/structs/metricsstructs.go
```diff
@@ -448,7 +448,7 @@ func (agg Aggregation) ShallowClone() *Aggregation {
 }
 
 func (agg Aggregation) IsAggregateFromAllTimeseries() bool {
-	return agg.AggregatorFunction == utils.Count || agg.AggregatorFunction == utils.Stdvar || agg.AggregatorFunction == utils.Stddev || agg.AggregatorFunction == utils.TopK || agg.AggregatorFunction == utils.BottomK
+	return agg.AggregatorFunction == utils.Stdvar || agg.AggregatorFunction == utils.Stddev || agg.AggregatorFunction == utils.TopK || agg.AggregatorFunction == utils.BottomK
 }
```
AndrewHess, line 452: I'm not quite sure what this function is for, from this documentation: https://prometheus.io/docs/prometheus/latest/querying/operators/#aggregation-operators I would think that `count` should be included.

Can you explain to me why excluding it is correct?

pkg/segment/results/mresults/seriesresult.go
```diff
@@ -820,6 +820,8 @@ func reduceEntries(entries []Entry, fn utils.AggregateFunctions, fnConstant floa
 			}
 		}
 	case utils.Count:
+		// Maybe do something here? return len(entries)?
+		ret += float64(len(entries))
```
AndrewHess, line 824: nit: is this comment still needed?

PR #2191, opened 2025-02-16
Alerts

SanketFZ:
Description

Added Structure in alertutils.go  ( Added APM Parameters )
and make changes API calls in 

Please review it and check if its right
# Checklist:


pkg/alerts/alertutils/alertutils.go
```diff
@@ -31,6 +31,7 @@ type AlertType uint8
 const (
 	AlertTypeLogs AlertType = iota + 1
 	AlertTypeMetrics
+	AlertTypeAPM
```
nkunal, line 34: never add a new enum type in the middle, it should be always at the end, otherwise it causes backwards compatability issue with existing data

pkg/alerts/alertutils/alertutils.go
```diff
@@ -56,6 +57,7 @@ type AlertDetails struct {
 	SilenceMinutes           uint64     `json:"silence_minutes"`
 	SilenceEndTime           uint64     `json:"silence_end_time"`
 	MetricsQueryParamsString string     `json:"metricsQueryParams"`
+	APMQueryParamsString     string     `json:"apmQueryParams"` // APM
```
nkunal, line 60: same comment here

SanketFZ: Ok I will make changes, will do the same as advised

SanketFZ:
_

PR #2194, opened 2025-02-17
Added tests for the InitSyncSegmeta function

Macbeth98:
# Description
- Refactored the `initSyncSegMetaForAllIds` function to take `GetAllSegmentsHook` as a parameter.
- Refactored and fixed the mock ingestion function: `InitMockColumnarMetadataStore`.
- Added a bulk version of the same that can ingest across myids and indexes.
- Made fixes for other test cases because of this refactoring.
- Added tests for `initSyncSegMetaForAllIds` with and without `GetAllSegmentsHook`

# Checklist:


pkg/segment/query/segquery_test.go
```diff
@@ -428,3 +427,96 @@ func Test_segQueryFilter(t *testing.T) {
...
+	segKeysToDelete := make(map[string]struct{})
+	count := 0
+
+	for segKey := range segKeysSet {
+		if count == numSegKeys {
+			break
+		}
+
+		segKeysToDelete[segKey] = struct{}{}
+		count++
```
AndrewHess, line 462: I think you should add another variable called `numSegKeysToDelete` and use that here instead of using `numSegKeys`. Right now the test is a little hard to read because you're making these two independent values be the same thing. This also makes your later assertion less clear:
```
	segmetadata.DeleteSegmentKeys(segKeysToDelete)

	allSegKeys = segmetadata.GetAllSegKeys()
	assert.Len(t, allSegKeys, totalSegKeysCount-numSegKeys)
```
but if you add this new variable or even do just the following it's much more clear
```
assert.Len(t, allSegKeys, totalSegKeysCount-len(segKeysToDelete))
```

pkg/segment/query/segquery_test.go
```diff
@@ -428,3 +427,96 @@ func Test_segQueryFilter(t *testing.T) {
...
+		assert.False(t, ok)
+	}
+
+	for segKey := range KeysToAdd {
+		_, ok := allSegKeys[segKey]
+		assert.True(t, ok)
+	}
+
+	err = os.RemoveAll(dir)
+	assert.Nil(t, err)
```
AndrewHess, line 521: nit: you don't have to do this. You made the dir from `TempDir()`, so it automatically gets deleted. You can keep it if you want though

pkg/segment/query/segquery_test.go
```diff
@@ -428,3 +427,96 @@ func Test_segQueryFilter(t *testing.T) {
...
+	segmetadata.DeleteSegmentKeys(segKeysToDelete)
+
+	allSegKeys = segmetadata.GetAllSegKeys()
+	assert.Len(t, allSegKeys, totalSegKeysCount-numSegKeys)
+
+	keysToNotAddCount := 2
+	keysToNotAdd := make(map[string]struct{})
+	KeysToAdd := make(map[string]struct{})
+
+	allSegKeysHook := func() (map[string]struct{}, error) {
```
AndrewHess, line 490: I'm not really sure what's happening with this `allSegKeysHook`. I mean I can read the code and understand it, but it's not obvious to me why we're doing this. Is this a completely separate test than the above test of deleting segkeys? 

Adding a few brief comments in this test would make it a lot easier to understand. They don't have to be long comments; something simple like "test initing segments with all but the first two" would be fine

pkg/segment/query/metadata/metamocks.go
```diff
@@ -22,27 +22,41 @@ import (
...
+	"github.com/siglens/siglens/pkg/virtualtable"
 	log "github.com/sirupsen/logrus"
 )
 
-// function to init mock server in memory. Should only be called by tests
-func InitMockColumnarMetadataStore(segBaseDir string, count int, numBlocks int, entryCount int) {
+func InitMockColumnarMetadataStoreInternal(myid int64, indexName string, count int, numBlocks int, entryCount int) ([]string, error) {
+	err := virtualtable.AddVirtualTable(&indexName, myid)
+	if err != nil {
+		return nil, fmt.Errorf("InitMockColumnarMetadataStore: AddVirtualTable failed err=%v", err)
```
AndrewHess, line 42: nit: wrong function name on line 42 with the name change

Also if this is an internal function, then maybe it should be private?

pkg/segment/query/metadata/metamocks.go
```diff
@@ -22,27 +22,41 @@ import (
...
 
 	for i := 0; i < count; i++ {
-		segkey := segBaseDir + fmt.Sprint(i)
-		bsumFname := segBaseDir + fmt.Sprint(i) + ".bsu"
+		segBaseDir := filepath.Join(vTableBaseDir, fmt.Sprint(i), "/")
+		err := os.MkdirAll(segBaseDir, 0755)
+		if err != nil {
+			return nil, fmt.Errorf("InitMockColumnarMetadataStore: MkdirAll failed err=%v", err)
+		}
+		segkey := filepath.Join(segBaseDir, fmt.Sprint(i))
```
AndrewHess, line 58: nit: there's an existing `config.GetSegKeyFromVTableDir()` you can use in place of line 58

pkg/segment/query/metadata/metamocks.go
```diff
@@ -93,7 +107,54 @@ func InitMockColumnarMetadataStore(segBaseDir string, count int, numBlocks int,
...
+
+		segKeys[i] = segkey
 	}
+
+	return segKeys, nil
+}
+
+// function to init mock server in memory. Should only be called by tests
+func InitMockColumnarMetadataStore(myid int64, indexName string, count int, numBlocks int, entryCount int) ([]string, error) {
+
```
AndrewHess, line 121: nit: why does this return a list of strings but the Bulk version returns a set of strings? I was going to suggest having this function just call the Bulk version so you remove the duplicate code between these two; but if you really do want them to return different types, maybe that refactoring wouldn't be worth it

Macbeth98: Yeah at first I made this function also return the set of strings instead of slice.
But this function is being used at many other places and usually only for one segKey.
Previously this function's input is segBaseDir, so only for that one segment. Now I changed it to vTableBasedDir and returning slice of strings, so that the access of the segKey from this will be easy.

AndrewHess (approved):

PR #2195, opened 2025-02-17
Fixed error in multiple queries not executing properly

Tominatoo:
# Description

All aggregations that require the entire TimeSeries now use a passed parameter of seriesEntriesMap instead of using r.DsResults (which should be set to nil after any aggregation). This is either calculated from r.DsResults on an initial aggregation, or calculated from r.Results on further aggregations.
All current expected functionality is still there, and count now works when using it with multiple queries, or using count on a subquery: count(group by (car_type) (testmetric0)).
Stdvar also works properly when using it on a subquery - though topk & bottomk still don't work properly on a subquery. No functionality has been lost from stdvar, topk, or bottomk, though. I just updated them to use seriesEntriesMap for consistency. I was told getting these 3 other functions to work properly on multiple queries wasn't in the scope of this task

# Testing

# Checklist:


AndrewHess (commented):
Looks good. Can you add a test for a query that now works with this change so we don't later regress? Our existing metrics_test.csv is probably a good spot, or you can put it elsewhere if you know a better spot

static/js/log-results-grid.js
AndrewHess, line 1: Are these your changes or from merging with `develop` updates? They probably don't belong in this PR if they're your changes

pkg/segment/results/mresults/metricresults.go
```diff
@@ -197,7 +197,24 @@ func (r *MetricsResult) AggregateResults(parallelism int, aggregation structs.Ag
 	// For some aggregations like sum and avg, we can compute the result from a single timeseries within a vector.
 	// However, for aggregations like count, topk, and bottomk, we must retrieve all the time series in the vector and can only compute the results after traversing all of these time series.
 	if aggregation.IsAggregateFromAllTimeseries() {
-		err := r.aggregateFromAllTimeseries(aggregation)
+		seriesEntriesMap := make(map[string]map[uint32][]RunningEntry, 0)
+
+		for grpID, ds := range r.DsResults {
+			if ds == nil {
+				continue
+			}
```
AndrewHess, line 205: nit: do we expect to ever have `ds == nil`? If not, maybe add a `log.Warnf()` just before the `continue`

AndrewHess (commented):
Looks pretty good. I'm confused about one thing: in your PR description you say topK/bottomK/stdvar are out of scope, but you do update the code for those in this PR. So did you fix them or not? If you did fix them, please add E2E tests for them like you did for count. If you didn't fix them, I'd agree they're out of scope, so you don't need any more changes for them

static/js/log-results-grid.js
Tominatoo: These are coming from this commit: [12f5e2c](https://github.com/siglens/siglens/pull/2195/commits/12f5e2caa77a6b1d3394261a5209c62dee32f442)

Where it seems github is automatically changing the format of these files. I don't know why this is doing that.

Tominatoo:
> Looks pretty good. I'm confused about one thing: in your PR description you say topK/bottomK/stdvar are out of scope, but you do update the code for those in this PR. So did you fix them or not? If you did fix them, please add E2E tests for them like you did for count. If you didn't fix them, I'd agree they're out of scope, so you don't need any more changes for them

I changed the code to use the same workflow as the count function.
Meaning, instead of using r.DsResults, they will all use the seriesEntriesMap parameter. This was to keep consistency in the code. This kept all working functionality intact for these other functions (computeExtremesKElements, computeAggStdvarOrStddev), but topk and bottomk still aren't working properly on a subquery, which is out of this scope. Me making these changes were not to fix them (although it did fix stdvar and stddev) but to ensure consistency of these functions that require the entire timeseries.

pkg/segment/results/mresults/metricresults.go
Tominatoo: No, I don't think ds == nil should ever occur. I changed it to return an error if it is nil, similar to how an error is handled if aggregateFromAllTimeseries returns an error.

pkg/segment/results/mresults/metricresults.go
```diff
@@ -201,7 +201,9 @@ func (r *MetricsResult) AggregateResults(parallelism int, aggregation structs.Ag
 
 		for grpID, ds := range r.DsResults {
 			if ds == nil {
-				continue
+				err := fmt.Errorf("Group %v has nonexistent downsample series", grpID)
```
AndrewHess, line 204: nit: prepend the function name to the error, so it's easier to debug

AndrewHess (approved):

PR #2196, opened 2025-02-17
[UX] Introduction of Hotkeys

Venkata-Sai-Vishwanath-robo:
# Description
This PR adds two helpful hotkeys to enhance user interaction and improve the overall experience.
| **Key Combination**       | **Action** |
|--------------------|------------|
| **Ctrl + Enter**         | Run Query |
| **/**   | Focus the folder and dashboard search input |

The implementation is designed to be easily extendable, allowing for the addition of more hotkeys in the future.

# Testing

# Checklist:


static/dashboards-home.html
```diff
@@ -81,7 +81,7 @@ <h1 class="myOrg-heading">Dashboards</h1>
             <div id="db-search-container">
                 <div>
                     <input type="text" id="search-input" class="form-control search-db-input"
-                        placeholder="Search for dashboards and folders">
+                        placeholder="Type / to search for dashboards and folders">
```
nkunal, line 84: we can search even without needing the `/` as soon as you start typing, the search happens

nkunal (commented):
The Ctrl+Enter on query builder is good enhancement to trigger the search. 
But requiring the / in the other two places, it doesn't seem like that is needed (see comments in line)
Thanks for you contribution!

static/dashboards-home.html
Venkata-Sai-Vishwanath-robo: I actually don't see that happening, Please let me know if I am missing something.
(Just for clarification: this feature is similar to the github's search where in if the user presses the '/' key the search bar gets the focus).

sonamgupta21 (commented):
Hey @Venkata-Sai-Vishwanath-robo ,

You can remove the `/` functionality as it is not needed for now.

For running a query, please update the shortcut to `Shift + Enter`.
There is no need to create a new file or function, as there is already a function `filterInputHandler()` to run the query on pressing `Enter` in the `event-handler.js` file. You can simply update that to `Shift + Enter`.

Thank you!

Venkata-Sai-Vishwanath-robo:
> Hey @Venkata-Sai-Vishwanath-robo ,
> 
> You can remove the `/` functionality as it is not needed for now.
> 
> For running a query, please update the shortcut to `Shift + Enter`. There is no need to create a new file or function, as there is already a function `filterInputHandler()` to run the query on pressing `Enter` in the `event-handler.js` file. You can simply update that to `Shift + Enter`.
> 
> Thank you!

Hello @sonamgupta21 

Sorry for reverting late.
I’ve created a new PR based on your suggestions. Currently, the keyup event listener is attached to the filter-input text area element. Would it be better to make the hotkey global instead?

Also, since Shift + Enter is commonly used to move to the next line, would it make sense to use Shift + Enter for a soft return and Enter to submit? Let me know your thoughts! 😊

PR #2215, opened 2025-02-19
Fix cluster stats column count in myorg page

nrkawane:
# Description

Column count listed under ‘Total’ does not much up to be addition of all column counts for various indices shown on the cluster stats (under My Org).

Screenshot 2025-02-20 211748

# Testing
1. Add test data (sigclient / 'Send Test Data' through UI)
2. Go To Cluster Stats under 'My Org'
3. Verify the Total column count equals the addition of column counts for all the listed indices


pkg/health/clusterStatsHandler.go
```diff
@@ -395,15 +397,13 @@ func getStats(myid int64, filterFunc func(string) bool, allSegMetas []*structs.S
 		if !exist {
 			allIndexCols[indexName] = make(map[string]struct{})
 			allIndexCols[indexName][tsKey] = struct{}{}
+			totalCols[tsKey+"-"+indexName] = struct{}{}
 		}
 		for col := range segMeta.ColumnNames {
 			allIndexCols[indexName][col] = struct{}{}
-			totalCols[col] = struct{}{}
+			totalCols[col+"-"+indexName] = struct{}{}
```
AndrewHess, line 404: Is `-` allowed in a column name? It's allowed in an index name, and if it's also allowed in a column name you can still have collisions (e.g., for the (col, index) pairs ("abc-foo", "ind") and ("abc", "foo-ind"))

AndrewHess (commented):
A few things:
1. I don't have enough context to understand exactly what this PR is fixing. Could you add some more info in the description?
2. If feasible, it would be good to add a unit test for what you're fixing so we don't regress later. It would also help clarify exactly what it is that you're fixing
3. nit: Your commit/PR title is really long. Usually we keep that fairly concise and add more info in the PR description as needed (or in the body of the commits is also fine)

AndrewHess (approved):

PR #2231, opened 2025-02-20
Change the default disk threshold and cleanup eviction

Macbeth98:
# Description
- Cleaned up Eviction hooks
- Changed the default disk threshold to 95%

# Checklist:


nkunal (dismissed):

AndrewHess (dismissed):
Looks good

Just FYI: there's a type `utils.WithDefault[T]` that you could use; we use it for a few config fields (I think your current changes are good since that's how we had it before, so I don't think it's worth changing to `WithDefault` unless you really want to). I've mostly been using `WithDefault` for boolean fields because parsing is a bit more annoying for those than for numbers

nkunal (approved):

PR #2244, opened 2025-02-21
K8s events index

Macbeth98:
# Description
- If the log record has `attributes.k8s.event`, then it will be set to new index `k8s_events_sig`.
- Made some changes to manage `PleArray` for multiple indexes.

# Testing
- Tested by writing two test cases:
    - Verifying the k8s events key and setting the index to `k8s_event_sig`
    - Testing the whole log ingestion by passing two k8s events log and two normal logs 

# Checklist:


pkg/otlp/logs.go
```diff
@@ -38,6 +38,9 @@ import (
 
 const defaultIndexName = "otel-logs"
 const indexNameAttributeKey = "siglensIndexName"
+const k8sKey = "k8s"
+const k8sEventsKey = "event"
+const k8sEventsIndexName = "k8s_events_sig"
```
AndrewHess, line 43: nit: change `k8s_events_sig` to use `-`, since we already have an index using hyphens (`otel-logs`), we should be consistent

pkg/otlp/logs.go
```diff
@@ -107,10 +110,39 @@ func ingestLogs(request *collogpb.ExportLogsServiceRequest, myid int64) (int, in
...
-	pleArray := make([]*segwriter.ParsedLogEvent, 0)
 	numTotalRecords := 0
 	numFailedRecords := 0
 
+	/// PLE Array Management Start ///
+	// Keeping this code block here to avoid using mutexes
+	var pleArrayMap = make(map[int][]*segwriter.ParsedLogEvent, 0)
+	var pleArrayMapIndex = 0
+
+	getPleArray := func() []*segwriter.ParsedLogEvent {
```
AndrewHess, line 121: Not sure what this comment means. Why would we need to use mutexes? I think you're saying if we moved these closures to be proper functions we'd need mutexes?

Macbeth98: I mean now I have these functions inside this `ingestLogs` function. If I have kept them out of the function block, then we should have to use Mutex as then the `pleArrayMap` and `pleArrayMapIndex` will become global vars.

But also I think we could still move these closures to proper functions, but we have to pass these variables as params. But I thought this should be fine, as this is only used for this function.

AndrewHess (approved):

pkg/otlp/logs.go
AndrewHess: Got it. This is fine

PR #2268, opened 2025-02-27
pauseMode configuration support

nrkawane:
By default it is false.
When set to false, the siglens will not start but instead simply print a message to the console/log every 2 minutes. This allows logging into pod to investigate environment and debug.

# Description
Summarize the change.

# Testing

# Checklist:


cmd/startup/startup.go
```diff
@@ -151,6 +152,15 @@ func Main() {
 	log.Infof("----- Siglens server type %s starting up.... ----- \n", nodeType.String())
 	log.Infof("----- Siglens server logging to %s ----- \n", logOut)
 
+	if config.GetRunningConfig().PauseModeConverted {
```
AndrewHess, line 155: nit: make a new function like `config.IsPauseModeEnabled()` and call that here. That's how we get most config values

pkg/config/config.go
```diff
@@ -1285,6 +1301,18 @@ func ExtractConfigData(yamlData []byte) (common.Configuration, error) {
...
+	if len(config.PauseMode) <= 0 {
+		config.PauseMode = "false"
+	}
+	pauseMode, err := strconv.ParseBool(config.PauseMode)
+	if err != nil {
+		log.Errorf("PauseMode: failed to parse PauseMode flag. Defaulting to false. Error: %v", err)
+		pauseMode = false
+		config.PauseMode = "false"
+	}
+	config.PauseModeConverted = pauseMode
```
AndrewHess, line 1314: FYI: this is valid and you don't have to change it, but personally I find it annoying to add all this for a boolean config field. There's a generic `WithDefault[T]` type that you can use to clean this up if you want to (e.g., see https://github.com/siglens/siglens/blob/develop/pkg/config/common/common.go#L58 and check how it's used)

server.yaml
```diff
@@ -49,4 +49,7 @@ queryTimeoutSecs: 300  # 5 minutes default
 
 # memoryLimits:
 #   lowMemoryMode: true  # Set to true to enable low memory mode
-#   maxUsagePercent: 80  # Percent of available RAM that siglens will occupy
\ No newline at end of file
+#   maxUsagePercent: 80  # Percent of available RAM that siglens will occupy
+
+## Pause SigLens from starting up. 
+#pauseMode: true
```
AndrewHess, line 55: Maybe renaming to `delayStartup` would be more clear? But pauseMode is also fine if we want to keep that

cmd/startup/startup.go
nrkawane: Fixed

pkg/config/config.go
nrkawane: It is indeed annoying. Lets keep it the way it is in this case - other configs follow the same pattern. In a new place I will use the generic one. Thanks

server.yaml
nrkawane: Actually we arent delaying. The SigLens never starts ... it simply pauses. Lets keep the name.

AndrewHess:
Looks good once you fix the failing config tests

AndrewHess (approved):

PR #2270, opened 2025-02-28
Fix agg over time Metric Queries

Macbeth98:
# Description
- Fixed the issue with Aggregation (count, sum, avg, min, max) Over time metric queries.
- Fixed the parser and the implementation.
- Now the results match with Prometheus.

# Testing
- Tested by taking the local copy of data directories of siglens and prometheus from the test instance that had K8s metrics ingested.
- Ran the queries and verified that the outputs are same.
- Updated the unit test cases.

# Checklist:


pkg/segment/query/metricsquery.go
```diff
@@ -165,7 +177,7 @@ func ApplyMetricsQuery(mQuery *structs.MetricsQuery, timeRange *dtu.MetricsTimeR
 	for mQuery.SubsequentAggs != nil {
 		if mQuery.SubsequentAggs.AggBlockType == structs.FunctionBlock {
 			mQuery.Function = *mQuery.SubsequentAggs.FunctionBlock
-			errors = mRes.ApplyFunctionsToResults(parallelism, mQuery.Function)
+			errors = mRes.ApplyFunctionsToResults(parallelism, mQuery.Function, timeRange)
```
AndrewHess, line 180: Should this be `finalTimeRange` too, or is `timeRange` correct?

pkg/segment/results/mresults/seriesresult_test.go
```diff
@@ -361,37 +402,38 @@ func Test_applyRangeFunctionAvg(t *testing.T) {
...
+	res, err := ApplyRangeFunction(timeSeries, structs.Function{RangeFunction: segutils.Avg_Over_Time, TimeWindow: 20, Step: 15}, timeRange)
 	assert.Nil(t, err)
 
-	assert.Len(t, res, 6)
+	/**
+	 * The result will be:
+	 Evalution starts at 1000 and evaluates at every 15 seconds with 20 seconds lookback at each evaluation
+	 => 1000: 1000-20 = 980 => 1.0
+	 => 1015: 1015-20 = 995 => (2.0 + 3.0 + 4.0 + 18.0) / 4 = 6.75
+	 => 1030: 1030-20 = 1010 => 18.0 + 2.5 / 2 = 10.25
```
AndrewHess, line 418: nit: 
```suggestion
	 => 1030: 1030-20 = 1010 => (18.0 + 2.5) / 2 = 10.25
```

AndrewHess (commented):
Looking good

Issue #2181, opened 2025-02-14
[STORY] Implement test metrics generation in sigclient for k8s


nkunal:
# Description

In https://github.com/siglens/siglens/tree/develop/tools/sigclient
add another option `-g, --generator` of `k8s`

We want fake metrics like the kube-state-metrics and node-exporter helm charts.

kube-state-metrics: add the metrics and corresponding tags documented here: https://github.com/kubernetes/kube-state-metrics/tree/main/docs/metrics 

node-exporter: add the metrics from this file:  

[node-exporter.txt](https://github.com/user-attachments/files/18802999/node-exporter.txt)

EthixLucifer:
@nkunal can i work on this issue:

The issue requested the addition of a new `-g, --generator` flag to the `sigclient` tool to generate fake metrics similar to `kube-state-metrics` and `node-exporter` Helm charts.  this is the approach i would like to inculcate fake ksm and node-exporter  metrics.

### 1. **Introducing the ****`-g, --generator`**** Flag**

- adding the `-g, --generator` flag to `root.go`, enabling users to specify `k8s` as a generator type.
- Supports additional sub-flags:
  - `--ksm`: Generates only `kube-state-metrics`.
  - `--node`: Generates only `node-exporter` metrics.
  - `--both`: Generates both metric types.

### 2. New File: **`k8s_generator.go`**

This file contains the logic to generate Kubernetes (`kube-state-metrics`) and `node-exporter` metrics.

#### **Generating ****`kube-state-metrics`**** (KSM) Metrics:**

```go
func (g *K8sGenerator) generateKSMMetric() map[string]interface{} {
    pod := g.pods[rand.Intn(len(g.pods))]
    return map[string]interface{}{
        "metric": "kube_pod_info",
        "value": 1,
        "tags": map[string]string{
            "pod": pod.name,
            "namespace": pod.namespace,
            "node": pod.node,
            "phase": pod.phase,
            "created_by_kind": "Deployment",
            "created_by_name": fmt.Sprintf("%s-deployment", g.faker.AppName()),
        },
        "timestamp": time.Now().Unix(),
    }
}
```

#### **Generating ****`node-exporter`**** Metrics:**

```go
func (g *K8sGenerator) generateNodeExporterMetric() map[string]interface{} {
    node := g.nodes[rand.Intn(len(g.nodes))]
    return map[string]interface{}{
        "metric": "node_cpu_seconds_total",
[...]

PR #2274, opened 2025-03-01
 [STORY] Implement test metrics generation in sigclient for k8s…

EthixLucifer:
**Fixes**: #2181  

---

## Testing Approach  
To ensure the new **k8s generator** works as expected, I performed the following tests:  

### **Unit Testing**  
- Verified the `generateKSMMetric` and `generateNodeExporterMetric` methods to ensure:  
  - Metrics have the correct structure and tags.  
  - Expected values are correctly generated.  

### **Integration Testing**  
- Ran the `sigclient` tool with the `-g k8s` flag and validated that:  
  - The generated metrics were successfully ingested into the destination server.  
  - Various metric generation modes functioned correctly:  
    - **Kubernetes metrics only** (`--ksm`).  
    - **Node-exporter metrics only** (`--node`).  
    - **Both Kubernetes and node-exporter metrics** (`--both`).  

### **Manual Testing**  
Executed the following commands to validate functionality:  

```bash
# Generate only Kubernetes metrics
go run main.go ingest esbulk -g k8s --ksm -d http://localhost:8081/elastic -t 100 --indexName k8sOnly

# Generate only node-exporter metrics
go run main.go ingest esbulk -g k8s --node -d http://localhost:8081/elastic -t 100 --indexName nodeexporter

# Generate both Kubernetes and node-exporter metrics
go run main.go ingest esbulk -g k8s --both -d http://localhost:8081/elastic -t 100 --indexName both

### DIFF
--- a/tools/sigclient/cmd/sigclient.go
+++ b/tools/sigclient/cmd/sigclient.go
@@ -118,8 +118,6 @@ var esBulkCmd = &cobra.Command{
 			log.Infof("minColumns : %+v\n", dataGeneratorConfig.MinColumns)
 		}
 
-		// ingest.StartIngestion(ingest.ESBulk, generatorType, dataFile, totalEvents, continuous, batchSize, dest, indexPrefix, indexName, numIndices, processCount, ts, 0, bearerToken, 0, eventsPerDay, dataGeneratorConfig)
-		// ingest.StartIngestion(ingest.ESBulk, generatorType, dataFile, totalEvents, continuous, batchSize, dest, indexPrefix, indexName, numIndices, processCount, ts, 0, bearerToken, 0, eventsPerDay, dataGeneratorConfig, false, false, true)
 		ingest.StartIngestion(ingest.ESBulk, generatorType, dataFile, totalEvents, continuous, batchSize, dest, indexPrefix, indexName, numIndices, processCount, ts, 0, bearerToken, 0, eventsPerDay, dataGeneratorConfig, ksmOnly, nodeOnly, bothMetrics)
 
 	},
@@ -166,9 +164,6 @@ var functionalTestCmd = &cobra.Command{
 		if doIngest {
 
 			dataGeneratorConfig := utils.InitFunctionalTestGeneratorDataConfig(numFixedCols, maxVariableCols)
-
-			// ingest.StartIngestion(ingest.ESBulk, "functional", "", totalEvents, false, batchSize, dest, indexPrefix,
-			// 	indexName, numIndices, processCount, true, 0, bearerToken, 0, 0, dataGeneratorConfig)
 
 			ingest.StartIngestion(ingest.ESBulk, "functional", "", totalEvents, false, batchSize, dest, indexPrefix, indexName, numIndices, processCount, true, 0, bearerToken, 0, 0, dataGeneratorConfig, false, false, true)
 
@@ -220,8 +215,7 @@ var performanceTestCmd = &cobra.Command{
 		go func(cancel context.CancelFunc) {
 			defer cancel()
 			// addTs should be false for performance testing
-			// ingest.StartIngestion(ingest.ESBulk, "performance", "", totalEvents, false, batchSize, dest, indexPrefix,
-			// 	indexName, numIndices, processCount, false, 0, bearerToken, 0, 0, dataGenConfig)
+
 			ingest.StartIngestion(ingest.ESBulk, "performance", "", totalEvents, false, batchSize, dest, indexPrefix, indexName, numIndices, processCount, false, 0, bearerToken, 0, 0, dataGenConfig, false, false, true)
 		}(cancel)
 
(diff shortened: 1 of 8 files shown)

### CODE REVIEW

File: tools/sigclient/main.go [hunk 1]
```diff
@@ -15,6 +15,8 @@
 // You should have received a copy of the GNU Affero General Public License
 // along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
+// updated package
+
 package main
```
AndrewHess, line 20: Why do you have this? Probably remove it

tools/sigclient/pkg/ingest/ingest.go
```diff
@@ -57,11 +57,9 @@ func (q IngestType) String() string {
 
 const PRINT_FREQ = 100_000
 
-// returns any errors encountered. It is the caller's responsibility to attempt retries
+// Send a request to the server and return any errors encountered.
 func sendRequest(iType IngestType, client *http.Client, lines []byte, url string, bearerToken string) error {
```
AndrewHess, line 61: nit: your revised first sentence looks good; but you should probably keep the second sentence

tools/sigclient/sigclient
AndrewHess, line 1: Remove this file

tools/sigclient/pkg/ingest/k8s_generator.go
```diff
@@ -0,0 +1,250 @@
...
+	}
+
+	// Default to alternating behavior if no flags are set
+	if g.metricCounter%2 == 0 {
+		return g.generateKSMMetric(), nil
+	}
+	return g.generateNodeExporterMetric(), nil
+}
+
+func (g *K8sGenerator) Init(...string) error { return nil }
```
AndrewHess, line 105: What's the point of this? Do you need `K8sGenerator` to implement an interface that requires this Init function?

AndrewHess (commented):
I ran this from your README
```
go run main.go ingest metrics -g k8s --both -d http://localhost:8081/otsdb -t 1000
```
but I got a bunch of `testmetric<N>` metrics ingested, instead of the ksm and node metrics 

Apart from that and some nits, this is looking pretty good so far

tools/sigclient/pkg/ingest/k8s_generator.go
EthixLucifer: No, as such Init function is not directly necessary, i added it as a no-op so if in future K8sgenerator requires some initializations before running, it could be done in the init function. let me know if you'd like me to remove it. currently it returns nill, so removing it wouldn't make much of difference,  let me know your POV.

tools/sigclient/cmd/sigclient.go
```diff
@@ -1,17 +1,17 @@
 // Copyright (c) 2021-2024 SigScalr, Inc.
-//
+
 // This file is part of SigLens Observability Solution
-//
+
```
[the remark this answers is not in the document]
EthixLucifer: done

EthixLucifer:
@AndrewHess could you review this PR

tools/sigclient/pkg/ingest/ingest.go
```diff
@@ -134,66 +142,105 @@ func generateESBody(recs int, actionLine string, rdr utils.Generator,
...
 	}
+
 	start := time.Now()
 	for i := uint64(0); i < cardinality; i++ {
 		currPayload, err := generateUniquePayload(rdr)
 		if err != nil {
-			log.Errorf("generatePredefinedSeries: failed to generate unique payload: %+v", err)
-			return err
+			log.Warnf("Skipping duplicate series after retries (total generated: %d)", i)
+			continue
```
AndrewHess, line 243: nit: keep the function name prepended to the log message. We follow this convention to make debugging a little easier

tools/sigclient/pkg/ingest/ingest.go
```diff
@@ -256,7 +298,7 @@ func runIngestion(iType IngestType, rdr utils.Generator, wg *sync.WaitGroup, url
 			bb = bytebufferpool.Get()
 		}
 		if iType == OpenTSDB {
-			payload, err = generateBodyFromPredefinedSeries(recsInBatch, uint64(len(preGeneratedSeries)))
+			payload, err = generateBodyFromPredefinedSeries(recsInBatch) // Updated call
```
AndrewHess, line 301: What is this comment for? Is this a TODO?

tools/sigclient/pkg/ingest/k8s_generator.go
```diff
@@ -77,163 +95,180 @@ func (g *K8sGenerator) GetLogLine() ([]byte, error) {
...
+		// Generate node-exporter only
+		metrics = append(metrics, g.generateNodeExporterMetric(baseTs))
+
+	default:
+		// Default alternating behavior
+		if g.metricCounter%2 == 0 {
+			metrics = append(metrics, g.generateKSMMetric(baseTs))
+		} else {
+			metrics = append(metrics, g.generateNodeExporterMetric(baseTs))
+		}
```
AndrewHess, line 122: I don't think there's a reason to do this. You can just return an error, unless I'm missing something

tools/sigclient/pkg/ingest/k8s_generator.go
```diff
@@ -77,163 +95,180 @@ func (g *K8sGenerator) GetLogLine() ([]byte, error) {
...
+	case 2:
+		return "kube_deployment_status_replicas"
+	case 3:
+		return "kube_service_info"
+	default:
+		return "kube_namespace_labels"
+	}
+}
+
+func (g *K8sGenerator) getKSMMetricValue(metricType int) interface{} {
```
AndrewHess, line 232: Doesn't this always return a `float64`? You should change the return type if so

Same comment for `getNodeExporterMetricValue()`

tools/sigclient/pkg/ingest/k8s_generator.go
AndrewHess, line 1: Since you're `switch`ing on raw ints (rather than an enum) in `generateNodeExporterMetric()` and the two helper functions, there's some coupling between those functions and it's a little harder to read since you're not using an enum.

I'd suggest one of these options:
1. Combine the two helper functions into one (so that only two functions are coupled instead of three)
2. Make an enum, and use the names instead of the raw integers

Similar comment for `generateKSMMetric()` and its two helpers

PR #2286, opened 2025-03-04
Add refactored metrics query structs

AndrewHess:
# Description
This adds some structs and interfaces for the new way we'll handle metrics queries. No integration has been done so far; that'll be in the next PR.

This adds a `timeseries` interface which acts as the unlabelled timeseries data. Several struct types implement this to create a timeseries based on existing ones, similar to what is needed for PromQL queries. The tags that are needed for PromQL are in a wrapper type `TaggedSeries`

# Testing
New unit tests

# Checklist:


pkg/metrics/timeseries.go
```diff
@@ -0,0 +1,525 @@
+// Copyright (c) 2021-2024 SigScalr, Inc.
```
nkunal, line 1: we should change the year in all new files

nkunal (dismissed):

nkunal (approved):

PR #2287, opened 2025-03-04
Fix Response format for Promql scalar queries

Macbeth98:
# Description
- When the queries are simple scalar queries (eg: 42, or (42 + 32), etc), the server is panics when queried through promql endpoint. 
- This PR fixes this issue and added support for scalar type queries response format. 
- Also fixes the bug with downsample value of interval, as the value should be based on step.

# Testing
- All unit test cases pass.
- Tested by running only the scalar queries with the compliance tool.

# Checklist:

### DIFF
--- a/pkg/segment/results/mresults/metricresults.go
+++ b/pkg/segment/results/mresults/metricresults.go
@@ -28,6 +28,7 @@ import (
 	"time"
 
 	parser "github.com/prometheus/prometheus/promql/parser"
+	"github.com/siglens/siglens/pkg/common/dtypeutils"
 	putils "github.com/siglens/siglens/pkg/integrations/prometheus/utils"
 	tsidtracker "github.com/siglens/siglens/pkg/segment/results/mresults/tsid"
 	"github.com/siglens/siglens/pkg/segment/structs"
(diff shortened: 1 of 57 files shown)


pkg/segment/results/mresults/metricresults.go
```diff
@@ -574,6 +574,39 @@ func (r *MetricsResult) GetResultsPromQl(mQuery *structs.MetricsQuery, pqlQueryt
...
 	}, nil
 }
+
+func (r *MetricsResult) GetResultsPromQlForScalarType(pqlQueryType parser.ValueType, startTime, endTime uint32, step uint32) (*structs.MetricsQueryResponsePromQl, error) {
+	if pqlQueryType != parser.ValueTypeScalar {
+		return nil, fmt.Errorf("GetResultsPromQlForScalarType: Unsupported PromQL query result type: %v", pqlQueryType)
+	}
+
+	var pqlData structs.Data
+	pqlData.ResultType = parser.ValueType("matrix")
```
AndrewHess, line 584: Shouldn't this be "scalar"?

AndrewHess (approved):

PR #2288, opened 2025-03-04
fix: promql response format for Instant queries

Macbeth98:
# Description
- Fixed the response for promql instant query API

# Testing
- Tested by verifying the results from Prometheus response and SigLens response

# Checklist:

### DIFF
--- a/pkg/segment/results/mresults/metricresults.go
+++ b/pkg/segment/results/mresults/metricresults.go
@@ -547,7 +547,7 @@ func getPromQLSeriesFormat(seriesId string) *structs.Result {
 		if idx == 0 {
 			keyValue = strings.SplitN(removeMetricNameFromGroupID(val), ":", 2)
 		} else {
-			keyValue = strings.Split(val, ":")
+			keyValue = strings.SplitN(val, ":", 2)
 		}
 
 		if len(keyValue) > 1 {
--- a/pkg/segment/results/mresults/metricresults.go
+++ b/pkg/segment/results/mresults/metricresults.go
@@ -34,6 +34,7 @@ import (
 	"github.com/siglens/siglens/pkg/segment/structs"
 	segutils "github.com/siglens/siglens/pkg/segment/utils"
 	"github.com/siglens/siglens/pkg/utils"
+	log "github.com/sirupsen/logrus"
 	"github.com/valyala/bytebufferpool"
 )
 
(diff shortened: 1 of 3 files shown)


pkg/segment/results/mresults/metricresults.go
```diff
@@ -535,6 +536,93 @@ func (r *MetricsResult) GetOTSDBResults(mQuery *structs.MetricsQuery) ([]*struct
...
+	var result structs.Result
+	var keyValue []string
+	result.Metric = make(map[string]string)
+	result.Metric["__name__"] = ExtractMetricNameFromGroupID(seriesId)
+	for idx, val := range tagValues {
+		if idx == 0 {
+			keyValue = strings.SplitN(removeMetricNameFromGroupID(val), ":", 2)
+		} else {
+			keyValue = strings.Split(val, ":")
+		}
```
AndrewHess, line 551: It looks like you moved this section of code, but also for `idx == 0` you changed `Split()` to `SplitN()`. Why do you need that change? I suspect this is a subtle issue, so it's probably worth adding a comment in the code

pkg/segment/results/mresults/metricresults.go
```diff
@@ -535,6 +536,93 @@ func (r *MetricsResult) GetOTSDBResults(mQuery *structs.MetricsQuery) ([]*struct
...
+			// TODO: Inspect on whether we should ensure that the timestamp is present in the results?
+
+			// In the case where the timestamp is not present in the results
+			if len(results) > 2 {
+				// If the results have more than 2 timestamps, this should not happen.
+				// As the startTime = timestamp - 1 and endTime = timestamp, and intervalSeconds = 1
+				// So, the results should have only 2 timestamps
+				log.Errorf("GetResultsPromQlInstantQuery: More than 2 timestamps found in the results for seriesId: %v", seriesId)
+				return nil, errors.New("error in fetching the results. Multiple timestamps found in the results")
+			}
```
AndrewHess, line 600: It seems odd to me that we'd get into this function (`GetResultsPromQlInstantQuery()`) with two timestamps. Shouldn't instant queries have just one timestamp?

pkg/segment/results/mresults/metricresults.go
```diff
@@ -535,6 +536,93 @@ func (r *MetricsResult) GetOTSDBResults(mQuery *structs.MetricsQuery) ([]*struct
...
+			if ok {
+				result.Value = []interface{}{timestamp, fmt.Sprintf("%v", floatValue)}
+				pqlData.Result = append(pqlData.Result, *result)
+				continue
+			}
+
+			// TODO: Inspect on whether we should ensure that the timestamp is present in the results?
+
+			// In the case where the timestamp is not present in the results
+			if len(results) > 2 {
```
AndrewHess, line 594: I'm unclear about what this TODO comment means. Mostly confused by "inspect on"

I'm pretty sure if you run an instant vector query for a timestamp and there's no value at that timestamp, Prometheus will choose the most recent value that came before that timestamp

Macbeth98: So currently when I similar query in Prometheus, where I know there is no data that was ingested at that time. Prometheus just gave a empty response. 

I did this here in this way, as our timestamps may not exactly match because of time bucketing.  So for our scenario the timestamp may or may not present. 

And if we want to fix this and implement how prometheus, then I assume, that we should not even reach the the code execution from this comment. But since I was not exactly sure I made this comment as a TODO.

pkg/segment/results/mresults/metricresults.go
Macbeth98: It is because of how we are doing the query. When we receive a timestamp from the UI, we do startTime as givemTimestamp-1 and endTime as givenTimestamp.  And the interval will become `1s`.

So depending on the data, there is a chance that there can two timestamps. One at startTime and other at endTime.

pkg/segment/results/mresults/metricresults.go
Macbeth98: Actually I was supposed to use the `SplitN` for both the cases.

AndrewHess (approved):

PR #2289, opened 2025-03-04
Promql range query fixes

Macbeth98:
# Description
- The PromQL results are not being sent in a sorted order. This PR fixes that.
- Made a temp Fixes for panics that occurred while executing queries through compliance tool.

# Testing
- Ran the test through compliance tool

# Checklist:

### DIFF
--- a/pkg/segment/results/mresults/metricresults.go
+++ b/pkg/segment/results/mresults/metricresults.go
@@ -547,7 +547,7 @@ func getPromQLSeriesFormat(seriesId string) *structs.Result {
 		if idx == 0 {
 			keyValue = strings.SplitN(removeMetricNameFromGroupID(val), ":", 2)
 		} else {
-			keyValue = strings.Split(val, ":")
+			keyValue = strings.SplitN(val, ":", 2)
 		}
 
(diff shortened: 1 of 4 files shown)


pkg/segment/results/mresults/metricresults.go
```diff
@@ -635,8 +635,12 @@ func (r *MetricsResult) GetResultsPromQl(mQuery *structs.MetricsQuery, pqlQueryt
 		for seriesId, results := range r.Results {
 			result := getPromQLSeriesFormat(seriesId)
 
+			result.Value = make([]interface{}, len(results))
+			index := 0
+
 			for k, v := range results {
-				result.Value = append(result.Value, []interface{}{int64(k), fmt.Sprintf("%v", v)})
+				result.Value[index] = []interface{}{int64(k), fmt.Sprintf("%v", v)}
+				index++
```
AndrewHess, line 643: FYI if you change `make([]interface{}, len(results))` to `make([]interface{}, 0, len(results))` you could keep using `append()` instead of tracking the index, and there's no performance hit. But this is also fine

AndrewHess (approved):
Looks good

PR #2290, opened 2025-03-05
Add support for PromQL without Aggregation operator

Macbeth98:
# Description
- Add support for PromQL without aggregation.

# Testing
- Tested on UI by running the below queries
   1. `sum without (instance) (demo_api_http_requests_in_progress)`
   2. `sum  without(instance) (avg_over_time(demo_api_http_requests_in_progress[3h:30m]))`
- Added few functional unit tests.
- TODO: Need to add tests for parser and other functional unit tests.

### Update
- Added Tests for parser and an e2e test.

# Checklist:

### DIFF
--- a/pkg/e2etests/metrics_e2e_test.go
+++ b/pkg/e2etests/metrics_e2e_test.go
@@ -1551,7 +1551,6 @@ func Test_SimpleMetricQueryGroupByWithout(t *testing.T) {
 
 	res := segment.ExecuteMetricsQuery(&metricQueryRequest[0].MetricsQuery, &metricQueryRequest[0].TimeRange, 0)
 	assert.NotNil(t, res)
-	fmt.Println(res)
 	assert.Equal(t, 2, len(res.Results))
 
 	for seriesId, seriesDp := range res.Results {
--- a/pkg/e2etests/metrics_e2e_test.go
+++ b/pkg/e2etests/metrics_e2e_test.go
@@ -1514,6 +1514,70 @@ func Test_SimpleMetricQuery_Regex_on_TagFilters_Plus_Filter_Plus_GroupByTag_v1(t
 	}
 }
 
+func Test_SimpleMetricQueryGroupByWithout(t *testing.T) {
+	defer cleanUp(t)
+	go query.PullQueriesToRun()
+
+	startTimestamp := dataStartTimestamp
+	allTimeSeries, _, _, _ := GetTestMetricsData(startTimestamp)
+
+	err := initTestConfig(t)
+	assert.Nil(t, err)
+
+	err = ingestTestMetricsData(allTimeSeries)
+	assert.Nil(t, err)
+
+	mSegs, err := rotateMetricsDataAndClearSegStore(true)
+	assert.Nil(t, err)
+	assert.Greater(t, len(mSegs), 0)
+
+	err = initializeMetricsMetaData()
+	assert.Nil(t, err)
+
+	timeRange := &dtypeutils.MetricsTimeRange{
+		StartEpochSec: uint32(startTimestamp),
+		EndEpochSec:   uint32(startTimestamp + 4600),
+	}
+
+	query := `sum without (shape, radius) (testmetric0)`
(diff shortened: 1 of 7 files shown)


pkg/e2etests/metrics_e2e_test.go
```diff
@@ -1514,6 +1514,70 @@ func Test_SimpleMetricQuery_Regex_on_TagFilters_Plus_Filter_Plus_GroupByTag_v1(t
...
+	assert.Equal(t, 1, len(metricQueryRequest))
+
+	expectedResults := map[string][]float64{
+		"testmetric0{color:red,size:small,type:solid": {10, 50, 60, 70},
+		"testmetric0{color:red,type:solid":            {40},
+	}
+
+	res := segment.ExecuteMetricsQuery(&metricQueryRequest[0].MetricsQuery, &metricQueryRequest[0].TimeRange, 0)
+	assert.NotNil(t, res)
+	fmt.Println(res)
```
AndrewHess, line 1554: Delete this print

pkg/segment/query/metricsquery.go
```diff
@@ -120,9 +120,18 @@ func ApplyMetricsQuery(mQuery *structs.MetricsQuery, timeRange *dtu.MetricsTimeR
...
 		for _, v := range mQuery.TagsFilters {
 			delete(allTagKeys, v.TagKey)
+			if v.IgnoreTag && !v.NotInitialGroup {
+				continue
+			}
+
+			filteredTags = append(filteredTags, v)
 		}
+
+		mQuery.TagsFilters = filteredTags
```
AndrewHess, line 133: nit: instead of deleting all tags and adding some back, just delete some tags

pkg/segment/structs/metricsstructs.go
```diff
@@ -84,7 +84,9 @@ type MetricsQuery struct {
 type Aggregation struct {
 	AggregatorFunction utils.AggregateFunctions //aggregator function
 	FuncConstant       float64
-	GroupByFields      []string // group by fields will be sorted
+	GroupByFields      []string            // group by fields will be sorted
+	Without            bool                // if set exclude the above group by fields
+	WithoutFieldsSet   map[string]struct{} // set of fields to exclude; only used when Without is set
```
AndrewHess, line 89: These comments don't quite make sense. I think they work like this: if `Without` is set, then group by all fields except those in `WithoutFieldsSet`; otherwise, group by all fields in `GroupByFields`. Is that right?

pkg/segment/query/metricsquery.go
Macbeth98: oh actually there is a mistake.. this line `filteredTags := make([]*structs.TagsFilter, 0, len(allTagKeys))`
should be: `filteredTags := make([]*structs.TagsFilter, 0, len(mQuery.TagFilters))`.

The `allTagKeys` is different. It is all the Tagkeys that are in the segment. 

And here I am filtering Tag Keys.

pkg/segment/structs/metricsstructs.go
Macbeth98: Hmm I will update the comments. But what you said is `true`.. and `WithoutFieldsSet` is basically `GroupByFields` in set format. I made this set here.. as this is easier to process later.

pkg/segment/structs/metricsstructs.go
```diff
@@ -88,7 +88,7 @@ type Aggregation struct {
 	FuncConstant       float64
 	GroupByFields      []string            // group by fields will be sorted
 	Without            bool                // if set exclude the above group by fields
-	WithoutFieldsSet   map[string]struct{} // set of fields to exclude; only used when Without is set
+	GroupByFieldsSet   map[string]struct{} // same as GroupByFields, but it is a map; only used or set when Without is set
```
AndrewHess, line 91: nit: I'd recommend not adding `GroupByFieldsSet` as a field here since it duplicates the data in `GroupByFields`. IMO the cost of the potential to create a bug either because they get out of sync or someone forgets to check `Without` first is higher than the benefit.

If you do want to use them as a set, I'd say just create that set on the fly when you need it. There shouldn't be more than a handful of fields we're gouping on in general, so performance is negligible. Also it looks like you're only using this in one spot, so you're not saving performance anyway by saving the result as a field here

AndrewHess (approved):

PR #2291, opened 2025-03-05
Jaeger grafana plugin api

VasudevRaut:
# Description
In this PR, Jaeger-Grafana-Plugin Introduces the following Services API endpoints:
Dependencies API
/api/dependencies?endTs={EpochMS}&lookback={EpochMS}
Traces API
GET /jaeger/api/traces?service=customer&start=1741114973191000&end=1741118573191000&lookback=custom

# Testing
`curl --location 'http://localhost:5122/jaeger/api/dependencies?endTs=1741197633019&lookback=604800000' \
--header 'Content-Type: application/json''`

`curl --location 'http://localhost:5122/jaeger/api/traces?service=frontend&start=1741191945220000&end=1741195545220000&lookback=custom' \
--header 'Content-Type: application/json'`

# Checklist:


pkg/segment/tracing/handler/jaegerHandler.go
```diff
@@ -293,8 +302,11 @@ func ProcessGetTracesSearch(ctx *fasthttp.RequestCtx, myid int64) {
 		return
 	}
 
+	startEpoch = convertEpochToMilliseconds(startEpoch)
+	endEpoch = convertEpochToMilliseconds(endEpoch)
+
 	searchRequestBody := structs.SearchRequestBody{
-		SearchText: "*",
+		SearchText: "service=" + service,
```
nkunal, line 309: if `service` var is nil, we should not do this and keep it as `*`

pkg/segment/tracing/handler/jaegerHandler.go
```diff
@@ -320,39 +332,50 @@ func ProcessGetTracesSearch(ctx *fasthttp.RequestCtx, myid int64) {
...
+			Spans:   spans,
+		}
 
+		processes := map[string]Process{
+			"p1": {ServiceName: "frontend"},
+			"p2": {ServiceName: "route"},
+			"p3": {ServiceName: "redis-manual"},
+			"p4": {ServiceName: "driver"},
+			"p5": {ServiceName: "customer"},
+			"p6": {ServiceName: "mysql"},
```
nkunal, line 367: remove this hardcoding, for now you can create an empty map and assign it

pkg/segment/tracing/handler/jaegerHandler.go
```diff
@@ -141,3 +179,293 @@ func ProcessGetOperations(ctx *fasthttp.RequestCtx, myid int64) {
...
+	spanEntry := Span{
+		TraceID:       traceID,
+		SpanID:        span.SpanID,
+		OperationName: span.OperationName,
+		References:    references,
+		StartTime:     int64(span.ActualStartTime),
+		Duration:      int64(span.Duration),
+		Tags:          spanTags,
+		Logs:          []string{},
+		ProcessID:     "p1",
```
nkunal, line 416: remove this harcoding

pkg/ast/pipesearch/searchHandler.go
```diff
@@ -144,7 +149,11 @@ func ParseSearchBody(jsonSource map[string]interface{}, nowTs uint64) (string, u
 		case uint64:
 			endEpoch = uint64(val)
 		case string:
-			endEpoch = utils.ParseAlphaNumTime(nowTs, string(val), nowTs)
+			var err error
+			endEpoch, err = strconv.ParseUint(val, 10, 64)
+			if err != nil {
+				endEpoch = utils.ParseAlphaNumTime(nowTs, string(val), nowTs)
+			}
```
AndrewHess, line 156: Why do you need this change? I might be missing something, but it seems like `endEpoch` would always be the same with and without this change. If this is actually a behavior change in some cases, then maybe that fix belongs inside of `ParseAlphaNumTime`?

pkg/segment/tracing/handler/jaegerHandler.go
```diff
@@ -141,3 +179,305 @@ func ProcessGetOperations(ctx *fasthttp.RequestCtx, myid int64) {
...
+func convertEpochToMilliseconds(epoch int64) int64 {
+	switch {
+	case epoch >= 1e9 && epoch < 1e10:
+		return epoch * 1000
+	case epoch >= 1e12 && epoch < 1e13:
+		return epoch
+	case epoch >= 1e15 && epoch < 1e16:
+		return epoch / 1000
+	default:
+		return 0
```
AndrewHess, line 481: nit: we should probably log and/or return an error in the `default`

pkg/segment/tracing/handler/jaegerHandler.go
```diff
@@ -2,24 +2,62 @@ package handler
...
 	"github.com/valyala/fasthttp"
 )
 
 type ResponseBody struct {
-	Data   []string `json:"data"`
-	Total  int      `json:"total"`
-	Limit  int      `json:"limit"`
-	Offset int      `json:"offset"`
-	Errors []string `json:"errors"`
+	Data   interface{} `json:"data"`
```
AndrewHess, line 17: We should avoid using `interface{}` when possible; it makes understanding the code a little harder.

I think you did this so we can have the ResponseBody return responses with two different types of data, but the type of data depends on the endpoint being called. If that's the case one options is to:
1. Make two new types that just have a Data field with the appropriate type
2. Remove `Data` from `ResponseBody` (and maybe rename it)
3. Make two more types that you'll actually use as the responses. Have these both embed the updated `ResponseBody` struct and the appropriate struct you created in step 1. Note: by "embed" I mean something specific that might just be in Go; look it up if you're not familiar with the term

pkg/segment/tracing/handler/jaegerHandler.go
```diff
@@ -2,24 +2,62 @@ package handler
...
+	SpanID  string `json:"spanID"`
+}
+
+type Tag struct {
+	Key   string      `json:"key"`
+	Type  string      `json:"type"`
+	Value interface{} `json:"value"`
+}
+
+const NoDependencyGraphsMessage = "no dependencies graphs have been generated"
```
AndrewHess, line 59: Making a constant for this is good. I get the feeling this value is coupled to something else (e.g., some error value in our codebase, or some message from jaeger); if so, it would be good to add a comment indicating that

Actually if it's an error within our codebase, you should have that define once somewhere, and use that constant in all places it's needed

pkg/segment/tracing/handler/jaegerHandler.go
```diff
@@ -141,3 +179,305 @@ func ProcessGetOperations(ctx *fasthttp.RequestCtx, myid int64) {
...
+		return
+	}
+
+	var allTraceData []TraceData
+	serviceProcesses := make(map[string]string)
+	for _, traces := range pipeSearchResponseOuter.Traces {
+		tracesId := traces.TraceId
+		searchGanttChartRequestBody := structs.SearchRequestBody{
+			SearchText: "trace_id=" + tracesId,
+			StartEpoch: "now-365d",
```
AndrewHess, line 343: Should this really be 1 year? I thought we recently changed this on the endpoint for our UI to be much shorter like 1 hour or 1 day.

pkg/segment/tracing/handler/jaegerHandler.go
```diff
@@ -141,3 +179,305 @@ func ProcessGetOperations(ctx *fasthttp.RequestCtx, myid int64) {
...
+		spanTags = append(spanTags, Tag{
+			Key:   k,
+			Type:  fmt.Sprintf("%T", v),
+			Value: v,
+		})
+	}
+
+	var references []Reference
+	references = append(references, Reference{
+		RefType: "CHILD_OF",
```
AndrewHess, line 405: FYI: generally, I think hardcoding is ok if that value is used just once, but if it's used more you should make a constant. I believe here's the only use of CHILD_OF, so you don't need to change this

AndrewHess (commented):
Looking pretty good. You should also add a few unit tests; at least for `computeStartTime()` which should be pretty easy

pkg/ast/pipesearch/searchHandler.go
VasudevRaut: If endEpoch comes in the format "1700000000", ParseAlphaNumTime does not handle it because it expects inputs like "now-10m" or "now-2h", not plain numeric epoch timestamps."

pkg/utils/errorutils.go
```diff
@@ -33,6 +33,9 @@ var qidToBatchErrorMapLock *sync.RWMutex
 const NIL_VALUE_ERR = "NIL_VALUE_ERR"
 const CONVERSION_ERR = "CONVERSION_ERR"
 
+// Error shown when no dependency graphs are available for a particular time frame
+const ErrNoDependencyGraphs = "no dependencies graphs have been generated"
```
AndrewHess, line 37: You should also use this here: https://github.com/siglens/siglens/blob/develop/pkg/segment/tracing/handler/tracehandler.go#L716

pkg/segment/tracing/handler/jaegerHandler_test.go
AndrewHess, line 1: TestComputeStartTime looks good. TestProcessGetDependencies and TestProcessGetTracesSearch aren't testing too much since you aren't ingesting any data in the test, so you don't really need them; but it's fine to keep them

AndrewHess (approved):

pkg/utils/errorutils.go
AndrewHess: This can be done in a separate PR @VasudevRaut

PR #2305, opened 2025-03-07
dont accumulate all tsids when only cardinality was asked, removes un…

nkunal:
…wanted map inserts

# Description
Summarize the change.

# Testing

# Checklist:

### DIFF
--- a/pkg/segment/reader/metrics/tagstree/tagstreereader.go
+++ b/pkg/segment/reader/metrics/tagstree/tagstreereader.go
@@ -306,7 +306,7 @@ func (attr *AllTagTreeReaders) FindTSIDS(mQuery *structs.MetricsQuery) (*tsidtra
 				return nil, err
 			}
 		} else {
-			_, _, rawTagValueToTSIDs, _, err := attr.GetMatchingTSIDs(mQuery.HashedMName, tf.TagKey, tf.HashTagValue, tf.TagOperator)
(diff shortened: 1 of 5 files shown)


pkg/segment/reader/metrics/tagstree/tagstreereader.go
```diff
@@ -430,7 +434,11 @@ func (ttr *TagTreeReader) GetMatchingTSIDs(mName uint64, tagValue uint64, tagOpe
 
 				for i := uint32(0); i < tsidCount; i++ {
 					tsid := utils.BytesToUint64LittleEndian(tagTreeBuf[treeOffset : treeOffset+8])
-					rawTagValueToTSIDs[valueAsStr][tsid] = struct{}{}
+					if tsidCard != nil {
+						tsidCard.AddRaw(tsid)
+					} else {
+						rawTagValueToTSIDs[valueAsStr][tsid] = struct{}{}
+					}
```
AndrewHess, line 441: IMO this is pretty unexpected behavior from the function signature/description. And this ~100 lines into a long function. I'd suggest one of these changes:
1. Update the comment for this function 
2. Also return the tsidCard, to make it more obvious that it's getting updated
3. Don't have this change at all, and have the caller iterater through the returned TSIDs and add them to the HLL if they want. Presumably you're doing it this way for performance, so maybe we don't want to do that

AndrewHess (approved):

PR #2307, opened 2025-03-07
Fix issue not adding some tags to tags tree

AndrewHess:
# Description
When adding tags to the tags tree for a certain TSID, sometimes we erroneously skipped adding the tags because we got a false positive from a bloom filter, so we thought the TSID was already added.

This just removes the bloom filter check, as `tth.addTags()` should already correctly handle a call for tags that are already there.

# Testing

# Checklist:

### DIFF
--- a/pkg/segment/writer/metrics/tagstree.go
+++ b/pkg/segment/writer/metrics/tagstree.go
@@ -115,8 +115,6 @@ Returns a bool indicating if this tsid is new
 Internally, will use the internal bloom to check if the tsid has already been added or not
 */
 func (tth *TagsTreeHolder) AddTagsForTSID(mName []byte, tags *TagsHolder, tsid uint64) error {
-	rawTSID := utils.Uint64ToBytesLittleEndian(tsid)
-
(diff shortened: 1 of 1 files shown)


pkg/segment/writer/metrics/tagstree.go
```diff
@@ -116,18 +116,18 @@ Internally, will use the internal bloom to check if the tsid has already been ad
...
+	tth.rwLock.Lock()
+	defer tth.rwLock.Unlock()
+
+	err := tth.addTags(mName, tags, tsid)
+	if err != nil {
+		log.Errorf("TagsTreeHolder.AddTagsForTSID: failed to add tags to tree. mName: %v, tsid: %v, tags holder: %+v; err=%v", mName, tsid, tags, err)
+		return err
 	}
+
+	tth.tagBloom.Add(rawTSID)
```
nkunal, line 129: if we are not using bloom then we can remove adding it too

nkunal (approved):

PR #2335, opened 2025-03-11
Add timeout support for metric queries

Macbeth98:
# Description
- Adds the timeout support for metrics query.
- Also ensures that the execution is stopped when a query is cancelled.

# Testing
- Tested by running queries with 30 seconds timeout and added a sleep of 1 minute in the query execution flow and verified that the queries are getting cancelled and execution is being stopped.

# NOTE:
- Although there can be a delay before the query execution is fully halted and summary is printed. For example, in the above test, even though the server tried to cancel the query at 30 second timeout, but since the query execution already in sleep for 1 minute, it will get cancelled fully after this 1 minute is done.
- You can think of this 1 minute sleep is inside some sort of for loop or some other execution. There are checks added at some places of the main query execution flow,  that will only proceed to the next step given that the query is not cancelled.
- After this 1 minute of sleep or some execution of function or `for` loop, the flow will check if query is cancelled. If it is then, it will return and will not proceed further.

# Checklist:


pkg/segment/query/metricsquery.go
```diff
@@ -162,6 +162,12 @@ func ApplyMetricsQuery(mQuery *structs.MetricsQuery, timeRange *dtu.MetricsTimeR
 	if mQuery.ExitAfterTagsSearch {
 		return mRes
 	}
+
+	if mQuery.IsQueryCancelled() {
+		mRes.AddError(fmt.Errorf("query cancelled"))
+		return mRes
+	}
```
AndrewHess, line 169: I was thinking for logs we had a better way of handling cancelled queries so we don't have to insert a check like this between each function call, but I'm not remembering exactly.

AndrewHess (dismissed):

pkg/segment/query/metricsquery.go
Macbeth98: So for logs, we setup a cleanup callback, which is `QueryProcessor.Cleanup()` and it will be called when a query is deleted. 

And  `QueryProcessor.Cleanup()` calls `dataProcessor.Cleanup()`, which for each dataProcessor sets the flag is cancelled to true. 

And in the main `dataProcessor.Fetch()` loop, we first check if the query is already in cleanup state or not.
You can check this code block: https://github.com/siglens/siglens/blob/develop/pkg/segment/query/processor/dataprocessor.go#L142-L163

But since metrics query execution is not structured in this format, I kept this check at few places, so as to ensure that the query execution is halted after each stage.

I could remove the two extra checks that I have added at `FindTsids` function or in the `applyMetricsOperatorOnSegments`. I added at those places also, as I thought these are more expensive functions.
And keep these checks only in `ApplyMetricsQuery` function.

AndrewHess (approved):

pkg/segment/query/metricsquery.go
AndrewHess: Ok, I think keeping them is good

PR #2336, opened 2025-03-12
Fixed interval calculation for log histograms to properly limit bucket size

VasudevRaut:
# Description
Fixed interval calculation for log histograms to properly limit bucket size

# Checklist:


AndrewHess (commented):
Looks good. You should also add a test in splParer_test.go with a query that would have been incorrectly parsed without this fix, but now gets correctly parsed. That way we won't regress later on

AndrewHess (approved):

PR #2366, opened 2025-03-16
usage stats to count active series counts, unrotated tagtrees

nkunal:
# Description
1. Use usageStats to count activeSeriesCounts
2. Some simplification of code
3. Add unrotated versions for each of the rotated version calls that were there for the TagsTreeReaders

# Testing

# Checklist:


pkg/integrations/prometheus/promql/metricsSearchHandler.go
```diff
@@ -1018,18 +1018,29 @@ func ProcessGetTagKeysWithMostSeriesRequest(ctx *fasthttp.RequestCtx, myid int64
...
+				log.Errorf("ProcessGetTagKeysWithMostSeriesRequest: Failed to get tsids for key %v", tagKey)
 				// continue, since we would want to still the rest of tagkeys
 				continue
 			}
 		}
+	}
 
+	// do unrotated
+	err = metrics.CountUnrotatedTSIDsForTagKeys(timeRange, myid, seriesCardMap)
+	if err != nil {
```
AndrewHess, line 1037: IMO, external packages like `promql` shouldn't need to know that tags trees have a concept of rotated vs unrotated. So I think a cleaner design would be to have the tags tree package just have a function like `CountTSIDsForTagKeys()`, and internally it can call helper functions for rotated and unrotated TSIDs as needed

(btw I'm assuming that doing this wouldn't have any performance implications; if it does, then maybe it does make sense to keep rotated vs unrotated public)

pkg/integrations/prometheus/promql/metricsSearchHandler.go
```diff
@@ -1107,23 +1119,45 @@ func ProcessGetTagPairsWithMostSeriesRequest(ctx *fasthttp.RequestCtx, myid int6
...
-					Key:       key,
-					Value:     value,
-					NumSeries: tsidCount,
-				}
-				seriesCounts = append(seriesCounts, keyAndCount)
+	// do unrotated
+	err = metrics.CountUnrotatedTSIDsForTagPairs(timeRange, myid, tagPairsCardMap)
+	if err != nil {
+		log.Errorf("ProcessGetTagPairsWithMostSeriesRequest: Failed to count tsids for unrotated")
+		// continue, since we would want to still the rest
```
AndrewHess, line 1148: Should this be a warning then? I'm not sure what our distinction is between warning vs error logs

pkg/segment/query/metricsquery.go
```diff
@@ -101,6 +102,7 @@ func ApplyMetricsQuery(mQuery *structs.MetricsQuery, timeRange *dtu.MetricsTimeR
 		finalTimeRange.StartEpochSec = timeRange.StartEpochSec - uint32(mQuery.LookBackToInclude)
 	}
 
+	// todo: switch this to use rotated segs only
 	mSegments, err := getAllRequestsWithinTimeRange(finalTimeRange, mQuery.OrgId, querySummary)
```
AndrewHess, line 106: Why? Don't we want `ApplyMetricsQuery()` to run on rotated and unrotated?

pkg/segment/query/metricsquery.go
```diff
@@ -461,29 +478,53 @@ func getRegexMatchedMetricNames(mSegSearchReq *structs.MetricsSearchRequest, reg
...
+		log.Errorf("GetSeriesCardinalityOverTimeRange: invalid time range, diff: %+v, start: %v, end: %v",
+			diff, timeRange.EndEpochSec, timeRange.StartEpochSec)
+		return 0, errors.New("Invalid time range")
+	}
+
+	pastXhours := max(1, diff/3600)
+	asCounts, err := usageStats.GetActiveSeriesCounts(pastXhours, myid)
 	if err != nil {
-		log.Errorf("GetSeriesCardinalityOverTimeRange: failed to get tags trees within time range %+v; err=%v", timeRange, err)
+		log.Errorf("GetSeriesCardinalityOverTimeRange: failed to get card, err: %v", err)
```
AndrewHess, line 491: nit: IMO in logs a little more verbosity is worth a little more clarity
```suggestion
		log.Errorf("GetSeriesCardinalityOverTimeRange: failed to get cardinality, err: %v", err)
```

pkg/segment/query/metricsquery.go
```diff
@@ -461,29 +478,53 @@ func getRegexMatchedMetricNames(mSegSearchReq *structs.MetricsSearchRequest, reg
...
-	for _, segmentTagTreeReader := range tagsTreeReaders {
-		for tagKey := range tagKeys {
-			_, err := segmentTagTreeReader.GetTSIDsForKey(tagKey, tsidCard)
-			if err != nil {
-				log.Errorf("GetSeriesCardinalityOverTimeRange: failed to get tsids for key %v; err=%v", tagKey, err)
-				return 0, err
-			}
+	slices.Sort(asCounts)
+	n := int(round(float64(k) * 0.95))
+	p95 := asCounts[n]
```
AndrewHess, line 503: When `float64(k) * 0.95)` isn't an integer, don't you want to linearly interpolate? We already have a function to find percentiles (with interpolation) and I think it has better time complexity than what you're doing: https://github.com/siglens/siglens/blob/develop/pkg/segment/tracing/utils/lineartimefinding.go#L28

pkg/segment/reader/metrics/tagstree/tagstreereader.go
```diff
@@ -498,99 +498,87 @@ func (attr *AllTagTreeReaders) GetAllTagPairs() (map[string]map[string]struct{},
...
 		}
 	}
 
-	return allTSIDs, nil
+	return nil
 }
 
 func (attr *AllTagTreeReaders) GetTSIDCountForTagPair(tagKey string,
-	tagValue string) (uint64, error) {
+	tagValue string, tsidCard *utils.GobbableHll) error {
```
AndrewHess, line 541: Having a function called `GetTSIDCountForTagPair` that doesn't return a number seems really weird to me. If you're inserting these into the HLL, maybe a better name would be `InsertTSIDsForTagPair`

pkg/segment/writer/metrics/metricssegment.go
```diff
@@ -1518,6 +1519,131 @@ func GetTagsTreeHolder(orgid int64, mid string) *TagsTreeHolder {
...
+		log.Errorf("CountUnrotatedTSIDsForTagKeys: failed to get unrotated metric segments for time range=%v, myid=%v, err=%v", tRange, myid, err)
+		return err
+	}
+
+	for _, segment := range unrotatedMetricSegments {
+		tagsTreeHolder := GetTagsTreeHolder(myid, segment.Mid)
+		if tagsTreeHolder != nil {
+			for tagkey, tkTree := range tagsTreeHolder.allTrees {
+				tsidCard, ok := seriesCardMap[tagkey]
+				if !ok {
```
AndrewHess, line 1539:
```suggestion
				if !ok  || tsidCard == nil {
```

pkg/segment/writer/metrics/metricssegment.go
```diff
@@ -1518,6 +1519,131 @@ func GetTagsTreeHolder(orgid int64, mid string) *TagsTreeHolder {
...
+	}
+
+	for _, segment := range unrotatedMetricSegments {
+		tagsTreeHolder := GetTagsTreeHolder(myid, segment.Mid)
+		if tagsTreeHolder == nil {
+			continue
+		}
+		for tagkey, tkTree := range tagsTreeHolder.allTrees {
+			tvaluesMap, ok := tagPairsMap[tagkey]
+			if !ok {
```
AndrewHess, line 1591: Also check if `tvaluesMap` is nil

nit: maybe rename `tvaluesMap` to just `values` or `valuesSet`

same comments on line 1631

pkg/segment/writer/metrics/metricssegment.go
```diff
@@ -1518,6 +1519,131 @@ func GetTagsTreeHolder(orgid int64, mid string) *TagsTreeHolder {
...
+				tagPairsMap[tagkey] = tvaluesMap
+			}
+
+			for _, allTi := range tkTree.rawValues {
+				for _, ti := range allTi {
+					tv := string(ti.tagValue)
+					_, ok := tvaluesMap[tv]
+					if !ok {
+						tvaluesMap[tv] = struct{}{}
+					}
```
AndrewHess, line 1602: nit: why not just
```suggestion
					tvaluesMap[tv] = struct{}{}
```

same comment on line 1641

pkg/segment/writer/metrics/metricssegment.go
AndrewHess, line 1: Add unit tests for these new functions would be good if it's not too much hassle

pkg/segment/writer/metrics/tagstree.go
```diff
@@ -520,3 +521,35 @@ func (tree *TagTree) encodeTagsTree() ([]byte, error) {
...
+	defer tt.rwLock.RUnlock()
+
+	for _, allTi := range tt.rawValues {
+		for _, ti := range allTi {
+			for _, tsid := range ti.matchingtsids {
+				tv := string(ti.tagValue)
+				tsidCard, ok := tvaluesMap[tv]
+				if !ok {
+					tsidCard = structs.CreateNewHll()
+					tvaluesMap[tv] = tsidCard
```
AndrewHess, line 549: Check for nil `tsidCard`

pkg/usageStats/stats.go
```diff
@@ -485,18 +485,62 @@ func UpdateQueryStats(queryCount uint64, respTime float64, orgid int64) {
...
+	statsFnames := getBaseStatsDirs(startEpoch, endEpoch, orgid)
+
+	for _, statsFile := range statsFnames {
+		filename := getStatsFilename(statsFile)
+		fd, err := os.OpenFile(filename, os.O_RDONLY, 0666)
+		if err != nil {
+			if !os.IsNotExist(err) {
+				log.Errorf("GetUsageStats: error opening stats file = %v, err= %v", filename, err)
+			}
+			continue
```
AndrewHess, line 501: Why don't we want to log anything if the file doesn't exist? Would we want at least a warning log?

EDIT: I see this code got moved, not created. So if you don't want to add extra logging, I think that's ok

pkg/segment/query/metricsquery.go
nkunal: ah didn't know we had a pkg for that

AndrewHess (approved):

PR #2372, opened 2025-03-18
Fix metric tag filter query

Macbeth98:
# Description
- Fixes the queries with Tag Filters.
- Before the response for a query in this format: `metric{tk1:tv1, ...tkn:tv1}`, would only return series displaying only the tags given in the query. 
- But with this PR, it will return all the labels and series.

# Testing
- All unit test cases pass
- Updated metrics cicd queries

# Checklist:

### DIFF
--- a/pkg/integrations/prometheus/promql/parser.go
+++ b/pkg/integrations/prometheus/promql/parser.go
@@ -344,6 +344,10 @@ func handleAggregateExpr(expr *parser.AggregateExpr, mQuery *structs.MetricsQuer
 
 	if len(expr.Grouping) > 0 {
 		mQuery.Groupby = true
+	} else {
+		mQuery.AggWithoutGroupBy = true
+		mQuery.SelectAllSeries = false
+		mQuery.GetAllLabels = false
 	}
 
 	mQuery.FirstAggregator.GroupByFields = sort.StringSlice(expr.Grouping)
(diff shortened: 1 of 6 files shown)


pkg/integrations/prometheus/promql/parser.go
```diff
@@ -734,8 +738,23 @@ func handleVectorSelector(mQueryReqs []*structs.MetricsQueryRequest, intervalSec
...
+				if tag.IsGroupByKey && !tag.NotInitialGroup {
+					mQuery.SelectAllSeries = false
+					break
+				}
+			}
+		}
+
+		if mQuery.SelectAllSeries && !mQuery.Groupby && !mQuery.AggWithoutGroupBy {
+			mQuery.GetAllLabels = true
+		}
```
AndrewHess, line 757: I'm quite confused by this. I'm not sure what you're trying to do; I'm confused for which promql queries `SelectAllSeries` should be true and for which it should be false

Also for setting `GetAllLabels`, I'm confused by this: `!mQuery.Groupby && !mQuery.AggWithoutGroupBy`. The query doesn't have a groupby but it also doesn't have a aggregation without a group by? That would imply it doesn't have an aggregation; is that what you're trying to say?

pkg/segment/structs/metricsstructs.go
```diff
@@ -78,6 +78,7 @@ type MetricsQuery struct {
 	GetAllLabels        bool // If set, the query should return all series for the matched metrics
 	Groupby             bool // flag to group by tags
 	GroupByMetricName   bool // flag to group by metric name
+	AggWithoutGroupBy   bool // flag to aggregate without group by
```
AndrewHess, line 81: It seems a bit odd to me to have this as a top-level field. Would it make sense to move it into the `Aggregation` and/or `MetricQueryAgg` types? If you keep it here, I think it will be confusing if this is ever true but some aggregation (either FirstAggregator or one in SubsequentAggs has a non-empty GroupByColumns)

pkg/integrations/prometheus/promql/parser.go
Macbeth98: `AggWithoutGrroupBy` flag represents, that there is an aggregation but is with group by. Example: `avg(metric{tk:tv})`

The first `if` statement says that if it is a group by metric name or an aggregation without group by and `len(tagFilters) > 0`, then we do not need to `selectAllSeries`, because we only need to get the series filtered by the given tags. But in the other case where there are no tag filters then SelectAllSeries should be true, otherwise we won't get all the series.

And regarding this `if` block

```


		if mQuery.SelectAllSeries && !mQuery.Groupby && !mQuery.AggWithoutGroupBy {
			mQuery.GetAllLabels = true
		}
```

We should get All Labels, for queries like this: `metric({tk1:tv1, ...})`, but when there is a group by or it is an aggregation without a group by, then there is no need to get all labels. So this condition handles that.

AndrewHess (approved):

PR #2379, opened 2025-03-19
Do not thrown err when tags tree file does not exist

Macbeth98:
# Description
- Checks if the error is due to file not found, if it is continue processing next tags, otherwise return error
- Reduced error log flooding

# Testing
- Tested by ingesting some metrics data for sometime.
- Ingest different types of metrics. Run the query and make sure new metrics can be queried.
- Kill the server and Ingest metrics again. And query to ensure there are no errors

# Checklist:

### DIFF
--- a/pkg/segment/reader/metrics/tagstree/tagstreereader.go
+++ b/pkg/segment/reader/metrics/tagstree/tagstreereader.go
@@ -102,6 +102,9 @@ func InitAllTagsTreeReader(tagsTreeBaseDir string) (*AllTagTreeReaders, error) {
 		// This also inserts the tagTreeReader into the tagTrees map.
 		_, err = attr.initTagsTreeReader(tagKey)
 		if err != nil {
+			if utils.IsNotExistError(err) {
+				continue
+			}
 			err = fmt.Errorf("InitAllTagsTreeReader: failed to initialize tag tree reader for tag key %s in base dir %v; err=%v", tagKey, tagsTreeBaseDir, err)
 			log.Errorf(err.Error())
 			return nil, err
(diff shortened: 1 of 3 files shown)


pkg/segment/reader/metrics/tagstree/tagstreereader.go
```diff
@@ -116,8 +119,7 @@ func (attr *AllTagTreeReaders) initTagsTreeReader(tagKey string) (*TagTreeReader
 
 	fd, err := os.OpenFile(fName, os.O_RDONLY, 0644)
 	if err != nil {
-		log.Errorf("initTagsTreeReader: failed to open file %s. Error: %v.", fName, err)
-		return nil, err
+		return nil, utils.NewErrorWithCode(os.ErrNotExist.Error(), fmt.Errorf("initTagsTreeReader: failed to open file %s", fName))
```
nkunal, line 122: should we check if the error is due file not exists and return that, instead of always returning this ?

Macbeth98: Ahh yes, we should.

pkg/utils/errorutils.go
```diff
@@ -225,3 +226,11 @@ func IsRPCUnavailableError(err error) bool {
 
 	return false
 }
+
+func IsNotExistError(err error) bool {
+	if ewc, ok := err.(*ErrorWithCode); ok {
+		return ewc.code == fs.ErrNotExist.Error()
+	}
+
+	return false
```
AndrewHess, line 235: On line 234, should you check `os.IsNotExist()` in case someone calls this on an error that isn't an `ErrorWithCode`? (or `err == s.ErrNotExist.Error()`? Not sure if that's different than `os.IsNotExist()`)

PR #2381, opened 2025-03-19
Added metrics WAL functionality added unit test for the WAL

VasudevRaut:
# Description
Added metrics WAL functionality added unit test for the WAL

# Testing

# Checklist:

### DIFF
--- a/pkg/segment/writer/metrics/wal/wal.go
+++ b/pkg/segment/writer/metrics/wal/wal.go
@@ -1,0 +1,199 @@
+package wal
+
+import (
+	"bytes"
+	"encoding/binary"
+	"errors"
+	"io"
+	"os"
+	"path/filepath"
+	"strings"
+
+	"github.com/siglens/siglens/pkg/config"
+	log "github.com/sirupsen/logrus"
+)
+
+type WALDatapoint struct {
+	Timestamp uint64
+	DpVal     float64
+	Tsid      uint64
+}
+
+type WAL struct {
+	file *os.File
+}
+
+func getBaseWalDir() (string, error) {
+	var sb strings.Builder
+	sb.WriteString(config.GetDataPath())
+	sb.WriteString(config.GetHostID())
+	sb.WriteString("/wal-ts/")
+	dirPath := sb.String()
+	err := os.MkdirAll(dirPath, 0755)
+	return dirPath, err
+}
+
+func NewWAL(filename string) (*WAL, error) {
+	dirPath, err := getBaseWalDir()
+	if err != nil {
+		log.Errorf("NewWAL : Failed to get base WAL directory: %v", err)
+		return nil, err
+	}
+
+	filePath := filepath.Join(dirPath, filename+".wal")
+	f, err := os.OpenFile(filePath, os.O_CREATE|os.O_WRONLY|os.O_APPEND, 0644)
+	if err != nil {
+		log.Errorf("NewWAL : Failed to open WAL file at path %s: %v", filePath, err)
+		return nil, err
+	}
+
+	return &WAL{file: f}, nil
+}
+
+func (w *WAL) AppendToWAL(dps []WALDatapoint) error {
+	blockBuf := &bytes.Buffer{}
+	err := dataCompression(dps, blockBuf)
+	if err != nil {
+		log.Errorf("AppendToWAL: dataCompression failed: %v", err)
+		return err
+	}
+
+	blockSize := uint32(blockBuf.Len())
+	if err := binary.Write(w.file, binary.LittleEndian, blockSize); err != nil {
+		log.Errorf("AppendToWAL: failed to write block size to WAL file: %v", err)
+		return err
+	}
+
+	_, err = w.file.Write(blockBuf.Bytes())
+	if err != nil {
+		log.Errorf("FlushWal: failed to write block content of size %d to WAL file: %v", len(blockBuf.Bytes()), err)
+		return err
+	}
+	return nil
+}
+
+func (w *WAL) Close() error {
+	return w.file.Close()
+}
+
+func DeleteWAL(filename string) error {
+	dirPath, err := getBaseWalDir()
+	filePath := filepath.Join(dirPath, filename+".wal")
+	if err != nil {
+		log.Errorf("DeleteWAL: failed to get base WAL directory: %v", err)
+		return err
+	}
+	return os.Remove(filePath)
+}
+
+type WalIterator struct {
+	file         *os.File
+	currentBlock []WALDatapoint
+	currentIndex int
+}
+
+func NewReaderWAL(filename string) (*WalIterator, error) {
+	dirPath, err := getBaseWalDir()
+	filePath := filepath.Join(dirPath, filename+".wal")
+	if err != nil {
+		log.Errorf("NewReaderWAL: failed to get base WAL directory: %v", err)
+		return nil, err
+	}
+
+	f, err := os.Open(filePath)
+	if err != nil {
+		log.Errorf("NewReaderWAL: failed to open WAL file at path %s: %v", filePath, err)
+		return nil, err
+	}
+	return &WalIterator{file: f}, nil
+}
+
+func (it *WalIterator) Next() (*WALDatapoint, bool, error) {
+	if it.currentIndex < len(it.currentBlock) {
+		dp := &it.currentBlock[it.currentIndex]
+		it.currentIndex++
+		return dp, true, nil
+	}
+
+	var blockSize uint32
+	err := binary.Read(it.file, binary.LittleEndian, &blockSize)
+	if errors.Is(err, io.EOF) {
+		log.Infof("Next: reached end of WAL file")
+		return nil, false, nil
+	} else if err != nil {
+		log.Errorf("Next: failed to read block size from WAL file: %v", err)
+		return nil, false, err
+	}
+
+	blockData := make([]byte, blockSize)
+	_, err = io.ReadFull(it.file, blockData)
+	if err != nil {
+		log.Errorf("Next: failed to read block data of size %d: %v", blockSize, err)
+		return nil, false, err
+	}
+
+	blockBuf := bytes.NewReader(blockData)
+	newBlock, err2 := dataDecompression(blockBuf)
+
+	if err2 != nil {
+		log.Errorf("Next: dataDecompression failed: %v", err2)
+		return nil, false, err2
+	}
+
+	if len(newBlock) == 0 {
+		log.Warnf("Next: decompressed block has zero datapoints")
+		return nil, false, nil
+	}
+
+	it.currentBlock = newBlock
+	it.currentIndex = 1
+
+	return &newBlock[0], true, nil
+}
+
+func (it *WalIterator) Close() error {
+	return it.file.Close()
+}
+
+func dataCompression(dps []WALDatapoint, blockBuf *bytes.Buffer) error {
+	for i, dp := range dps {
+		if err := binary.Write(blockBuf, binary.LittleEndian, dp.Timestamp); err != nil {
+			log.Errorf("dataCompression: failed to write Timestamp at index %d: %v", i, err)
+			return err
+		}
+		if err := binary.Write(blockBuf, binary.LittleEndian, dp.DpVal); err != nil {
+			log.Errorf("dataCompression: failed to write DpVal at index %d: %v", i, err)
+			return err
+		}
+		if err := binary.Write(blockBuf, binary.LittleEndian, dp.Tsid); err != nil {
+			log.Errorf("dataCompression: failed to write Tsid at index %d: %v", i, err)
+			return err
+		}
+	}
+	return nil
+}
+
+func dataDecompression(blockBuf *bytes.Reader) ([]WALDatapoint, error) {
+	var newBlock []WALDatapoint
+	index := 0
+
+	for blockBuf.Len() > 0 {
+		var dp WALDatapoint
+		if err := binary.Read(blockBuf, binary.LittleEndian, &dp.Timestamp); err != nil {
+			log.Errorf("dataDecompression: failed to read Timestamp at index %d: %v", index, err)
+			return nil, err
+		}
+		if err := binary.Read(blockBuf, binary.LittleEndian, &dp.DpVal); err != nil {
+			log.Errorf("dataDecompression: failed to read DpVal at index %d: %v", index, err)
+
+			return nil, err
+		}
+		if err := binary.Read(blockBuf, binary.LittleEndian, &dp.Tsid); err != nil {
+			log.Errorf("dataDecompression: failed to read Tsid at index %d: %v", index, err)
+			return nil, err
+		}
+		newBlock = append(newBlock, dp)
+		index++
+	}
+	return newBlock, nil
+}
--- a/pkg/segment/writer/metrics/wal/wal_test.go
+++ b/pkg/segment/writer/metrics/wal/wal_test.go
@@ -1,0 +1,119 @@
+package wal
+
+import (
+	"os"
+	"path/filepath"
+	"testing"
+
+	"github.com/siglens/siglens/pkg/config"
+	"github.com/stretchr/testify/assert"
+)
+
+func init() {
+	config.SetDataPath(os.TempDir() + "/test-wal/")
+	runningConfig := config.GetTestConfig(os.TempDir())
+	config.SetConfig(runningConfig)
+}
+
+func cleanUpWalFile() {
+	os.RemoveAll(config.GetDataPath())
+}
+
+func TestWALAppendAndRead(t *testing.T) {
+	filename := "testwal"
+
+	cleanUpWalFile()
+	defer cleanUpWalFile()
+
+	wal, err := NewWAL(filename)
+	assert.NoError(t, err)
+	defer wal.Close()
+
+	datapoints := []WALDatapoint{
+		{Timestamp: 1001, DpVal: 12.5, Tsid: 1},
+		{Timestamp: 1002, DpVal: 13.7, Tsid: 2},
+		{Timestamp: 1003, DpVal: 15.1, Tsid: 3},
+	}
+
+	err = wal.AppendToWAL(datapoints)
+	assert.NoError(t, err)
+
+	it, err := NewReaderWAL(filename)
+	assert.NoError(t, err)
+	defer it.Close()
+
+	var readDatapoints []WALDatapoint
+
+	for {
+		dp, ok, err := it.Next()
+		assert.NoError(t, err)
+		if !ok {
+			break
+		}
+		readDatapoints = append(readDatapoints, *dp)
+	}
+
+	assert.Equal(t, len(datapoints), len(readDatapoints))
+
+	for i, dp := range datapoints {
+		assert.Equal(t, dp.Timestamp, readDatapoints[i].Timestamp)
+		assert.Equal(t, dp.DpVal, readDatapoints[i].DpVal)
+		assert.Equal(t, dp.Tsid, readDatapoints[i].Tsid)
+	}
+}
+
+func GetWalFilePath(filename string) (string, error) {
+	dirPath, err := getBaseWalDir()
+	filePath := filepath.Join(dirPath, filename+".wal")
+	return filePath, err
+}
+
+func TestNewWAL(t *testing.T) {
+	filename := "test_newwal"
+	defer cleanUpWalFile()
+
+	wal, err := NewWAL(filename)
+	assert.NoError(t, err)
+	assert.NotNil(t, wal)
+	defer wal.Close()
+
+	filePath, _ := GetWalFilePath(filename)
+	_, err = os.Stat(filePath)
+	assert.NoError(t, err)
+}
+
+func TestDeleteWAL(t *testing.T) {
+	filename := "test_deletewal"
+
+	wal, err := NewWAL(filename)
+	assert.NoError(t, err)
+	wal.Close()
+
+	err = DeleteWAL(filename)
+	assert.NoError(t, err)
+
+	filePath, _ := GetWalFilePath(filename)
+	_, err = os.Stat(filePath)
+	assert.True(t, os.IsNotExist(err))
+}
+
+func TestAppendToWAL(t *testing.T) {
+	filename := "test_appendwal"
+	defer cleanUpWalFile()
+
+	wal, err := NewWAL(filename)
+	assert.NoError(t, err)
+	defer wal.Close()
+
+	dps := []WALDatapoint{
+		{Timestamp: 1010, DpVal: 20.0, Tsid: 10},
+	}
+
+	err = wal.AppendToWAL(dps)
+	assert.NoError(t, err)
+
+	filePath, _ := GetWalFilePath(filename)
+	info, err := os.Stat(filePath)
+	assert.NoError(t, err)
+	assert.Greater(t, info.Size(), int64(0))
+}


pkg/segment/writer/metrics/wal/wal.go
```diff
@@ -0,0 +1,297 @@
...
+)
+
+type WALDatapoint struct {
+	Timestamp uint64
+	DpVal     float64
+	Tsid      uint64
+}
+
+type WAL struct {
+	file          *os.File
```
nkunal, line 22: nit: maybe call this fd ?

pkg/segment/writer/metrics/wal/wal.go
```diff
@@ -0,0 +1,297 @@
...
+/*
+This function appends the block to the file sequentially—each second, a new block is added.
+
+File Format:
+BlockLen:               4 Bytes
+ZstdEncode following block:
+    NumOfDatapoints (N): 4 Bytes
+    BinaryEncodeAllTimestamps
+    BinaryEncodeAllDpVals
+    BinaryEncodeAllTsid
```
nkunal, line 53: nit: change `BinaryEncodeAll..` to `BinaryEncodedAll..`

pkg/segment/writer/metrics/wal/wal.go
```diff
@@ -0,0 +1,297 @@
...
+		return nil, false, nil
+	} else if err != nil {
+		log.Errorf("Next: failed to read block size from WAL file: %v", err)
+		return nil, false, err
+	}
+
+	blockData := make([]byte, blockSize)
+	_, err = io.ReadFull(it.file, blockData)
+	if err != nil {
+		log.Errorf("Next: failed to read block data of size %d: %v", blockSize, err)
```
nkunal, line 133: add filename here

pkg/segment/writer/metrics/wal/wal.go
```diff
@@ -0,0 +1,297 @@
...
+	}
+
+	timestamps := make([]uint64, N)
+	dpVals := make([]float64, N)
+	tsids := make([]uint64, N)
+	for i, dp := range dps {
+		timestamps[i] = dp.Timestamp
+		dpVals[i] = dp.DpVal
+		tsids[i] = dp.Tsid
+	}
```
nkunal, line 177: you don't need to copy over all the these things agains, you can just iterate over the dps eblow

pkg/segment/writer/metrics/wal/wal.go
```diff
@@ -0,0 +1,297 @@
...
+
+	for _, ts := range timestamps {
```
nkunal, line 179: here you can do 
```
for _ dp := range dps {
       binary.Write(rawBlockBuf, binary.LittleEndian, dp.Timestamp)
```
I understand you are trying to save iteration of the for loops, but allocating these temp vars is a huge mem pressure.

pkg/segment/writer/metrics/wal/wal.go
```diff
@@ -0,0 +1,297 @@
...
+}
+
+func compressDataWithZSTD(data []byte) ([]byte, error) {
+	encoder, err := zstd.NewWriter(nil)
+	if err != nil {
+		log.Errorf("zstdCompressBytes: failed to create zstd encoder: %v", err)
+		return nil, err
+	}
+	defer encoder.Close()
+	return encoder.EncodeAll(data, nil), nil
```
nkunal, line 224: make this a class function, and the second param you pass to EncodeAll, pass a class variable encodeBuf, take this buf from the buffer pool and keep reusing .
see how the segwriter is doing it here : 
https://github.com/siglens/siglens/blob/4ea5e452855dbc720a20289a01a6bfdb1c540677/pkg/segment/writer/segwriter.go#L974

pkg/segment/writer/metrics/wal/wal.go
```diff
@@ -0,0 +1,297 @@
...
+	_, err = compressedBlockBuf.Write(compressedData)
+	if err != nil {
+		log.Errorf("EncodeWALBlock: failed to write compressed data to buffer: %v", err)
+		return err
+	}
+	return nil
+}
+
+func compressDataWithZSTD(data []byte) ([]byte, error) {
+	encoder, err := zstd.NewWriter(nil)
```
nkunal, line 218: the encoder can be just a module var, and does not need to be clsoed every time.
see reference here:
https://github.com/siglens/siglens/blob/develop/pkg/segment/writer/segwriter.go#L72

pkg/segment/writer/metrics/wal/wal.go
```diff
@@ -0,0 +1,272 @@
...
+		log.Errorf("decodeWALBlock: failed to read datapoint count: %v", err)
+		return nil, err
+	}
+
+	timestamps := make([]uint64, N)
+	dpVals := make([]float64, N)
+	tsids := make([]uint64, N)
+
+	for i := 0; i < int(N); i++ {
+		err := binary.Read(rawReader, binary.LittleEndian, &timestamps[i])
```
nkunal, line 239: then in here you would do :
` binary.Read(rawReader, binary.LittleEndian, &it.readDps[i].timestamp)`

pkg/segment/writer/metrics/wal/wal.go
```diff
@@ -0,0 +1,272 @@
...
+		if err != nil {
+			log.Errorf("decodeWALBlock: failed to read timestamp: %v", err)
+			return nil, err
+		}
+	}
+
+	for i := 0; i < int(N); i++ {
+		err := binary.Read(rawReader, binary.LittleEndian, &dpVals[i])
```
nkunal, line 247: same comment here

pkg/segment/writer/metrics/wal/wal.go
```diff
@@ -0,0 +1,272 @@
...
+		if err != nil {
+			log.Errorf("decodeWALBlock: failed to read dpVal: %v", err)
+			return nil, err
+		}
+	}
+
+	for i := 0; i < int(N); i++ {
+		err := binary.Read(rawReader, binary.LittleEndian, &tsids[i])
```
nkunal, line 255: same comment here

pkg/segment/writer/metrics/wal/wal.go
```diff
@@ -0,0 +1,272 @@
...
+		if err != nil {
+			log.Errorf("decodeWALBlock: failed to read tsid: %v", err)
+			return nil, err
+		}
+	}
+
+	dps := make([]walDatapoint, N)
+	for i := 0; i < int(N); i++ {
```
nkunal, line 263: you won't need this for loop

pkg/segment/writer/metrics/wal/wal.go
```diff
@@ -0,0 +1,272 @@
...
+		dps[i] = walDatapoint{
+			timestamp: timestamps[i],
+			dpVal:     dpVals[i],
+			tsid:      tsids[i],
+		}
+	}
+
+	return dps, nil
```
nkunal, line 271: here you would return `it.readDps[0:N]`

pkg/segment/writer/metrics/wal/wal.go
```diff
@@ -0,0 +1,303 @@
...
+func (w *WAL) AppendToWAL(dps []WalDatapoint) error {
+	err := w.encodeWALBlock(dps)
+	if err != nil {
+		log.Errorf("AppendToWAL: dataCompression failed: %v", err)
+		return err
+	}
+
+	checksum := crc32.ChecksumIEEE(w.encodeBuf)
+
+	blockSize := uint32(len(w.encodeBuf) + 4)
```
nkunal, line 80: add a comment what this 4 is for

pkg/segment/writer/metrics/wal/wal.go
```diff
@@ -0,0 +1,303 @@
...
+	}
+
+	_, err = w.fd.Write(w.encodeBuf)
+	if err != nil {
+		log.Errorf("AppendToWAL: failed to write block content of size %d to WAL file: %v", len(w.encodeBuf), err)
+		return err
+	}
+
+	w.totalDps += uint32(len(dps))
+	w.encodedSize += uint64(4 + blockSize)
```
nkunal, line 101: add a comment what this 4 is for

pkg/segment/writer/metrics/wal/wal.go
```diff
@@ -0,0 +1,303 @@
...
+	var blockSize uint32
+	err := binary.Read(it.fd, binary.LittleEndian, &blockSize)
+	if errors.Is(err, io.EOF) {
+		return nil, false, nil
+	} else if err != nil {
+		log.Errorf("WalIterator Next: failed to read block size from WAL file: %v", it.fd.Name())
+		return nil, false, err
+	}
+
+	if blockSize < 4 {
```
nkunal, line 168: this 4 is used in multiple places, maybe define its as const

pkg/segment/writer/metrics/wal/wal.go
```diff
@@ -0,0 +1,303 @@
...
+	rawReader := bytes.NewReader(it.readBuf)
+
+	var N uint32
+	err = binary.Read(rawReader, binary.LittleEndian, &N)
+	if err != nil {
+		log.Errorf("decodeWALBlock: failed to read datapoint count: %v", err)
+		return err
+	}
+
+	it.readDps = toputils.ResizeSlice(it.readDps, int(N+uint32(len(it.readDps))))
```
nkunal, line 276: everytime you read a new block, you have to reuse the it.readDps , you don't need to maintain the old datapoints. other wise you will reun out of memory.

pkg/segment/utils/segconsts.go
```diff
@@ -65,6 +65,8 @@ var MAX_BYTES_METRICS_BLOCK uint64 = 1e+8         // 100MB
 var METRICS_SEARCH_ALLOCATE_BLOCK uint64 = 1.5e+8 // 150MB
 var MAX_BYTES_METRICS_SEGMENT uint64 = 1e+10      // 10GB
 var MAX_ACTIVE_SERIES_PER_SEGMENT = 10_000_000
+var MAX_BYTES_WAL_FILE uint64 = 134217728 // 128MB
```
nkunal, line 68: nit : rename to `MAX_WAL_FILE_SIZE_BYTES

pkg/segment/utils/segconsts.go
```diff
@@ -65,6 +65,8 @@ var MAX_BYTES_METRICS_BLOCK uint64 = 1e+8         // 100MB
...
+var MAX_WALDATPOINT_SIZE_TO_FLUSH = 10000
```
nkunal, line 69: rename to `WAL_BLOCK_FLUSH_SIZE`

pkg/segment/writer/metrics/metricssegment.go
```diff
@@ -531,6 +554,11 @@ func EncodeDatapoint(mName []byte, tags *TagsHolder, dp float64, timestamp uint3
 			return err
 		}
 	}
+	err = mSeg.mBlock.appendToWALBuffer(walDataPoint)
+	if err != nil {
+		log.Errorf("EncodeDatapoint : Failed to buffer WAL datapoint: %v", err)
```
nkunal, line 559: don;t log this, it could create flood, just return the error

pkg/segment/writer/metrics/metricssegment.go
```diff
@@ -1698,3 +1753,64 @@ func FindTagValuesUnrotated(tRange *dtu.MetricsTimeRange, mQuery *structs.Metric
...
+				log.Errorf("AppendWalDataPoint : Failed to rotate WAL file: %v", err)
+				return err
+			}
+		}
+		mb.walDPBuffer = mb.walDPBuffer[:0]
+	}
+	return nil
+}
+
+func (mb *MetricsBlock) RotateWAL() error {
```
nkunal, line 1777: this does not need to be exported

pkg/segment/writer/metrics/metricssegment.go
```diff
@@ -145,11 +147,17 @@ Every 5s, this metrics buffer should persist to disk and will create / update tw
...
-	sortedTsids    []uint64
-	mBlockSummary  *structs.MBlockSummary
-	blkEncodedSize uint64 // total encoded size of the block
+	tsidLookup      map[uint64]int
+	allSeries       []*TimeSeries
+	sortedTsids     []uint64
+	mBlockSummary   *structs.MBlockSummary
+	blkEncodedSize  uint64 // total encoded size of the block
+	currentWal      *wal.WAL
+	walDPBuffer     []wal.WalDatapoint
```
nkunal, line 156: nit: rename this to `dpsInWalMem`

pkg/segment/writer/metrics/metricssegment.go
```diff
@@ -145,11 +147,17 @@ Every 5s, this metrics buffer should persist to disk and will create / update tw
...
+	allWALFD        []*wal.WAL
```
nkunal, line 157: rename to `allWALs` and put it below `currentWal`
add a comment to what this means

pkg/segment/writer/metrics/metricssegment.go
```diff
@@ -145,11 +147,17 @@ Every 5s, this metrics buffer should persist to disk and will create / update tw
...
+	currentWALIndex uint64
+	segID           uint64
+	shardID         string
```
nkunal, line 160: for now keep the name of this as mId, we are going to do a renaming of mid once everywhere later

pkg/segment/writer/metrics/metricssegment.go
```diff
@@ -531,6 +554,11 @@ func EncodeDatapoint(mName []byte, tags *TagsHolder, dp float64, timestamp uint3
 			return err
 		}
 	}
+	err = mSeg.mBlock.appendToWALBuffer(walDataPoint)
```
nkunal, line 557: a better way may be to expose a func : `mSeg.mBlock.addDpToWal(timestamp, dp, tsid)`
that way you dont have to allocate this `wal.WalDatapoint` on stack.

pkg/segment/writer/metrics/metricssegment.go
```diff
@@ -1698,3 +1753,64 @@ func FindTagValuesUnrotated(tRange *dtu.MetricsTimeRange, mQuery *structs.Metric
 
 	return nil
 }
+
+func (mb *MetricsBlock) appendToWALBuffer(dp wal.WalDatapoint) error {
```
nkunal, line 1757: continuing the comment from line 557, then in this function, you will get those 3 params. 

you would then just do 
```
mb.walDPBuffer[dpIdx].Timestamp = ts
mb.walDPBuffer[dpIdx].dpVal = dpVal
mb.walDPBuffer[dpIdx].tsid = tsid
dpIdx++
```
THis will improve mem efficiency, and keep reusing the buffer

pkg/segment/writer/metrics/metricssegment.go
```diff
@@ -554,9 +552,8 @@ func EncodeDatapoint(mName []byte, tags *TagsHolder, dp float64, timestamp uint3
 			return err
 		}
 	}
-	err = mSeg.mBlock.appendToWALBuffer(walDataPoint)
+	err = mSeg.mBlock.appendToWALBuffer(uint64(timestamp), dp, tsid)
```
nkunal, line 555: dont cast , just make this function accept int64

pkg/segment/writer/metrics/metricssegment.go
```diff
@@ -1754,27 +1751,32 @@ func FindTagValuesUnrotated(tRange *dtu.MetricsTimeRange, mQuery *structs.Metric
...
-	if len(mb.walDPBuffer) > utils.MAX_WALDATPOINT_SIZE_TO_FLUSH {
-		err := mb.currentWal.AppendToWAL(mb.walDPBuffer)
+func (mb *MetricsBlock) appendToWALBuffer(timestamp uint64, dp float64, tsid uint64) error {
+	mb.lock.Lock()
+	defer mb.lock.Unlock()
+	mb.dpsInWalMem[mb.dpIdx].Timestamp = timestamp
+	mb.dpsInWalMem[mb.dpIdx].DpVal = dp
+	mb.dpsInWalMem[mb.dpIdx].Tsid = tsid
+	mb.dpIdx++
+	if len(mb.dpsInWalMem) >= utils.WAL_BLOCK_FLUSH_SIZE {
```
nkunal, line 1761: check the `dpIdx` here

pkg/segment/writer/metrics/metricssegment.go
```diff
@@ -1754,27 +1751,32 @@ func FindTagValuesUnrotated(tRange *dtu.MetricsTimeRange, mQuery *structs.Metric
...
+		err := mb.currentWal.AppendToWAL(mb.dpsInWalMem)
```
nkunal, line 1762: don;t you need to do `mb.dpsInWalMem[0:mb.dpIdx]` here to pick the correct amount of DPs

cmd/startup/startup.go
```diff
@@ -380,6 +380,13 @@ func startIngestServer(serverAddr string) {
 			}
 		}
 	}()
+
+	go func() {
```
nkunal, line 384: we should do this inline and not a goroutine. Because we dont want to start accepting new requests until we have rcovered the wal

cmd/startup/startup.go
```diff
@@ -380,6 +380,13 @@ func startIngestServer(serverAddr string) {
...
+		err := metrics.RecoverWALData()
+		if err != nil {
+			log.Errorf("Filed to recover wal files %v", err)
```
nkunal, line 387: `Failed to.. err: %v`
prefix function name here

pkg/segment/writer/metrics/metricssegment.go
```diff
@@ -567,6 +568,183 @@ func EncodeDatapoint(mName []byte, tags *TagsHolder, dp float64, timestamp uint3
...
 }
 
+type walFilesInfo struct {
+	mId      string
+	segID    uint64
+	blockNo  uint64
+	walFiles []string
+}
+
+func collectWALFileInfo(baseDir string) (map[string]*walFilesInfo, error) {
```
nkunal, line 578: rename this to `extractWALFileInfo`

pkg/segment/writer/metrics/wal/wal.go
```diff
@@ -0,0 +1,334 @@
...
+	fNameWithPath string
+	totalDps      uint32
+	encodedSize   uint64
+	encodeBuf     []byte
+	rawBlockBuf   *bytes.Buffer
+	checksumBuf   []byte
+}
+
+var encoder, _ = zstd.NewWriter(nil)
+var decoder, _ = zstd.NewReader(nil)
```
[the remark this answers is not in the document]
VasudevRaut: At the time of writing, through the WAL caller module, we have added a lock so that only one thread can use the encoder. Additionally, during reading, we are not reading concurrently, so a lock is not needed.

pkg/segment/writer/metrics/wal/wal.go
```diff
@@ -0,0 +1,334 @@
...
+func (w *WAL) Close() error {
+	if w.fd != nil {
+		return w.fd.Close()
+	}
+	return errors.New("file descriptor is nil")
+}
+
+func (w *WAL) DeleteWAL() error {
+	if w.fd != nil {
+		_ = w.fd.Close()
```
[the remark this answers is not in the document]
VasudevRaut: If w.fd is already closed elsewhere, calling Close()  again may result in an error
Even if Close() fails, we proceed with os.Remove(w.fNameWithPath) that's why ignore it

pkg/segment/writer/metrics/wal/wal.go
```diff
@@ -0,0 +1,334 @@
...
+}
+
+/*
+This function appends the block to the file, a new block is added.
+
+File Format:
+version:					1 Bytes
+	BlockLen:               4 Bytes
+	Checksum:				4 Bytes
+	ZstdEncoding of following block:
```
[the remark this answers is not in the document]
VasudevRaut: /*
This function appends a new block to the file.

File Format:
	Version:                1 Byte  // File format version

	// Repeating Blocks Structure
	[Block] {
		BlockLen:               4 Bytes  // Size of this block (excluding this field)
		Checksum:               4 Bytes  // CRC32 checksum for data integrity
		ZstdEncoded Block:
			NumOfDatapoints (N): 4 Bytes  // Number of datapoints in this block
			BinaryEncodedAllTimestamps   // Encoded timestamps for all datapoints
			BinaryEncodedAllDpVals       // Encoded values for all datapoints
			BinaryEncodedAllTsid         // Encoded Time-Series IDs
	}

	Multiple such blocks are appended continuously.
*/

pkg/segment/writer/metrics/wal/wal.go
```diff
@@ -0,0 +1,334 @@
...
+	}
+
+	_, err = w.fd.Write(w.encodeBuf)
+	if err != nil {
+		log.Errorf("AppendToWAL: failed to write block content of size %d to WAL file: %v", len(w.encodeBuf), err)
+		return err
+	}
+
+	w.totalDps += uint32(len(dps))
+	w.encodedSize += uint64(UINT32_SIZE + blockSize) //UINT32_SIZE : 4 bytes for storing block size prefix
```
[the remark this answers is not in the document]
VasudevRaut: Adding 4-byte UINT32 (blockSize field) size to encodedSize, excluded from blockSize.

pkg/segment/writer/metrics/wal/wal.go
```diff
@@ -155,7 +157,7 @@ func (w *WAL) GetWALStats() (string, uint32, uint64) {
 	return w.fNameWithPath, w.totalDps, w.encodedSize
 }
 
-func NewReaderWAL(baseDir, filename string) (*WalIterator, error) {
+func NewWALReader(baseDir, filename string) (*WalIterator, error) {
```
AndrewHess, line 160: nit: since you changed `NewWAL()` to take one parameter, it probably makes sense to do the same here

pkg/segment/writer/metrics/metricssegment.go
```diff
@@ -145,11 +148,19 @@ Every 5s, this metrics buffer should persist to disk and will create / update tw
...
+	tsidLookup      map[uint64]int
+	allSeries       []*TimeSeries
+	sortedTsids     []uint64
+	mBlockSummary   *structs.MBlockSummary
+	blkEncodedSize  uint64     // total encoded size of the block
+	currentWal      *wal.WAL   // Active WAL file
+	allWALs         []*wal.WAL // List of WAL files
+	dpsInWalMem     []wal.WalDatapoint
+	dpIdx           uint64 // Next write position in dpsInWalMem
+	currentWALIndex uint64
```
AndrewHess, line 160: Since you're adding a lot of WAL-related fields, it might make sense to make a new struct type and have this `MetricsBlock` have that struct as a field. But since you're more familiar with this code than I am, do whatever you think is going to be best; if you think the current structure is best, then keep it

pkg/segment/writer/metrics/metricssegment.go
```diff
@@ -145,11 +148,19 @@ Every 5s, this metrics buffer should persist to disk and will create / update tw
...
+	segID           uint64
+	mId             string
+	lock            sync.Mutex
```
AndrewHess, line 163: I think this lock is only used for appending a WAL block. If so, maybe rename/comment it to make that clear (or if you move it into a new struct type for WAL things, keeping it as `lock` makes sense

pkg/segment/writer/metrics/metricssegment.go
```diff
@@ -542,6 +568,194 @@ func EncodeDatapoint(mName []byte, tags *TagsHolder, dp float64, timestamp uint3
...
+		fileName := file.Name()
+
+		parts := strings.Split(fileName, "_")
+		if len(parts) < 6 {
+			continue
+		}
+
+		mId := parts[1]
+		segIDStr := parts[3]
+		blockNoStr := parts[5]
```
AndrewHess, line 600: You should add a comment somewhere (around here, at the top of this function, etc.) describing what a WAL filename should be, since it's encoding some info

pkg/segment/writer/metrics/metricssegment.go
```diff
@@ -542,6 +568,194 @@ func EncodeDatapoint(mName []byte, tags *TagsHolder, dp float64, timestamp uint3
...
+	}
+
+	for _, fileData := range walFilesData {
+		mBlock := initMetricsBlock(fileData.mId, fileData.segID, fileData.blockNo)
+		isWalFileEmpty := true
+		for _, walFileName := range fileData.walFiles {
+			walIterator, err := wal.NewWALReader(baseDir, walFileName)
+			if err != nil {
+				log.Errorf("RecoverWALData :Failed to create WAL reader for file %s: %v", walFileName, err)
+				return err
```
AndrewHess, line 665: If we get an error with some WAL files or some blocks in a WAL file, we should still ingest whatever is valid.

pkg/segment/writer/metrics/metricssegment.go
```diff
@@ -1698,3 +1939,74 @@ func FindTagValuesUnrotated(tRange *dtu.MetricsTimeRange, mQuery *structs.Metric
...
+	defer mb.lock.Unlock()
+	mb.dpsInWalMem[mb.dpIdx].Timestamp = timestamp
+	mb.dpsInWalMem[mb.dpIdx].DpVal = dp
+	mb.dpsInWalMem[mb.dpIdx].Tsid = tsid
+	mb.dpIdx++
+	if int(mb.dpIdx) >= utils.WAL_BLOCK_FLUSH_SIZE {
+		err := mb.currentWal.Append(mb.dpsInWalMem[0:mb.dpIdx])
+		if err != nil {
+			log.Errorf("AppendWalDataPoint : Failed to append datapoints to WAL: %v", err)
+			return err
```
AndrewHess, line 1954: You should check `int(mb.dpIdx) >= utils.WAL_BLOCK_FLUSH_SIZE` before you insert into `dpsInWalMem`. Without that, if we return an error on 1954 and then call `appendToWALBuffer()` again, we can get a panic at 1946 because `dpIdx` is too large

pkg/segment/writer/metrics/metricssegment.go
```diff
@@ -1698,3 +1939,74 @@ func FindTagValuesUnrotated(tRange *dtu.MetricsTimeRange, mQuery *structs.Metric
...
+	mb.currentWALIndex++
+	return mb.initNewWAL()
+}
+
+func (mb *MetricsBlock) initNewWAL() error {
+	if mb.currentWal != nil {
+		err := mb.currentWal.Close()
+		if err != nil {
+			log.Errorf("initNewWAL : Failed to close current WAL: %v", err)
+		}
```
AndrewHess, line 1978: nit: I think you're not returning the error here because we can still gracefully handle this case. If so, change `log.Errorf` to `log.Warnf`

pkg/segment/writer/metrics/metricssegment.go
AndrewHess, line 1: I thought we also planned on flushing the WAL to disk every second. Did our plan change?

AndrewHess (commented):
Looking pretty good

pkg/segment/writer/metrics/metricssegment.go
```diff
@@ -1941,53 +1948,81 @@ func FindTagValuesUnrotated(tRange *dtu.MetricsTimeRange, mQuery *structs.Metric
...
+
 	return nil
 }
 
+func timeBasedWalDPSFlush() {
+	for {
+		time.Sleep(WAL_DPS_FLUSH_SLEEP_DURATION * time.Second)
+		for _, ms := range GetAllMetricsSegments() {
+			if ms.mBlock.walState.dpIdx > 0 {
+				ms.mBlock.walState.lock.Lock()
```
AndrewHess, line 1983: I think you should take the lock before you check `dpIdx`; otherwise we can still execute the logic inside this `if` when `dpIdx == 0`. I'm not sure if that would cause any problems

PR #2397, opened 2025-03-21
fix: min/max duration parsing into nano-seconds for search traces

### LINKED ISSUE #574: [BUG] Min and Max Duration search doesn't work for traces

bhorsujal:
# Description
This commit introduces a `durationToNanosecond()` function that converts duration strings (in seconds, milliseconds, or microseconds) into nanoseconds. It validates the input format and uses showToast() to display an error if the conversion fails.

Fixes #574

# Testing
Verified the conversion of min/max values using the showToast() function to ensure that inputs (e.g., 1.2s, 100ms, 500µs) are accurately converted to nanoseconds.
Ran make test to confirm that all tests pass without issues.

# Checklist:

- [x] I have formatted the code, if applicable.


bhorsujal:
Thank you so much for reviewing and merging the PR.

PR #2401, opened 2025-03-21
Added Usage Stats Screen

sonamgupta21:
# Description
**Frontend Changes:**
  1. Added a new "Usage Stats" link in the navbar.
  2. Displays logs volume, traces volume, active series count for metrics, and datapoints.
  3. Implemented timepicker (supports both relative and custom time ranges).
  4. Added an option to select granularity (hourly, daily, monthly).
  5. Graphs update dynamically when time range or granularity is changed.

**Backend Changes:**
  1. `api/clusterIngestStats` API now supports fetching stats for a custom time range.
  2. Added Granularity in Request Body:
  If no granularity is provided, the backend auto adjusts the granularity according to the time range. (used for the cluster page).
  3. Added support for fetching stats with monthly granularity.

image
image
image

# Testing

# Checklist:

### DIFF
--- a/pkg/health/clusterStatsHandler.go
+++ b/pkg/health/clusterStatsHandler.go
@@ -259,72 +259,176 @@ func ProcessClusterIngestStatsHandler(ctx *fasthttp.RequestCtx, orgId int64) {
 		httpResp.ChartStats[k]["MetricsBytesCount"] = float64(entry.MetricsBytesCount)
 		httpResp.ChartStats[k]["TraceBytesCount"] = float64(entry.TraceBytesCount)
 		httpResp.ChartStats[k]["TraceSpanCount"] = entry.TraceSpanCount
+		httpResp.ChartStats[k]["ActiveSeriesCount"] = entry.ActiveSeriesCount
 	}
 	utils.WriteJsonResponse(ctx, httpResp)
 }
 
-func parseAlphaNumTime(inp string, defValue uint64) (uint64, usageStats.UsageStatsGranularity) {
-	granularity := usageStats.Daily
-	sanTime := strings.ReplaceAll(inp, " ", "")
-	retVal := defValue
+func parseIngestionStatsRequest(jsonSource map[string]interface{}) (uint64, usageStats.UsageStatsGranularity) {
+	defaultPastHours := uint64(7 * 24) // 7 days default
 
+	startEpoch, hasStart := jsonSource["startEpoch"]
+	endEpoch, hasEnd := jsonSource["endEpoch"]
+	granularity, hasGranularity := jsonSource["granularity"]
+
+	getDefault := func() (uint64, usageStats.UsageStatsGranularity) {
+		if hasGranularity {
+			return defaultPastHours, parseGranularity(granularity)
+		}
+		return defaultPastHours, determineGranularity(defaultPastHours)
+	}
+
+	if !hasStart || !hasEnd || startEpoch == nil || endEpoch == nil {
+		return getDefault()
+	}
+
+	// Relative time format (now-Xh)
+	if startStr, ok := startEpoch.(string); ok && strings.Contains(startStr, "now-") {
+		pastHours, _ := parseAlphaNumTime(startStr, defaultPastHours)
+		if hasGranularity {
+			return pastHours, parseGranularity(granularity)
+		}
+		return pastHours, determineGranularity(pastHours)
+	}
+
+	// Parse timestamps
+	startTs := parseTimestamp(startEpoch)
+	endTs := parseTimestamp(endEpoch)
+
+	// Validate timestamps
+	if startTs == -1 || endTs == -1 || endTs <= startTs {
+		return getDefault()
+	}
+
+	// Calculate hours difference
+	hours := uint64((endTs - startTs) / 3600)
+	if hasGranularity {
+		return hours, parseGranularity(granularity)
+	}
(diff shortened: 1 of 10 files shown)


pkg/health/clusterStatsHandler.go
```diff
@@ -259,72 +259,176 @@ func ProcessClusterIngestStatsHandler(ctx *fasthttp.RequestCtx, orgId int64) {
...
+
+	getDefault := func() (uint64, usageStats.UsageStatsGranularity) {
+		if hasGranularity {
+			return defaultPastHours, parseGranularity(granularity)
+		}
+		return defaultPastHours, determineGranularity(defaultPastHours)
+	}
+
+	if !hasStart || !hasEnd || startEpoch == nil || endEpoch == nil {
+		return getDefault()
```
nkunal, line 282: nit: a little hard to follow the code, you probably dont need the getDefault func enclosure, you can just write that code in here

pkg/usageStats/stats.go
```diff
@@ -573,6 +575,18 @@ func GetUsageStats(pastXhours uint64, granularity UsageStatsGranularity, orgid i
...
+
+		endMonth := time.Date(endEpoch.Year(), endEpoch.Month(), 1, 0, 0, 0, 0, endEpoch.Location())
+
+		for !runningTs.After(endMonth) {
+			bucketInterval = runningTs.Format("2006-01")
+			resultMap[bucketInterval] = &ReadStats{}
+
+			runningTs = time.Date(runningTs.Year(), runningTs.Month()+1, 1, 0, 0, 0, 0, runningTs.Location())
+		}
 	}
```
AndrewHess, line 590: nit: add an `else` and return an error there, since that would be an unknown `granularity`

same comment at line 610

pkg/health/clusterStatsHandler.go
```diff
@@ -259,72 +259,176 @@ func ProcessClusterIngestStatsHandler(ctx *fasthttp.RequestCtx, orgId int64) {
...
+	}
+
+	// Handle relative time format (now-Xh)
+	if startStr, ok := startEpoch.(string); ok && strings.Contains(startStr, "now-") {
+		pastHours, _ := parseAlphaNumTime(startStr, defaultPastHours)
+		if hasGranularity {
+			return pastHours, parseGranularity(granularity)
+		}
+		return pastHours, determineGranularity(pastHours)
+	}
```
AndrewHess, line 289: You're not checking `endEpoch` at all here? That seems like a bug

pkg/health/clusterStatsHandler.go
```diff
@@ -259,72 +259,176 @@ func ProcessClusterIngestStatsHandler(ctx *fasthttp.RequestCtx, orgId int64) {
...
+	case 0:
+		return usageStats.Hourly
+	case 1:
+		return usageStats.Daily
+	case 2:
+		return usageStats.ByMinute
+	case 3:
+		return usageStats.Monthly
+	default:
+		return usageStats.Daily
```
AndrewHess, line 365: Is this because the granularity is defined like this:
```
const (
	Hourly UsageStatsGranularity = iota + 1
	Daily
	ByMinute
	Monthly
)
```
?

If so, you're off by 1 on these cases (since `iota + 1` is 1), and I think it would be better to do
```
case int(usageStats.Hourly):
    return usageStats.Hourly
...
```

pkg/usageStats/stats.go
```diff
@@ -607,6 +609,8 @@ func GetUsageStats(pastXhours uint64, granularity UsageStatsGranularity, orgid i
 			bucketInterval = rStat.TimeStamp.Truncate(time.Duration(intervalMinutes) * time.Minute).Format("2006-01-02T15:04")
 		} else if granularity == Monthly {
 			bucketInterval = rStat.TimeStamp.Format("2006-01")
+		} else {
+			return nil, fmt.Errorf("GetUsageStats: unexpected unknown granularity value: %v", granularity)
```
AndrewHess, line 613: nit: remove either "unexpected" or "unknown"

pkg/health/clusterStatsHandler.go
```diff
@@ -279,8 +279,11 @@ func parseIngestionStatsRequest(jsonSource map[string]interface{}) (uint64, usag
...
-	// Handle relative time format (now-Xh)
-	if startStr, ok := startEpoch.(string); ok && strings.Contains(startStr, "now-") {
+	// Handle relative time format
+	startStr, startIsString := startEpoch.(string)
+	endStr, endIsString := endEpoch.(string)
+
+	if startIsString && endIsString && strings.Contains(startStr, "now-") && endStr == "now" {
 		pastHours, _ := parseAlphaNumTime(startStr, defaultPastHours)
 		if hasGranularity {
 			return pastHours, parseGranularity(granularity)
```
AndrewHess, line 289: I think there's still some edge cases. Like what if `end="now-1h"` instead of just `"now"`? Or what if `end="now"` and `start` is an absolute time?

To simplify this, I don't think you need this `if` block at line 286 at all, because you already parse the start and end times on lintes 295-296. But one thing to fix: there you parse them with `parseTimestamp()`, which on line 414 requires relative times to contain `"now-"`—you should probably change that to `"now"`. You also might want to add some unit tests to make sure you're correctly handling all the cases you expect to

AndrewHess (approved):

PR #2415, opened 2025-03-25
Promql instant query Support

Macbeth98:
# Description
- Implements the Instant Query for Metrics similar to Prometheus, except the reset of data when the `up` metric for the instance is `0`.

# Testing
- Tested by ingesting data into  Prometheus and SigLens
- Ran same instant query through Grafana. 
- Verified that the values are same for some queries.
- Added a new unit test case.

# Checklist:

### DIFF
--- a/pkg/e2etests/metrics_e2e_test.go
+++ b/pkg/e2etests/metrics_e2e_test.go
@@ -1769,7 +1769,7 @@ func Test_SimpleMetricInstantQuery(t *testing.T) {
 	assert.JSONEq(t, string(marshalExpectedResp), string(resp))
 
 	// Now since this is more than the DEFAULT_LOOKBACK_FOR_INSTANT_QUERIES(5m), the query will not return any results.
-	endTime = startTimestamp + 4600
+	endTime = startTimestamp + 3600 + promql.DEFAULT_LOOKBACK_FOR_INSTANT_QUERIES + 1
 
 	fastCtx = &fasthttp.RequestCtx{}
 	fastCtx.Request.PostArgs().Add("query", query)
--- a/pkg/e2etests/metrics_e2e_test.go
+++ b/pkg/e2etests/metrics_e2e_test.go
@@ -1732,7 +1732,7 @@ func Test_SimpleMetricInstantQuery(t *testing.T) {
 	err = initializeMetricsMetaData()
 	assert.Nil(t, err)
 
-	endTime := startTimestamp + 3601
+	endTime := startTimestamp + 3600
 	query := `(testmetric0)`
 
 	fastCtx := &fasthttp.RequestCtx{}
@@ -1744,7 +1744,6 @@ func Test_SimpleMetricInstantQuery(t *testing.T) {
 
 	resp := fastCtx.Response.Body()
 	assert.NotNil(t, resp)
-	fmt.Println(string(resp))
 
 	expectedPromQLResp := &structs.MetricsPromQLInstantQueryResponse{
 		Status: "success",
@@ -1768,6 +1767,27 @@ func Test_SimpleMetricInstantQuery(t *testing.T) {
 	marshalExpectedResp, err := json.Marshal(expectedPromQLResp)
 	assert.Nil(t, err)
 	assert.JSONEq(t, string(marshalExpectedResp), string(resp))
+
+	// Now since this is more than the DEFAULT_LOOKBACK_FOR_INSTANT_QUERIES(5m), the query will not return any results.
+	endTime = startTimestamp + 4600
+
+	fastCtx = &fasthttp.RequestCtx{}
+	fastCtx.Request.PostArgs().Add("query", query)
+	fastCtx.Request.PostArgs().Add("time", strconv.Itoa(int(endTime)))
+
+	promql.ProcessPromqlMetricsSearchRequest(fastCtx, 0)
+	assert.Equal(t, 200, fastCtx.Response.StatusCode())
+
+	expectedPromQLResp = &structs.MetricsPromQLInstantQueryResponse{
+		Status: "success",
+		Data: &structs.PromQLInstantData{
+			ResultType: parser.ValueTypeVector,
+		},
+	}
+
+	marshalExpectedResp, err = json.Marshal(expectedPromQLResp)
+	assert.Nil(t, err)
+	assert.JSONEq(t, string(marshalExpectedResp), string(fastCtx.Response.Body()))
 }
 
 func Test_metricsPersistAfterGracefulRestart(t *testing.T) {
--- a/pkg/e2etests/metrics_e2e_test.go
+++ b/pkg/e2etests/metrics_e2e_test.go
@@ -28,11 +28,13 @@ import (
 	"os"
 	"path/filepath"
 	"sort"
+	"strconv"
 	"strings"
 	"sync"
 	"testing"
 	"time"
 
+	"github.com/prometheus/prometheus/promql/parser"
 	"github.com/siglens/siglens/pkg/common/dtypeutils"
 	"github.com/siglens/siglens/pkg/config"
 	"github.com/siglens/siglens/pkg/integrations/prometheus/promql"
@@ -41,6 +43,7 @@ import (
 	segmetadata "github.com/siglens/siglens/pkg/segment/metadata"
 	"github.com/siglens/siglens/pkg/segment/query"
 	"github.com/siglens/siglens/pkg/segment/results/mresults"
+	"github.com/siglens/siglens/pkg/segment/structs"
 	"github.com/siglens/siglens/pkg/segment/utils"
 	"github.com/siglens/siglens/pkg/segment/writer"
 	"github.com/siglens/siglens/pkg/segment/writer/metrics"
@@ -48,6 +51,7 @@ import (
 	log "github.com/sirupsen/logrus"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
+	"github.com/valyala/fasthttp"
 )
 
 /*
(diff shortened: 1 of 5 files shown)


pkg/e2etests/metrics_e2e_test.go
```diff
@@ -1705,6 +1709,67 @@ func Test_SimpleMetricQueryUnrotatedBlockData(t *testing.T) {
...
+	fastCtx := &fasthttp.RequestCtx{}
+	fastCtx.Request.PostArgs().Add("query", query)
+	fastCtx.Request.PostArgs().Add("time", strconv.Itoa(int(endTime)))
+
+	promql.ProcessPromqlMetricsSearchRequest(fastCtx, 0)
+	assert.Equal(t, 200, fastCtx.Response.StatusCode())
+
+	resp := fastCtx.Response.Body()
+	assert.NotNil(t, resp)
+	fmt.Println(string(resp))
```
AndrewHess, line 1747: nit: remove this

pkg/e2etests/metrics_e2e_test.go
```diff
@@ -1705,6 +1709,67 @@ func Test_SimpleMetricQueryUnrotatedBlockData(t *testing.T) {
...
+	assert.Nil(t, err)
+
+	mSegs, err := rotateMetricsDataAndClearSegStore(true)
+	assert.Nil(t, err)
+	assert.Greater(t, len(mSegs), 0)
+
+	err = initializeMetricsMetaData()
+	assert.Nil(t, err)
+
+	endTime := startTimestamp + 3601
```
AndrewHess, line 1735: Where is 3601 coming from?

Also, you should test that when querying with timestamp more than `DEFAULT_LOOKBACK` past the last timestamp returns no data

pkg/segment/results/mresults/seriesresult.go
```diff
@@ -118,6 +121,24 @@ func (s *Series) GetIdx() int {
...
+		}
+		if ts == s.entries[s.idx].downsampledTime {
+			fmt.Println("AddEntry: duplicate timestamp found", ts, "dp:", dp, "series:", s.grpID.String(), "s.entries[s.idx].dpVal:", s.entries[s.idx].dpVal)
+		}
+		if ts > s.entries[s.idx].downsampledTime {
+			s.entries[s.idx].downsampledTime = ts
+			s.entries[s.idx].dpVal = dp
+		}
+
+		s.idx++
```
AndrewHess, line 138: Why do you decrement and then increment, instead of not changing `s.idx` at all?

pkg/segment/results/mresults/seriesresult.go
```diff
@@ -118,6 +121,24 @@ func (s *Series) GetIdx() int {
...
 func (s *Series) AddEntry(ts uint32, dp float64) {
+	if s.isInstantQuery {
+		// if instant query, we only need the latest value per series
+		if s.idx > 0 {
+			// Decrement the index to overwrite the previous entry
+			s.idx--
+		}
+		if ts == s.entries[s.idx].downsampledTime {
+			fmt.Println("AddEntry: duplicate timestamp found", ts, "dp:", dp, "series:", s.grpID.String(), "s.entries[s.idx].dpVal:", s.entries[s.idx].dpVal)
+		}
```
AndrewHess, line 132: Don't print; log if needed. Is it needed here?

pkg/integrations/prometheus/promql/metricsSearchHandler.go
```diff
@@ -54,6 +54,8 @@ import (
 	"github.com/valyala/fasthttp"
 )
 
+const DEFAULT_LOOKBACK uint32 = 5 * 60 // 5 mins
```
nkunal, line 57: nit: maybe rename this to DEFAULT_LOOKBACK_FOR_INSTANT_QUERIES

pkg/integrations/prometheus/promql/metricsSearchHandler.go
```diff
@@ -180,7 +182,9 @@ func ProcessPromqlMetricsSearchRequest(ctx *fasthttp.RequestCtx, myid int64) {
 		}
 	}
 
-	metricQueryRequest, pqlQuerytype, queryArithmetic, err := ConvertPromQLToMetricsQuery(searchText, endTime-1, endTime, myid)
+	startTime := endTime - DEFAULT_LOOKBACK
+	log.Infof("qid=%v, ProcessPromqlMetricsSearchRequest: InstantQuery; searchString=[%v] startEpochs=[%v] endEpochs=[%v]", qid, searchText, startTime, endTime)
```
nkunal, line 186: do we need this log line ? if not we can remove it.

pkg/segment/results/mresults/seriesresult.go
nkunal, line 132: remove this

pkg/segment/results/mresults/seriesresult.go
```diff
@@ -145,14 +166,39 @@ func (s *Series) sortEntries() {
...
+
+		// Decrement the index to overwrite the previous entry, since for instant only one entry is allowed
+		s.idx--
+		toJoin.idx--
+
+		maxTsEntry := s.entries[s.idx]
+		if maxTsEntry.downsampledTime < toJoin.entries[toJoin.idx].downsampledTime {
+			maxTsEntry = toJoin.entries[toJoin.idx]
+		}
```
AndrewHess, line 182: Are we really supposed to downsample before merging series? In that case it's not clear which value to choose if they downsample to the same timestamp, even if one of them was ingested later than the other. But I haven't checked how Prometheus handles this case

pkg/integrations/prometheus/promql/metricsSearchHandler.go
Macbeth98: Yes, We are not logging any info, we do for every other query. So I added this line that shows the query information.

pkg/e2etests/metrics_e2e_test.go
Macbeth98: yeah got it. The data that is ingested in this series has the highest timestamp at `3600`.

pkg/segment/results/mresults/seriesresult.go
Macbeth98: So the normal query flow expects the `idx` to be next value or next slot to insert items. But since, for instant query, there will only be one timestamp, which will be at `0` index. But `idx` will be at 1. So I am decrementing it, but to not disrupt the normal flow, I have incremented it back again.

pkg/segment/results/mresults/seriesresult.go
Macbeth98: No I will remove it.

pkg/segment/results/mresults/seriesresult.go
Macbeth98: For instant query, we are not doing any downsample, but since this flow utilizes the same flow as range query, the variable name is still downsample time.

pkg/segment/results/mresults/seriesresult.go
Macbeth98: I have refactored this part of the code.

pkg/e2etests/metrics_e2e_test.go
```diff
@@ -1768,6 +1767,27 @@ func Test_SimpleMetricInstantQuery(t *testing.T) {
 	marshalExpectedResp, err := json.Marshal(expectedPromQLResp)
 	assert.Nil(t, err)
 	assert.JSONEq(t, string(marshalExpectedResp), string(resp))
+
+	// Now since this is more than the DEFAULT_LOOKBACK_FOR_INSTANT_QUERIES(5m), the query will not return any results.
+	endTime = startTimestamp + 4600
```
AndrewHess, line 1772: just reference `DEFAULT_LOOKBACK_FOR_INSTANT_QUERIES` directly. If we want to change `DEFAULT_LOOKBACK_FOR_INSTANT_QUERIES`, we shouldn't have to change this test

pkg/segment/results/mresults/seriesresult.go
```diff
@@ -145,14 +168,35 @@ func (s *Series) sortEntries() {
...
+		// Decrement the index to overwrite the previous entry, since for instant only one entry is allowed
+		sIdx := s.idx - 1
+		toJoinIdx := toJoin.idx - 1
+
+		maxTsEntry := s.entries[sIdx]
+		if maxTsEntry.downsampledTime < toJoin.entries[toJoinIdx].downsampledTime {
+			maxTsEntry = toJoin.entries[toJoinIdx]
+		}
+
+		s.entries[s.idx] = maxTsEntry
```
AndrewHess, line 186: On line 186, isn't `s.idx` a bug? Shouldn't it be `sIdx`?

nit: on line 178, defining `sIdx := s.idx - 1` is a bit confusing; since it's only used in the next ~5 lines I think it's fine if you want to keep it

pkg/segment/results/mresults/seriesresult.go
```diff
@@ -118,6 +121,26 @@ func (s *Series) GetIdx() int {
...
+		idx := s.idx
+
+		// if instant query, we only need the latest value per series
+		if idx > 0 {
+			// Decrement the index to overwrite the previous entry
+			idx--
+		} else if idx == 0 {
+			// Increment the index to the next entry
+			s.idx++
+		}
```
AndrewHess, line 134: Now I'm confused. What's the expected state after this function finishes (for an instant query)? I thought it would be `s.idx` is always 1, and the new value is either added (if none existed before) or overwrites the previous one (if it's at a later timestamp than the previous).

But from this `if` chain, sometimes you exit the function with `s.idx == 0` and sometimes with `s.idx == 1`. Maybe you need unit tests

pkg/segment/results/mresults/seriesresult.go
Macbeth98: Yes that should be `sIdx`

pkg/segment/results/mresults/seriesresult.go
Macbeth98: Yes `s.idx` will always be one, except for the initial first time, it will be `0`, then you add a value at `0`, now the `s.idx` will be 1 and it will stay `1`.

AndrewHess (approved):

PR #2433, opened 2025-03-29
fix: PromQL Metrics Nested Queries with Binary Expr Plus other Misc fixes

Macbeth98:
# Description
- Fixes issue when `label_replace` function does not have source label.
- Fixes issue with `last_over_time` function
- Fixes parser for Nested queries containing Binary Expr.
- Added the ability to execute Metric Query Aggs Chain after an Binary Operation is executed.

# Testing
- Able to parse and execute the following query:
```
  sum by (cluster, namespace) (
    group by (cluster, namespace, workload) (
          (
            label_replace(
              label_replace(
                kube_pod_owner{cluster=~".+",namespace=~".+",owner_kind=~".+",owner_name=~".+",pod=~".+"},
                "workload",
                "$1",
                "owner_name",
                "^(.*?)(-[a-z0-9]{9,10})?$"
              ),
              "workload_type",
              "$1",
              "owner_kind",
              "(.*)"
            )
          )
        or
          (
            label_replace(
              label_replace(
                kube_pod_owner{cluster=~".+",namespace=~".+",owner_kind="",pod=~".+"},
                "workload",
                "$1",
                "pod",
                "^(.*?)(-[a-z0-9]{9,10})?$"
              ),
              "workload_type",
              "pod",
              "",
              ""
            )
          )
      or
        (
          label_replace(
            label_replace(
              kube_pod_owner{cluster=~".+",namespace=~".+",owner_kind="Node",pod=~".+"},
              "workload",
              "$1",
              "pod",
              "^(.*?)(-[a-z0-9]{9,10})?$"
            ),
            "workload_type",
            "staticpod",
            "",
            ""
          )
        )
    )
  )
or 
  (
      last_over_time(
        group by (cluster, namespace) (
          kube_namespace_status_phase{cluster=~".+",namespace=~".+",phase="Active"} == 1
        )[1h:]
      )
    -
      1
  )
  ```
- The results are same as Prometheus, we got some extra series included because of this Binary Expr: `(kube_namespace_status_phase{cluster=~".+", namespace=~".+", phase="Active"} == 1`. 
- But removing this block : `(last_over_time(group by (cluster, namespace) (kube_namespace_status_phase{cluster=~".+", namespace=~".+", phase="Active"} == 1)[1h:]) - 1)` gives the exact result as Prometheus
- Updated unit tests and added new unit test.
- Fixed most of the unit tests that were failing.
- **Still probably need to E2E tests**

# Checklist:

### DIFF
--- a/pkg/integrations/prometheus/promql/parser.go
+++ b/pkg/integrations/prometheus/promql/parser.go
@@ -771,15 +771,23 @@ func handleBinaryExpr(expr *parser.BinaryExpr, mQueryReqs []*structs.MetricsQuer
 	myid := mQueryReqs[0].MetricsQuery.OrgId
 	timeRange := mQueryReqs[0].TimeRange
 
-	arithmeticOperation := structs.QueryArithmetic{}
+	binaryOperation := structs.QueryArithmetic{}
 	var lhsRequest, rhsRequest []*structs.MetricsQueryRequest
 	var lhsQueryArth, rhsQueryArth []*structs.QueryArithmetic
 	lhsIsVector := false
 	rhsIsVector := false
 
+	if len(mQueryReqs) > 0 && mQueryReqs[0].MetricsQuery.SubsequentAggs != nil {
+		// This means there is a common group by on multiple metrics separated by operators.
+		// So, we need to append the aggregations to the current BinaryOperation, so that
+		// once the binary operation is done, we can apply the aggregations on the result.
+		// And we can discard the current aggregation in the mQueryReqs[0]
+		appendMetricAggsToTheMQuery(&binaryOperation, &mQueryReqs[0].MetricsQuery)
+	}
+
 	if constant, ok := expr.LHS.(*parser.NumberLiteral); ok {
-		arithmeticOperation.ConstantOp = true
-		arithmeticOperation.Constant = constant.Val
+		binaryOperation.ConstantOp = true
+		binaryOperation.Constant = constant.Val
 	} else {
 		lhsRequest, _, lhsQueryArth, err = parsePromQLQuery(expr.LHS.String(), timeRange.StartEpochSec, timeRange.EndEpochSec, myid)
 		if err != nil {
@@ -787,23 +795,23 @@ func handleBinaryExpr(expr *parser.BinaryExpr, mQueryReqs []*structs.MetricsQuer
 		}
 		if len(lhsRequest) > 0 {
 			lhsIsVector = true
-			arithmeticOperation.LHS = lhsRequest[0].MetricsQuery.QueryHash
+			binaryOperation.LHS = lhsRequest[0].MetricsQuery.QueryHash
 		}
 		if len(lhsQueryArth) > 0 {
-			arithmeticOperation.LHSExpr = lhsQueryArth[0]
+			binaryOperation.LHSExpr = lhsQueryArth[0]
 		}
 	}
 
 	if constant, ok := expr.RHS.(*parser.NumberLiteral); ok {
-		if arithmeticOperation.ConstantOp {
+		if binaryOperation.ConstantOp {
 			// This implies that both LHS and RHS are constants.
-			arithmeticOperation.RHSExpr = &structs.QueryArithmetic{
+			binaryOperation.RHSExpr = &structs.QueryArithmetic{
 				ConstantOp: true,
 				Constant:   constant.Val,
 			}
 		} else {
-			arithmeticOperation.ConstantOp = true
-			arithmeticOperation.Constant = constant.Val
+			binaryOperation.ConstantOp = true
+			binaryOperation.Constant = constant.Val
 		}
 	} else {
 		rhsRequest, _, rhsQueryArth, err = parsePromQLQuery(expr.RHS.String(), timeRange.StartEpochSec, timeRange.EndEpochSec, myid)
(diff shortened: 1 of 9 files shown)

### CODE REVIEW

File: pkg/integrations/prometheus/promql/parser.go [hunk 1]
```diff
@@ -771,81 +771,83 @@ func handleBinaryExpr(expr *parser.BinaryExpr, mQueryReqs []*structs.MetricsQuer
...
-	} else if mQueryReqs[0].MetricsQuery.HashedMName == uint64(0) {
-		// This means there is a common group by on multiple metrics separated by operators.
-		// So, we need to append the aggregations to each of the Request in LHS and RHS.
-		// And we can discard the current aggregation in the mQueryReqs[0]
-		rhsRequest = appendMetricAggsToTheMQuery(rhsRequest, &mQueryReqs[0].MetricsQuery)
-		mQueryReqs = appendMetricAggsToTheMQuery(lhsRequest, &mQueryReqs[0].MetricsQuery)
-	} else {
+	if mQueryReqs[0].MetricsQuery.HashedMName > 0 || mQueryReqs[0].MetricsQuery.SubsequentAggs != nil {
+		// This means that the current MQueryReqs is not empty and we need to append the new request
 		mQueryReqs = append(mQueryReqs, lhsRequest...)
```
AndrewHess, line 835: nit: I'm not sure why you need to check the `hashedMName`, but `HashedMName == 0` is probably a slightly better way—if we change the type from unsigned to signed, then having `>` would cause issues, but `==` wouldn't

pkg/integrations/prometheus/promql/parser.go
```diff
@@ -771,81 +771,83 @@ func handleBinaryExpr(expr *parser.BinaryExpr, mQueryReqs []*structs.MetricsQuer
...
+		mQueryReqs = lhsRequest
 	}
+
 	mQueryReqs = append(mQueryReqs, rhsRequest...)
 
 	if expr.VectorMatching != nil && len(expr.VectorMatching.MatchingLabels) > 0 {
-		if putils.IsLogicalOperator(arithmeticOperation.Operation) {
-			return []*structs.MetricsQueryRequest{}, []*structs.QueryArithmetic{}, fmt.Errorf("convertPqlToMetricsQuery: Grouping modifiers can only be used for comparison and arithmetic %T", expr)
-		}
+		// TODO: Fix for Logical operators
```
AndrewHess, line 843: nit: might be worth adding a little more info to the TODO, unless you're going to fix this really soon

pkg/integrations/prometheus/utils/operatorutils.go
```diff
@@ -180,7 +180,7 @@ func SetFinalResult(queryOp *structs.QueryArithmetic, finalResult map[string]map
 			}
 		}
 	default:
-		log.Errorf("SetFinalResult: does not support using this operator: %v", queryOp.Operation)
+		log.Debugf("SetFinalResult: does not support using this operator: %v", queryOp.Operation)
 	}
```
AndrewHess, line 184: Why do we want to silently fail? I'd think we'd want to return an error, and log the error somewhere

same comment line 196

pkg/segment/segexecution.go
```diff
@@ -230,17 +232,38 @@ func processQueryArithmeticNodeOp(queryOp *structs.QueryArithmetic, resMap map[u
...
 
 	finalResult := make(map[string]map[uint32]float64)
 
 	resultLHS, leftOk := resMap[queryOp.LHS]
 	resultRHS, rightOk := resMap[queryOp.RHS]
 	swapped := false
 
+	var referenceMetricRes *mresults.MetricsResult
+
+	returnfunc := func(scalarValuePt *float64, err error) (map[string]map[uint32]float64, *float64, error) {
```
AndrewHess, line 249: nits:
1. capitalize (so rename to `returnFunc`)
2. both places you call this, you call with `(nil, nil)` parameters. So maybe you don't need parameters? Or are you still working on this function?

pkg/integrations/prometheus/promql/parser_test.go
```diff
@@ -906,55 +906,25 @@ func Test_parsePromQLQuery_NestedQueries_v8(t *testing.T) {
...
 	assert.Equal(t, mHashedMName1, mQueryReqs[0].MetricsQuery.HashedMName)
 	assert.Equal(t, mHashedMName2, mQueryReqs[1].MetricsQuery.HashedMName)
-	assert.False(t, mQueryReqs[0].MetricsQuery.SelectAllSeries)
-	assert.False(t, mQueryReqs[1].MetricsQuery.SelectAllSeries)
-	assert.True(t, mQueryReqs[0].MetricsQuery.Groupby)
-	assert.True(t, mQueryReqs[1].MetricsQuery.Groupby)
+	assert.True(t, mQueryReqs[0].MetricsQuery.SelectAllSeries)
+	assert.True(t, mQueryReqs[1].MetricsQuery.SelectAllSeries)
+	assert.False(t, mQueryReqs[0].MetricsQuery.Groupby)
+	assert.False(t, mQueryReqs[1].MetricsQuery.Groupby)
```
AndrewHess, line 912: Why don't we want this query
```
sum by (app, proc) ( instance_memory_limit_bytes - instance_memory_usage_bytes )
```
to get parsed to have a groupby?

pkg/integrations/prometheus/promql/parser.go
Macbeth98: `HashedMName` here represents hashed metric name. So if `HashedMName` is there, it means a query exists. But there can be scenario, where subsequent aggs exists but metric name is not there due to path of tree that the parser is walking. And also the other way around, where just metric name is there but no aggs.

pkg/integrations/prometheus/promql/parser.go
Macbeth98: Yes my next one of the two PRs, will be this fix.

pkg/integrations/prometheus/utils/operatorutils.go
Macbeth98: Hmm this error was flooding the logs. Actually I will revert this change. And will fix this in the PR of the above mentioned TODO.

pkg/segment/segexecution.go
Macbeth98: Hmm, yeah I can remove this, initially I though of calling this function by default at all return cases, but I can remove them.

pkg/integrations/prometheus/promql/parser_test.go
Macbeth98: So now, the parser will output two queries and one arithmetic operations. 

Previously the above query was parsed like this 

```
sum by (app, proc) ( instance_memory_limit_bytes) - sum by (app, proc) ( instance_memory_usage_bytes )
```
so this is subtraction between two group by queries, so group by is set as `true`.

But now, this is parsed exactly the same as how the query is. First do the subtraction between two metric queries and then do a group by on the resulting series. So the individual metric queries do not have group by.

PR #2442, opened 2025-04-01
Implemented WAL support for metric names

VasudevRaut:
# Description
Summarize the change.

# Testing

# Checklist:

### DIFF
--- a/pkg/segment/writer/metrics/metricssegment.go
+++ b/pkg/segment/writer/metrics/metricssegment.go
@@ -143,9 +143,9 @@ type MetricsSegment struct {
 }
 
 type mNameWalState struct {
-	metricsName []string           // metric names seen across segment
-	wal         *wal.MetricNameWal // Active WAL file
-	lock        sync.Mutex
+	metricsNames []string           // metric names seen across segment
+	wal          *wal.MetricNameWal // Active WAL file
+	lock         sync.Mutex
 }
 
 /*
@@ -391,7 +391,7 @@ func InitMetricsSegment(orgid int64, mId string) (*MetricsSegment, error) {
 	return &MetricsSegment{
 		mNamesMap: make(map[string]bool, 0),
 		mNameWalState: mNameWalState{
-			metricsName: make([]string, 0),
+			metricsNames: make([]string, 0),
 		},
 		currBlockNum: 0,
 		mBlock: &MetricsBlock{
@@ -518,7 +518,7 @@ func EncodeDatapoint(mName []byte, tags *TagsHolder, dp float64, timestamp uint3
 	if !mSeg.mNamesMap[string(mName)] {
 		mSeg.mNamesMap[string(mName)] = true
 		mSeg.mNameWalState.lock.Lock()
-		mSeg.mNameWalState.metricsName = append(mSeg.mNameWalState.metricsName, string(mName))
+		mSeg.mNameWalState.metricsNames = append(mSeg.mNameWalState.metricsNames, string(mName))
 		mSeg.mNameWalState.lock.Unlock()
 	}
 	mSeg.rwLock.Unlock()
@@ -1977,7 +1977,7 @@ func (mb *MetricsBlock) appendToWALBuffer(timestamp uint32, dp float64, tsid uin
 	defer mb.dpWalState.lock.Unlock()
 
 	if int(mb.dpWalState.dpIdx) >= utils.WAL_BLOCK_FLUSH_SIZE {
-		err := mb.dpWalState.currentWal.AppendDataPoints(mb.dpWalState.dpsInWalMem[0:mb.dpWalState.dpIdx])
+		err := mb.dpWalState.currentWal.Append(mb.dpWalState.dpsInWalMem[0:mb.dpWalState.dpIdx])
 		if err != nil {
 			log.Errorf("AppendWalDataPoint : Failed to append datapoints to WAL: %v", err)
 			return err
@@ -2006,7 +2006,7 @@ func timeBasedWalDPSFlush() {
 		for _, ms := range GetAllMetricsSegments() {
 			ms.mBlock.dpWalState.lock.Lock()
 			if ms.mBlock.dpWalState.dpIdx > 0 {
-				err := ms.mBlock.dpWalState.currentWal.AppendDataPoints(ms.mBlock.dpWalState.dpsInWalMem[0:ms.mBlock.dpWalState.dpIdx])
+				err := ms.mBlock.dpWalState.currentWal.Append(ms.mBlock.dpWalState.dpsInWalMem[0:ms.mBlock.dpWalState.dpIdx])
 				if err != nil {
 					log.Warnf("timeBasedWalDPSFlush : Failed to append datapoints to WAL: %v", err)
 				}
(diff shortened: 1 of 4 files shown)


pkg/segment/writer/metrics/wal/wal.go
```diff
@@ -333,3 +337,202 @@ func (it *WalIterator) decodeWALBlock(blockBuf *bytes.Reader) error {
...
+				|-- [Metric Name] {
+					StringLength:  2 Bytes   // Length of the metric name (uint16)
+					StringData:    Variable  // Actual metric name (UTF-8 encoded)
+				}
+	}
+
+	Multiple such blocks are appended continuously.
+*/
+
+func (w *WAL) AppendMName(mName []string) error {
```
AndrewHess, line 380: nit: since this is a list
```suggestion
func (w *WAL) AppendMName(mNames []string) error {
```

pkg/segment/writer/metrics/wal/wal.go
```diff
@@ -333,3 +337,202 @@ func (it *WalIterator) decodeWALBlock(blockBuf *bytes.Reader) error {
...
+		return err
+	}
+
+	_, err = w.fd.Write(w.encodedBuf)
+	if err != nil {
+		log.Errorf("AppendMName: failed to write block content of size %d to WAL file: %v", len(w.encodedBuf), err)
+		return err
+	}
+
+	w.totalDps += uint32(len(mName))
```
AndrewHess, line 409: nit: now the name `totalDps` doesn't quite make sense, since we're storing metric names in this WAL

pkg/segment/writer/metrics/wal/wal.go
```diff
@@ -333,3 +337,202 @@ func (it *WalIterator) decodeWALBlock(blockBuf *bytes.Reader) error {
...
+		it.currentIndex++
+		return &it.readMetricsNames[it.currentIndex-1], nil
+	}
+
+	var blockSize uint32
+	err := binary.Read(it.fd, binary.LittleEndian, &blockSize)
+	if errors.Is(err, io.EOF) {
+		return nil, nil
+	} else if err != nil {
+		log.Printf("MNameWalIterator Next: failed to read block size from WAL file: %v", err)
```
AndrewHess, line 452: We don't use `log.Printf`—use Errorf/Warnf/Infof as appropriate

same comment elsewhere

pkg/segment/writer/metrics/metricssegment.go
```diff
@@ -137,6 +139,12 @@ type MetricsSegment struct {
 	rwLock          *sync.RWMutex   // read write lock for access
 	datapointCount  uint64          // total number of datapoints across all series in the block
 	Orgid           int64
+	mNameWalState   mNameWalState
+}
+
+type mNameWalState struct {
+	walMetricsName []string // metric names seen across segment
+	currentWal     *wal.WAL // Active WAL file
```
AndrewHess, line 147: minor naming nits:
1. `walMetricsName` should be `metricNames`; from the context of the struct name, we already know it's for a WAL
2. `currentWal` can be `wal`, since you're not storing any other WALs in this struct

pkg/segment/writer/metrics/metricssegment.go
```diff
@@ -234,6 +243,7 @@ func initOrgMetrics(orgid int64) error {
 			return err
 		}
 		err = mSeg.mBlock.initNewWAL()
+		err = mSeg.initNewMNameWAL()
```
AndrewHess, line 246: nit: since you have two types of WALs now, maybe `initNewWAL` should be renamed

P.S. your IDE should make it really easy to rename something in one place and every other place it's used gets automatically updated to the new name

pkg/segment/writer/metrics/metricssegment.go
```diff
@@ -2044,3 +2066,183 @@ func (mb *MetricsBlock) deleteWALFiles() {
...
+func timeBasedMetricsNameWalFlush() {
+	for {
+		time.Sleep(METRICS_NAME_WAL_FLUSH_SLEEP_DURATION * time.Second)
+		for _, ms := range GetAllMetricsSegments() {
+			if len(ms.mNameWalState.walMetricsName) > 0 {
+				err := ms.mNameWalState.currentWal.AppendMName(ms.mNameWalState.walMetricsName)
+				if err != nil {
+					log.Warnf("timeBasedMetricsNameWalFlush : Failed to append datapoints to WAL: %v", err)
+				}
+				ms.mNameWalState.walMetricsName = ms.mNameWalState.walMetricsName[:0]
```
AndrewHess, line 2105: If we get `err != nil`, this will log a warning but then delete the data—that's not good. If we get data corruption/loss, we should `Errorf` instead of `Warnf`. I'd suggest you keep it as `Warnf`, but skip line 2105 when we get an error

pkg/segment/writer/metrics/metricssegment.go
```diff
@@ -2044,3 +2066,183 @@ func (mb *MetricsBlock) deleteWALFiles() {
...
+
+	for _, file := range files {
+		if file.IsDir() {
+			continue
+		}
+
+		fileName := file.Name()
+
+		// Expected WAL filename format: "shardId_<shard>_segId_<segID>_.wal"
+		parts := strings.Split(fileName, "_")
```
AndrewHess, line 2155: Nice comment

pkg/segment/writer/metrics/metricssegment.go
```diff
@@ -2044,3 +2066,183 @@ func (mb *MetricsBlock) deleteWALFiles() {
...
+			err = deleteWalFile(mNameWalDir, walFileName)
+			if err != nil {
+				log.Warnf("RecoverMNameWALData : Failed to delete wal file %s: %v", walFileName, err)
+			}
+		}
+
+		if !isWalFileEmpty {
+			err := ms.FlushMetricNames()
+			if err != nil {
+				log.Warnf("RecoverMNameWALData :Failed to flush Metrics Name for shardID=%s, segID=%d,: %v",
```
AndrewHess, line 2240: CICD is failing here. This `%s` should either be `%d` (for numbers) or `%v` (for any type)

pkg/segment/writer/metrics/wal/wal.go
AndrewHess, line 1: I find it odd that you can append datapoints (existing method) and metric names (new method) to the same WAL, but as far as I understand, your intention is to keep separate WALs for this; and if someone did append both of these to one WAL, I'm not sure they'd be able to read the elements correctly without knowing which type the next block was for.

If it's relatively quick/easy to do, it might make sense to have two different WAL struct types, or make a WAL interface and two implementations of that, or make a WAL type with the core functionality and two more for the specifics of DPs vs metrics, or ...

pkg/segment/writer/metrics/metricssegment.go
```diff
@@ -143,8 +143,9 @@ type MetricsSegment struct {
 }
 
 type mNameWalState struct {
-	walMetricsName []string // metric names seen across segment
-	currentWal     *wal.WAL // Active WAL file
+	metricsName []string           // metric names seen across segment
```
AndrewHess, line 146: nit: 
```suggestion
	metricNames []string           // metric names seen across segment
```

pkg/segment/writer/metrics/wal/wal.go
```diff
@@ -377,40 +416,59 @@ File Format:
 	Multiple such blocks are appended continuously.
 */
 
-func (w *WAL) AppendMName(mName []string) error {
+func (w *MetricNameWal) AppendMNames(mName []string) error {
```
AndrewHess, line 419: nit: now that this is a method on `MetricNameWal`, you could rename the function to simply `Append` (unless you plan on having other types of appends for this WAL type later). You can also keep what you have if you prefer

similar comment for `AppendDataPoints()`

AndrewHess (commented):
A few minor nits

AndrewHess (approved):

AndrewHess:
@VasudevRaut your code is failing the linter. You should run `make lint` (or `make pr`, which runs the linter and other things) before pushing code

AndrewHess (approved):

PR #2478, opened 2025-04-08
Remove empty columns,value send to the UI

VasudevRaut:
# Description
✅ Removed empty columns being sent to the UI

✅ Fixed failing E2E tests

# Testing

# Checklist:


pkg/segment/query/segquery.go
```diff
@@ -147,6 +147,16 @@ func GetNodeAndQueryTypes(sNode *structs.SearchNode, aggs *structs.QueryAggregat
...
 }
 
+func IsLogsQuery(firstAgg *structs.QueryAggregators) bool {
+	for curAgg := firstAgg; curAgg != nil; curAgg = curAgg.Next {
+		_, qType := GetNodeAndQueryTypes(&structs.SearchNode{}, curAgg)
+		if qType != structs.RRCCmd {
+			return false
+		}
+	}
+	return true
```
AndrewHess, line 157: I'm not sure this does what you want. For example for a query like `weekday=Monday | eval x=latency/100 | stats avg(x) ` I think `IsLogsQuery()` would return false, but is that what you want?

You should add unit tests for this function

AndrewHess (commented):
1. Why do we only want to do remove empty values for logs queries? Why not also for stats queries?
2. You should add unit tests. e.g., if you add tests for `AsResult()` it will both make it more clear what this PR is intending to do, and help ensure we don't regress on this later

pkg/segment/query/processor/queryprocessor_test.go
AndrewHess, line 1: FYI: these tests would be a bit more clear if you gave the query instead of the AST structure. For example, see `colusage_test.go`. What you have will work though

AndrewHess (approved):

PR #2484, opened 2025-04-09
Add helper to read/write files in checksummed chunks

AndrewHess:
# Description
See title. The purpose is to detect when we read corrupted files. The reader is also backward compatible

# Testing
New unit tests

# Checklist:

### DIFF
--- a/pkg/utils/checksumfile.go
+++ b/pkg/utils/checksumfile.go
@@ -101,10 +101,13 @@ func (csf *checksumFile) ReadAt(buf []byte, offset int64) (int, error) {
 			offset, csf.fd.Name(), err)
 	}
 
-	if magic, err := readUint32(csf.fd); err != nil {
+	magic, err := readUint32(csf.fd)
+	if err != nil {
 		return 0, fmt.Errorf("checksumFile.ReadAt: Cannot read magic number from file %v, err=%v",
 			csf.fd.Name(), err)
-	} else if magic != magicNumber {
+	}
+
+	if magic != magicNumber {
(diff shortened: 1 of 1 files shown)


pkg/utils/checksumfile.go
```diff
@@ -0,0 +1,165 @@
...
+	_, err := csf.fd.Seek(offset, 0)
+	if err != nil {
+		return 0, fmt.Errorf("checksumFile.ReadAt: Cannot seek to offset %d in file %v, err=%v",
+			offset, csf.fd.Name(), err)
+	}
+
+	if magic, err := readUint32(csf.fd); err != nil {
+		return 0, fmt.Errorf("checksumFile.ReadAt: Cannot read magic number from file %v, err=%v",
+			csf.fd.Name(), err)
+	} else if magic != magicNumber {
```
nkunal, line 107: nit: this does not need to be in the else part ? since we have a return at the above if block

nkunal (commented):
looks fine. One minor comment

pkg/utils/checksumfile.go
AndrewHess: I'm not sure what you mean

AndrewHess: Do you mean change it to this:
```
magic, err := readUint32(csf.fd)
if err != nil {
	return 0, fmt.Errorf("checksumFile.ReadAt: Cannot read magic number from file %v, err=%v",
		csf.fd.Name(), err)
}

if magic != magicNumber {
```

nkunal (approved):

PR #2504, opened 2025-04-12
code simplification, mem improvements

nkunal:
# Description
- Remove unused byte param
- Use string interning to reduce repeated creation of cname string

# Testing

# Checklist:

### DIFF
--- a/pkg/segment/reader/microreader/microreader.go
+++ b/pkg/segment/reader/microreader/microreader.go
@@ -116,8 +116,7 @@ func ReadBlockSummaries(fileName string) ([]*structs.BlockSummary,
 		offset += 2
 		bmh := &structs.BlockMetadataHolder{
 			BlkNum:            blkNum,
-			ColumnBlockOffset: make(map[string]int64, numCols),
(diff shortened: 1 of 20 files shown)


pkg/segment/reader/microreader/microreader.go
```diff
@@ -138,7 +154,100 @@ func ReadBlockSummaries(fileName string, rbuf []byte) ([]*structs.BlockSummary,
 		blockSummaries = append(blockSummaries, blkSumm)
 	}
 
-	return blockSummaries, allBmh, rbuf, nil
+	return blockSummaries, allBmh, nil
+}
+
+func ReadBlockSumOnly(fileName string) ([]*structs.BlockSummary, error) {
```
AndrewHess, line 160: The diff between this function and `ReadBlockSummaries()` only shows ~10 lines that are meaningfully changed. I'd recommend only having `ReadBlockSummaries()` but add a flag for whether the `BlockMetadataHolder`s should be read. That way if we have to change something later (e.g., update error handling), we only have to change one function

AndrewHess (approved):

PR #2513, opened 2025-04-16
-Fix issue with incorrect bucket timestamps

VasudevRaut:
# Description
-Fix issue with incorrect bucket timestamps when custom time range is selected.  
-Updated backend to return timestamps in epoch format instead of formatted strings.

# Testing

# Checklist:


pkg/usageStats/stats.go
```diff
@@ -533,9 +533,9 @@ func readUsageStats(startEpoch, endEpoch time.Time, orgid int64) ([]*ReadStats,
 }
 
 // Calculate total bytesCount,linesCount and return hourly / daily / minute count
-func GetUsageStats(pastXhours uint64, granularity UsageStatsGranularity, orgid int64) (map[string]*ReadStats, error) {
+func GetUsageStats(pastXhours uint64, granularity UsageStatsGranularity, orgid int64, endTs int64) (map[string]*ReadStats, error) {
 
-	endEpoch := time.Now()
+	endEpoch := time.Unix(endTs, 0)
```
nkunal, line 538: seem slike `endTs` is always set to `time.Now` in line 276 of clysterStatsHandler then we do we have to pass it around so many functions, why don't we just the endEpoch to time.Now right here

nkunal (approved):

PR #2546, opened 2025-04-21
Read/write log files in checksummed blocks

VasudevRaut:
# Description

# Testing


pkg/segment/writer/agiletreewriter.go
```diff
@@ -172,8 +176,8 @@ func (stb *StarTreeBuilder) encodeMetadata(strMFd *os.File) (uint32, error) {
 	// Now we know the size of the metadata, so we can write it. The value we
 	// write doesn't include the 4 bytes we use to store the value.
 	// We need to write from the 2nd byte, since the first byte is the file type.
-	_, err = strMFd.WriteAt(utils.Uint32ToBytesLittleEndian(metadataSize-4), 1)
-	if err != nil {
+	binary.LittleEndian.PutUint32(metaBuf.Bytes()[0:4], metadataSize-4)
+	if err := csf.AppendPartialChunk(metaBuf.Bytes()); err != nil {
```
AndrewHess, line 180: You return right after this without flushing the chunk. This might be unexpected for future callers, so I'd recommend either
1. Flushing the chunk—if it makes sense to do so
2. Adding a comment at the top of this method saying the caller is responsible for flushing the chunk.

pkg/segment/reader/segread/segstatsreader.go
AndrewHess, line 1: This doesn't really make sense. You:
1. Read the whole file
2. Open the file again (and I think you forgot to close it)
3. Stat the file to get the size
4. Do this `_, err = csf.Fd.ReadAt(magicNumBuf, 0)` to check if the file is the checksum format or the old format. But I think that will never give you the magic number (I might be wrong about that though) because I think even if it is the checksum format it'll give you the actual data that was encoded, not the metadata like magic number, etc.
5. Discard the whole buffer you read in step 1
6. Read the whole file back into the buffer

I think you can skip step 1, do step 2 (and defer the file close), either stat the file or make a new method like `ReadAt` that also resizes the buffer, and then do step 6. I don't think you need to know if this file is the checksum format or the old format; if you don't, you also don't need the added complexity of knowing the metadata takes up 12 bytes or making the magic number public. If you do need that complexity, you should make a constant (in checksumfile.go for the metadata size) and rename `MagicNumber` to be more specific—but hopefully we just don't need knowledge of the checksum file format in this package

pkg/segment/reader/segread/rollupreader.go
```diff
@@ -166,11 +168,13 @@ func (rur *RollupReader) GetDayRollups() (map[uint16]map[uint64]*writer.RolledRe
...
 	allBlks map[uint16]map[uint64]*writer.RolledRecs, qid uint64) error {
 
+	f := toputils.ChecksumFile{Fd: fd}
+
 	offset := int64(0)
 	var blkNum uint16
 	bbBlkNum := make([]byte, 2) // blkNum (2)
 	var bKey uint64
-	bbBKey := make([]byte, 8) // for bucket key timestamp
+	bbBKey := make([]byte, 9) // for bucket key timestamp + RR_ENC_BITSET
```
AndrewHess, line 177: Looks like `RR_ENC_BITSET` is an enum value. Did you mean `RR_ENC_TYPE`?

pkg/segment/writer/segstore.go
```diff
@@ -1197,7 +1206,7 @@ func (ss *SegStore) flushBloomIndex(cname string, bi *BloomIndex) uint64 {
 		bi.HistoricalCount = bi.HistoricalCount[streamIdHistory-utils.BLOOM_SIZE_HISTORY:]
 
 	}
-	return uint64(dataBuf.Len())
+	return uint64(dataBuf.Len() + 4)
```
AndrewHess, line 1209: Why plus 4? Probably worth adding a comment in the code

pkg/segment/writer/segstore.go
```diff
@@ -1263,7 +1281,7 @@ func (segstore *SegStore) flushBlockRangeIndex(cname string, ri *RangeIndex) uin
 		return 0
 	}
 	fr.Close()
-	return uint64(packedLen)
+	return uint64(packedLen + 4)
```
AndrewHess, line 1284: Same here

pkg/segment/writer/segstore.go
```diff
@@ -1414,68 +1432,98 @@ func (ss *SegStore) writeWipTsRollups(cname string) error {
...
-		if _, err = fd.Write(toputils.Uint16ToBytesLittleEndian(uint16(mrSize))); err != nil {
+		if err := csf.AppendPartialChunk(toputils.Uint16ToBytesLittleEndian(mrSize)); err != nil {
 			log.Errorf("writeSingleRup: blkNum=%v bkey=%v mrsize write failed fname=%v, err=%v",
 				blkNum, bkey, fname, err)
 			return err
 		}
 
-		// write actual bitset
-		err = cb.WriteTo(fd)
+		err = csf.Flush()
```
AndrewHess, line 1505: nit: since you flush right before this, instead of writing a partial chunk and flushing immediately, you can use the `AppendChunk()` method

Same comment elsewhere

PR #2552, opened 2025-04-22
Gomemlimit config

VasudevRaut:
# Description

# Testing
added configuration in the server.yaml

>   GOMEMLIMIT : 314572800

Ingest the record 
> go run main.go ingest esbulk -n 1 -g benchmark -d http://localhost:8081/elastic -t 40000

run the bash script in the loop

> #!/bin/bash
while true; do
  curl -s -o /dev/null http://localhost:5122/api/search -d '{"state":"query","searchText":"blah *","startEpoch":"now-15d","endEpoch":"now","indexName":"ind-0","queryLanguage":"Splunk QL"}'
  sleep 1
done

Verify if memory increases or decreases, and ensure it stays close to 300MB


pkg/config/config.go
```diff
@@ -1324,6 +1326,25 @@ func SetConfig(config common.Configuration) {
...
+	if os.Getenv("GOMEMLIMIT") != "" {
+		return
+	}
+
+	if cfg == nil || cfg.MemoryConfig.GoMemLimit == 0 {
+		return
+	}
+
+	limit := strconv.FormatUint(cfg.MemoryConfig.GoMemLimit, 10)
+	if err := os.Setenv("GOMEMLIMIT", limit); err != nil {
```
AndrewHess, line 1340: I'm not sure if Go will use the new env variable value, or if it only reads it when the program starts. If you tested this and it works correctly, you can ignore this comment. Otherwise, you might need to use https://pkg.go.dev/runtime/debug#SetMemoryLimit

AndrewHess (commented):
Looks good, but there's one thing I'm not sure about.

Were you able to test this somehow? Your PR description doesn't mention testing

pkg/config/config.go
VasudevRaut: @AndrewHess    
### Testing

added configuration in the server.yaml

> GOMEMLIMIT : 314572800


Ingest the record

> go run main.go ingest esbulk -n 1 -g benchmark -d http://localhost:8081/elastic -t 40000


run the bash script in the loop

> #!/bin/bash
> while true; do
> curl -s -o /dev/null http://localhost:5122/api/search -d '{"state":"query","searchText":"blah *","startEpoch":"now-15d","endEpoch":"now","indexName":"ind-0","queryLanguage":"Splunk QL"}'
> sleep 1
> done

Verify if memory increases or decreases, and ensure it stays close to 300MiB

AndrewHess: And you also check that if you don't set GOMEMLIMIT in the config, then the memory goes above 300MiB with that workflow, right?

AndrewHess (dismissed):

PR #2561, opened 2025-04-24
segreader now uses pooled memory for raw block buffer

nrkawane:
# Description
Use of fixed length buffer is replaced with flexible buffer pool. 

Step 1. Run siglens  
`go run cmd/siglens/main.go`

Step 2. Ingest logs (run the tools/sigclient)
`go run main.go ingest esbulk --minColumns 150 -p 4 -t 100000 -d http://localhost:8081/elastic`

Step 3. Run query for more than 1min (ctrl + c to quit)
`while true; do curl -s -o /dev/null  http://localhost:5122/api/search -d '{"state":"query","searchText":"blah*","startEpoch":"now-15d","endEpoch":"now","indexName":"*","queryLanguage":"Splunk QL"}'; sleep 1; done`

Step 4. Run profiler 
`go tool pprof http://localhost:5122/debug/pprof/heap`

Observe the following output 
```
Showing top 10 nodes out of 57
      flat  flat%   sum%        cum   cum%
  647.64MB 57.39% 57.39%   647.64MB 57.39%  github.com/siglens/siglens/pkg/segment/reader/segread/segreader.glob..func1
     256MB 22.68% 80.07%      256MB 22.68%  github.com/klauspost/compress/zstd.(*fastBase).ensureHist (inline)
  170.60MB 15.12% 95.19%   170.60MB 15.12%  github.com/siglens/siglens/pkg/utils.ResizeSlice[go.shape.uint8] (inline)
   19.16MB  1.70% 96.89%    19.16MB  1.70%  github.com/klauspost/compress/zstd.encoderOptions.encoder
    3.84MB  0.34% 97.23%   259.84MB 23.02%  github.com/klauspost/compress/zstd.(*doubleFastEncoder).Encode
         0     0% 97.23%    70.55MB  6.25%  github.com/fasthttp/router.(*Router).Handler
         0     0% 97.23%   281.20MB 24.92%  github.com/klauspost/compress/zstd.(*Encoder).EncodeAll
         0     0% 97.23%    19.16MB  1.70%  github.com/klauspost/compress/zstd.(*Encoder).initialize
         0     0% 97.23%      256MB 22.68%  github.com/klauspost/compress/zstd.(*fastBase).addBlock
         0     0% 97.23%     5.69MB   0.5%  github.com/siglens/siglens/cmd/startup.startQueryServer.func1

```

Step 5. Additional details 
```
(pprof) list segreader.glob..func1
Total: 795.43MB
ROUTINE ======================== github.com/siglens/siglens/pkg/segment/reader/segread/segreader.glob..func1 in /home/nilesh/src/siglens/pkg/segment/reader/segread/segreader/segreader.go
  416.76MB   416.76MB (flat, cum) 52.39% of Total
         .          .     38:   New: func() interface{} {
         .          .     39:           // The Pool's New function should generally only return pointer
         .          .     40:           // types, since a pointer can be put into the return interface
         .          .     41:           // value without an allocation:
  416.76MB   416.76MB     42:           slice := make([]byte, 0, utils.WIP_SIZE)
         .          .     43:           return &slice
         .          .     44:   },
         .          .     45:}
         .          .     46:
         .          .     47:var FileReadBufferPool = sync.Pool{

```

After the fix is applied, repeat steps 1 - 4 and observe the following profiler output
```
Dropped 84 nodes (cum <= 3.16MB)
Showing top 10 nodes out of 54
      flat  flat%   sum%        cum   cum%
  258.19MB 40.82% 40.82%   258.19MB 40.82%  github.com/siglens/siglens/pkg/memorypool.newItem
     256MB 40.47% 81.29%      256MB 40.47%  github.com/klauspost/compress/zstd.(*fastBase).ensureHist (inline)
   58.70MB  9.28% 90.57%    58.70MB  9.28%  github.com/siglens/siglens/pkg/utils.ResizeSlice[go.shape.uint8]
   20.53MB  3.24% 93.81%    20.53MB  3.24%  github.com/klauspost/compress/zstd.encoderOptions.encoder
    7.12MB  1.13% 94.94%     7.12MB  1.13%  github.com/siglens/siglens/pkg/segment/results/blockresults.InitBlockResults
    4.80MB  0.76% 95.69%     4.80MB  0.76%  github.com/bits-and-blooms/bitset.New
    4.06MB  0.64% 96.34%     4.06MB  0.64%  github.com/siglens/siglens/pkg/utils.glob..func1
    2.82MB  0.45% 96.78%     3.32MB  0.52%  github.com/klauspost/compress/zstd.(*blockDec).decodeLiterals
    1.92MB   0.3% 97.09%   257.92MB 40.77%  github.com/klauspost/compress/zstd.(*doubleFastEncoder).Encode
         0     0% 97.09%     4.28MB  0.68%  github.com/bits-and-blooms/bloom/v3.(*BloomFilter).Copy
```


pkg/segment/reader/segread/segreader/segreader.go
```diff
@@ -34,25 +36,50 @@ import (
...
+	pool4K   = memorypool.NewMemoryPool(0, S_4_KB)
+	pool32K  = memorypool.NewMemoryPool(0, S_32_KB)
+	pool64K  = memorypool.NewMemoryPool(0, S_64_KB)
+	pool128K = memorypool.NewMemoryPool(0, S_128_KB)
+	pool256K = memorypool.NewMemoryPool(0, S_256_KB)
+	pool512K = memorypool.NewMemoryPool(0, S_512_KB)
+	pool1M   = memorypool.NewMemoryPool(0, S_1_MB)
+	pool2M   = memorypool.NewMemoryPool(0, S_2_MB)
+
+	UncompressedReadBufferPool = sync.Pool{
```
nkunal, line 63: you can remove this pool now

pkg/segment/reader/segread/segreader/segreader.go
```diff
@@ -169,13 +247,21 @@ func (sfr *SegmentFileReader) Close() error {
...
 }
 
-func (sfr *SegmentFileReader) ReturnBuffers() {
-	UncompressedReadBufferPool.Put(&sfr.currRawBlockBuffer)
+func (sfr *SegmentFileReader) ReturnBuffers() error {
+	// UncompressedReadBufferPool.Put(&sfr.currRawBlockBuffer)
 	FileReadBufferPool.Put(&sfr.currFileBuffer)
+	err := PutBufToPool(sfr.currRawBlockBuffer)
+	if err != nil {
+		return fmt.Errorf("Error putting Raw Block Buffer back to pool")
```
nkunal, line 262: nit: instead of `fmt.Errorf("Error putting Raw Block Buffer back to pool")` I would do 
`fmt.Errorf("ReturnBuffers: Error putting Raw Block Buffer back to pool, err: %v", err)`

pkg/segment/reader/segread/segreader/segreader.go
```diff
@@ -226,6 +312,7 @@ func (sfr *SegmentFileReader) loadBlockUsingBuffer(blockNum uint16) (bool, error
 	if cOffAndLen.Length == 0 {
 		return false, fmt.Errorf("SegmentFileReader.loadBlockUsingBuffer: offset was 0, colName: %v, cnameIdx: %v, fname: %v", sfr.ColName, cnameIdx, sfr.fileName)
 	}
+	sfr.currRawBlockBuffer = GetBufFromPool(int64(cOffAndLen.Length))
```
nkunal, line 315: the `loadBlockUsing..` can be called multiple times. so you should fist check if `currRawBlockBuffer` is not nil, then check the size of it and see if it can fit the cOffAndLen.Lnegth, if not then return that buffer and get a new one.
otherwise we would leak

pkg/segment/reader/segread/segreader/segreader.go
nrkawane: > you can remove this pool now

So we typically dont need larger size buffers - 1M, 2M ?

nkunal: depends on the dataset the customer is sending, the max for this use case will never exceed WIP_LIMIT, which is at 2 MB

pkg/segment/reader/segread/multicolreader.go
```diff
@@ -353,11 +351,13 @@ func (mcsr *MultiColSegmentReader) ExtractValueFromColumnFile(colKeyIndex int, b
...
 
 func (mcsr *MultiColSegmentReader) returnBuffers() {
-
 	if mcsr.allFileReaders != nil {
 		for _, reader := range mcsr.allFileReaders {
 			if reader != nil {
-				reader.ReturnBuffers()
+				err := reader.ReturnBuffers()
+				if err != nil {
+					log.Errorf("returnBuffers: Error returning buffer back to memory: %v", err)
```
nkunal, line 359: nit: I would put `MultiColSegmentReader.returnBuffers: ...` as the log line to better more clarity. But you can keep what you have

pkg/segment/reader/segread/segreader/segreader.go
```diff
@@ -34,25 +36,52 @@ import (
...
+	pool4K   = memorypool.NewMemoryPool(0, S_4_KB)
+	pool32K  = memorypool.NewMemoryPool(0, S_32_KB)
+	pool64K  = memorypool.NewMemoryPool(0, S_64_KB)
+	pool128K = memorypool.NewMemoryPool(0, S_128_KB)
+	pool256K = memorypool.NewMemoryPool(0, S_256_KB)
+	pool512K = memorypool.NewMemoryPool(0, S_512_KB)
+	pool1M   = memorypool.NewMemoryPool(0, S_1_MB)
+	pool2M   = memorypool.NewMemoryPool(0, S_2_MB)
+
+	UncompressedReadBufferPool = sync.Pool{
```
nkunal, line 65: you should remove this pool now, since is not used

pkg/segment/reader/segread/segreader/segreader.go
```diff
@@ -34,25 +36,52 @@ import (
...
+		New: func() interface{} {
+			// The Pool's New function should generally only return pointer
+			// types, since a pointer can be put into the return interface
+			// value without an allocation:
+			slice := make([]byte, 0, utils.WIP_SIZE)
+			return &slice
+		},
+	}
+
+	FileReadBufferPool = sync.Pool{
```
nkunal, line 75: you should remove this as well

pkg/segment/reader/segread/segreader/segreader.go
```diff
@@ -83,6 +112,57 @@ type SegmentFileReader struct {
...
+	case S_256_KB:
+		return pool256K.Put(buf)
+	case S_512_KB:
+		return pool512K.Put(buf)
+	case S_1_MB:
+		return pool1M.Put(buf)
+	case S_2_MB:
+		return pool2M.Put(buf)
+	default:
+		// TODO - log error??
```
nkunal, line 161: this is not an error, since we have been asked for a larger pool, we took from the pool by make cmd and while returning, we let GC take care of it

pkg/segment/reader/segread/segreader/segreader.go
```diff
@@ -154,7 +234,7 @@ func InitNewSegFileReader(fd *os.File, colName string, allBlocksToSearch map[uin
 		allBlocksToSearch:     allBlocksToSearch,
 		currOffset:            0,
 		currFileBuffer:        *FileReadBufferPool.Get().(*[]byte),
```
nkunal, line 236: you still need to replace this with the pool

pkg/segment/reader/segread/segreader/segreader.go
```diff
@@ -169,13 +249,21 @@ func (sfr *SegmentFileReader) Close() error {
...
+	if err != nil {
+		return err
+	}
 	return sfr.currFD.Close()
 }
 
-func (sfr *SegmentFileReader) ReturnBuffers() {
-	UncompressedReadBufferPool.Put(&sfr.currRawBlockBuffer)
+func (sfr *SegmentFileReader) ReturnBuffers() error {
+	// UncompressedReadBufferPool.Put(&sfr.currRawBlockBuffer)
```
nkunal, line 260: remove commented code

pkg/segment/reader/segread/segreader/segreader.go
```diff
@@ -83,6 +112,57 @@ type SegmentFileReader struct {
 	allBmi             *structs.AllBlksMetaInfo
 }
 
+func GetBufFromPool(size int64) []byte {
```
nkunal, line 115: the GetBuf.. and PutBuf.. dont need to be public since they are not used outside of this module

pkg/segment/reader/segread/segreader/segreader.go
```diff
@@ -169,13 +282,31 @@ func (sfr *SegmentFileReader) Close() error {
...
+	}
 	return sfr.currFD.Close()
 }
 
-func (sfr *SegmentFileReader) ReturnBuffers() {
-	UncompressedReadBufferPool.Put(&sfr.currRawBlockBuffer)
-	FileReadBufferPool.Put(&sfr.currFileBuffer)
+func (sfr *SegmentFileReader) ReturnBuffers() error {
+	var errorMessages []string
+	err := PutBufToPool(sfr.currFileBuffer)
```
nkunal, line 294: nit: if someone just constucts the `InitSegmentReader` but never ends up calling `loadBlockUsing..` which actually initializes this buffer ptrs then these ptrs would be nil, and we may log an error. so perhaps we should if it is not nil, only then call the Put method

pkg/segment/reader/segread/segreader/segreader.go
```diff
@@ -169,13 +282,31 @@ func (sfr *SegmentFileReader) Close() error {
...
+	if err != nil {
+		errorMessages = append(errorMessages, fmt.Sprintf("Segreader.ReturnBuffers: Error putting buffer back to pool, err: %v", err))
+	}
+
+	err = PutBufToPool(sfr.currRawBlockBuffer)
```
nkunal, line 299: same comment here

pkg/segment/reader/segread/timereader.go
```diff
@@ -229,9 +236,15 @@ func (trr *TimeRangeReader) Close() error {
 }
 
 func (trr *TimeRangeReader) returnBuffers() {
-	segreader.UncompressedReadBufferPool.Put(&trr.blockUncompressedBuffer)
-	segreader.FileReadBufferPool.Put(&trr.blockReadBuffer)
 	rawTimestampsBufferPool.Put(&trr.blockTimestamps)
+	err := segreader.PutBufToPool(trr.blockReadBuffer)
```
nkunal, line 240: same comment as earlier about nil ptr

pkg/segment/reader/segread/timereader.go
```diff
@@ -229,9 +236,15 @@ func (trr *TimeRangeReader) Close() error {
...
+	if err != nil {
+		log.Errorf("TimeReader.returnBuffers: Error putting buffer back to pool, err: %v", err)
+	}
+	err = segreader.PutBufToPool(trr.blockUncompressedBuffer)
```
nkunal, line 244: same comment here

nkunal (commented):
Looks great, just a couple of minor comments.
Also in the PR description could you add :
1) The test cmds (ingestion and query) that you ran
2) Before this change:  profiling mem usage
3) After this change:  profiling mem usage

pkg/segment/reader/segread/segreader/segreader.go
nrkawane: There are are multiple paths to PutBuftoPool (from other modules as well) that would return nil if the input is nil (buffer not initialized. That should result in the expected behavior (not showing error for un-initialized buffer)

pkg/segment/reader/segread/timereader.go
```diff
@@ -418,9 +425,8 @@ func ReadAllTimestampsForBlock(blkNums map[uint16]struct{}, segKey string,
 				break
 			}
 		}
-		buffer := *segreader.UncompressedReadBufferPool.Get().(*[]byte)
+		buffer := segreader.GetBufFromPool(int64(segreader.COMPRESSION_FACTOR * blkLen))
 		rawChunk, err := readChunkFromFile(fd, buffer, blkLen, firstBlkOff)
```
AndrewHess, line 429: Why do you multiply by `COMPRESSION_FACTOR` here? It looks like you get a buffer of size `COMPRESSION_FACTOR*blkLen`, then read `blkLen` bytes into it, then iterate through some of it (I think only up to `blkLen`, but I'm not totally sure) and then put the buffer back in the pool

pkg/segment/reader/segread/timereader.go
```diff
@@ -434,6 +440,11 @@ func ReadAllTimestampsForBlock(blkNums map[uint16]struct{}, segKey string,
 			allReadJob <- &timeBlockRequest{tsRec: rawBlock, blkNum: currBlk, numRecs: numRecs}
 			readOffset += readLen
 		}
+		err = segreader.PutBufToPool(rawChunk)
```
AndrewHess, line 443: `rawBlock` (line 440) is a slice of `rawChunk`. You need to wait until the message you put on the channel on line 440 is done with `rawBlock` before you can put `rawChunk` back in the pool, to avoid memory corruption

Before your changes, `rawChunk` was put back in a pool under a `defer` statement, which was safe because it happens after the `readerWG.Wait()` call on line 450

pkg/segment/reader/segread/timereader.go
nrkawane: During test runs it was observed that decoding typically needs reallocation 90% of the time. @nkunal suggested going with a buffer 8x times. That brought down reallocation to almost never.

pkg/segment/reader/segread/timereader.go
nrkawane: Can you suggest a good pattern for handling that without ignoring the error and 'defer'ing it still?

AndrewHess: A few options that come to mind:
1. Once you're done with this `rawChunk` send a special message on the channel. That message indicates to the worker receiving it that it should `PutBufToPool(rawChunk)`—so then you do the PutBuf in a worker and not here.
2. You could keep track of how many messages you put on the channel for each `rawChunk`, then have each worker increment an atomic number when they're done processing messages. Then once that atomic counter reaches a threshold, you know it's safe to `PutBufToPool(rawChunk)`. Not sure what a good way to monitor the counter would be though (maybe callbacks?).

Of those options, I think option 1 will be a bit more clean, but they were just the first things that came to mind; maybe you'll have a better idea.

I'm assuming that once a message on the channel is fully processed, that part of `rawChunk` is never used again. I think that's right, but you should double check.

I'm just realizing you might be asking "how can I defer it but still handle the error". If that's what you're asking, see https://go.dev/play/p/4T3Zpw85Zj2. You could `err = PutBufToPool(rawChunk); log(err)` inside the closure

nrkawane: It was indeed the later. Fixed.

AndrewHess (approved):

PR #2563, opened 2025-04-24
fix for reading dictenc columns when some blocks were missing it

nkunal:
# Description

# Testing


pkg/segment/writer/packer.go
```diff
@@ -1452,22 +1453,17 @@ func EncodeBlocksum(allBmi *AllBlksMetaInfo, bsum *BlockSummary,
 
 	clen := 0
 	numCols := uint16(0)
-	for cname := range allBmi.CnameDict {
+	for cname, cnameIdx := range bmiCnameIdxDict {
+		if bmiColOffLen[cnameIdx].Length == 0 {
```
AndrewHess, line 1457: Might be worth checking `cnameIdx < len(bmiColOffLen)` to avoid a panic. I assume this _should_ never happen

PR #2567, opened 2025-04-25
Fix  Fast Parse Float function 

VasudevRaut:
# Description
Fix  Fast Parse Float function 
fix log output issue

# Testing

### DIFF
--- a/pkg/utils/numutils.go
+++ b/pkg/utils/numutils.go
@@ -29,10 +29,12 @@ func FastParseFloat(val []byte) (float64, error) {
 	}
 
 	var sign float64 = 1
-	var i int
+	i := 0
 
 	if val[i] == '-' {
 		sign = -1
+		i++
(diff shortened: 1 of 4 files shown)


pkg/utils/numutils.go
```diff
@@ -62,9 +64,50 @@ func FastParseFloat(val []byte) (float64, error) {
...
+		}
+
+		multiplier := 1.0
+		for j := 0; j < exp; j++ {
+			if expSign == 1 {
+				multiplier *= 10
+			} else {
+				multiplier /= 10
+			}
+		}
```
AndrewHess, line 104: I think using whatever `math.Pow()` function Go has may be better (it might be optimized, might avoid issues for really large and really small numbers). But I'm not totally sure. What you have is probably good enough

AndrewHess (approved):
Looks good

AndrewHess (approved):

PR #2576, opened 2025-04-27
fix circle ci python 3.8 issue

nkunal:
# Description

The path-filtering orb is not working , it gives a error since it is no longer supported on python 3.8 version.
Temporarily removing the path-filtering orb and letting circle ci run all the time.

# Testing

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,16 +1,127 @@
 version: 2.1
-
-setup: true
-
 orbs:
-  path-filtering: circleci/path-filtering@1.1.0
+  browser-tools: circleci/browser-tools@1.4.4
 
+parameters:
+  run-tests:
+    type: boolean
+    default: true
+
+jobs:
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -1,16 +1,127 @@
...
+
+      - run:
+          name: Run Functional Test
+          command: |
+            sleep 5  # Wait for Siglens to start
+            cd tools/sigclient
+            go run main.go functional -d http://localhost:8081/elastic -f functionalQueries/functionalQueries.yml -q localhost:5122
 
-orbs:
-  path-filtering: circleci/path-filtering@1.1.0
```
AndrewHess, line 6: I think we can fix this issue by reverting this PR and changing this orb version to the latest (1.3.0; see https://circleci.com/developer/orbs/orb/circleci/path-filtering). 1.3.0 uses Python 3.9

PR #2640, opened 2025-05-09
Ignore SSL Cert validation in webhook

nrkawane:
* Configures a http client to ignore ssl certificate validation / check
* Use the http client when verifying webhook config (contact point) and when invoking webhook to send alerts

### DIFF
--- a/pkg/alerts/alertsHandler/alertsHandler.go
+++ b/pkg/alerts/alertsHandler/alertsHandler.go
@@ -298,7 +298,13 @@ func ProcessTestContactPointRequest(ctx *fasthttp.RequestCtx) {
 }
 
 func testWebhookURL(webhookURL string) error {
-	resp, err := http.Get(webhookURL)
+	client := alertutils.GetCertErrorForgivingHttpClient()
+	req, err := http.NewRequest("GET", webhookURL, nil)
+	if err != nil {
+		log.Errorf("testWebhookURL: failed to create a test webhook request - %v", err)
+		return err
+	}
+	resp, err := client.Do(req)
 	if err != nil {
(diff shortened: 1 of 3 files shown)


pkg/alerts/alertsHandler/alertsHandler.go
```diff
@@ -298,7 +298,13 @@ func ProcessTestContactPointRequest(ctx *fasthttp.RequestCtx) {
 }
 
 func testWebhookURL(webhookURL string) error {
-	resp, err := http.Get(webhookURL)
+	client := alertutils.GetCertErrorForgivingHttpClient()
+	req, err := http.NewRequest("GET", webhookURL, nil)
+	if err != nil {
+		log.Errorf("testWebhookURL: failed to create a test webhook request - %v", err)
```
nkunal, line 304: nit: typically we use the format 
```
log.Errorf("....., request, err: %v", err)
```
two things
1) prefix the var name, for e.g. : err 
2) use a colon or equals seperator after the var name instead of a dash, just to keep it uniform across our entire code base. One could argue dash is also fine, but then everywhere we should just use the same separator

pkg/alerts/alertutils/alertutils.go
```diff
@@ -288,3 +290,13 @@ func (alert *AlertDetails) DecodeQueryParamFromBase64() error {
 
 	return nil
 }
+
+func GetCertErrorForgivingHttpClient() *http.Client {
+	return &http.Client{
+		Transport: &http.Transport{
```
nkunal, line 296: can you verify if the user uses `http://some-address` instead of tls, this will still work

nrkawane: Verified.

AndrewHess (approved):

PR #2642, opened 2025-05-09
UI for adding extra headers

VaishRS0624:
# Description
image
Screenshot 2025-05-09 at 7 35 00 PM
Screenshot 2025-05-09 at 7 35 20 PM
Screenshot 2025-05-09 at 7 35 44 PM

# Testing

### DIFF
--- a/cmd/startup/startup.go
+++ b/cmd/startup/startup.go
@@ -353,7 +353,9 @@ func ShutdownSiglensServer(gotSigusr1 bool) {
 	alertsHandler.Disconnect()
 	writer.WaitForSortedIndexToComplete()
 
-	queriesContextCancelFn()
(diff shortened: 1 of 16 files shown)


sonamgupta21 (commented):
1. After saving the headers, there is a large gap between them.
image

2. If I add a new contact type under the same contact point, the buttons to add or save the header and value do not work.
image

3. If I add two webhooks under the same contact point—one with headers and one without—and save and return, I see headers displayed for both in the UI.
4. Maybe, users should not be allowed to add a new header until the previous one is completely filled.

fivemtebex:
Interesting solution to this problem! I've been working on similar challenges recently.

sonamgupta21 (approved):
Looks good.

PR #2645, opened 2025-05-10
Clean up imports

AndrewHess:
# Description
The important changes are in Makefile, .golangci.yml, and playwright-test.yml. Everything else is just renaming.

1. Require common util packages to be imported with specific aliases. See .golangci.yml and the [documentation](https://golangci-lint.run/usage/linters/#importas)
2. Have `make build` compile the PEG file if it changed
3. Have `make lint` show all the lint errors, instead of early exiting


Makefile
```diff
@@ -43,6 +43,7 @@ gofmt :
 	~/go/bin/goimports -w .
 
 %.go: %.peg
+	$(GO) install github.com/mna/pigeon@latest
```
nkunal, line 46: do we always want to install latest or should we pin it to some version ?

PR #2651, opened 2025-05-13
Traces Search Optimization & UI Improvements

sonamgupta21:
# Description
- Search:
  - Initial Fetch: Always fetch page 1 on new search. (backend sends max 50 records per fetch)
  - Infinite Scrolling: Load additional pages on scroll. 

- Dropdown:
  - Dropdown data is now fetched only on click (not on page load), reducing unnecessary calls.
  - Corrected operation retrieval logic based on selected service name.

- Removed timepicker component (as it was used only in search traces and not functioning correctly).
- Improved loading experience by showing “Loading traces…” during new searches.
- Deleted unused download.css.
- Updated styling of the download popup for a more consistent UI.
- Refactored code for better readability.


static/js/search-traces.js
```diff
@@ -16,167 +16,138 @@
...
+        defaultValue: $('#operation-span-name').text() || 'All',
+        dataUpdate: true,
+        clickedHead: async function () {
+            return await fetchColumnValues('name', $('#service-span-name').text());
+        },
+    });
 }
 
+async function fetchColumnValues(column, serviceFilter = null) {
+    let searchText = `SELECT DISTINCT \`${column}\` FROM \`traces\``;
```
nkunal, line 143: not in this PR, but in the next one, we should replace these SQL queries with SPL

nkunal (approved):

PR #2668, opened 2025-05-15
fix query update for websocket when runtimechart is true

VasudevRaut:
# Description
Previously, when performing a histogram search with `runTimechart: true, fields like percent_complete, state, total_events_searched, total_possible_events, qtype, and sortByTimestampAtDefault` were only present inside the timechartUpdate object. As a result, the frontend was unable to access them and displayed these values as undefined.

Fix Applied:
I have updated the backend to promote these fields from timechartUpdate to the root level of the response. This ensures that the frontend can directly access and display them correctly during histogram (timechart) searches.

# Testing


AndrewHess (approved):
What does this fix?

PR #2680, opened 2025-05-16
fix: send empty time interval buckets in histogram backend response

VasudevRaut:
# Description
Previously, missing time intervals were not included in the bucket response sent to the UI. 
Now, the backend generates and sends empty buckets for all expected time intervals, ensuring consistent histogram rendering on the frontend.

# Testing

### DIFF
--- a/pkg/ast/pipesearch/searchHandler.go
+++ b/pkg/ast/pipesearch/searchHandler.go
@@ -24,6 +24,8 @@ import (
 	"strings"
 	"time"
 
+	sutils "github.com/siglens/siglens/pkg/segment/utils"
+
 	"github.com/fasthttp/websocket"
 	"github.com/siglens/siglens/pkg/alerts/alertutils"
 	"github.com/siglens/siglens/pkg/ast/pipesearch/multiplexer"
@@ -499,6 +501,14 @@ func RunQueryForNewPipeline(conn *websocket.Conn, qid uint64, root *structs.ASTN
 			}
 
 			if isAsync {
+				if runTimechartQuery {
+					err = getFinalBuckets(queryStateData.CompleteWSResp.TimechartComplete, root.TimeRange)
+					if err != nil {
+						log.Errorf("RunQueryForNewPipeline: failed to get final buckets err: %v", err)
+						return nil, false, root.TimeRange, err
+					}
+				}
+
 				wErr := conn.WriteJSON(queryStateData.CompleteWSResp)
 				if wErr != nil {
 					log.Errorf("qid=%v, RunQueryForNewPipeline: failed to write json to websocket, err: %v", qid, wErr)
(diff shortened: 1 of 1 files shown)


pkg/ast/pipesearch/searchHandler.go
```diff
@@ -593,6 +603,149 @@ func RunQueryForNewPipeline(conn *websocket.Conn, qid uint64, root *structs.ASTN
 	}
 }
 
+func getFinalBuckets(complete *structs.PipeSearchCompleteResponse, timeRange *dtu.TimeRange) error {
```
nkunal, line 606: 1) This func is 150+ lines, suggest breaking it up
2) This func is not "gettting" anything but rather it is filling up the response, so perhaps name it differently, usually I follow `actionNoun` pattern, for eg:  `fillAllBuckets` or `populateMissingBuckets`
3) nit: rename `complete` to be `completeResp`

pkg/ast/pipesearch/searchHandler.go
```diff
@@ -593,6 +603,149 @@ func RunQueryForNewPipeline(conn *websocket.Conn, qid uint64, root *structs.ASTN
...
+
+	spanOptions, err := structs.GetDefaultTimechartSpanOptions(timeRange.StartEpochMs, timeRange.EndEpochMs, 0)
+	if err != nil {
+		return fmt.Errorf("failed to get default timechart span options: %w", err)
+	}
+
+	interval := spanOptions.SpanLength.Num
+	timeScaler := spanOptions.SpanLength.TimeScalr
+
+	resultMap := make(map[string]*structs.BucketHolder)
```
nkunal, line 616: nit: rename this to be `buckHolderMap` or something else, reslutMap is too generic, and the reviewer/reader has to back up in the code to understand what result map it is

pkg/ast/pipesearch/searchHandler.go
```diff
@@ -593,6 +603,138 @@ func RunQueryForNewPipeline(conn *websocket.Conn, qid uint64, root *structs.ASTN
...
+	case sutils.TMHour:
+		return time.Duration(interval) * time.Hour, nil
+	case sutils.TMDay:
+		return time.Duration(interval) * 24 * time.Hour, nil
+	case sutils.TMMonth:
+		// Month handled separately
+		return 0, nil
+	default:
+		return 0, fmt.Errorf("unsupported time scaler: %v", scaler)
+	}
```
AndrewHess, line 650: This TMMonth handling looks odd to me. I'd suggest either
1. Do it properly
2. Remove it so it falls into the `default` and returns an error

Even if you don't need to handle it in this function because the caller has some `if` for `TMMonth` and this duration isn't used in that case, doing option 1 or 2 keeps the code cleaner. E.g., with option 1, as someone reading the code I just have to remember "getDuration handles seconds through months" instead of "getDuration handles seconds through days, and for months it returns 0 but not an error, so whenever I see getDuration being used I should remember to check if the month case is going to cause issues". And this is going to be harder to remember and much more prone to bugs when we inevitably need to touch this part of the code again

pkg/ast/pipesearch/searchHandler.go
```diff
@@ -593,6 +603,138 @@ func RunQueryForNewPipeline(conn *websocket.Conn, qid uint64, root *structs.ASTN
...
+				return nil, fmt.Errorf("unsupported time scaler: %v", timeScaler)
+			}
+			bucketInterval := strconv.FormatInt(alignedTs.UnixMilli(), 10)
+			bucketHolderMap[bucketInterval] = createEmptyBucket(bucketInterval, completeResp)
+			runningTs = runningTs.Add(duration)
+		}
+	} else {
+		runningTs = time.Date(runningTs.Year(), runningTs.Month(), 1, 0, 0, 0, 0, runningTs.Location())
+		for !runningTs.After(endEpoch) {
+			bucketInterval := strconv.FormatInt(runningTs.UnixMilli(), 10)
```
AndrewHess, line 678: I don't fully understand what's different about this case (TMMonth). Maybe a comment in the code about what we're trying to do would be helpful

PR #2687, opened 2025-05-19
fix(retention): delete corresponding segment directories from S3 duri…

VasudevRaut:
# Description
fix(retention): ensure segment directories are deleted from S3 during retention-based cleanup

Previously, when segments were deleted locally due to retention, the corresponding S3 directories were not removed. This change adds logic to identify and delete those directories from S3, preventing orphaned data and saving storage.

hyperion [PR](https://github.com/sigscalr/hyperion/pull/2919)

# Testing


pkg/retention/retention.go
```diff
@@ -350,7 +349,10 @@ func DeleteSegmentData(segmentsToDelete map[string]*structs.SegMeta) {
 
 		// Delete segment files from s3
 		dirPath := segMetaEntry.SegmentKey
-		filesToDelete := fileutils.GetAllFilesInDirectory(path.Dir(dirPath) + "/")
+		filesToDelete, err := blob.GetAllFilesInDirectory(path.Dir(dirPath) + "/")
+		if err != nil {
+			log.Errorf("DeleteSegmentData: Failed to list files in S3 directory: %v, error: %v", path.Dir(dirPath), err)
```
AndrewHess, line 354: nit: say "blob" or "remote" instead of "S3", since it's not specific to S3

pkg/retention/retention.go
```diff
@@ -406,7 +408,10 @@ func DeleteMetricsSegmentData(mmetaFile string, metricSegmentsToDelete map[strin
 
 		// Delete segment files from s3
 		dirPath := metricsSegmentMeta.MSegmentDir
-		filesToDelete := fileutils.GetAllFilesInDirectory(path.Dir(dirPath) + "/")
+		filesToDelete, err := blob.GetAllFilesInDirectory(path.Dir(dirPath) + "/")
+		if err != nil {
+			log.Errorf("DeleteSegmentData: Failed to list files in S3 directory: %v, error: %v", path.Dir(dirPath), err)
```
AndrewHess, line 413: Same comment

PR #2698, opened 2025-05-20
Fix race condition during suffix file creation for metric segment sha…

VasudevRaut:
# Description
Resolved issue where multiple threads creating new metric segments on bootup could cause suffix file renaming to fail due to concurrent writes.

# Testing

### DIFF
--- a/pkg/segment/writer/suffix/suffix.go
+++ b/pkg/segment/writer/suffix/suffix.go
@@ -22,6 +22,7 @@ import (
 	"fmt"
 	"os"
 	"path"
+	"sync"
 
 	"github.com/siglens/siglens/pkg/config"
 	"github.com/siglens/siglens/pkg/hooks"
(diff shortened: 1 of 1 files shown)


pkg/segment/writer/suffix/suffix.go
```diff
@@ -65,6 +68,9 @@ func writeSuffix(fileName string, entry *entry) error {
 		return err
 	}
 
+	suffixWriteLock.Lock()
+	defer suffixWriteLock.Unlock()
```
AndrewHess, line 72: This is going to cause waiting even in cases when it's not necessary, like calling writeSuffix() on two different files. I think this is going to be ok because we call writeSuffix() relatively infrequently (can you double check that?). But it's probably worth mentioning in the code.

If we need to, we could get around this (e.g., use file locks), but I think that added complexity isn't required

PR #2705, opened 2025-05-21
UI: Add Thousands Separator for Big Number & Numeric Columns

sonamgupta21:
# Description
- Added thousands separator to the Big Number visualization in dashboards.
- Added a configurable formatting button with an option to toggle the use of thousands separators for numeric columns (respect the data type as it is and only format values that are already numbers) of logs screen table. 
    - Default view - no separators
    - Save user preference in cokkie. 

image
image


nkunal (commented):
could we switch the default to be comma-seperated ?

nkunal (approved):

PR #2710, opened 2025-05-21
Delete unnecessary allocations

AndrewHess:
# Description
For `url_c1=http* | timechart span=1m count` we would create a slice of equally-spaced buckets (1 minute apart in this case) for each block. This caused some unnecessary allocations to keep creating this slice. Moreover, we only used the slice to determine which bucket a particular timestamp belongs in. Since the buckets are equally spaced, this PR replaces the slice with a struct, and we can find the correct bucket for a timestamp with simple arithmetic instead of binary search.

# Testing
Added unit tests to ensure no behavior change.

pprof reported about 50% fewer bytes allocated, and the query ran about 5% faster.

### DIFF
--- a/pkg/segment/aggregations/timechartagg.go
+++ b/pkg/segment/aggregations/timechartagg.go
@@ -36,7 +36,9 @@ type scorePair struct {
 	index         int
 }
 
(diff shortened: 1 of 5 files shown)


nkunal (approved):
FANTASTIC !!!

PR #2711, opened 2025-05-21
Reduce allocations for combining buckets

AndrewHess:
# Description
In timechart/stats queries we have multiple buckets whose values get aggregated into corresponding buckets for other segments. This was happening in `Reduce()`, which uses `CValueEnclosure` parameters. This caused unnecessary allocations when assigning the value (e.g., `e.Cval = e.Cval + other.Cval`) because `Cval` is an `interface{}` and Go allocates a new `interface{}` on each assignment (see https://github.com/golang/go/issues/65348#issuecomment-1915376922).

This PR avoids those allocations for some numeric aggregations by using our `Number` type which avoids using `interface{}` but can still handle float and integer values.

Now we have
```
	rawVal sutils.CValueEnclosure // raw value
	number *sutils.Number         // If this is not nil, it's always correct
	dirty  bool                   // If true, "number" is correct but "rawVal" is not
```
The idea behind this is we updated `number` multiple times, and then only copy it over to `rawVal` when we need to use `rawVal`, since writing `rawVal` will cause an allocation when we set `rawVal.CVal`

# Testing
For `url_c1=http* | timechart span=1m count`:
Before: 743 MB, 1.79 s
After: 485 MB, 1.77 s (it's a small difference, but I ran these multiple times and on average it was slightly faster with the changes in this PR)

For `* | timechart span=1m count`
Before: 464 MB, 1.23 s
After: 90 MB, 1.09 s

### DIFF
--- a/pkg/segment/results/blockresults/runningstats.go
+++ b/pkg/segment/results/blockresults/runningstats.go
@@ -26,6 +26,7 @@ import (
 	"github.com/siglens/siglens/pkg/segment/structs"
 	sutils "github.com/siglens/siglens/pkg/segment/utils"
 	"github.com/siglens/siglens/pkg/utils"
+	log "github.com/sirupsen/logrus"
 )
 
 type RunningBucketResults struct {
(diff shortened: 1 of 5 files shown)


pkg/segment/results/blockresults/runningstats.go
```diff
@@ -44,12 +45,31 @@ type SerializedRunningBucketResults struct {
...
+func (rs *runningStats) syncRawValue() {
+	if !rs.dirty || rs.number == nil {
+		return
+	}
+
+	rs.dirty = false
+
+	err := rs.number.ToCVal(&rs.rawVal)
+	if err != nil {
+		log.Error(ErrFailedConversion)
```
nkunal, line 68: Should we log this since this is in fast path ?

nkunal (approved):

nkunal (approved):

PR #2716, opened 2025-05-22
Synthetic traces

vsharma-va:
# Description
Added ability to ingest synthetic traces. 
Endpoint - `otlp/v1/sampletraces`
## Changes ->
1. traces.go - exported the function `HandleTraceIngestionResponse`
2. dataset.go - added two structs - `TraceFaker` (keeps track of `gofaker.Faker`) `SyntheticTraceState` (Going to be used to implement links and get the Dependency Graph working)
3. sampledataset.go - `processSingleTrace()` - Builds a single tree (The tree can be symmetric or asymmetric)
`ProcessSyntheticTraceRequest()` - gets the control from fasthttp and runs goroutines to build multiple traces
`generateSpan()` - creates a randomized span
`generateTraceState()` - creates a randomized trace state
4. `entryHandlers.go` + `server.go` -> changes to make the endpoint available

## Remaining Steps ->

~~1. Generating links between the traces to enable Dependency Graph~~
~~2. Randomize start time and end time for each trace and its spans~~
3. Controlling requests per minute parameter for individual traces

sample
trace sample


pkg/sampledataset/dataset.go
```diff
@@ -32,6 +32,23 @@ type Generator interface {
 	GetRawLog() (map[string]interface{}, error)
 }
 
+type SyntheticTraceState struct {
+	TraceIds *[]string
+	SpanIds  *map[string][]string
```
AndrewHess, line 37: Don't use pointers to slices or maps. Go already uses pointers for these when passing to functions, so you don't need to add another pointer

pkg/sampledataset/dataset.go
```diff
@@ -32,6 +32,23 @@ type Generator interface {
...
+}
+
+type TraceFaker struct {
+	Faker *gofakeit.Faker
+	Seed  int64
+}
```
AndrewHess, line 43: nit: this doesn't look specific to traces, so I'd recommend removing "Trace" from the name. Later, maybe we'll want to reuse this struct for non-trace things, and having it called `TraceFaker` will be misleading in that case

pkg/sampledataset/sampledataset.go
```diff
@@ -68,6 +74,135 @@ func populateActionLines(idxPrefix string, indexName string, numIndices int) []s
...
+
+	f := InitTraceFaker(0)
+	depth := 7
+	width := 2
+	startTime := time.Now().UnixNano()
+	totalDurationOfTrace := int64(time.Millisecond * 100)
+	endTime := startTime + totalDurationOfTrace
+	pleArray := make([]*segwriter.ParsedLogEvent, 0)
+	traceId := strings.ReplaceAll(f.Faker.HexUint32(), "0x", "")
+	log.Errorf("ProcessSyntheticTraceRequest: %v", traceId)
```
AndrewHess, line 98: Remove this log

pkg/sampledataset/sampledataset.go
```diff
@@ -68,6 +74,135 @@ func populateActionLines(idxPrefix string, indexName string, numIndices int) []s
...
+	var jsParsingStackbuf [utils.UnescapeStackBufSize]byte
+
+	idxToStreamIdCache := make(map[string]string)
+	cnameCacheByteHashToStr := make(map[uint64]string)
+
+	f := InitTraceFaker(0)
+	depth := 7
+	width := 2
+	startTime := time.Now().UnixNano()
+	totalDurationOfTrace := int64(time.Millisecond * 100)
```
AndrewHess, line 94: We may want these as configurations. It's probably lower priority, but if we move this to sigclient it's pretty easy to add them as command line args

pkg/sampledataset/sampledataset.go
```diff
@@ -68,6 +74,135 @@ func populateActionLines(idxPrefix string, indexName string, numIndices int) []s
...
+		levelParent[0] = []string{parentSpanId}
+	}
+	for i := 1; i < depth; i++ {
+		for k := range levelParent[i-1] {
+			for j := 0; j < width; j++ {
+				if f.Faker.Rand.Float32() >= 0.9 && !generateSymmetricTree {
+					continue
+				} else {
+					currSpanId := strings.ReplaceAll(f.Faker.HexUint32(), "0x", "")
+					g, _ := generateSpan(traceId, currSpanId, levelParent[i-1][k], serviceName, f, (*spanArray[i-1])["name"].(string), startTime+int64(time.Millisecond*5), endTime-int64(time.Millisecond*10))
```
AndrewHess, line 116: 1. What does generateSymmetricTree mean?
2. There's several magic numbers here (0.9, 5ms, 10ms). Generally you should have a variable for those (or a constant near the top of the file) to help clarify what those numbers mean

AndrewHess (commented):
Most of our tooling for ingesting fake data is in /tools/sigclient. Should this be moved there? Or did we talk about this previously and decide not to have it in sigclient?

vsharma-va:
> Most of our tooling for ingesting fake data is in /tools/sigclient. Should this be moved there? Or did we talk about this previously and decide not to have it in sigclient?

From what I understood from the ticket in [jira](https://sigscalr.atlassian.net/browse/SIG-2737) is that we want a feature that is similar to `Send Sample Test Data` in the Log Ingestion page.

pkg/sampledataset/dataset.go
vsharma-va: Understood, I have removed unnecessary pointers from `SyntheticTraceState` struct

pkg/sampledataset/dataset.go
vsharma-va: yeah, I have renamed it to a more generic name `FakerState`

pkg/sampledataset/sampledataset.go
vsharma-va: Yeah that would be good. I will try and consolidate width and depth into one variable. Because as it exists right now if values are set incorrectly (for eg `depth := 15` `width := 4`) it will generate a massive tree with more than 200,000 spans for every single trace

pkg/sampledataset/sampledataset.go
vsharma-va: 1. `generateSymmetricTree` - the function `processSingleTrace` generates a symmetric tree with height = `depth` value and each node will have children = `width`. So as to introduce randomness/asymmetry (if needed) the variable `generateSymmetricTree` is present. I have added a commit which makes it more obvious with a few comments
2. magic numbers have been moved up and have accompanying comments to explain exactly what they do

pkg/sampledataset/sampledataset.go
```diff
@@ -88,63 +88,118 @@ func processSingleTrace(orgId int64, wg *sync.WaitGroup, generateSymmetricTree b
...
+			sts.SpanIds[traceId] = []string{parentSpanId}
+			sts.ServiceName[traceId] = serviceName
+			sts.StartEndTime[parentSpanId] = []int64{startTime, endTime}
+		}
+		for i := 1; i < depth; i++ {
+			for k := range levelParent[i-1] {
+				for j := 0; j < width; j++ {
+					if 1-f.Faker.Rand.Float32() <= probToDropNode {
+						continue
 					} else {
```
AndrewHess, line 142: nit: `1-rand()` has the same distribution as `rand()`, so I think the `1-` is unnecessary and should be removed because it adds a little complexity for no benefit.

Or let me know if I'm missing something

pkg/sampledataset/sampledataset.go
```diff
@@ -68,6 +74,196 @@ func populateActionLines(idxPrefix string, indexName string, numIndices int) []s
...
+		ServiceName:  make(map[string]string),
+	}
+
+	for n := 0; n < numTracesToGenerate; n++ {
+		wg.Add(1)
+		go processSingleTrace(orgId, &wg, &sts, false)
+	}
+
+	wg.Wait()
+	log.Printf("ProcessSyntheticTraceRequest: All %d trace generations completed.", numTracesToGenerate)
```
AndrewHess, line 228: We don't use `log.Printf`. Change to `Infof`, `Debugf`, or remove this

pkg/sampledataset/sampledataset.go
```diff
@@ -68,6 +74,196 @@ func populateActionLines(idxPrefix string, indexName string, numIndices int) []s
...
+	span["kind"] = "SPAN_KIND_CLIENT"
+	span["start_time"] = parentStartTime
+	span["end_time"] = parentEndTime
+	span["duration"] = span["end_time"].(int64) - span["start_time"].(int64)
+	span["dropped_attributes_count"] = 0
+	span["dropped_events_count"] = 0
+	span["dropped_links_count"] = 0
+	span["status"] = "STATUS_CODE_OK"
+	// TODO ADD COLUMN FOR EACH ATTRIBUTE
+	return &span, nil
```
AndrewHess, line 253: Are you planning on doing this TODO in this PR?

vsharma-va: Yeah, I have implemented it in the commit [778477d](https://github.com/siglens/siglens/pull/2716/commits/778477dc8f1f1c9a72f0cdcdef53f243e09e23fa)

AndrewHess (approved):

PR #2731, opened 2025-05-27
SPL: Implementation for latest_time function

vsharma-va:
# Description

1. segstatsreader.go - added a new function `GetSegLatestTs()` which maintains latest seen timestamp
2. segresults.go - added a new switch case to add LatestTime's result to `segStatsResults` + updated `GetSegmentStatsResult` to handle `SS_DT_UNSIGNED_NUM`
3. searchaggs.go - updated to keep track of latest timestamp for every column
4. segstats.go - new function `AddSegStatsUNIXTime()` to extract latest timestamp can be updated to handle earliest timestamp as well
Added changes to implement latest_time for groupby queries as well


pkg/segment/results/blockresults/runningstats.go
```diff
@@ -166,6 +167,13 @@ func (rr *RunningBucketResults) AddMeasureResults(runningStats *[]runningStats,
 				batchErr.AddError("RunningBucketResults.AddMeasureResults:MinMax", err)
 			}
 			i += step
+		case sutils.LatestTime:
+			isLatestTime := measureFunc == sutils.LatestTime
```
AndrewHess, line 171: nit: `isLatestTime` is always true, so you don't need to compute it. You fell into the `sutils.LatestTime` case of `switch measureFunc`

pkg/segment/results/blockresults/runningstats.go
```diff
@@ -383,7 +391,14 @@ func (rr *RunningBucketResults) ProcessReduce(runningStats *[]runningStats, e su
...
 	} else {
-		(*runningStats)[i].rawVal = retVal
+		// otherwise the frontend adds commas to the epoch time returned by latest_time
+		if aggFunc != sutils.LatestTime {
+			(*runningStats)[i].rawVal = retVal
+		} else {
+			retVal.Dtype = sutils.SS_DT_STRING
+			retVal.CVal = strconv.FormatUint(retVal.CVal.(uint64), 10)
+			(*runningStats)[i].rawVal = retVal
+		}
```
AndrewHess, line 401: I don't like this change. Can we avoid it?

If we do what you have, then for queries with additional commands after the `stats` (e.g., `* | stats earliest(foo) as early | where early<100`) it:
1. may cause some issues because now `early` is a string instead of number
2. will cost some type conversions to evaluate `early<100` if we do that numerically

pkg/segment/results/segresults/segresults.go
```diff
@@ -554,6 +564,15 @@ func (sr *SearchResults) GetSegmentStatsResults(skEnc uint32, humanizeValues boo
...
 			}
+		case sutils.SS_DT_UNSIGNED_NUM:
+			// if uint is returned then the frontend will add commas to it(humanize it?)
+			strVal, err := aggVal.GetString()
+			if err != nil {
+				log.Errorf("GetSegmentStatsResults: failed to convert string slice to string, qid: %v, err: %v", sr.qid, err)
+				bucketHolder.MeasureVal[mfName] = ""
+			} else {
+				bucketHolder.MeasureVal[mfName] = strVal
+			}
```
AndrewHess, line 575: Why is this significantly different than the signed number case a few lines above?
```
		case sutils.SS_DT_SIGNED_NUM:
			if humanizeValues {
				measureVal = humanize.Comma(aggVal.CVal.(int64))
			}
			bucketHolder.MeasureVal[mfName] = measureVal
```

pkg/segment/structs/segstructs.go
```diff
@@ -1461,8 +1462,8 @@ var unsupportedStatsFuncs = map[sutils.AggregateFunctions]struct{}{
 	sutils.Earliest:     {},
 	sutils.EarliestTime: {},
 	sutils.Latest:       {},
-	sutils.LatestTime:   {},
-	sutils.StatsRate:    {},
+	// sutils.LatestTime:   {},
```
AndrewHess, line 1465: nit: remove it instead of commenting it out

pkg/segment/results/segresults/segresults.go
vsharma-va: Same reason as the previous review. Should I just let it be a `uint` and it will be handled in the frontend later?

pkg/segment/results/segresults/segresults.go
```diff
@@ -549,6 +549,8 @@ func (sr *SearchResults) GetSegmentStatsResults(skEnc uint32, humanizeValues boo
 				measureVal = humanize.CommafWithDigits(measureVal.(float64), 3)
 			}
 			bucketHolder.MeasureVal[mfName] = measureVal
+		case sutils.SS_DT_UNSIGNED_NUM:
+			fallthrough
 		case sutils.SS_DT_SIGNED_NUM:
 			if humanizeValues {
 				measureVal = humanize.Comma(aggVal.CVal.(int64))
```
AndrewHess, line 556: I think the `fallthrough` will cause a panic on line 556 during the type assertion; for `SS_DT_UNSIGNED_NUM` I'm pretty sure we store a uint64, not int64

vsharma-va: Yes you are right, but there is no implementation of humanize.Comma for uint64. Should I write one or convert uint64 to int64 (might cause issues?).

AndrewHess: If it's quick/easy to properly handle large uint64s, do that. Otherwise, I think it's fine to cast it to a int64 (I mean, type assert it to a uint64 and then cast that to int64) and perhaps log a message if the uint64 is larger than the max int64

pkg/segment/results/segresults/segresults.go
```diff
@@ -529,6 +529,27 @@ func (sr *SearchResults) GetRemoteInfo(remoteID string, inrrcs []*sutils.RecordR
...
+		case 2:
+			parts[j] = "0" + parts[j]
+		case 1:
+			parts[j] = "00" + parts[j]
+		}
+		v = v / 1000
+		j--
+	}
+	parts[j] = strconv.FormatUint(v, 10)
+	return strings.Join(parts, ",")
```
AndrewHess, line 554: This all looks odd to me. I was expecting `humanizeUints()` to do something like:
```
result = ""
while num > 0:
  prepend comma to result when necessary
  prepend (num % 1000) with leading zeros (except if num<1000 ?) to result
  num /= 1000
```

What you have might be fine, but you should add unit tests for it. Also it's not clear to me why `parts` has 7 elements (max number we handle?), and I think you might end up formatting 12345 as `,,,,,,12,345`, but maybe I'm wrong

AndrewHess (approved):

PR #2734, opened 2025-05-28
Dashboard Folder Management: Rename & Move Support

sonamgupta21:
# Description
Editable Folder Name in UI:
    - Added an edit icon next to the folder name on the Dashboards screen.
    - Clicking the icon turns the folder name into an input field.
    - Users can edit the name and press Enter or click outside to save the changes.
image
image

Move Dashboard Between Folders:
    - Added a dropdown to select and update the folder for a dashboard.
    - User can move a dashboard from one folder to another and save the updated settings.
 
image

Backend Changes:
    - Added a `refreshFolderMetadata()` function:
    - When a folder name is updated, the next time any dashboard from that folder is accessed via `getDashboard()`, the code will detect stale folder metadata and automatically refresh it.

### DIFF
--- a/pkg/server/ingest/server.go
+++ b/pkg/server/ingest/server.go
@@ -67,8 +67,7 @@ func (hs *ingestionServerCfg) Close() {
 }
 
 func (hs *ingestionServerCfg) Run() (err error) {
-
-	//Register all the method handlers here
+	// Register all the method handlers here
 	ingest.InitIngestionMetrics()
 	writer.InitWriterNode()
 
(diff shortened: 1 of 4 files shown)


pkg/dashboards/dashboards.go
```diff
@@ -254,11 +237,68 @@ func getDashboard(id string, myid int64) (map[string]interface{}, error) {
 		return nil, err
 	}
 
+	if err := refreshFolderMetadata(id, detailDashboardInfo, myid); err != nil {
+		log.Warnf("getDashboard: Failed to refresh folder metadata for dashboard %s: %v", id, err)
```
nkunal, line 241: nit: prefix var names before their values.  a better msg would be :
```
... for dashboard, id: %v, err: %v", id, err)
```

pkg/dashboards/dashboards.go
```diff
@@ -254,11 +237,68 @@ func getDashboard(id string, myid int64) (map[string]interface{}, error) {
...
+	if !exists {
+		log.Warnf("refreshFolderMetadata: dashboard %s not found in structure", id)
+		return nil
+	}
+
+	folderID := dashboardItem.ParentID
+	currentFolder, exists := structure.Items[folderID]
+
+	if !exists {
+		log.Warnf("refreshFolderMetadata: folder %s not found", folderID)
```
nkunal, line 268: %v is better format specifier than %s, since the former would work for any data type

pkg/dashboards/dashboards.go
```diff
@@ -254,11 +237,68 @@ func getDashboard(id string, myid int64) (map[string]interface{}, error) {
...
+		return nil
+	}
+
+	currentPath := buildFolderPath(folderID, structure)
+
+	// Check if already up-to-date
+	if folderData, ok := dashboardDetails["folder"].(map[string]interface{}); ok {
+		if storedPath, ok := folderData["path"].(string); ok && storedPath == currentPath {
```
nkunal, line 276: same comment as above

nkunal (commented):
looks fine, just one minor comment

nkunal (approved):

PR #2735, opened 2025-05-28
Fix longer functionl tests

VasudevRaut:
# Description
Fix failing longer functional test due to query type detection logic change

This PR fixes a failing longer functional test which broke after [recent changes ](https://github.com/siglens/siglens/pull/2478/files#diff-9154167dc8f81cbf3a45488271a9fbe19275a5f90efcfb1a64436fa5478f0060R150)in how we classify query types.

Previously, if a query started with a stats keyword, we considered it a stats-query. However, the logic was later updated to classify a query as stats or aggs if any part of the query contains the respective keywords—not just the starting clause.

Due to this change, the query type identified during evaluation has shifted, and as a result, the query response format also changed, causing existing tests to fail. This PR updates the expected results to reflect the new query classification logic.

# Testing


tools/sigclient/longerFunctionalQueries/eval_groupby_2col_2stats.json
```diff
@@ -1,50 +1,74 @@
...
+            {"CVal": 1, "Dtype": 2}
+          ],
+          "MeasureVal": {
+            "count(*)": 17106,
+            "sum(latitude)": 19979.46791800002
+          }
+        }
+      ],
+      "measureFunctions": ["count(*)", "sum(latitude)"],
+      "percent_complete": 99.98326079678607,
```
AndrewHess, line 59: This seems odd
1. why is it not 100?
2. does this get validated?

tools/sigclient/longerFunctionalQueries/eval_groupby_2col_2stats.json
```diff
@@ -1,50 +1,74 @@
...
+      "qid": 6,
+      "qtype": "aggs-query",
+      "sortByTimestampAtDefault": false,
+      "state": "COMPLETE",
+      "totalMatched": {
+        "relation": "eq",
+        "value": 0
+      },
```
AndrewHess, line 67: Is this supposed to be 0? I thought we give the number of records that matched the initial filter even for `stats` queries

tools/sigclient/longerFunctionalQueries/eval_stats_sort.json
```diff
@@ -1,61 +1,257 @@
...
-            "hobby_http",
-            "cnt"
-        ],
-        "columnsOrder": [
-            "cnt",
-            "hobby_http"
-        ]
+        "measureFunctions": ["cnt"],
+        "percent_complete": 100,
+        "qid": 1,
```
AndrewHess, line 244: Probably remove `qid`, since it might change from run to run. Hopefully we're already not testing on it

AndrewHess (commented):
Can you add some info in the PR description about why you're changing these? Were the expected results never setup correctly?

AndrewHess:
> Can you add some info in the PR description about why you're changing these? Were the expected results never setup correctly?

Great update to the PR description!

AndrewHess (approved):

PR #2738, opened 2025-05-29
Upgrading OTEL collector config to the latest

nrkawane:
# Description
The OTEL collector's logging exporter has been replaced with debug exporter. Changed the config we need to use with newer version as per [the details](https://github.com/open-telemetry/opentelemetry-collector/issues/11337#issue-2563588147) of the change.

### DIFF
--- a/ssmetrics-otel-collector-config.yaml
+++ b/ssmetrics-otel-collector-config.yaml
@@ -14,12 +14,12 @@ processors:
   batch:
 
 exporters:
-  logging:
-    loglevel: debug
+  debug:
+    verbosity: detailed
   prometheusremotewrite:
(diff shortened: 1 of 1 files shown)


ssmetrics-otel-collector-config.yaml
```diff
@@ -14,10 +14,10 @@ processors:
   batch:
 
 exporters:
-  logging:
-    loglevel: debug
+  debug:
+    verbosity: detailed
   prometheusremotewrite:
-    endpoint: 'http://siglens:8081/promql/api/v1/write'
+    endpoint: 'http://localhost:8081/promql/api/v1/write'
```
nkunal, line 20: you forgot here

PR #2741, opened 2025-05-29
SPL: Implemented percentile function "perc<val>(<field>)" using tdigest

vsharma-va:
# Description
Implemented Percentile function.
## Details -
1.  Uses [t-digest](https://www.sciencedirect.com/science/article/pii/S2665963820300403#sec3).
2. Compression value is based on splunks [docs](https://help.splunk.com/en/splunk-cloud-platform/search/spl2-search-reference/statistical-and-charting-functions/aggregate-functions#usage-11). => if cardinality < 1000 each individual value will have its own buckets i.e. no compression
3. Using a go library for t-digest implementation -> can be found [here](https://github.com/caio/go-tdigest)

## Remaining Steps -
~~1. Implement it for timechart queries~~
2. Some groupby query flow which I may have missed


pkg/ast/spl/spl.peg
AndrewHess, line 1: The [documentation](https://docs.splunk.com/Documentation/Splunk/9.1.1/SearchReference/Aggregatefunctions#Description_12) says
> perc\<percentile>(<value>) or the abbreviation p\<percentile>(<value>)

so you should also handle `p` just like `perc`

Also, you should add unit tests in splParser_test.go

pkg/segment/aggregations/evalaggs.go
```diff
@@ -769,6 +771,123 @@ func ComputeAggEvalForList(measureAgg *structs.MeasureAggregator, sstMap map[str
 	return nil
 }
 
+func ComputeAggEvalForPerc(measureAgg *structs.MeasureAggregator, sstMap map[string]*structs.SegStats, measureResults map[string]sutils.CValueEnclosure, runningEvalStats map[string]interface{}) error {
+	fields := measureAgg.ValueColRequest.GetFields()
+	fieldToValue := make(map[string]sutils.CValueEnclosure)
+	// result := 0.0
```
AndrewHess, line 777: nit: delete commented-out code

pkg/segment/aggregations/evalaggs.go
```diff
@@ -769,6 +771,123 @@ func ComputeAggEvalForList(measureAgg *structs.MeasureAggregator, sstMap map[str
...
+		err = PerformEvalAggForPerc(measureAgg, countStat.Count, td, fieldToValue)
+		if err != nil {
+			return fmt.Errorf("PerformEvalAggForPerc: Error while performing eval agg for perc, err: %v", err)
+		}
+	} else {
+		sst, ok := sstMap[fields[0]]
+		if !ok {
+			return fmt.Errorf("ComputeAggEvalForPerc: sstMap did not have segstats for field %v, measureAgg: %v", fields[0], measureAgg.String())
+		}
+		length := len(sst.Records)
```
AndrewHess, line 808: nit: this is more clear
```suggestion
		numRecords := len(sst.Records)
```

pkg/segment/aggregations/evalaggs.go
```diff
@@ -769,6 +771,123 @@ func ComputeAggEvalForList(measureAgg *structs.MeasureAggregator, sstMap map[str
...
+				return fmt.Errorf("PerformEvalAggForPerc: Error while performing eval agg for perc, err: %v", err)
+			}
+			fieldToValue = make(map[string]sutils.CValueEnclosure)
+		}
+	}
+
+	runningEvalStats[measureAgg.String()] = td
+	percValStr := measureAgg.Param
+	if len(percValStr) != 0 {
+		percValFlt, err := strconv.ParseFloat(percValStr, 64)
```
AndrewHess, line 825: nit: instead of `strconv.ParseFloat` use our `utils.FastParseFloat()`. `ParseFloat()` creates a new error any time the string can't be converted to a number, and if that error happens a lot for a query, it can cause significant heap usage and slow down the query. There's a few relatively recent PRs that replace ParseFloat with FastParseFloat for this reason

pkg/segment/results/blockresults/blockresult.go
```diff
@@ -203,6 +205,18 @@ func convertRequestToInternalStats(req *structs.GroupByRequest, usedByTimechart
...
+				curId, err := aggregations.SetupMeasureAgg(m, &allConvertedMeasureOps, sutils.Perc, &allReverseIndex, colToIdx, idx)
+				if err != nil {
+					log.Errorf("convertRequestToInternalStats: Erroro while adding measure agg in running stats for perc, err: %v", err)
+				}
+				idx = curId
+				continue
+			} else {
+				// TODO: handle timechart
+				mFunc = sutils.Perc
+			}
```
AndrewHess, line 219: I don't quite understand this comment; maybe add more info. Why do we need extra handling for timechart commands with percentile? Above, for `case sutils.Sum, sutils.Max, sutils.Min, sutils.List:` we also have just
```
			} else {
				mFunc = m.MeasureFunc
			}
```
so why don't we also need special timechart handling for that if we need it for percentile?

pkg/segment/utils/segconsts.go
```diff
@@ -111,6 +111,10 @@ const MS_IN_DAY = 86_400_000 // 24 * 60 * 60 * 1000
 // https://docs.splunk.com/Documentation/SplunkCloud/9.1.2312/SearchReference/Multivaluefunctions
 const MAX_SPL_LIST_SIZE = 100
 
+// splunk uses nearest rank for less than 1000 cardinality and tdigest for > 1000 cardinality
+// https://help.splunk.com/en/splunk-cloud-platform/search/spl2-search-reference/statistical-and-charting-functions/aggregate-functions#differences-between-splunk-and-excel-percentile-algorithms-0
+const TDIGESTCOMPRESSION = 1000
```
AndrewHess, line 116: nit
```suggestion
const TDIGEST_COMPRESSION = 1000
```

pkg/segment/utils/segconsts.go
```diff
@@ -111,6 +111,10 @@ const MS_IN_DAY = 86_400_000 // 24 * 60 * 60 * 1000
 // https://docs.splunk.com/Documentation/SplunkCloud/9.1.2312/SearchReference/Multivaluefunctions
 const MAX_SPL_LIST_SIZE = 100
 
+// splunk uses nearest rank for less than 1000 cardinality and tdigest for > 1000 cardinality
+// https://help.splunk.com/en/splunk-cloud-platform/search/spl2-search-reference/statistical-and-charting-functions/aggregate-functions#differences-between-splunk-and-excel-percentile-algorithms-0
```
AndrewHess, line 115: nit: we follow SPL, not SPL2, so you should link to different docs

pkg/utils/gobutils.go
AndrewHess, line 1: You should add a unit test in gobutils_test.go. At least test something like `original == decode(encode(original))`

AndrewHess (commented):
One thing I'm seeing: several places, you parse the `MeasureAggregator.Param` to a float and then have to handle the error case; but _I think_ `MeasureAggregator.Param` is only used for the percentile, so it would probably be better to change the `Param` type to a float and do the conversion just once (perhaps in the PEG)

pkg/segment/aggregations/evalaggs.go
vsharma-va: Understood. But due to changing the `param` for `measureAggs` to float64 this conversion is no longer required and I have removed it

pkg/segment/results/blockresults/blockresult.go
vsharma-va: Sorry for the confusion, I forgot to remove this todo

pkg/segment/utils/segconsts.go
vsharma-va: I have moved this declaration to `CreateNewTDigest` function in `gobutils.go`, because go was throwing an error for circular imports. Let me know if I should define it somewhere else?

pkg/utils/gobutils.go
vsharma-va: Added a few unit tests in `gobutils_test.go`

pkg/ast/spl/tests/splParser_test.go
```diff
@@ -10409,6 +10434,10 @@ func Test_MVExpand_MissingField(t *testing.T) {
 	assert.NotNil(t, err)
 }
 
+func Test_Percentile_Alternative(t *testing.T) {
+
+}
```
AndrewHess, line 10439: Implement or remove this

pkg/ast/spl/tests/splParser_test.go
```diff
@@ -10345,10 +10364,16 @@ func getAggFunctions() []sutils.AggregateFunctions {
...
-		sutils.Perc, sutils.ExactPerc, sutils.UpperPerc,
+		sutils.Perc, TestPercAlternate, sutils.ExactPerc, sutils.UpperPerc,
 	}
 }
 
+const TestPercAlternate sutils.AggregateFunctions = 100
+
+var alternateAggTestingMap = map[sutils.AggregateFunctions]sutils.AggregateFunctions{
+	TestPercAlternate: sutils.Perc,
+}
```
AndrewHess, line 10375: 1. nit: in general, keep global definitions near the top of the file
2. hardcoding `TestPercAlternate = 100` seems bad. If we eventually get to 100 on other sutils.AggregateFunctions, then there could be a collision. Do we really need `TestPercAlternate`? If we do, then choose a number that won't collide

AndrewHess (dismissed):

PR #2744, opened 2025-05-29
optimize GetFields 

nkunal:
# Description

1) Optimize GetFields
`GetFields` was getting calculated on each iteration of running stats, even though it operates on the incoming query expression.
With this PR, this query `URL != "" | stats avg(eval(len(URL))) as l, count as c by CounterID | where c > 100000 | sort 25 -l`
Before:  `8.5 s`
Now : `7.8 s`

# Testing


AndrewHess (approved):
Looks good. We'll probably want similar caching for the other `...Expr` types. Or maybe they all flow through the `ValueExpr` type (?) so it would be sufficient to add it to `ValueExpr.GetFields()`

nkunal:
> Looks good. We'll probably want similar caching for the other `...Expr` types. Or maybe they all flow through the `ValueExpr` type (?) so it would be sufficient to add it to `ValueExpr.GetFields()`

yeah in future we can add that, based on if they show up on go profiler

PR #2758, opened 2025-06-04
SPL: implemented sigfig eval function

vsharma-va:
# Description
Implemented sigfig mathematical eval function. Docs can be found [here](https://docs.splunk.com/Documentation/Splunk/9.1.1/SearchReference/MathematicalFunctions#sigfig.28.26lt.3Bnum.26gt.3B.29).
Added two new helper functions -
1. `adjustResultToReqSigFigs` - returns the value after appropriate rounding
2. `mulDivSigFigHelper` - returns the value after appropriate rounding for multiplication/division. Can be done using `strings.Replace` + conversion to int to get mantissa length but I don't know which one would be faster

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -68,9 +68,10 @@
       }
     },
     "node_modules/@eslint/eslintrc/node_modules/brace-expansion": {
-      "version": "1.1.11",
-      "resolved": "https://registry.npmjs.org/brace-expansion/-/brace-expansion-1.1.11.tgz",
-      "integrity": "sha512-iCuPHDFgrHX7H2vEI/5xpz07zSHB00TpugqhmYtVmMO6518mCuRMoOYFldEBl0g187ufozdaHgWKcYFb61qGiA==",
+      "version": "1.1.12",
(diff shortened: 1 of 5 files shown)


AndrewHess:
> Docs can be found [here](https://help.splunk.com/en/splunk-cloud-platform/search/spl2-search-reference/evaluation-functions/mathematical-functions#ariaid-title12).

You should link to SPL docs, not SPL2 docs

pkg/segment/structs/mathfuncstest/mathfuncs_test.go
```diff
@@ -0,0 +1,149 @@
+package mathfuncstest
+
+import (
```
AndrewHess, line 3: 1. Add our normal copyright/license comment to the top of the file. Make sure you use 2021-2025
2. Since you're testing code from evaluationstructs.go, rename this file/dir so it's easier for someone looking for these tests to find them. I think some decent options are `structs/evaluationtests/mathfuncs_test.go`, `structs/tests/evaluationstructs_test.go`, or similar

PR #2774, opened 2025-06-11
Fix SPL wildcard to match multiline values

gaurikarkhile001:
### Problem
Wildcard SPL expressions like `.*google.*` or `*google*` did not match fields containing newlines (e.g., "watching\nhttps://google.com").

### Testing Done
  Unit tests added  
  Verified with multiline ingestion (e.g., `m["hobby"] = "watching\nhttps://google.com"`)  
  All tests passed via `go test ./pkg/regex/...`


pkg/regex/regex.go
```diff
@@ -38,15 +38,20 @@ type simpleRegex struct {
...
+	// fallback if pattern contains newline — multiline match not supported by simpleRegex
+	if bytes.Contains([]byte(pattern), []byte("\n")) {
+		return regexp.Compile(pattern)
+	}
+
+	// Check if the pattern contains a single '.' (not part of '.*') which indicates
+	// a single-character wildcard that our simple regex can't handle
+	if containsSingleDotWildcard(pattern) {
+		return regexp.Compile(pattern)
+	}
```
AndrewHess, line 53: Why does your regex on line 41 have the dot here: `[a-zA-Z0-9_:/.\-]`? Unless I'm missing something, I think you'd get the same behavior by removing the dot there and then you don't need this `if` block here—seems cleaner to me that way

pkg/regex/regex.go
```diff
@@ -61,7 +66,26 @@ func New(pattern string) (Regex, error) {
...
 }
 
+// containsSingleDotWildcard checks if the pattern contains a '.' that isn't part of '.*',
+// which indicates a single-character wildcard that our simple regex can't handle
+func containsSingleDotWildcard(pattern string) bool {
+	for i := 0; i < len(pattern); i++ {
+		if pattern[i] == '.' {
+			// Check if this dot is followed by '*' (making it '.*')
+			if i+1 < len(pattern) && pattern[i+1] == '*' {
+				continue // This is a .* wildcard, which our simple regex can handle
```
AndrewHess, line 76: 1. You'll get an out-of-bounds panic on a string that ends with `.`
2. not sure if this is important, but you're not handling escaping (i.e., `\.*`). If you do need to handle that, you'll also need to be careful about `\\.*` (escaping the `\`, so the `.` is not escaped)

pkg/regex/regex.go
```diff
@@ -61,7 +66,26 @@ func New(pattern string) (Regex, error) {
...
+			// This is a single dot wildcard, which our simple regex can't handle
+			return true
+		}
+	}
+	return false
+}
+
 func (r *simpleRegex) Match(buf []byte) bool {
+	// Normalize line breaks to a space for matching
+	normalized := bytes.ReplaceAll(buf, []byte("\n"), []byte(" "))
```
AndrewHess, line 87: Is this valid? My initial hunch is that for some regex queries, this is going to cause false positives/negatives

AndrewHess (approved):

PR #2779, opened 2025-06-12
SPL: Implement median

vsharma-va:
# Description
Median works exactly the same as `perc<val>(<field>)` where `val=50`. Docs can be found [here](https://docs.splunk.com/Documentation/Splunk/9.1.1/SearchReference/Aggregatefunctions#median.28.26lt.3Bvalue.26gt.3B.29)

# Testing
Modified tests in splParser_test.go

### DIFF
--- a/pkg/segment/query/processor/dataprocessor.go
+++ b/pkg/segment/query/processor/dataprocessor.go
@@ -829,25 +829,8 @@ func NewStatsDP(options *structs.StatsExpr) *DataProcessor {
 	}
 }
 
-// Note: this has side-effects.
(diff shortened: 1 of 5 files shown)


vsharma-va:
Right now, I have changed the spl.peg file which essentially takes median and converts it into `perc50(<field>)` during parsing. The `StrEnc` remains as `median(<field>)`.
Let me know if I should change it and handle it manually during calculations of aggs in different flows?

AndrewHess (approved):

AndrewHess (approved):

PR #2780, opened 2025-06-12
SPL: Implement eval commands mvdedup + mvappend

vsharma-va:
# Description
Implementation of `mvdedup` & `mvappend`. Docs can be found [here](https://docs.splunk.com/Documentation/Splunk/9.1.1/SearchReference/MultivalueEvalFunctions#mvappend.28.26lt.3Bvalues.26gt.3B.29)

# Testing
Added individual functional tests + updated `multi_value.json` functional test to include `mvappend`

### DIFF
--- a/pkg/segment/structs/evaluationstructs.go
+++ b/pkg/segment/structs/evaluationstructs.go
@@ -1134,6 +1134,63 @@ func handleSplit(self *MultiValueExpr, fieldToValue map[string]sutils.CValueEncl
 	return stringsList, nil
 }
 
+func handleMVDedup(self *MultiValueExpr, fieldToValue map[string]sutils.CValueEnclosure) ([]string, error) {
+	if self.MultiValueExprParams == nil || len(self.MultiValueExprParams) != 1 || self.MultiValueExprParams[0] == nil {
+		return []string{}, fmt.Errorf("MultiValueExpr.Evaluate: mvdedup requires one multiValueExpr argument")
+	}
+	mvSlice, err := self.MultiValueExprParams[0].Evaluate(fieldToValue)
+	if utils.IsNilValueError(err) {
+		return nil, err
+	} else if err != nil {
+		return []string{}, fmt.Errorf("TextExpr.EvaluateText -> handleMVDedup: %v", err)
+	}
+
+	seen := make(map[string]string, len(mvSlice))
+	dedupedSlice := []string{}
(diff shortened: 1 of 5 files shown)


pkg/segment/structs/evaluationstructs.go
```diff
@@ -1134,6 +1134,63 @@ func handleSplit(self *MultiValueExpr, fieldToValue map[string]sutils.CValueEncl
...
 
+func handleMVDedup(self *MultiValueExpr, fieldToValue map[string]sutils.CValueEnclosure) ([]string, error) {
+	if self.MultiValueExprParams == nil || len(self.MultiValueExprParams) != 1 || self.MultiValueExprParams[0] == nil {
+		return []string{}, fmt.Errorf("MultiValueExpr.Evaluate: mvdedup requires one multiValueExpr argument")
+	}
+	mvSlice, err := self.MultiValueExprParams[0].Evaluate(fieldToValue)
+	if utils.IsNilValueError(err) {
+		return nil, err
+	} else if err != nil {
+		return []string{}, fmt.Errorf("TextExpr.EvaluateText -> handleMVDedup: %v", err)
```
AndrewHess, line 1145: nit
```suggestion
		return []string{}, fmt.Errorf("handleMVDedup: %v", err)
```

similar comment in `handleMVAppend`

tools/sigclient/functionalQueries/multi_value.json
```diff
@@ -51,40 +63,50 @@
                     "bc",
                     "123"
                 ],
-                "split_join": "a?bc?123"
+                "split_join": "a?bc?123",
+                "timestamp": 1749740369865
```
AndrewHess, line 67: I think we don't want/need the timestamp. Presumably we're not validating the timestamp; if we are, it seems like this test would fail when we run the suite later?

tools/sigclient/functionalQueries/multi_value.json
AndrewHess, line 1: I'd lean towards not changing this test, unless this change revealed some bugs that the other E2E tests you added didn't catch. I don't really understand why some of the records get new expected values (or maybe the order of the records is just different?), and IMO the other E2E tests you added should be sufficient

vsharma-va: The order of the records is just different. Rest of the things should be same as before.
I will revert this file to the original version then

PR #2785, opened 2025-06-16
Add SPL Multivalue eval functions -mvmap, mvrange

gaurikarkhile001:
### Description
This PR adds implementation and test coverage for the following SPL multivalue functions: [docs](https://help.splunk.com/en/splunk-cloud-platform/search/spl2-search-reference/evaluation-functions/multivalue-eval-functions)

- `mvmap`: Applies a mapping function to each element of a multivalue field.
- `mvrange`: Generates a multivalue field consisting of a sequence of numbers.

### Testing
- Added functional test cases in `multi_value.json`
- Updated `mvrange.json` and `mvmap.json` functional test


pkg/segment/structs/evaluationstructs.go
```diff
@@ -1145,6 +1145,97 @@ func handleSplit(self *MultiValueExpr, fieldToValue map[string]sutils.CValueEncl
...
+		var dtype sutils.SS_DTYPE
+		if fval, err := strconv.ParseFloat(val, 64); err == nil {
+			cval = fval
+			dtype = sutils.SS_DT_FLOAT
+		} else {
+			cval = val
+			dtype = sutils.SS_DT_STRING
+		}
+
+		localMap["value"] = sutils.CValueEnclosure{CVal: cval, Dtype: dtype}
```
AndrewHess, line 1180: This looks odd to me. Why do you hardcode `"value"`? I'm not sure if it's safe to change it to something else, and I'm not sure if somewhere else is expecting a key called `"value"`

I'd suggest making a constant and using that constant everywhere instead of hardcoding. This will
1. Make it more obvious where (and hopefully why) this key is being used
2. Prevent typos in the future
3. Make the function of this key name more obvious, if you name the constant well

pkg/segment/structs/evaluationstructs.go
```diff
@@ -1145,6 +1145,97 @@ func handleSplit(self *MultiValueExpr, fieldToValue map[string]sutils.CValueEncl
...
+
+		res, err := self.ValueExprParams[0].EvaluateToString(localMap)
+		if err != nil {
+
+			if dtype == sutils.SS_DT_STRING {
+				res = fmt.Sprintf("%v10", val)
```
AndrewHess, line 1186: ?
```suggestion
				res = fmt.Sprintf("%v", val)
```

AndrewHess (commented):
You have a test for `range=mvrange(1,11,2)`, but [this documentation](https://docs.splunk.com/Documentation/Splunk/9.1.1/SearchReference/MultivalueEvalFunctions#Basic_examples_3) mentions `mvrange(1514834731,1524134919,"7d")` can be used for a 7-day step; you should add an automated test for something similar.

pkg/segment/structs/evaluationstructs.go
```diff
@@ -1339,6 +1339,158 @@ func handleMVAppend(self *MultiValueExpr, fieldToValue map[string]sutils.CValueE
...
+	step := int64(1)
+	if len(self.NumericExprParams) >= 3 {
+		raw, err := self.NumericExprParams[2].Evaluate(fieldToValue)
+		if err != nil {
+			return nil, fmt.Errorf("step error: %v", err)
+		}
+		var stepVal interface{} = raw
+		switch v := stepVal.(type) {
+
+		case float64:
```
AndrewHess, line 1406: nit: delete this blank line; we usually have the first `case` on the line immediately following `switch`

pkg/segment/structs/evaluationstructs.go
```diff
@@ -1339,6 +1339,158 @@ func handleMVAppend(self *MultiValueExpr, fieldToValue map[string]sutils.CValueE
...
+		return nil, fmt.Errorf("handleMVRange: step cannot be zero")
+	}
+	startRaw, err := self.NumericExprParams[0].Evaluate(fieldToValue)
+	if err != nil {
+		return nil, fmt.Errorf("handleMVRange: start value error: %v", err)
+	}
+	var startVal interface{} = startRaw
+
+	var start int64
+	switch v := startVal.(type) {
```
AndrewHess, line 1434: There's no purpose in making `startVal`

Same for `endVal`

pkg/segment/structs/evaluationstructs.go
```diff
@@ -1339,6 +1339,158 @@ func handleMVAppend(self *MultiValueExpr, fieldToValue map[string]sutils.CValueE
...
+	if (step > 0 && start >= end) || (step < 0 && start <= end) {
+		return []string{}, nil
+	}
+
+	var result []string
+	for i := start; (step > 0 && i < end) || (step < 0 && i > end); i += step {
+		result = append(result, strconv.FormatInt(i, 10))
+	}
+
+	return result, nil
```
AndrewHess, line 1479: When `step == 0` and `start != end`, your code will return no elements. What does Splunk do? Does it return no elements or one element at `start`?

pkg/segment/structs/evaluationstructs.go
```diff
@@ -1339,6 +1339,158 @@ func handleMVAppend(self *MultiValueExpr, fieldToValue map[string]sutils.CValueE
...
+	if strings.HasSuffix(s, "d") {
+		daysStr := strings.TrimSuffix(s, "d")
+		days, err := strconv.Atoi(daysStr)
+		if err != nil {
+			return 0, err
+		}
+		return time.Hour * 24 * time.Duration(days), nil
+	}
+	return 0, fmt.Errorf("unsupported duration format: %s", s)
+}
```
AndrewHess, line 1492: I'm pretty sure Splunk handles other things besides `d`. You should check, and then implement and test anything else that's required

PR #2793, opened 2025-06-17
Sig 2804 spl stats distinct count takes excessive memory for eval queries

t-sunk:
# Description
With the previous implementation, `*|stats dc(eval(xyz))` would create a map with one element for every unique value of field `xyz`. The current implementation now uses the HyperLogLog (HLL) algorithm to calculate approximate cardinality.

I have changed all calls to `dc` except calls involving [streamstats](https://docs.splunk.com/Documentation/Splunk/9.4.2/SearchReference/Streamstats) with either `window` or `time_window` arguments. These require that we have some way to insert elements into the window (which we can do with HLL), and some way to remove them from the window (which we cannot do with HLL).

# Testing
This passes the pre-existing tests for cardinality. This is because for smaller cardinalities, this algorithm computes exact cardinality. Our tests do not cover any cases where the cardinality is large.

Testing for correctness on larger cardinalities: On the test logs, `*|stats dc(latency)` (implemented earlier) gives me $19,963$ and  `*|stats dc(eval(latency))` (implemented now) gives `20,007`. The correct answer here should be `19,983`; meaning both of the calculated answers are within $0.12$% of the actual answer.

# Performance
There was a 100x slowdown in earlier commits; this was fixed in commit [1d46c0e](https://github.com/siglens/siglens/pull/2793/commits/1d46c0e34e52ab1541723119e53c0abd9a665767).


AndrewHess (commented):
Looks decent overall. You mentioned a 100x slowdown, which is quite severe. I was able to debug and patch this like this:
[diff.patch](https://github.com/user-attachments/files/20804419/diff.patch)
(you can run `git apply <file>` to get the changes)
I only tried it on the simple query you mentioned in the description, but it seems to have basically negated any slowdown. You should do something like that unless there's some edge cases that would make it give incorrect results

pkg/segment/aggregations/evalaggs.go
```diff
@@ -816,27 +817,58 @@ func ComputeAggEvalForVarp(measureAgg *structs.MeasureAggregator, sstMap map[str
...
+	} else {
+		if measureAgg.ValueColRequest.BooleanExpr != nil {
+			boolResult, err := measureAgg.ValueColRequest.BooleanExpr.Evaluate(fieldToValue)
+			if err != nil {
+				return fmt.Errorf("PerformAggEvalForCardinality: there are some errors in the eval function that is inside the values function: %v", err)
+			}
+			if boolResult {
+				// sentinel value, this corresponds to all bits set, which makes it a good sentinel value for this
+				// implementation of HLL
+				hll.AddRaw(uint64(0xffffffffffffffff))
```
AndrewHess, line 836: This is a good comment. I'm still unclear why this is a good sentinel value though, compared to anything else. A little more info the comment (link to docs that state this, or describe why this value makes a good sentinel) would be nice. Or if it actually don't matter, you can just say we chose this arbitrarily.

AndrewHess (approved):
One minor comment; you can merge as is if you want

AndrewHess (approved):

PR #2806, opened 2025-06-20
Add SPL multivalue eval functions - mvfilter

gaurikarkhile001:
# Description
Implemented mvfilter function to filter multivalue fields based on a condition. [Docs](https://docs.splunk.com/Documentation/Splunk/9.1.1/SearchReference/MultivalueEvalFunctions)

# Testing
Added individual functional test to include mvfilter

### DIFF
--- a/pkg/segment/structs/evaluationstructs.go
+++ b/pkg/segment/structs/evaluationstructs.go
@@ -1340,7 +1340,7 @@ func handleMVAppend(self *MultiValueExpr, fieldToValue map[string]sutils.CValueE
 }
 
 func (self *BoolExpr) ExtractSingleField() string {
-	
+
 	fields := self.GetFields()
 	if len(fields) == 0 {
 		return ""
--- a/pkg/segment/structs/evaluationstructs.go
+++ b/pkg/segment/structs/evaluationstructs.go
@@ -1339,6 +1339,99 @@ func handleMVAppend(self *MultiValueExpr, fieldToValue map[string]sutils.CValueE
 	return finalMVSlice, nil
 }
 
+func (self *BoolExpr) ExtractSingleField() string {
+	if self.IsTerminal {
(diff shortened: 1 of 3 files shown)


pkg/segment/structs/evaluationstructs.go
```diff
@@ -1339,6 +1339,99 @@ func handleMVAppend(self *MultiValueExpr, fieldToValue map[string]sutils.CValueE
 	return finalMVSlice, nil
 }
 
+func (self *BoolExpr) ExtractSingleField() string {
```
AndrewHess, line 1342: If I understand the purpose of this function, it can be done much simpler:
```
fields := self.GetFields()
if len(fields) != 1 {
    return "", false
}

return fields[0], true
```

gaurikarkhile001: While this looks simpler, it's not functionally equivalent to the current code for several reasons:

- The current function returns a single string, not a (string, bool) tuple
- The current function has more complex logic that handles specific cases, particularly for non-terminal expressions
- It can extract a field even if multiple fields are used, as long as the condition references one primary field

AndrewHess: In that case, it sounds like your only criticism of my snippet is checking uniqueness. So I'd recommend updating my snippet to check if GetFields() returns only one unique value.

Alternatively, you could update all the `GetFields()` functions to return only unique fields, but I think that's quite a bit more work and probably not worth it.

I would highly prefer a ~10 line snippet like mine compared to the ~40 lines of nested `if` statements you already have, because it takes me significantly more work to:
1. understand what the nested ifs are intending to do
2. get confident that there's not missed edge cases, and remember that we might have to update this is if later we add a new enum value (similar to `VEMStringExpr`, but something else)

AndrewHess (approved):

PR #2807, opened 2025-06-20
SPL: Implemented tojson search command

vsharma-va:
# Description
Implemented `tojson` search command. Supporting grammer + a new processor.
Docs can be found [here](https://help.splunk.com/en/splunk-enterprise/search/spl-search-reference/9.2/search-commands/tojson).

# Testing
~~Working on tests~~ (done)

### DIFF
--- a/benchmark_test.go
+++ b/benchmark_test.go
@@ -64,6 +64,7 @@ var json = jsoniter.ConfigFastest
 var loadDataBytes0 = []byte(`{"index" : { "_index" : "bidx-0"} }
 {"address":"91982 Plain side, New Orleans, North Dakota 65104","app_name":"Oxcould","app_version":"4.3.15","batch":"batch-89","city":"New Orleans","country":"Chile","first_name":"Luigi","gender":"male","group":"group 2","hobby":"Cosplaying","http_method":"HEAD","http_status":404,"ident":"23a1949c-c32d-47ab-a573-47859fac0e76","image":"https://picsum.photos/381/329","job_company":"TopCoder","job_description":"Dynamic","job_level":"Markets","job_title":"Liaison","last_name":"Tromp","latency":2891953,"latitude":33.139514,"longitude":114.767227,"question":"Sustainable gentrify yr meditation Godard salvia vice migas drinking fanny pack?","ssn":"660889936","state":"North Dakota","street":"91982 Plain side","url":"https://www.internationalextend.info/networks","user_agent":"Mozilla/5.0 (Macintosh; Intel Mac OS X 10_6_5 rv:3.0) Gecko/1915-11-28 Firefox/35.0","user_color":"DodgerBlue","user_email":"graceheaney@beatty.io","user_phone":"2597778030","weekday":"Saturday","zip":"65104"}
 `)
+
 var loadDataBytes1 = []byte(`{"index" : { "_index" : "bidx-1"} }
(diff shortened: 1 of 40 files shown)


pkg/ast/spl/spl.peg
```diff
@@ -5390,9 +5409,149 @@ AppendBlock <- PIPE CMD_APPEND options:(AppendCmdOption SPACE?)* subsearch:Subse
...
+    }
+    gobRegex := utils.GobbableRegex{}
+    var err error
+    if regex.RawString != "" {
+        err = gobRegex.SetRegex(regex.RawString)
+    } else {
+        err = gobRegex.SetRegex(regex.FieldName)
+    }
+    if err != nil {
+        return nil, fmt.Errorf("spl peg: Regex compile: %v", err)
```
AndrewHess, line 5429: Give more context, e.g.:
```suggestion
        return nil, fmt.Errorf("spl peg: Regex compile for tojson: %v", err)
```

pkg/segment/structs/segstructs.go
```diff
@@ -366,6 +367,37 @@ type AppendCmdOption struct {
...
+	FieldsDtypes    []*ToJsonFieldsDtypeOptions
+	DefaultType     *ToJsonFieldsDtypeOptions
+	FillNull        bool
+	IncludeInternal bool
+	OutputField     string
+	AllFields       bool
+}
+
+type ToJsonCmdOption struct {
+	OptionType string
```
AndrewHess, line 380: In the PEG parser, it looks like this field is only supposed to take a few values. If that's the case, an enum is probably a better option than string for this

EDIT: I think this struct is only used in the PEG parser? If so, probably move the definition there

pkg/ast/spl/spl.peg
```diff
@@ -5390,9 +5409,149 @@ AppendBlock <- PIPE CMD_APPEND options:(AppendCmdOption SPACE?)* subsearch:Subse
...
+OutputFieldOption <- "output_field" EQUAL strVal:String {
+    return &structs.ToJsonCmdOption {
+        OptionType: "output_field",
+        Value: strVal.(string),
+    }, nil
+}
+
+ToJsonBlock <- PIPE CMD_TOJSON options:(ToJsonOption)* {
+    defaultDtypeRegex := utils.GobbableRegex{}
+    err := defaultDtypeRegex.SetRegex(`\*`)
```
AndrewHess, line 5503: I don't understand this. Why are you setting the regex to match a literal `*` char?

pkg/segment/structs/segstructs.go
```diff
@@ -366,6 +367,37 @@ type AppendCmdOption struct {
...
+type ToJsonDtypes uint8
+
+const (
+	TJ_None ToJsonDtypes = iota
+	TJ_Auto
+	TJ_Bool
+	TJ_Json
+	TJ_Num
+	TJ_Str
+	TJ_PP // post process
```
AndrewHess, line 398: nit: IMO the characters saved is not worth the extra cognitive cost of remembering what PP stands for
```suggestion
	TJ_PostProcess
```
I'm also not sure what post process means in this context, so it might be worth adding more info in the comment

AndrewHess (commented):
You should also add some E2E tests (e.g., functional test suite tests)

pkg/ast/spl/spl.peg
vsharma-va: I forgot to change it. I will fix it in a commit

pkg/segment/structs/segstructs.go
```diff
@@ -1538,7 +1570,7 @@ var unsupportedEvalFuncs = map[string]struct{}{
 	"isnum":            {},
 	"isnotnull":        {},
 	"spath":            {},
-	"eventcount":       {},
+	// "eventcount":       {},
```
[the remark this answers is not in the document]
vsharma-va: I will restore it

PR #2810, opened 2025-06-21
Check if access to an index is allowed

nrkawane:
* Add a hook to check authorisation to access index and filter out unauthorised index 
* Updated callers and test to pass on appropriate request context (or nil in case it is irrelevant)


AndrewHess (commented):
Looks good so far

AndrewHess (approved):

PR #2820, opened 2025-06-26
SPL: informational functions - ismv, isobject

vsharma-va:
# Description
Implemented `isobject`, `ismv`. Docs can be found [here](https://help.splunk.com/en/splunk-enterprise/search/spl-search-reference/9.4/evaluation-functions/informational-functions#ariaid-title6)


pkg/segment/structs/evaluationstructs.go
```diff
@@ -704,14 +704,46 @@ func (self *BoolExpr) evaluateToCValueEnclosure(fieldToValue map[string]sutils.C
...
+			var v any
+			err = json.Unmarshal([]byte(val), &v)
+			if err != nil {
+				return getBoolCValueEnclosure(false), nil
+			} else {
+				return getBoolCValueEnclosure(true), nil
+			}
+		case "ismv":
+			err := fmt.Errorf("error while evaluating ismv")
+			if self.LeftValue.NumericExpr != nil {
```
AndrewHess, line 721: 1. Why do you expect the `NumericExpr` to be non-nil? Couldn't it be a non-numeric value?
2. Do you need to check if `LeftValue != nil`, or is that already guaranteed to be true if we get here?

pkg/segment/structs/evaluationstructs.go
```diff
@@ -704,14 +704,46 @@ func (self *BoolExpr) evaluateToCValueEnclosure(fieldToValue map[string]sutils.C
...
+				fields := self.GetFields()
+				if len(fields) == 1 {
+					val, ok := fieldToValue[fields[0]]
```
AndrewHess, line 724: I think you're checking that you got exactly one arg. Why do you do that here but not in `isobject`, since they both require one arg?

pkg/segment/structs/evaluationstructs.go
vsharma-va: 1. If it is not `NumericExpr` then the value provided to this function is not a field. Therefore we should return an error
2. It is guaranteed, otherwise the spl parser would throw an error

AndrewHess: 1. I found that pretty confusing. I had to dig into spl.peg to figure out why this was a NumericExpr instead of a StringExpr when you expect a field name. I'd recommend either handling both cases (and anything else that could give a field name), or adding a comment in your code saying why this should be a NumericExpr.

   Currently this is coupled to the parser (the parser tries to use a NumericExpr before trying to use a StringExpr), but as far as I can tell a StringExpr would also be valid if we changed the parser (in a valid way) later; but this code is confusing without that non-obvious context.

2. Sounds good

vsharma-va: 1. Yeah, So currently whenever a field is given to these functions, it gets put inside `NumericExpr`. To make sure it doesn't break if the parser changes in the future, I have removed the `NumericExpr != nil` check, since `GetFields()` already handles all the cases and returns the fields based on the type of the expression

AndrewHess (approved):

PR #2824, opened 2025-06-26
Improve sorting performance with small limit

AndrewHess:
# Description
To evaluate queries with a `sort` command where only the top `N` results are kept and `N` is small (< 1000), we use a heap to do the sorting. When the heap is at capacity, we were pushing each item and then popping an item. With this PR, we only push if we're not going to immediately pop the same item.

# Testing
Decreased this query time by about 15%
```
SearchPhrase != "" | sort 10 str(EventTime) | fields SearchPhrase, EventTime
```


nkunal (approved):
Sweet!!

PR #2841, opened 2025-07-01
Eval command cannot be used to assign boolean results to a field

t-sunk:
# Description
Queries such as 
```
* | eval fast=(latency<100000)
```
would be accepted by the grammar, but the result would be treated as a string.
This is now rejected by the grammar.


AndrewHess (approved):
FYI quoting the docs would have been helpful in the parser or PR description

From https://help.splunk.com/en/splunk-enterprise/search/spl-search-reference/9.4/search-commands/eval
> The result of an eval expression cannot be a Boolean.

nrkawane (approved):
Simply approving regenerated spl.go

PR #2861, opened 2025-07-07
SPL: Implementation of all bitwise functions

vsharma-va:
# Description
Implementation of all bitwise functions. Docs can be found [here](https://help.splunk.com/en/splunk-enterprise/search/spl-search-reference/9.4/evaluation-functions/bitwise-functions#ariaid-title7)

### DIFF
--- a/pkg/segment/structs/evaluationstructs.go
+++ b/pkg/segment/structs/evaluationstructs.go
@@ -507,6 +507,10 @@ var timeFormatReplacements = []struct {
 
 var ErrFloatMissingField = fmt.Errorf("Missing field")
 var ErrFloatFieldNull = fmt.Errorf("field was null")
+var ErrBitwiseOpNotInt = fmt.Errorf("NumericExpr.Evaluate: bitwise operation only works on integers")
+var ErrBitwiseOpExceedsSizeLimit = fmt.Errorf("NumericExpr.Evaluate: bitwise operation can only take integers of upto 2^53-1 bits")
+var ErrBitwiseOpNoNeg = fmt.Errorf("NumericExpr.Evaluate: bitwise operation only works on positive integers")
+var ErrBitwiseOpShiftExceedsLimit = fmt.Errorf("NumericExpr.Evaluate: bitwise shift operations can have a max shift of %v", MAX_BIT_SHIFT_SIZE)
 
 var ErrWithCodeConversionErr = utils.NewErrorWithCode(utils.CONVERSION_ERR,
 	sutils.ErrFloatConversionFailed)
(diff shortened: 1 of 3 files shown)


pkg/segment/structs/evaluationstructs.go
```diff
@@ -507,6 +507,10 @@ var timeFormatReplacements = []struct {
 
 var ErrFloatMissingField = fmt.Errorf("Missing field")
 var ErrFloatFieldNull = fmt.Errorf("field was null")
+var ErrBitwiseOpNotInt = fmt.Errorf("NumericExpr.Evaluate: bitwise operation only works on integers")
+var ErrBitwiseOpExceedsSizeLimit = fmt.Errorf("NumericExpr.Evaluate: bitwise operation can only take integers of upto 2^53-1 bits")
```
AndrewHess, line 511: Did you mean this?
```suggestion
var ErrBitwiseOpExceedsSizeLimit = fmt.Errorf("NumericExpr.Evaluate: bitwise operation can only take integers of up to 53 bits")
```

pkg/segment/structs/evaluationstructs.go
```diff
@@ -517,6 +521,9 @@ var ErrWithCodeFloatMissingField = utils.NewErrorWithCode(utils.NIL_VALUE_ERR,
 var ErrWithCodeFieldNull = utils.NewErrorWithCode(utils.NIL_VALUE_ERR,
 	ErrFloatFieldNull)
 
+const MAX_BIT_OPERATION_NUM_SIZE = 1<<53 - 1
+const MAX_BIT_SHIFT_SIZE = 53
```
AndrewHess, line 525: Link to the splunk docs for why we're using this, to make it more clear to future readers

pkg/segment/structs/evaluationstructs.go
```diff
@@ -3097,6 +3104,19 @@ func (self *NumericExpr) evaluateWithSigfig(expr *NumericExpr, fieldToValue map[
 	}
 }
 
+func handleBitwiseOperationInput(left float64, right float64) (uint64, uint64, error) {
+	if left < 0 || right < 0 {
+		return 0, 0, ErrBitwiseOpNoNeg
+	}
+	if left-math.Floor(left) != 0 || right-math.Floor(right) != 0 {
+		return 0, 0, ErrBitwiseOpNotInt
+	}
```
AndrewHess, line 3113: nit:
1. I'd suggest changing the function name from `handle...` to `isValid...` or similar. Generally when I see a function called `handle...` I think "it's going to do some operation" but when I see `isValid...` I think "it's going to check something, but not really do meaningful work"
2. Also here, link to splunk docs to make it clear why we're enforcing these restrictions

AndrewHess (commented):
You should add some parser tests in splParser_test.go

AndrewHess (approved):

AndrewHess (approved):

PR #2862, opened 2025-07-07
Fix float formatting issues

t-sunk:
# Description
This fixes two issue with floating point numbers in the parser tests:
1. A number (e.g. `0.0005`) may be formatted using scientific notation (`5e-4`) when it is converted to string using `fmt.Sprintf()`. This results in expressions such as `eval(5e-4)`. SigLens does not currently support the use of `e` in floats.
2. [While generating the tests](https://github.com/t-sunk/siglens/blob/9384a28b1c9870e2239df69b61df8f3d2906fa96/pkg/ast/spl/tests/splParser_test.go#L10176), we use [`strconv.FormatFloat()`](https://pkg.go.dev/strconv#FormatFloat), which is meant for use with [`strconv.ParseFloat()`](https://pkg.go.dev/strconv#ParseFloat). However we were using [a different function](https://github.com/siglens/siglens/blob/9ccb4c73a2a383cacf1a49819d59d31817b6c1ff/pkg/ast/spl/spl.peg#L4010) to parse this string to a float within the grammar, which gave inconsistent results.

# Testing
Running the test `Test_Aggs` 100 times, it passes:
```
cd ~/siglens/pkg/ast/spl/tests/
go test -v -run '^Test_Aggs$' -count=100
```

### DIFF
--- a/pkg/ast/spl/spl.go
+++ b/pkg/ast/spl/spl.go
@@ -25,6 +25,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"io/ioutil"
 	"math"
 	"os"
 	"regexp"
@@ -544,13 +545,13 @@ var g = &grammar{
 			pos:  position{line: 529, col: 1, offset: 14941},
 			expr: &choiceExpr{
 				pos: position{line: 529, col: 10, offset: 14950},
-				alternatives: []any{
+				alternatives: []interface{}{
 					&actionExpr{
 						pos: position{line: 529, col: 10, offset: 14950},
 						run: (*parser).callonStart2,
 						expr: &seqExpr{
 							pos: position{line: 529, col: 10, offset: 14950},
-							exprs: []any{
+							exprs: []interface{}{
 								&labeledExpr{
 									pos:   position{line: 529, col: 10, offset: 14950},
 									label: "indexBlock",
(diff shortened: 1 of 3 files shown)


AndrewHess (approved):
Looks good. I'm a little confused why there was a problem; in the description you say
> [While generating the tests](https://github.com/t-sunk/siglens/blob/9384a28b1c9870e2239df69b61df8f3d2906fa96/pkg/ast/spl/tests/splParser_test.go#L10176), we use [strconv.FormatFloat()](https://pkg.go.dev/strconv#FormatFloat), which is meant for use with [strconv.ParseFloat()](https://pkg.go.dev/strconv#ParseFloat). However we were using [a different function](https://github.com/siglens/siglens/blob/9ccb4c73a2a383cacf1a49819d59d31817b6c1ff/pkg/ast/spl/spl.peg#L4010) to parse this string to a float within the grammar, which gave inconsistent results.

and mention issues with our parser for `5e-4`. But [here](https://github.com/siglens/siglens/blob/9ccb4c73a2a383cacf1a49819d59d31817b6c1ff/pkg/utils/numutils_test.go#L43) we have a test case for our float parser that I believe would handle `5e-4`

t-sunk:
We support using scientific notation for the string passed to `utils.FastParseFloat` [here](https://github.com/siglens/siglens/blob/9ccb4c73a2a383cacf1a49819d59d31817b6c1ff/pkg/utils/numutils.go#L72), but in our `spl.peg` grammar [here](https://github.com/siglens/siglens/blob/9ccb4c73a2a383cacf1a49819d59d31817b6c1ff/pkg/ast/spl/spl.peg#L4497) we do not allow `e` in the format for float. 

So for `5e-4`, the parser does not recognise it as a float-string, and thus `utils.FastParseFloat` is not called.

PR #2869, opened 2025-07-09
SPL: Implement stats estdc, estdc_error

t-sunk:
# Description
Estdc is implemented as an alias to dc, since that is an estimate anyways.
Estdc_error is calculated based on [this function](https://pkg.go.dev/github.com/siglens/go-hll#Hll.RelativeError).

A few details:
1. Hash collisions have not been taken into consideration for estdc_error here, as they're only significant when we have $\geq 10^{15}$ records
2. In the docs for the HLL library [here](https://pkg.go.dev/github.com/siglens/go-hll#readme-hashing), it has been recommended that MurmurHash be used for hashing, but we use `xxhash` instead. 
3. We store the hashes explicitly upto `5120` hashes with default settings. This also means we have zero error until then. After this point, we switch to using HLLs and the error is around `0.4%` with default settings.
Screenshot 2025-07-09 231924

# Testing
Added functional tests and one unit test.
Created a new file for functional tests as all existing files would give 0 as the result for estdc_error.


tools/sigclient/functionalQueries/stats_estdcerr.json
```diff
@@ -0,0 +1,30 @@
...
+        "measureFunctions": [
+            "dcerror"
+        ],
+        "measure": [
+            {
+                "GroupByValues": [
+                    "*"
+                ],
+                "MeasureVal": {
+                    "dcerror": 0.004058444380522224
```
AndrewHess, line 18: Will this test be flaky? Or do we have some leniency for the expected result? Could the computed result vary from run to run (e.g., if the HLL processes records in a different order)?

pkg/segment/aggregations/timechartagg.go
```diff
@@ -394,6 +394,22 @@ func MergeVal(eVal *sutils.CValueEnclosure, eValToMerge sutils.CValueEnclosure,
...
 			}
 		}
+	case sutils.EstdcError:
+		if useAdditionForMerge {
+			aggFunc = sutils.Sum
+		} else {
+			if hllToMerge != nil {
+				err := hll.StrictUnion(hllToMerge.Hll)
+				if err != nil {
+					batchErr.AddError("MergeVal:HLL_STATS", err)
```
AndrewHess, line 404: I'm not sure what this error means

pkg/segment/query/processor/streamstatscommand.go
```diff
@@ -850,6 +860,16 @@ func performMeasureFunc(currIndex int, ssResults *structs.RunningStreamStatsResu
 			CVal:  strValue,
 		}
 
+		ssResults.Window.PushBack(&structs.RunningStreamStatsWindowElement{Index: currIndex, Value: cvalue, TimeInMilli: timestamp})
+	case sutils.EstdcError:
+		// for windowed stream stats, we compute exact cardinality
+		// thus the error is always 0.0
+		ssResults.CurrResult.CVal = 0.0
```
AndrewHess, line 867: Is this true even for large windows?

t-sunk: Yes, this is true. For windowed stats, we need to be able to perform Insertion into the window and Removal from the window. We cannot perform removal using HLL so we don't use that here. We instead store a set of strings. So the error is always zero.

pkg/segment/aggregations/timechartagg.go
t-sunk: StrictUnion gives an error if the two HLLs we're merging have different settings/configurations. I picked that error message because it is what we use for `dc` when we face this issue.

tools/sigclient/functionalQueries/stats_estdcerr.json
t-sunk: I'm looking into why this failed the tests. Here is what I expect to happen:
The computed result will not vary from run to run. This just depends on the default HLL settings. 
Since we have exact float values in other places ([example](https://github.com/t-sunk/siglens/blob/b9973dc8b701e94cf3901546346094c171450c61/tools/sigclient/functionalQueries/stats_group_by_eval.json#L63)), I expect we don't need leniency here either.

t-sunk: Commit [b9973dc](https://github.com/siglens/siglens/pull/2869/commits/b9973dc8b701e94cf3901546346094c171450c61) failed because it used an older version of `go-hll`. It's unrelated to this test.

AndrewHess (approved):

PR #2873, opened 2025-07-11
mv_to_json fixes

t-sunk:
# Description
Fixes 3 issues related to mv_to_json ([docs](https://help.splunk.com/en/splunk-enterprise/search/spl-search-reference/9.3/evaluation-functions/multivalue-eval-functions#ariaid-title15))
1. mv_to_json no longer returns a multivalued field. This function takes a multivalued field as an argument, but the result is a single string. Earlier, this resulted in a multivalued field of length 1.
2. There were some issues when mv_to_json was used with `infer_types` set to true when the first value was null. Sample query: ```* | eval json_arr = mv_to_json_array(split("null,true",","), true()) 
| fields json_arr```
This has been fixed
3. mv_to_json with `infer_types` set to true would not infer JSON objects or arrays previously. (Splunk's documents state that it should do so, but it's not very explicit). This is now fixed.

There are still a few issues:
1. For the second argument to this function, we only accept the strings `true()` or `false()`. Any boolean expression should be valid here. This issue affects other commands as well.
2. I test for this in UTs, but I don't know how to run a query where I test that `mv_to_json` with `infer_types = true` works when we have JSON objects. There is currently a bug where we don't unescape strings properly and I couldn't find a way to create a JSON object as a string in a mv field.
Here's what I have tried:
- `split("{\"key\": 1}", "x")`
- `spath("{\"a\": {\"key\": 1}}", "a")`
- `mvappend("{\"key\": 1}", "true")`

# Testing
Updated parser and functional tests
Added unit tests


tools/sigclient/functionalQueries/mv_to_json_array.json
```diff
@@ -6,68 +6,52 @@
...
                 "ident": "452922de-6bb7-4552-9dcf-e76360d78795",
-                "inferred": [
-                    "[\"Grace\",\"Frank\",true,false,null,12,24,\"Doorride\",7446604,-24.899319,154.051647]"
-                ],
-                "non_inferred": [
-                    "[\"Grace\",\"Frank\",\"true\",\"false\",\"null\",\"12\",\"24\",\"Doorride\",\"7446604\",\"-24.899319\",\"154.051647\"]"
-                ]
+                "inferred": "[null,null,true,false,null,12,24,null,7446604,-24.899319,154.051647]",
+                "non_inferred": "[\"Grace\",\"Frank\",\"true\",\"false\",\"null\",\"12\",\"24\",\"Doorride\",\"7446604\",\"-24.899319\",\"154.051647\"]",
+                "timestamp": 1752224399759
```
AndrewHess: If the timestamp changes from run to run, exclude it here. I assume we don't actually verify the timestamp anyway.

Same comment a couple places in this file

AndrewHess (commented):
Looks pretty good

tools/sigclient/functionalQueries/mv_to_json_array.json
t-sunk: Yes, I believe this is correct. I have removed it.

AndrewHess (approved):

PR #2879, opened 2025-07-15
Fix: Mvsort modifies input field

t-sunk:
# Description
Previously, for a multivalue field `mv` with the values `["b", "a"]`, when we run the query
```
* | eval mv = split("ba", "")
| eval sorted = mvsort(mv)
```
The multivalued field `mv` would be modified to `["a", "b"]`. This has been fixed.

# Testing
Updated mvsort.json to include such a case.


AndrewHess (approved):
Nice

PR #2882, opened 2025-07-15
Support-SPL-boolean-literal-functions

t-sunk:
# Description
1. Updates grammar to use BoolExpr instead of strings "true()" or "false()". This affects results when the parameter is an expression. Sample query:
```
* | eval mv = split("123", ""),
inferTypes = 1, 
json = mv_to_json_array(mv, inferTypes == 1)
```
2. Made corresponding changes to the functions mv_to_json_array and tojson

# Testing
This code passes existing tests which already used true() and false() but parsed it very differently.
I have modified the parser tests to work correctly here.


AndrewHess (approved):
Looks good

PR #2894, opened 2025-07-24
Do not default to * when index name is not specified in a query

nrkawane:


AndrewHess (dismissed):

pkg/ast/pipesearch/searchHandler.go
```diff
@@ -76,7 +76,7 @@ func ParseSearchBody(jsonSource map[string]interface{}, nowTs uint64) (string, u
 
 	iText, ok := jsonSource[KEY_INDEX_NAME]
 	if !ok || iText == "" {
-		indexName = "*"
+		log.Errorf("ParseSearchBody: indexName is not specified!")
```
nkunal, line 79: Log qid here if its available

nrkawane: Callers of the function dont always have it. Eg [listIndices](https://github.com/siglens/siglens/blob/d82d2f9b23969b18b794e701c85996d6b2a0a249/pkg/ast/pipesearch/listIndicesHandler.go#L159).

AndrewHess (approved):

